Issue #40, opened 2017-09-19
create notebook images with renga installed


rokroskar:
make several general notebook images with renga pre-installed based on https://hub.docker.com/r/jupyter/

PR #50, opened 2017-09-25
notebooks: run from an existing file

### LINKED ISSUE #52: link notebook code to an execution
We need a way to link notebook code in the graph

jirikuncar:
addresses #40 and closes #52

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM jupyter/minimal-notebook:latest
+
+COPY . /code/renga
+RUN pip install --no-cache /code/renga
+
+ENTRYPOINT ["tini", "--", "renga", "notebooks", "run"]
--- a/renga/cli/notebooks.py
+++ b/renga/cli/notebooks.py
@@ -24,6 +24,7 @@ from binascii import hexlify
 
 import click
 
+import renga
 from renga import errors
 from renga.cli._options import option_endpoint
 
@@ -70,12 +71,14 @@ def list(config, all, endpoint):
 @click.option('--engine', default='docker')
 @click.option(
     '--image',
-    default='jupyter/minimal-notebook:latest',
+    default='rengahub/minimal-notebook:latest',
     help='Notebook image to use.')
+@click.option('--bucket', type=int, envvar='RENGA_BUCKET_ID')
+@click.option('--file', type=int, envvar='RENGA_FILE_ID')
 @option_endpoint
 @with_config
 @click.pass_context
-def launch(ctx, config, engine, image, endpoint):
+def launch(ctx, config, engine, image, bucket, file, endpoint):
     """Launch a new notebook."""
     cfg = config.get('project', config)['endpoints'][endpoint]
     if ':' not in image:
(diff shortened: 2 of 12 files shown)


renga/cli/notebooks.py
```diff
@@ -68,10 +70,12 @@ def is_notebook(context):
 
 @notebooks.command()
 @click.option('--engine', default='docker')
+@click.option('--bucket', type=int, envvar='RENGA_BUCKET_ID')
```
rokroskar: why use environment variables? not sure it's necessary/useful

we should, however, add an option here to add the code file from disk if `--file` is a path instead of an int

renga/cli/notebooks.py
```diff
@@ -68,10 +70,12 @@ def is_notebook(context):
...
+@click.option('--file', type=int, envvar='RENGA_FILE_ID')
 @option_endpoint
 @with_config
 @click.pass_context
-def launch(ctx, config, engine, endpoint):
+def launch(ctx, config, engine, bucket, file, endpoint):
```
rokroskar: use something other than `file`?

renga/cli/notebooks.py
```diff
@@ -94,11 +98,44 @@ def launch(ctx, config, engine, endpoint):
...
             labels=['renga.notebook.token={0}'.format(notebook_token)],
             endpoint=endpoint)
         cfg = ctx.obj['config'].get('project', config)['endpoints'][endpoint]
         cfg['notebook'] = context.id
 
-    execution = context.run(engine=engine)
-    click.echo(execution.url)
+    execution = context.run(
+        engine=engine,
+        environment={
```
rokroskar: the code used in an execution should be a part of the context not the execution; these should be a part of, and come from, the context spec

renga/cli/notebooks.py
```diff
@@ -94,11 +98,44 @@ def launch(ctx, config, engine, endpoint):
...
+def run(bucket, file, notebook_args):
+    """Open an existing Jupyter notebook."""
+    client = renga.from_env()
+
+    if len(notebook_args) > 1:
+        notebook, *notebook_args = notebook_args
+    else:
+        notebook = 'jupyter-notebook'
+
+    with open('code.ipynb', 'wb') as code:
```
rokroskar: we have the renga client, why not retrieve the actual filename?

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM jupyter/minimal-notebook:latest
+
+USER root
```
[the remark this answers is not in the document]
jirikuncar: yes :)

renga/cli/notebooks.py
jirikuncar: It will be refactored one we have an code node connecting to the context.

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM jupyter/minimal-notebook:latest
```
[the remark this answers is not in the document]
jirikuncar: eventually it would be nice to have something which mirrors `jupyter/docker-stacks`.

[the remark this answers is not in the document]
rokroskar: 👍  exactly

PR #51, opened 2017-09-25
api: use explorer to retrieve file

rokroskar:
use the explorer API to retrieve a file

relates to #52 because it makes it possible to get the file name via the API

### DIFF
--- a/renga/models/storage.py
+++ b/renga/models/storage.py
@@ -23,7 +23,7 @@ import requests
 
 from renga import errors
 
-from ._datastructures import Collection, Model
+from ._datastructures import Collection, LazyResponse, Model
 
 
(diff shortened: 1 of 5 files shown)


renga/models/storage.py
```diff
@@ -139,13 +139,12 @@ def __init__(self, bucket, **kwargs):
...
 
-    def __getitem__(self, resource_id):
+    def __getitem__(self, file_id):
         """Return a file object."""
-        # FIXME use explorer api
-        return self.Meta.model(
-            {
-                'id': resource_id,
-            }, client=self._client, collection=self)
+        return File(
```
jirikuncar: `self.Meta.model`

PR #93, opened 2018-02-05
Cwl

jirikuncar:


renga/cli/workflow.py
```diff
@@ -0,0 +1,52 @@
...
+@click.option('--revision', default='HEAD')
+@click.argument('path', type=click.Path(exists=True, dir_okay=False), nargs=-1)
+@pass_repo
+def create(repo, revision, path):
+    """Create a workflow description for a file."""
+    graph = Graph(repo)
+    for p in path:
+        graph.add_file(p, revision=revision)
+
+    click.echo(
```
rokroskar: shouldn't we save the workflow? give an option to specify the name so it's easier to re-execute?

tests/test_cli.py
```diff
@@ -100,6 +101,25 @@ def test_run_simple(runner):
...
+            try:
+                cli.cli.main(
+                    args=('run', 'wc', 'data.csv'),
+                    prog_name=runner.get_default_prog_name(cli.cli),
+                )
+            except SystemExit as e:
+                assert e.code in {None, 0}
+
+    result = runner.invoke(cli.cli, ['workflow', 'create', 'counted.txt'])
+    assert result.exit_code == 0
```
rokroskar: check that the generated workflow is actually correct?

renga/cli/workflow.py
jirikuncar: sure. but later. I need to solve issue with detecting `basedir`.

tests/test_cli.py
jirikuncar: Once I implement reader for workflows and https://github.com/SwissDataScienceCenter/renga-python/pull/93#discussion_r166132578 is solved.

PR #94, opened 2018-02-07
Gitlab runner

jirikuncar:


Dockerfile.renga-python
```diff
@@ -0,0 +1,7 @@
+FROM python:3.6-alpine
+
+COPY . /code/renga
+WORKDIR /code/renga
+RUN pip install --no-cache -e .[all]
```
rokroskar: don't we want to skip docs at least? bloats the image

jirikuncar: we can make it configurable to create image flavors in the future ...

PR #111, opened 2018-02-19
cli: add status command

### LINKED ISSUE #107: Implement `renga status` that checks the validity of data provenance
- [x] all CWL outputs should be produced from latest version of inputs

--
addresses https://github.com/SwissDataScienceCenter/renga/issues/117

jirikuncar:
- [x] add tests

image

(closes #107)


renga/cli/_graph.py
```diff
@@ -128,6 +131,76 @@ def _output_keys(self):
...
+        for filepath, _ in index.entries.keys():
+            if not filepath.startswith('.renga') and \
+                    filepath not in {'.gitignore', }:
+                self.add_file(filepath, revision=revision)
+                current_files.add(filepath)
+
+        # Prepare status info for each file.
+        self._need_update()
+
+        graph_files = list(sorted(
```
rokroskar: a few things here:

* do you need to turn `graph_files` into a list? it could be very long. 
* since you are doing a groupby in the next step, the sort here doesn't seem necessary either.

renga/cli/_graph.py
```diff
@@ -128,6 +131,76 @@ def _output_keys(self):
...
+            ((commit, filepath) for (commit, filepath) in self.G
+             if filepath in current_files),
+            key=itemgetter(1)
+        ))
+
+        status = {'up-to-date': {}, 'outdated': {}, 'multiple-versions': {}}
+
+        for filepath, files in groupby(graph_files, itemgetter(1)):
```
rokroskar: isn't `files` actually `commits` or `shas`?

renga/cli/_graph.py
```diff
@@ -128,6 +131,76 @@ def _output_keys(self):
         """Return a list of the output keys."""
         return [n for n, d in self.G.out_degree() if d == 0]
 
+    def _need_update(self):
```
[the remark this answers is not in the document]
jirikuncar: The result would be the same since I have multiple starting points.

renga/cli/_graph.py
jirikuncar: `keys` ...

renga/cli/_graph.py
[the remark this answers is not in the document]
rokroskar: right, but you can eliminate most of L139-L154. And why not use an off-the-shelf algorithm if it's there already? Would be easier to drop in a more efficient graph library later.

PR #112, opened 2018-02-21
cli: refactored graph display for log command

### LINKED ISSUE #106: Fix bugs with provenance calculation
- [x] fix DAG display when the same file is used as input in multiple steps

--
addresses https://github.com/SwissDataScienceCenter/renga/issues/117

### LINKED ISSUE #107: Implement `renga status` that checks the validity of data provenance

jirikuncar:
closes #106

### DIFF
--- a/renga/cli/_ascii.py
+++ b/renga/cli/_ascii.py
@@ -1,0 +1,326 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2018 - Swiss Data Science Center (SDSC)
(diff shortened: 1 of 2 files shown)


rokroskar:
@jirikuncar can we just not use the pager then? just do `click.echo`?

jirikuncar:
> can we just not use the pager then? just do `click.echo`?

It gets annoying if you always have to scroll up.

Issue #104, opened 2018-02-16
provide a template LFS configuration for the newly initialized project


rokroskar:
~- [ ] `renga init` should also create a `.gitattributes` file that will sync the `data/` directory.~
* cli functions like `renga run` should add appropriate paths to `git lfs` for tracking
  - [ ] `renga run` invocations should add the outputs to `.gitattributes` by default -- provide a flag to override the behavior (done in #116)
  - [ ] `renga datasets add` should add all files except for `metadata.json`

addresses https://github.com/SwissDataScienceCenter/renga/issues/117

PR #116, opened 2018-02-23
LFS support for renga run

rokroskar:
- initialize LFS on project init
- track CWL tool outputs in LFS

waiting on #115 before adding the integration for datasets

---
addresses #104


renga/cli/init.py
```diff
@@ -68,12 +68,13 @@ def store_directory(ctx, param, value):
         exists=True, writable=True, file_okay=False, resolve_path=True))
 @click.option('--name', callback=validate_name)
 @click.option('--force', is_flag=True)
+@click.option('--lfs/--nolfs', is_flag=True, default=True)
```
jirikuncar: `--lfs/--no-lfs`

PR #123, opened 2018-02-28
dataset: removed repo from model

jirikuncar:


renga/models/dataset.py
```diff
@@ -256,26 +249,23 @@ def _add_from_url(self, path, url, nocopy=False, **kwargs):
         # make the added file read-only
         mode = dst.stat().st_mode & 0o777
         dst.chmod(mode & ~(stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH))
+
+        result = str(dst.relative_to(repo.path / self.path).as_posix())
```
rokroskar: - `as_posix()` already returns a string
- don't need `repo.path` here, use `dst.relative_to(self.path.absolute()).as_posix()` instead?

renga/models/dataset.py
```diff
@@ -208,25 +198,28 @@ def add_data(self, url, datadir=None, git=False, **kwargs):
...
-                    self.files.update(self._add_from_git(self.path, url, t))
+                    self.files.update(
+                        self._add_from_git(repo, self.path, url, t))
         else:
-            self.files.update(self._add_from_url(self.path, url, **kwargs))
+            self.files.update(
+                self._add_from_url(repo, self.path, url, **kwargs))
 
-    def _add_from_url(self, path, url, nocopy=False, **kwargs):
+    def _add_from_url(self, repo, path, url, nocopy=False, **kwargs):
```
rokroskar: don't think `repo` is needed here at all

renga/models/dataset.py
```diff
@@ -319,35 +309,27 @@ def _add_from_git(self, path, url, target):
...
 
         # grab all the authors from the commit history
-        repo = git.Repo(submodule_path.absolute().as_posix())
+        git_repo = git.Repo(submodule_path.absolute().as_posix())
         authors = set(
             Author(name=commit.author.name, email=commit.author.email)
-            for commit in repo.iter_commits(paths=target))
+            for commit in git_repo.iter_commits(paths=target))
 
+        result = str(dst.relative_to(repo.path / self.path).as_posix())
```
rokroskar: same as above -- shouldn't need `repo.path` or `str()`

renga/models/dataset.py
jirikuncar: We do in https://github.com/SwissDataScienceCenter/renga-python/pull/123/files#diff-ceb5f0488f1df82397418d3727d4eb35R231

rokroskar: but that's recursing, we can remove it in both places

PR #137, opened 2018-03-06
cli: do not leave .renga dir in an existing repository unless --force 

spMohanty:


renga/api/repository.py
```diff
@@ -156,8 +157,16 @@ def init_repository(
             if self.git is None:
                 self.git = GitRepo.init(str(path))
         else:
-            assert self.git is None, 'Git repo already exists.'
-            self.git = GitRepo.init(str(path))
+            if self.git is not None:
+                # Cleanup
+                self.renga_path.rmdir()
+                raise QuestionableGitOperation(
+                    'Git repo already exists. Please use the `--force` flag ' +
```
jirikuncar: You can simply rely on Python parser to join static strings, hence please remove the `+`.

renga/api/repository.py
```diff
@@ -156,8 +157,16 @@ def init_repository(
             if self.git is None:
                 self.git = GitRepo.init(str(path))
         else:
-            assert self.git is None, 'Git repo already exists.'
-            self.git = GitRepo.init(str(path))
+            if self.git is not None:
+                # Cleanup
+                self.renga_path.rmdir()
```
jirikuncar: I am not sure if it's a good idea to remove the renga path completely.

renga/api/repository.py
```diff
@@ -149,15 +151,34 @@ def init_repository(
...
+                    'You are trying to initialize a Renga ' +
+                    'repository in an already existing Renga repository. ' +
+                    'If you are sure you know what you are doing, please ' +
+                    'try again with the `--force` flag.'
+                )
+            else:
+                raise
 
         path = self.path.absolute()
         if force:
```
jirikuncar: I would really like to keep the code simple, hence I would propose following:
```python
if force:
    self.renga_path.mkdir(parents=True, exist_ok=force)
    if self.git is None:
        self.git = GitRepo.init(str(path))
else:
    if self.renga_path.exists():
        raise UsageError('You are ...')
    ...
```

renga/api/repository.py
spMohanty: In case of a valid (and working)`renga` repository, If there is an already existing `.renga` directory, then the control never actually reaches this line as [this](https://github.com/SwissDataScienceCenter/renga-python/blob/development/renga/api/repository.py#L152) throws a `FileExistsError` error.
So if the control actually reaches here, then it means that it was created only in the previous `GitRepo.init` call, so I think it should be safe to remove the renga path completely I think ?

jirikuncar: please see https://github.com/SwissDataScienceCenter/renga-python/pull/137/files#r172632507

renga/api/repository.py
```diff
@@ -149,15 +150,35 @@ def init_repository(
...
+            self.renga_path.mkdir(parents=True, exist_ok=force)
             if self.git is None:
                 self.git = GitRepo.init(str(path))
         else:
-            assert self.git is None, 'Git repo already exists.'
-            self.git = GitRepo.init(str(path))
+            if self.renga_path.exists():
+                raise UsageError(
+                    'You are trying to initialize a Renga '
```
jirikuncar: The description should be added to `renga.cli.init`. It is specific to the command line interface, but here we are implementing generic API for Renga repository.

PR #145, opened 2018-03-08
docs: split models and improve formatting

jirikuncar:


docs/index.rst
```diff
@@ -17,17 +17,60 @@
...
+For more information about the Renga API `see its documentation
+<https://renga.readthedocs.org/latest/developer/index.html>`_.
+
+Use the Renga command line
+--------------------------
+
+Interaction with the platform can take place via the command-line
+interface (CLI).
+
+Start by creating for folder where you want to keep your Renga project:
```
rokroskar: This is now duplicated in the README and here -- is that intentional?

rokroskar: we should find a way to not duplicate this text in the "first steps" in the main `renga` repo

jirikuncar: I would remove it from README and just point to docs directly ...

PR #150, opened 2018-03-16
cli: fix detection of dirty files

jirikuncar:
Solves problem with files generated by a workflow.

### DIFF
--- a/renga/cli/_graph.py
+++ b/renga/cli/_graph.py
@@ -97,17 +97,97 @@ class Graph(object):
                     os.path.normpath(basedir / input_.default.path), input_.id
                 )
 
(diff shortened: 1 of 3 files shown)


renga/cli/_graph.py
```diff
@@ -97,18 +97,98 @@ def iter_file_inputs(self, tool, basedir):
...
+                is_step=True,
+            )
+
+            step_tool = self.G.nodes[tool_key]['tool']
+
+            for input_path, input_id in self.iter_file_inputs(
+                step_tool, basedir
+            ):
+                if input_path in commit.stats.files:
+                    #: Check intermediate commited files
```
rokroskar: comimited --> committed

PR #160, opened 2018-04-26
Show git status

jirikuncar:
Can we rely on an output of `git status`?

```console
$ touch x
$ renga run touch y
Error: The repository is dirty. Please use the "git" command to clean it.

On branch master
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	x

nothing added to commit but untracked files present (use "git add" to track)

Once you have added the untracked files, commit them with "git commit".
```

---
addresses https://github.com/SwissDataScienceCenter/renga/issues/186

### DIFF
--- a/renga/cli/_git.py
+++ b/renga/cli/_git.py
@@ -24,6 +24,8 @@ from contextlib import contextmanager
 import click
 from git import Repo
 
+from renga import errors
+
 GIT_KEY = 'renga.git'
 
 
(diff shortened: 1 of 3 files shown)


rokroskar:
@jirikuncar I think in addition to showing the message from `git` we need to also follow with something like

`Once you have added the untracked files, commit them with "git commit"`

PR #162, opened 2018-05-17
global: renga->renku

jirikuncar:


rokroskar:
does coveralls need to be reconfigured for renku?

PR #166, opened 2018-05-18
use reentry to workaround pkg_resources slowness

asteven:
Use https://github.com/DropD/reentry to speed up the cli.

I noticed this short lag during a renku demo and thought that looks familiar. We also use click and click-plugins in some projects so decided to have a look.

### before
```
(renku-orig) [22:30:31] crius:~% timeit renku               
Usage: renku [OPTIONS] COMMAND [ARGS]...

  Check common Renku commands used in various situations.

Options:
  --version            Print version number.
  --config PATH        Location of client config files.
  --config-path        Print application config path.
  --path <path>        Location of a Renku repository.  [default: .]
  --renku-home <path>  Location of Renku directory.  [default: .renku]
  -h, --help           Show this message and exit.

Commands:
  dataset     Handle datasets.
  deactivate  Deactivate environment for tracking work on a...
  help        Show help message and exit.
  init        Initialize a project.
  log         Show logs for a file.
  run         Tracking work on a specific problem.
  runner      Simplify running of CI scripts.
  status      Show a status of the repository.
  update      Update existing files by rerunning their...
  workflow    Workflow operations.
  workon      Activate environment for tracking work on a...
0.81
(renku-orig) [22:30:34] crius:~% 
```

### after
```
(renku) [22:31:16] crius:~% timeit renku
Usage: renku [OPTIONS] COMMAND [ARGS]...

  Check common Renku commands used in various situations.

Options:
  --version            Print version number.
  --config PATH        Location of client config files.
  --config-path        Print application config path.
  --path <path>        Location of a Renku repository.  [default: .]
  --renku-home <path>  Location of Renku directory.  [default: .renku]
  -h, --help           Show this message and exit.

Commands:
  dataset     Handle datasets.
  deactivate  Deactivate environment for tracking work on a...
  help        Show help message and exit.
  init        Initialize a project.
  log         Show logs for a file.
  run         Tracking work on a specific problem.
  runner      Simplify running of CI scripts.
  status      Show a status of the repository.
  update      Update existing files by rerunning their...
  workflow    Workflow operations.
  workon      Activate environment for tracking work on a...
0.59
(renku) [22:31:17] crius:~% 
```

This is not a huge but still noticable speedup.
Then again maybe it's just premature optimization ;-)

Note that when installing using pip in editable mode ('pip install -e') instead of installing from a wheel you may also want to look at
https://github.com/pypa/setuptools/issues/510
and
https://pypi.org/project/fastentrypoints/

PR #169, opened 2018-05-23
global: removal of pkg_resources

jirikuncar:
(addresses #166)


jirikuncar:
@rokroskar you need to install `renku` using wheel.

Check this out:
```console
$ python3.7 -X importtime -c 'import renku.cli' 
```

rokroskar:
ah right, I forgot about that... ok will try

PR #185, opened 2018-05-31
cli: add multiple files to a dataset at once

### LINKED ISSUE #181: handle globs in renku dataset add
We should do something with e.g.

```
renku dataset add ../path/to/data/*
```

cc/ @lusamino

jirikuncar:
(closes #181)

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -335,7 +335,11 @@ def test_streams_and_args_names(runner, capsys):
     assert result.exit_code == 0
 
 
-def test_datasets(data_file, data_repository, runner):
+def test_datasets(
+    data_file,
+    data_repository,
+    runner,
+):
     """Test importing data into a dataset."""
     # create a dataset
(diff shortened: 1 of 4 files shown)


rokroskar:
tests?

jirikuncar:
> tests?

they work :D or do you mean I should write one? (of course I will ;))

tests/test_cli.py
```diff
@@ -372,6 +376,28 @@ def test_datasets(data_file, data_repository, runner):
...
+    assert os.stat('data/dataset/metadata.yml')
+
+    paths = []
+    for i in range(3):
+        new_file = tmpdir.join('file_{0}'.format(i))
+        new_file.write(str(i))
+        paths.append(str(new_file))
+
+    # add data
+    result = runner.invoke(cli.cli, ['dataset', 'add', 'dataset'] + paths)
```
rokroskar: this is just testing that the command runs without an error.... maybe also test that the paths are where they should be?

PR #187, opened 2018-05-31
setup: add pydot dependency to fix gitlab runner

ableuler:
Currently the runner fails when trying to build the dot files for the graph due to a missing library (pydot). Couldn't test if this actually fixes it, but it's my best guess...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,10 @@ extras_require = {
         'Sphinx>=1.6.3',
         'renku-sphinx-theme>=0.1.0',
     ],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -76,6 +76,7 @@
     'filelock>=3.0.0',
     'gitpython>=2.1.8',
     'networkx>=2.1',
+    'pydot>=1.2.4',
```
jirikuncar: Please move it to the `runner` section on line 50 (PS: end the list with comma).

PR #193, opened 2018-06-07
Login to the proxy/gateway

ebiiii:
I had to patch a few things around to not have the config loading crashes (and prevent me from testing the interaction with the proxy/gateway), but it doesn't save the received token anywhere. :(  I would need some help with that.


ebiiii:
I would also need help with the formatter used (yapf).
When I run it, it tells me about some unknown config option (which seems to exists however!!). If I comment the corresponding line in .style.yapf, it works. But then travis doesn't use the same options ?

`% yapf -d renku/cli/login.py
yapf: Unknown style option "ALLOW_SPLIT_BEFORE_DICT_VALUE"
% yapf -d renku/cli/login.py
%
`

rokroskar:
Make sure you're using latest yapf? That option was added recently. Maybe we need to fix the dependency version?

ebiiii:
Thanks, I found the issue. :-)  `yapf` was using the (outdated) version outside of my virtualenv, whereas `python -m yapf` triggered the right one.

Now I just need to figure out how to have the settings written to the local repo or are we keeping the central config file ?

jirikuncar:
@ebiiii I don't think we should write the config to the git repository. It would make it hard to communicate with multiple endpoints from one repository.

jirikuncar:
Another option, I wanted to try is to add custom section to the `.git/config`.

rokroskar:
What's wrong with the system we had in place in the previous version of the cli?

jirikuncar:
> What's wrong with the system we had in place in the previous version of the cli?

In the previous version the configuration file was not part of Git repository (there was no git repo) and stayed always on users machine.

rokroskar:
Right, but this is mostly for things like access tokens so it should only live on the user's machine, no? Or what else is supposed to go in there?

ebiiii:
We have at least three options here:
- central config file (like ~/.config/renku/config.yaml or similar)
- repository specific, but not tracked (like in .git/config or other sibling files)
- repository specific and trakced (like .lfsconfig, .gitattributes, ...)

Each have different purposes and I agree that the last one is not meant for authentication tokens. So the question is do we want to have "per-repository" remote configuration or do we deal with multiple remote in the central config as was done so far?

In the end we may have to split the configuration over those three layers.

renku/cli/login.py
```diff
@@ -35,66 +35,46 @@
...
-@click.option('--password-stdin', is_flag=True)
+@click.option('--url', default='{endpoint}/auth')
 @click.option('--default', is_flag=True)
 @with_config
 @click.pass_context
-def login(
-    ctx, config, endpoint, url, client_id, username, password, password_stdin,
-    default
-):
+def login(ctx, config, endpoint, url, client_id, default):
```
ableuler: I get an exception here about a non-existing positional argument (client_id). It is not used in the code though so simply removing the argument fixed the issue for me.

rokroskar:
I think we should use the previous system for storing tokens -- the path was OS-dependent, e.g. `~/Library/Application Support/renku/config` or something like that on MacOS and `~/.config/renku` on linux etc. That system already assumed that the user might use multiple endpoints so that logic is already taken care of as well.

rokroskar:
@jirikuncar should we keep the config location it as it was previously?

jirikuncar:
The CLI relies purely on Git authentication for now.

PR #197, opened 2018-06-14
cli: update with siblings

### LINKED ISSUE #195: Cannot update papermill artifact notebook
# Summary

Renku update fails with an exception:
```
schema_salad.validate.ValidationException: .renku/workflow/fa22570fc5b148788d864607dcf54903.cwl:1:1: Object
                                                          `.renku/workflow/fa22570fc5b148788d864607dcf54903.cwl`
                                                          is not valid because
                                                            tried `Workflow` but
                                                              * missing required field `inputs`
                                                              * missing required field `outputs`
                                                              * missing required field `steps`
```

# Steps to recreate

Download and extract [update-test.zip](https://github.com/SwissDataScienceCenter/renku-python/files/2093380/update-test.zip).

Do the following in the update test folder:

```
renku init .
renku run papermill 01_notebook.ipynb 01_notebook_run.ipynb -p output_pickle_file output.pkl
renku run papermill 02_notebook.ipynb 02_notebook_run.ipynb -p input_pickle_file output.pkl -p output_csv summary.csv
```

Now, change something in 01_notebook.ipynb (e.g., the size of the random array).

```
git add 01_notebook.ipynb
git commit -m"Updated notebook."
renku status
```
You will see that several files are out of sync.
```
renku update output.pkl
```
01_notebook_run.ipynb is still out of sync.
```
renku update 01_notebook_run.ipynb
```
This fails with an error like the above.

jirikuncar:
closes #195


ciyer:
I'm testing this out. I noticed the following:
- renku show siblings
    - without any file arguments: returns no output.
    - on a file without any siblings: returns no output
    - on multiple files: does not make clear which files are siblings of what.

I think the output should be structured as
[paramter path]
\t-[sibling 1]
\t-[sibling 2] ...

No path provided as an argument should either produce a sibling report for every file OR an error requesting that at least one path is specified -- pick whichever is easier to implement.

ciyer:
renku update path1

- If the path has a sibling, the output prints out usage information, but the usage of the command is correct. It should just print out the error that siblings are missing.

jirikuncar:
> without any file arguments: returns no output.

See the note http://click.pocoo.org/5/arguments/#variadic-arguments

jirikuncar:
> on a file without any siblings: returns no output

It should return at least the original arguments.

renku/cli/show.py
```diff
@@ -0,0 +1,81 @@
...
+
+@click.group()
+def show():
+    """Show information about objects in current repository."""
+
+
+@show.command()
+@click.option('--revision', default='HEAD')
+@click.argument(
+    'paths', type=click.Path(exists=True, dir_okay=False), nargs=-1
```
ciyer: There is no output when no paths are provided. It should produce some output: either siblings for all files or an error that at least one file should be specified.

renku/cli/show.py
```diff
@@ -0,0 +1,81 @@
...
+    """Show siblings for given paths."""
+    graph = Graph(client)
+    graph.build_status(revision=revision)
+
+    nodes = {graph.add_file(path, revision=revision) for path in paths}
+    siblings_ = set()
+    for node in nodes:
+        siblings_ |= graph.siblings(node)
+
+    for _, path in siblings_:
```
ciyer: The output should indicate which set of files the siblings are siblings of.

rokroskar: some CLIs have an option like `--raw` for cases like this.

renku/cli/show.py
```diff
@@ -0,0 +1,87 @@
...
+from ._client import pass_local_client
+from ._git import with_git
+from ._graph import Graph
+
+
+@click.group()
+def show():
+    """Show information about objects in current repository.
+
+    NOTE: The command produces machine readable output.
```
rokroskar: I would argue that the default should be human-readable and we can provide a `--raw` option (or similar)

jirikuncar: @rokroskar we can create an issue for general formatting. I don't think I have time to fix it before Thursday as there are more urgent issues with the workflow update.

jirikuncar: (btw it's already possible to see the hierarchy by using `renku log`)

rokroskar: sure, makes sense -- please make said issue.

PR #226, opened 2018-06-29
brew: generate formula

jirikuncar:


renku/version.py
```diff
@@ -23,4 +23,4 @@
 
 from __future__ import absolute_import, print_function
 
-__version__ = '0.1.0.dev20180613'
+__version__ = '0.1.0.dev20180728'
```
rokroskar: this release is in the future!

PR #230, opened 2018-07-03
Snap

leafty:
This one build and includes python 3.6

How to build:
```bash
docker run -it -v "$PWD:$PWD" -w "$PWD" snapcore/snapcraft snapcraft
```


leafty:
Still not good :/

PR #233, opened 2018-07-04
cli: initial image pull implementation

jirikuncar:

PR #236, opened 2018-07-06
cli: set and query config values

jirikuncar:
addresses #233


renku/cli/config.py
```diff
@@ -0,0 +1,67 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Get and set Renku repository or global options.
+
+Set values
+~~~~~~~~~~
+
+You can set the registry URL to a custom value with:
```
rokroskar: This doc string reads a bit weird -- it says above "Get and set repository or global options" then here it says "set the registry URL". I know that registry and repository aren't the same thing, but someone reading this might think they are being used as synonyms. It might be better to say something like "You can set various Renku configuration options, for example the image registry URL, with a command like: "

renku/cli/image.py
```diff
@@ -17,17 +17,43 @@
 # limitations under the License.
 """Manipulate images related to the Renku project.
 
+Configure registry
+~~~~~~~~~~~~~~~~~~
+
+First, obtain an access token for the registry from the GitLab by going to
```
rokroskar: drop "the" in "the GitLab"

renku/cli/image.py
```diff
@@ -17,17 +17,43 @@
 # limitations under the License.
 """Manipulate images related to the Renku project.
 
+Configure registry
```
rokroskar: "Configure the image registry"

The target demographic of Renku may not know that registry and repository are two different things or that the registry even exists.

renku/cli/image.py
```diff
@@ -17,17 +17,43 @@
...
+If the image has indeed been built and pushed to the registry, you should be
+able to fetch it with:
 
 .. code-block:: console
 
     $ renku image pull
 
-This pulls an image that was built for the current commit.
+This pulls an image that was built for the current commit. You can also fetch
+a image build for a specific commit with:
```
rokroskar: an image... built

PR #238, opened 2018-07-10
cwl: support for initial work dir requirement

### LINKED ISSUE #228: cli: prepare directory structure for output files
- [x] the existing directory structure which overlaps with output globs should be created for each step in `update` or `rerun`
- [x] include tests for script that expect an output directory to exist

jirikuncar:
(closes #228)

### DIFF
--- a/renku/models/cwl/process_requirements.py
+++ b/renku/models/cwl/process_requirements.py
@@ -1,0 +1,87 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2018 - Swiss Data Science Center (SDSC)
+# A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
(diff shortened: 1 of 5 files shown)


mitch-volpi:
Tested and now working, using `renku rerun --with-siblings output.file`

renku/models/cwl/process_requirements.py
```diff
@@ -0,0 +1,91 @@
...
+                    if input_ is not None:
+                        paths.append(input_.default)
+                elif glob:
+                    paths.append(glob)
+            # TODO add Directory
+
+        requirement = cls()
+
+        for path in paths:
+            directory = Path(str(path)).parent
```
rokroskar: is this assuming `path` is always pointing to a file?

PR #249, opened 2018-07-17
cli: install git hooks

### LINKED ISSUE #201: cli: pre commit hook
Warn user if she is editing an output file.

jirikuncar:
* Adds pre-commit hooks for checking if some output files are being overwritten.

(closes #201)


renku/data/pre-commit.sh
```diff
@@ -0,0 +1,47 @@
...
+if [ -z "$(which renku)" ]; then
+  echo 'renku not on path; can not format. Please install renku:'
+  # TODO add system detection and recommend brew for macOS.
+  echo '    pip install renku'
+  exit 2
+fi
+
+MODIFIED_OUTPUTS=$(renku show outputs "${MODIFIED_FILES[@]}")
+if [ "$MODIFIED_OUTPUTS" ]; then
+  echo 'You are trying to update some output files.'
```
ciyer: Reword: "You are trying to update workflow outputs."

renku/data/pre-commit.sh
```diff
@@ -0,0 +1,47 @@
...
+  echo
+  echo 'Modified outputs:'
```
ciyer: I think "Modified workflow outputs" or "Modified files" is clearer

renku/data/pre-commit.sh
```diff
@@ -0,0 +1,47 @@
...
+  for file in "${MODIFIED_OUTPUTS[@]}"; do
+    echo "  $file"
+  done
+  echo
+  echo 'If you are sure, use "git commit --no-verify".'
```
ciyer: Reword: "To commit anyway, use ..."

ciyer:
The semantics of this feature are not totally clear when you commit without verifying. After that. ```renku status``` reports "All files were generated from the latest inputs." But ```renku rerun``` throws a ValidationException.

jirikuncar:
@ciyer when you modify an output you can't `rerun` it anymore. You could maybe do something like:

```
$ edit output.txt
$ git commit output.txt --no-verify -m 'output edited'
$ renku rerun --revision HEAD~1 output.txt
```

renku/data/pre-commit.sh
jirikuncar: The files can be generated by a tool. I will reword it to `... update generated files.`

renku/data/pre-commit.sh
jirikuncar: 👍 `Modified files`

ciyer:
@jirikuncar Ok, I see that. The error I get is the same as when I try to update or rerun a file that was not generated by a workflow. I will create a separate issue for that.

PR #258, opened 2018-07-20
docker: install git-lfs

### LINKED ISSUE #255: Dockerfile: install Git LFS in notebooks
- [ ] make sure all notebook images have Git LFS installed by default

jirikuncar:
(closes #255)


rokroskar:
@jirikuncar I don't think we should be building the notebook images here anymore -- this should be moved to https://github.com/SwissDataScienceCenter/renku/pull/325

PR #260, opened 2018-07-23
cli: add git related variables in .gitlab-ci.yml

### LINKED ISSUE #259: set CI variables
There are a few variables that should be added for fine-tuning the gitlab-ci builds:

```yaml
variables:
  GIT_STRATEGY: clone
  GIT_SSL_NO_VERIFY: "true"
  GIT_SUBMODULE_STRATEGY: recursive
  GIT_LFS_SKIP_SMUDGE: 1
```

Once this is done, the gitlab-ci in `renku-ui` should also be changed.

jirikuncar:
(closes #259)

### DIFF
--- a/renku/cli/.gitlab-ci.yml
+++ b/renku/cli/.gitlab-ci.yml
@@ -1,4 +1,10 @@
-# renku: v1.dev20180619
+# renku: v1.dev20180723
+
+variables:
+  GIT_STRATEGY: fetch
+  GIT_SSL_NO_VERIFY: "true"
+  GIT_SUBMODULE_STRATEGY: recursive
+  GIT_LFS_SKIP_SMUDGE: 1
 
 stages:
   - build


renku/cli/.gitlab-ci.yml
```diff
@@ -1,4 +1,10 @@
-# renku: v1.dev20180619
+# renku: v1.dev20180723
+
+variables:
+  GIT_STRATEGY: clone
+  GIT_SSL_NO_VERIFY: "true"
```
rokroskar: not sure the quotes are needed?

leafty:
Is it ok to see if fetch works fine?

renku/cli/.gitlab-ci.yml
```diff
@@ -1,4 +1,10 @@
-# renku: v1.dev20180619
+# renku: v1.dev20180723
+
+variables:
+  GIT_STRATEGY: clone
```
rokroskar: @leafty is probably right, this should be changed to fetch

rokroskar:
@leafty yes, you're right if we skip smudge the fetch should be sufficient

renku/cli/.gitlab-ci.yml
jirikuncar: yes, because we want string "true" not YAML boolean `true`.

renku/cli/.gitlab-ci.yml
jirikuncar: changed

rokroskar:
@leafty you've tested this setup on staging?

leafty:
I tried on internal.renku.ch with just
```yaml
variables:
  GIT_SSL_NO_VERIFY: "true"
```

rokroskar:
Ok, but without `GIT_LFS_SKIP_SMUDGE: 1` you'll still pull from LFS, no?

leafty:
Sure, but with the fetch strategy only the changed files are downloaded, not all of them.

PR #275, opened 2018-08-22
docs: snap installation

### LINKED ISSUE #271: document using snap to install renku
No snap documentation to be found anywhere.

jirikuncar:
(closes #271)


leafty:
The snap is not working afaik. The simplest way to make it work is to use `--clasic``

jirikuncar:
The Snap packages needs to be tested first. The preferred way is to go with Linuxbrew.

PR #277, opened 2018-08-23
Release v0.2.0rc1

jirikuncar:
- removes legacy models and api


rokroskar:
Do we want to add `tags: true` to the script deploy on travis? That will build the image, correct? But I suppose it won't automagically give use the git tag as the image tag...

jirikuncar:
Let's push docker images for each commit for now.

PR #340, opened 2018-10-31
Make dot output simpler by default

### LINKED ISSUE #339: simplified dot output

In order to give a useful graph for display in the current UI, we need a simplified dot file 

`renku log --simple --format=dot`

This would mean removing all `type` predicates and associated objects; all associations need to be collapsed and the direct edge between the two entities created; `location` predicates need to be folded into the object node.

rokroskar:
# Description

Provides an option for a simplified dot graph generation showing only inputs, outputs, and processes.

Fixes #339 

## Type of change

- New feature (non-breaking change which adds functionality)

## Checklist

## Example output

simple

simple

### DIFF
--- a/renku/cli/_format.py
+++ b/renku/cli/_format.py
@@ -17,6 +17,8 @@
 # limitations under the License.
 """Serializers for graph data."""
 
+import functools
+
 import click
 
 
@@ -62,12 +64,17 @@ def dot(graph, simple=False, landscape=True):
     g.bind('wfprov', 'http://purl.org/wf4ever/wfprov#')
 
     if simple:
-        rdf2dot_simple(g, sys.stdout, landscape)
+        _rdf2dot_simple(g, sys.stdout, landscape)
         return
     rdf2dot(g, sys.stdout)
 
 
-def rdf2dot_simple(g, stream, landscape=True):
+dot_landscape = functools.partial(dot, landscape=True)
+
+dot_portrait = functools.partial(dot, landscape=False)
+
+
+def _rdf2dot_simple(g, stream, landscape=True):
     """Create a simple graph of processes and artifacts."""
     from itertools import chain
 
@@ -145,8 +152,8 @@ def rdf2dot_simple(g, stream, landscape=True):
         node_path = path_re.match(node).groupdict()
         stream.write(
             '\t"{commit}:{path}" '
-            '[shape=box label=<{comment}<br/>'
-            '#{commit}<br/>{path}>] \n'.format(
+            '[shape=box label="{comment}<br/>'
+            '#{commit}<br/>{path}"] \n'.format(
                 comment=content['comment'],
                 commit=node_path['commit'][:5],
                 path=node_path.get('path') or ''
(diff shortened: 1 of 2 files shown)


rokroskar:
@jirikuncar did the refactor we discussed, I think it's gtg

renku/cli/log.py
```diff
@@ -86,9 +86,15 @@
     default=False,
     help='Display commands without output files.'
 )
+@click.option(
```
jirikuncar: I would rather see options:

```
dot
dot-landscape
dot-full
dot-full-landscape
```

jirikuncar: Keeping in mind that `dot` should be reasonable for UI.

renku/cli/log.py
```diff
@@ -86,9 +86,15 @@
...
+    '--simple',
+    is_flag=True,
+    default=False,
+    help='Generate a simplified graph.'
+)
 @click.argument('paths', type=click.Path(exists=True), nargs=-1)
 @pass_local_client
-def log(client, revision, format, no_output, paths):
+def log(client, revision, format, no_output, simple, paths):
```
jirikuncar:
```suggestion
def log(client, revision, format, no_output, paths):
```

renku/cli/log.py
```diff
@@ -101,4 +107,4 @@ def log(client, revision, format, no_output, paths):
     # NOTE shall we warn when "not no_output and not paths"?
     graph.build(paths=paths, revision=revision, can_be_cwl=no_output)
 
-    FORMATS[format](graph)
+    FORMATS[format](graph, simple=simple)
```
jirikuncar:
```suggestion
    FORMATS[format](graph)
```

renku/cli/_format.py
```diff
@@ -41,7 +43,7 @@ def _jsonld(graph, format, *args, **kwargs):
     return json.dumps(output, indent=2)
 
 
-def dot(graph):
+def dot(graph, simple=False, landscape=True):
```
jirikuncar:
```suggestion
def dot(graph, simple=True, landscape=True):
```

renku/cli/_format.py
```diff
@@ -55,20 +57,130 @@ def dot(graph):
...
     )
+
+    g.bind('prov', 'http://www.w3.org/ns/prov#')
+    g.bind('wfdesc', 'http://purl.org/wf4ever/wfdesc#')
+    g.bind('wf', 'http://www.w3.org/2005/01/wf/flow#')
+    g.bind('wfprov', 'http://purl.org/wf4ever/wfprov#')
+
+    if simple:
+        _rdf2dot_simple(g, sys.stdout, landscape)
+        return
```
jirikuncar:
```suggestion
    else:
```

renku/cli/_format.py
```diff
@@ -55,20 +57,130 @@ def dot(graph):
...
     rdf2dot(g, sys.stdout)
```
jirikuncar:
```suggestion
        rdf2dot(g, sys.stdout)
```

renku/cli/_format.py
```diff
@@ -17,10 +17,12 @@
 # limitations under the License.
 """Serializers for graph data."""
 
+import functools
+
 import click
 
 
-def ascii(graph):
+def ascii(graph, **kwargs):
```
jirikuncar: I would remove the `**kwargs` since we don't have generic options that would apply to all formatters.

rokroskar:
@jirikuncar please re-review, I think it's gtg.

renku/cli/_format.py
jirikuncar: @rokroskar do we still need `**kwargs` everywhere?

rokroskar: oops no, I forgot to remove them there.

PR #343, opened 2018-11-02
lfs: make directory in gitattributes recursive

### LINKED ISSUE #342: failure to add files in subdirectory to git-lfs correctly
**Describe the bug**

If the output directory is specified on the command line, it is added to `.gitattributes` as 

```
path/to/dir filter=lfs diff=lfs merge=lfs -text
```

Note from the [git-lfs docs](https://git-scm.com/docs/gitattributes):

> patterns that match a directory do not recursively match paths inside that directory (so using the trailing-slash path/ syntax is pointless in an attributes file; use path/** instead)

This means that whatever files are produced in the output directory are not currently being added to LFS. 

**Expected behavior**

The `.gitattributes` file should instead include

```
path/to/dir/** filter=lfs diff=lfs merge=lfs -text
```

rokroskar:
# Description

Fixes #342 

## Type of change

*Bug fix* (non-breaking change which fixes an issue)

### DIFF
--- a/renku/api/repository.py
+++ b/renku/api/repository.py
@@ -358,12 +358,17 @@ class RepositoryApiMixin(object):
         if HAS_LFS and self.git.config_reader(config_level='repository'
                                               ).has_section('filter "lfs"'):
             # FIXME create configurable filter and respect .gitattributes
-            paths = [
-                path for path in paths if not str(path).endswith('.ipynb')
-            ]
+            track_paths = []
+            for path in paths:
+                path = Path(path)
+                if path.is_dir():
+                    track_paths.append(str(path.joinpath('**')))
+                    continue
+                if not str(path).endswith('.ipynb'):
+                    track_paths.append(str(path))
 
             call(
-                ['git', 'lfs', 'track'] + list(paths),
+                ['git', 'lfs', 'track'] + track_paths,
                 stdout=PIPE,
                 stderr=STDOUT,
                 cwd=str(self.path),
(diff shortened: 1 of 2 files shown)


rokroskar:
```
$ renku run python python_scripts/Preprocess_delay_data.py --delay-data data/vbz-delay/2017/ --out-fold data/vbz-delay-preprocessed/
$ git show
...
+++ b/data/vbz-delay-preprocessed/fahrzeitensollist2017010120170107.csv.pkl.gz
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:6adbc23e70f42e9cdbe83f96c046359e536a704793009d1c56eedb4bf0d6e79e
+size 14951298
$ tail -n1 .gitattributes
data/vbz-delay-preprocessed/** filter=lfs diff=lfs merge=lfs -text
```

rokroskar:
```
$ renku run python src/clean_data.py data/zhvelo data/preprocessed/zhbikes.parquet
$ git show
diff --git a/.gitattributes b/.gitattributes
new file mode 100644
index 0000000..95c541f
--- /dev/null
+++ b/.gitattributes
@@ -0,0 +1 @@
+data/preprocessed/zhbikes.parquet filter=lfs diff=lfs merge=lfs -text
...
diff --git a/data/preprocessed/zhbikes.parquet b/data/preprocessed/zhbikes.parquet
new file mode 100644
index 0000000..3de547e
--- /dev/null
+++ b/data/preprocessed/zhbikes.parquet
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:af010cbda62ee69fdd21c7413e447a845f4e1daeeed342dbb87cdfdbd794f41e
+size 2922679
```

rokroskar:
So the ✋ test seems to be ok. But we should add an 🚘 test for this.

rokroskar:
@jirikuncar I think this is gtg, can you have a look again?

PR #347, opened 2018-11-05
lfs: check if lfs is installed and fail by default otherwise

### LINKED ISSUE #302: Data is not automatically tracked by git-lfs in projects created from the UI
**Describe the bug**
renku dataset add has different behavior depending on whether the project is created from the UI, or with the renku CLI.

**To Reproduce**
renku --version (local): 0.2.0rc2.dev20180824

Create a new project from the UI, clone the project locally, and in the local repo, create a data set and add data to it with the renku command line. The new data is not tracked by git lfs (there's no .gitattributes file).

**Expected behavior**
I expect the data to be tracked by git lfs. This is the behavior we get when adding data to a project created with renku init.

### LINKED ISSUE #346: Insist on LFS being present and configured when doing renku commands

`renku run` will happily execute and add outputs to the git repo directly if no LFS is configured. We should check that LFS is there and warn the user if that is not the case. 

The default should be to make the user very aware that they might be adding data straight to git -- so if no LFS is installed or it is not configured for the repo, we should raise an exception. It should be possible to override this behavior with e.g. a `--no-lfs` flag to `renku run/update/rerun`. 

Doing something more clever. Still thinking. 

**Additional context**

This was a *huge* problem during our tutorials because it can cause changes that are difficult to revert and manifest in subtle ways but can e.g. crash containers and/or laptops.

rokroskar:
# Description

Fixes #346 
Closes #302

## Type of change

* Bug fix (non-breaking change which fixes an issue)
* New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have performed a self-review of my own code.


renku/api/repository.py
```diff
@@ -353,10 +354,15 @@ def init_external_storage(self, force=False):
...
         )
 
-    def track_paths_in_storage(self, *paths):
+    def lfs_installed(self):
+        """Check that Large File Storage is installed."""
+        return HAS_LFS and self.git.config_reader(
+            config_level='repository'
+        ).has_section('filter "lfs"')
+
+    def track_paths_in_storage(self, *paths, lfs=True):
```
jirikuncar:
```suggestion
    def track_paths_in_storage(self, *paths, use_external_storage=True):
```

renku/api/repository.py
```diff
@@ -353,10 +354,15 @@ def init_external_storage(self, force=False):
...
         """Track paths in the external storage."""
-        if HAS_LFS and self.git.config_reader(config_level='repository'
-                                              ).has_section('filter "lfs"'):
+        if lfs and self.lfs_installed():
```
jirikuncar:
```suggestion
        if use_external_storage and self.external_storage_installed():
```

renku/api/repository.py
```diff
@@ -368,6 +374,8 @@ def track_paths_in_storage(self, *paths):
                 stderr=STDOUT,
                 cwd=str(self.path),
             )
+        elif lfs:
```
jirikuncar:
```suggestion
        elif use_external_storage:
```

renku/api/repository.py
```diff
@@ -368,6 +374,8 @@ def track_paths_in_storage(self, *paths):
...
+            raise errors.LFSNotInstalled(self.git)
```
jirikuncar:
```suggestion
            raise errors.ExternalStorageNotInstalled(self.git)
```

renku/cli/dataset.py
```diff
@@ -122,9 +122,15 @@ def create(client, name):
     multiple=True,
     help='Target path in the git repo.'
 )
+@click.option(
```
```suggestion
@click.option(
    'use_external_storage',
    '--external-storage/--no-external-storage',
    ' /-S',
    is_flag=True,
    default=True,
    help='Configure the file storage service.'
)
```

renku/cli/dataset.py
```diff
@@ -122,9 +122,15 @@ def create(client, name):
...
+    '--lfs/--no-lfs',
+    is_flag=True,
+    default=True,
+    help='Allow command without LFS configured.'
+)
 @pass_local_client
 @with_git()
-def add(client, name, urls, nocopy, relative_to, target):
+def add(client, name, urls, nocopy, relative_to, target, lfs):
```
```suggestion
def add(client, name, urls, nocopy, relative_to, target, use_external_storage):
```

renku/cli/dataset.py
```diff
@@ -137,6 +143,7 @@ def add(client, name, urls, nocopy, relative_to, target):
                         nocopy=nocopy,
                         target=target,
                         relative_to=relative_to,
+                        lfs=lfs
```
```suggestion
                        use_external_storage=use_external_storage,
```

renku/cli/run.py
```diff
@@ -118,10 +118,16 @@
     callback=lambda _, __, values: [int(value) % 256 for value in values],
     help='Allowed command exit-code.',
 )
+@click.option(
```
jirikuncar: use_external_storage

renku/cli/run.py
```diff
@@ -118,10 +118,16 @@
...
+    '--lfs/--no-lfs',
+    is_flag=True,
+    default=True,
+    help='Allow command without LFS configured.'
+)
 @click.argument('command_line', nargs=-1, type=click.UNPROCESSED)
 @pass_local_client
 @with_git(clean=True, up_to_date=True, commit=True, ignore_std_streams=True)
-def run(client, no_output, success_codes, command_line):
+def run(client, no_output, success_codes, lfs, command_line):
```
jirikuncar:
```suggestion
def run(client, no_output, success_codes, use_external_storage, command_line):
```

renku/cli/run.py
```diff
@@ -140,7 +146,7 @@ def run(client, no_output, success_codes, command_line):
     )
 
     with client.with_workflow_storage() as wf:
-        with factory.watch(client, no_output=no_output) as tool:
+        with factory.watch(client, no_output=no_output, lfs=lfs) as tool:
```
jirikuncar:
```suggestion
        with factory.watch(client, no_output=no_output, use_external_storage=use_external_storage) as tool:
```

renku/errors.py
```diff
@@ -202,3 +202,29 @@ def __init__(self, returncode, success_codes=None):
 
 class NotFound(APIError):
     """Raise when an API object is not found."""
+
+
+class LFSNotInstalled(RenkuException, click.ClickException):
```
jirikuncar:
```suggestion
class ExternalStorageNotInstalled(RenkuException, click.ClickException):
```

renku/models/cwl/command_line_tool.py
```diff
@@ -208,7 +208,7 @@ def generate_tool(self):
         )
 
     @contextmanager
-    def watch(self, repo=None, no_output=False):
+    def watch(self, repo=None, no_output=False, lfs=True):
```
jirikuncar:
```suggestion
    def watch(self, repo=None, no_output=False, use_external_storage=True):
```

renku/models/cwl/command_line_tool.py
```diff
@@ -264,7 +264,7 @@ def watch(self, repo=None, no_output=False):
             tool.inputs = list(inputs.values())
             tool.outputs = outputs
 
-            repo.track_paths_in_storage(*paths)
+            repo.track_paths_in_storage(*paths, lfs=lfs)
```
jirikuncar:
```suggestion
            repo.track_paths_in_storage(*paths, use_external_storage=use_external_storage)
```

tests/test_cli.py
```diff
@@ -604,6 +604,29 @@ def test_show_inputs(tmpdir_factory, project, runner, run):
             } == set(result.output.strip().split('\n'))
 
 
+def test_lfs_config(isolated_runner):
```
jirikuncar:
```suggestion
def test_configuration_of_external_storage(isolated_runner):
```

renku/api/repository.py
rokroskar: shouldn't we go the other way (call it `lfs` everywhere instead of `external_storage`?

jirikuncar: We had the discussion about using technology agnostic names in the past ...

rokroskar: Indeed, I remember it well, but at some point I think we might just be causing confusion atm. I'm open to being convinced otherwise though.

jirikuncar: We can provide multiple storage options in the future.

renku/api/repository.py
```diff
@@ -353,10 +354,15 @@ def init_external_storage(self, force=False):
             cwd=str(self.path.absolute()),
         )
 
-    def track_paths_in_storage(self, *paths):
+    def external_storage_installed(self):
```
jirikuncar:
```suggestion
    @property
    def external_storage_installed(self):
```

```suggestion
    def is_external_storage_installed(self):
```

renku/errors.py
```diff
@@ -202,3 +202,29 @@ def __init__(self, returncode, success_codes=None):
...
 class NotFound(APIError):
     """Raise when an API object is not found."""
+
+
+class ExternalStorageNotInstalled(RenkuException, click.ClickException):
+    """Raise when LFS is required but not found or installed in the repo."""
+
+    def __init__(self, repo):
+        """Build a custom message."""
+        msg = 'Git-LFS is either not installed or not configured '\
```
jirikuncar: can you use brackets instead of `\`?

renku/errors.py
```diff
@@ -202,3 +202,29 @@ def __init__(self, returncode, success_codes=None):
...
+
+        super(ExternalStorageNotInstalled, self).__init__(msg)
+
+
+class UninitializedProject(RenkuException, click.ClickException):
+    """Raise when a project does not seem to have been initialized yet."""
+
+    def __init__(self, repo_path):
+        """Build a custom message."""
+        msg = '{repo_path} does not seem to be a Renku project.\n'\
```
jirikuncar: see https://github.com/SwissDataScienceCenter/renku-python/pull/347/files#r231187454

rokroskar:
@jirikuncar I refactored the `--no-external-storage` option to occur at the top level. At the moment this requires that the flag is specified before the subcommands, which is OK in my opinion.

rokroskar:
@jirikuncar note also that in 3dda9af I removed the explicit check for LFS to be configured locally -- IMO there is no need and it is counter-intuitive as that is not how git settings typically work.

rokroskar:
Turns out that testing the lack of `filter.lfs` section is not easy because it may require sudo on the developer's machine to remove the section from the system git configuration. So unless we feel ok using sudo (which may break tests because it will ask for a password) I don't really have a good way of testing this.

jirikuncar:
@rokroskar we can mock the `read_config` method.

rokroskar:
true, though that sounds like a lot of work :)

jirikuncar:
@rokroskar I have a patch for you

renku/cli/dataset.py
```diff
@@ -136,7 +136,7 @@ def add(client, name, urls, nocopy, relative_to, target):
                         url,
                         nocopy=nocopy,
                         target=target,
-                        relative_to=relative_to,
+                        relative_to=relative_to
```
```suggestion
                        relative_to=relative_to,
```

renku/cli/init.py
```diff
@@ -97,23 +98,15 @@ def store_directory(ctx, param, value):
 )
 @click.option('--name', callback=validate_name)
 @click.option('--force', is_flag=True, help='Override project templates.')
-@click.option(
```
jirikuncar: we could keep it here for backward compatibility and include it in `evolve`

renku/api/datasets.py
```diff
@@ -174,7 +174,7 @@ def _add_from_url(self, dataset, path, url, nocopy=False, **kwargs):
         mode = dst.stat().st_mode & 0o777
         dst.chmod(mode & ~(stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH))
 
-        self.track_paths_in_storage(str(dst.relative_to(self.path)))
+        self.track_paths_in_storage(str(dst.relative_to(self.path)), **kwargs)
```
```suggestion
        self.track_paths_in_storage(str(dst.relative_to(self.path)))
```

rokroskar:
@jirikuncar please re-review -- issues addressed, tests passing.

PR #353, opened 2018-11-14
WIP: Update run.py

LiliGasser:
# Description

I've made changes to how files and folders are detected. 
I left three questions in there marked with ??? at the beginning of the line

## Type of change

## Checklist

### DIFF
--- a/renku/cli/run.py
+++ b/renku/cli/run.py
@@ -28,32 +28,47 @@ Tracking exection of your command line script is done by simply adding the
 * input files if linked to existing files in the repository,
 * output files if modified or created while running the command.
 
+Input and output files and folders can only be detected, if they are passed as
+arguments to ``renku run``.
+
 .. note:: If there were uncommitted changes then the command fails.
    See :program:`git status` for details.
 
-Detecting input files
-~~~~~~~~~~~~~~~~~~~~~
+Detecting input files and folders
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-An argument is identified as an input file only if its path matches an existing
-file in the repository. There might be several situations when the detection
-might not work as expected:
+Any file, passed as an argument to ``renku run``, which was not changed during 
+the execution is identified as an input file. The identification only works if 
+the path associated with the argument matches an existing file in the repository. 
 
-* If the file is modified during the execution, then it is stored as an output;
-* If the path points to a directory, then it is stored as a string option;
-* The input file is not defined as the tool argument.
+??? how are folders detected?
 
-Detecting output files
-~~~~~~~~~~~~~~~~~~~~~~
+There might be several situations when the detection might not work as expected
+if:
+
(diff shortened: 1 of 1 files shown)


renku/cli/run.py
```diff
@@ -28,32 +28,47 @@
...
 
+Input and output files and folders can only be detected, if they are passed as
+arguments to ``renku run``.
+
 .. note:: If there were uncommitted changes then the command fails.
    See :program:`git status` for details.
 
-Detecting input files
-~~~~~~~~~~~~~~~~~~~~~
+Detecting input files and folders
```
rokroskar: maybe we can refer to these collectively as "paths"? So this would be "Detecting input paths"

renku/cli/run.py
```diff
@@ -28,32 +28,47 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Any file, passed as an argument to ``renku run``, which was not changed during 
+the execution is identified as an input file. The identification only works if 
+the path associated with the argument matches an existing file in the repository. 
+
+??? how are folders detected?
```
rokroskar: same as files -- see above, we should just refer to them as paths

renku/cli/run.py
```diff
@@ -28,32 +28,47 @@
...
 
-* If the file is modified during the execution, then it is stored as an output;
-* If the path points to a directory, then it is stored as a string option;
-* The input file is not defined as the tool argument.
+* a file is modified during the execution, it will be stored as an output;
+* an input file is not passed as an argument to ``renku run``.
 
-Detecting output files
-~~~~~~~~~~~~~~~~~~~~~~
+Detecting output files and folders
```
rokroskar: --> output paths

renku/cli/run.py
```diff
@@ -28,32 +28,47 @@
...
-ends with an error: ``Error: There are not any detected outputs in the
-repository.`` In order to resolve it remove any proposed input file from the
-list first.
+There might be situations where an existing output file was not changed when the 
+command has been executed with different parameters. This file is then recognized
+as an input file.
+
+1. If there are no other output files, the execution ends with an error: 
+``Error: There are not any detected outputs in the repository.`` In order to 
+resolve it remove any proposed input file from the list first.
```
rokroskar: I believe it's "output file" ?

renku/cli/run.py
```diff
@@ -28,32 +28,47 @@
...
+??? which list?
+
+2. If there are other output files, the execution does not end with the above error.
+??? What to do here? Maybe I want to keep the file as it is and another change to the
```
rokroskar: I believe in that case you would need to use the `--no-output` option? @jirikuncar can you confirm?

jirikuncar: The confusing part is if you have 2 outputs and only one is detected because the other one was not changed. Then you won't get an error but also no warning.

jirikuncar:
integrated in #357

PR #354, opened 2018-11-14
cli: named workflows

jirikuncar:
Allows adding names to CWL files.

### DIFF
--- a/renku/cli/workflow.py
+++ b/renku/cli/workflow.py
@@ -15,9 +15,23 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Workflow operations."""
+"""Manage the set of CWL files created by ``renku`` commands.
+
+With no arguments, shows a list of captured CWL files. Several subcommands
+are available to perform operations on CWL files.
+
+Reference tools and workflows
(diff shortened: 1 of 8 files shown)


rokroskar:
Thinking about how such a feature would be used:

1. some combination of steps is used to produce an output; the user wants to remember this and creates a named workflow (e.g. "generate cumulative plot"
2. the user wants to "remember" a single important step, so they want to name it, e.g. "corpus word count"

How would I go about figuring out which cwl file I need to name? If I do 

```
$ renku workflow
e2fe342b7cd54636ac4f6455dd673729_python.cwl:
f32897546b0d4e1db84671a17280a0f1_python.cwl:
```

it's not obvious what they actually do.

renku/cli/workflow.py
```diff
@@ -15,22 +15,128 @@
...
+are available to perform operations on CWL files.
+
+Reference tools and workflows
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Managing large number of tools and workflows with automatically generated
+names may be cumbersome. The names can be added to the last executed
+``run``, ``rerun`` or ``update`` command by running
+``renku workflow set-name <name>``. The name can be added to an arbitrary
+file in ``.renku/workflow/*.cwl`` anytime later.
```
rokroskar:
```suggestion
file in ``.renku/workflow/*.cwl`` anytime later with ``renku workflow set-name <name> <cwl-path>``.
```

PR #358, opened 2018-11-15
cli: automatically initialize submodules

### LINKED ISSUE #355: automatically initialize submodules

When a project is cloned with a submodule, it's typically necessary to do 

```
git submodule update --init
```

If this has not been done, the symbolic links will point to nonexistent files. We should detect if this is the case and automatically fix it. 

`renku run` and other commands should do a submodule check and resolve the situation accordingly. 

related to https://github.com/SwissDataScienceCenter/renku/issues/441

rokroskar:
# Description

- [x] automatically initialize submodules
- [ ] when adding datasets via a submodule, set relative path if needed (when the two remote paths are on the same server)

closes #355
closes https://github.com/SwissDataScienceCenter/renku/issues/441

## Type of change

## Checklist


tests/test_cli.py
```diff
@@ -581,6 +581,46 @@ def test_relative_git_import_to_dataset(tmpdir, runner, project):
...
+
+    imported_woop = Path(project) / 'data' / 'foo' / woop.name
+    assert imported_woop.exists()
+
+    dst_project = Path(str(tmpdir_factory.mktemp('dst_project')))
+    subprocess.call(['git', 'clone', project, dst_project])
+    dst_woop = Path(dst_project) / 'data' / 'foo' / 'woop'
+    assert not dst_woop.exists()
+
+    with open(Path(dst_project) / 'woop.wc', 'w') as stdout:
```
jirikuncar:
```suggestion
    with (Path(dst_project) / 'woop.wc').open('w') as stdout:
```

tests/test_cli.py
```diff
@@ -581,6 +581,46 @@ def test_relative_git_import_to_dataset(tmpdir, runner, project):
...
+        with contextlib.redirect_stdout(stdout):
```
jirikuncar: this can be replaced by: `assert 0 == run(args=(...), stdout=stdout)`

jirikuncar: the `with Path` block is not necessary if you use the proposed code above.

renku/api/repository.py
```diff
@@ -104,6 +104,10 @@ def __attrs_post_init__(self):
         #: Create an instance of a Git repository for the given path.
         try:
             self.git = Repo(str(self.path))
+            # initialize submodules
```
jirikuncar: Wouldn't be better to move it to `with_git` decorator or create a new Click option with callback that can be used only with commands where the submodule init is necessary?

rokroskar: I thought about moving it to `with_git` but thought that since the submodules are in most cases an integral part of the project, it should just always be done. It incurs very minimal overhead after the first call.

tests/test_cli.py
rokroskar: thanks, much better :)

renku/api/repository.py
jirikuncar:
```suggestion
            # FIXME refactor submodule initialisation to avoid side-effects
```

jirikuncar: Let's keep it for now with a `FIXME` comment.

rokroskar: 👍

renku/api/_git.py
```diff
@@ -0,0 +1,102 @@
+# A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
```
jirikuncar: can we move this file to `renku/models/urls.py`?

renku/cli/_docker.py
```diff
@@ -24,90 +24,7 @@
...
-    @property
-    def image(self):
-        """Return image name."""
-        img = self.hostname
-        if self.owner:
-            img += '/' + self.owner
-        if self.name:
-            img += '/' + self.name
-        return img
+from renku.api._git import GitURL
```
jirikuncar:
```suggestion
from renku.models.urls import GitURL
```

rokroskar:
@jirikuncar thanks for the tip with the test -- it's passing now

tests/test_cli.py
```diff
@@ -581,6 +582,41 @@ def test_relative_git_import_to_dataset(tmpdir, runner, project):
...
+
+    imported_woop = Path(project) / 'data' / 'foo' / woop.name
+    assert imported_woop.exists()
+
+    dst_project = Path(str(tmpdir_factory.mktemp('dst_project')))
+    subprocess.call(['git', 'clone', project, str(dst_project)])
+    dst_woop = Path(dst_project) / 'data' / 'foo' / 'woop'
+    assert not dst_woop.exists()
+
+    assert runner.invoke(
```
jirikuncar:
```suggestion
    result = runner.invoke(
```

tests/test_cli.py
```diff
@@ -581,6 +582,41 @@ def test_relative_git_import_to_dataset(tmpdir, runner, project):
...
+        cli.cli, [
+            '--path',
+            str(dst_project), 'run', '--no-output', 'wc',
+            str(dst_woop.absolute())
+        ]
+    )
+
```
jirikuncar:
```suggestion
    assert 0 == result.exit_code
    
```

PR #369, opened 2018-11-26
api: refactoring client to support isolation

### LINKED ISSUE #359: Running several instances of `renku run` at the same time
Simply, some simulations might take long, and it would be desirable to have the possibility of running several renku at the same time, and leave then running, taking care of avoiding any collisions

In principle we might think that the inputs will be the same, but each output file will have a different suffix, to avoid collisions. Then, simply allow this possibility.

jirikuncar:
- Adds support for `renku run --isolation <CMD>`.
- Renames `client.git` to `client.repo`.
- Adds supports for `python -m renku` (tip: `time python -m renku` vs `time renku`)

*closes #359*


rokroskar:
what happens with submodules?

jirikuncar:
@rokroskar it should all work well, since a new client is created for the isolated environment.

PR #372, opened 2018-11-27
models: allow for '.' and '-' in repo path

### LINKED ISSUE #371: broken username regex
This string doesn't match any regex in the git repo parsing: `https://user:pass@example.com/prefix/owner.name/repo.git`

*Reported-by: @lusamino*

rokroskar:
closes #371 

- Bug fix (non-breaking change which fixes an issue)
- I have added tests to cover my changes.
- I have run ``./run-tests.sh`` locally.

### DIFF
--- a/renku/models/_git.py
+++ b/renku/models/_git.py
@@ -41,12 +41,12 @@ _RE_PORT = r':(?P<port>\d+)'
 
 _RE_PATHNAME = (
     r'/(?P<pathname>'
-    r'(((?P<owner>[\w-]+)/)?(?P<name>[\w\-]+)(\.git)?)?)'
+    r'(((?P<owner>[\w\-\.]+)/)?(?P<name>[\w\-\.]+)(\.git)?)?)'
 )
 
 _RE_PREFIXED_PATHNAME = (
-    r'/(?P<pathname>(([\w\-\~]+)/)+'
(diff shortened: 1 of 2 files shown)


renku/models/_git.py
```diff
@@ -96,7 +103,7 @@ class GitURL(object):
     password = attr.ib(default=None)
     port = attr.ib(default=None)
     owner = attr.ib(default=None)
-    name = attr.ib(default=None)
+    name = attr.ib(default='', converter=filter_repo_name)
```
jirikuncar:
```suggestion
    name = attr.ib(default=None, converter=filter_repo_name)
```

renku/models/_git.py
```diff
@@ -81,6 +81,13 @@ def _build(*parts):
 )
 
 
+def filter_repo_name(repo_name):
+    """Remove the .git extension from the repo name."""
+    if repo_name.endswith('.git'):
```
jirikuncar:
```suggestion
    if repo_name is not None and repo_name.endswith('.git'):
```

renku/models/_git.py
```diff
@@ -81,6 +81,13 @@ def _build(*parts):
...
+        return repo_name.split('.git')[0]
```
jirikuncar:
```suggestion
        return repo_name[:-len('.git')]
```

Issue #387, opened 2019-01-21
renku update failing with "Error collecting output for parameter 'output_0'"


LiliGasser:
**Describe the bug**
A `renku update --with-siblings` leads to the following error:
```
[job step_1] Job error:
Error collecting output for parameter 'output_0':
.renku/workflow/d508c7c5f5914d498e5ad4c660d799b9_python.cwl:50:7: Did not find output file with glob pattern: '['data/politicians/lastnames/1941_lastnames.pickle']'
[job step_1] completed permanentFail
[step step_1] Output is missing expected field file:///home/lili/NLP_DemocraSci/democrasci_preprocwp1/.renku/workflow/aa59806b8a46486e8126a83f5dc50c61.cwl#step_1/output_50
[step step_1] Output is missing expected field file:///home/lili/NLP_DemocraSci/democrasci_preprocwp1/.renku/workflow/aa59806b8a46486e8126a83f5dc50c61.cwl#step_1/output_31
[step step_1] Output is missing expected field file:///home/lili/NLP_DemocraSci/democrasci_preprocwp1/.renku/workflow/aa59806b8a46486e8126a83f5dc50c61.cwl#step_1/output_117
[step step_1] Output is missing expected field file:///home/lili/NLP_DemocraSci/democrasci_preprocwp1/.renku/workflow/aa59806b8a46486e8126a83f5dc50c61.cwl#step_1/output_73
[step step_1] Output is missing expected field file:///home/lili/NLP_DemocraSci/democrasci_preprocwp1/.renku/workflow/aa59806b8a46486e8126a83f5dc50c61.cwl#step_1/output_62
[step step_1] Output is missing expected field file:///home/lili/NLP_DemocraSci/democrasci_preprocwp1/.renku/workflow/aa59806b8a46486e8126a83f5dc50c61.cwl#step_1/output_2
..... many more
[step step_1] Output is missing expected field file:///home/lili/NLP_DemocraSci/democrasci_preprocwp1/.renku/workflow/aa59806b8a46486e8126a83f5dc50c61.cwl#step_1/output_54
[step step_1] Output is missing expected field file:///home/lili/NLP_DemocraSci/democrasci_preprocwp1/.renku/workflow/aa59806b8a46486e8126a83f5dc50c61.cwl#step_1/output_56
[...]

PR #390, opened 2019-01-25
cli: remove outputs before run

jirikuncar:
# Description

Automatically removes specified outputs before executing a `renku run` command.

Addresses #387 

## Type of change

## TODO

- [x] write tests

### DIFF
--- a/renku/api/_git.py
+++ b/renku/api/_git.py
@@ -75,6 +75,17 @@ def _clean_streams(repo, mapped_streams):
                 fp.write(blob.data_stream.read())
 
 
+def _expand_directories(paths):
+    """Expand directory with all files it contains."""
+    for path in paths:
+        path_ = Path(path)
+        if path_.is_dir():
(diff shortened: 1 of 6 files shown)


renku/api/_git.py
```diff
@@ -75,6 +75,17 @@ def _clean_streams(repo, mapped_streams):
                 fp.write(blob.data_stream.read())
 
 
+def _expand_directories(paths):
```
rokroskar: this could get expensive and/or slow

renku/cli/run.py
```diff
@@ -133,6 +133,12 @@
 
 
 @click.command(context_settings=dict(ignore_unknown_options=True, ))
+@click.option(
+    'outputs',
+    '--output',
+    multiple=True,
+    help='Force a path to be an output.',
```
rokroskar:
```suggestion
    help='Force a path to be considered an output.',
```

Issue #370, opened 2018-11-27
Create SPARQL query for common graph traversals


rokroskar:
- [ ] `renku status`
- [ ] `renku update`

Panaetius:
Done in the Graph PoC

PR #393, opened 2019-02-01
models: export path and parent directories

jirikuncar:
# Description

Exports missing triples and adds `path` as `schema.org/name` for `Entities`.

Addresses #370

### DIFF
--- a/renku/models/provenance/activities.py
+++ b/renku/models/provenance/activities.py
@@ -51,6 +51,7 @@ def _nodes(output, parent=None):
 @jsonld.s(
     type='prov:Activity',
     context={
+        'schema': 'http://schema.org/',
         'prov': 'http://www.w3.org/ns/prov#',
         'rdfs': 'http://www.w3.org/2000/01/rdf-schema#',
     },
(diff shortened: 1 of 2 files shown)


renku/models/provenance/entities.py
```diff
@@ -35,7 +35,12 @@ class CommitMixin:
 
     commit = attr.ib(default=None, kw_only=True)
     client = attr.ib(default=None, kw_only=True)
-    path = attr.ib(default=None, kw_only=True, converter=_str_or_none)
+    path = jsonld.ib(
+        context='schema:name',
```
rokroskar:
```suggestion
        context='prov:atLocation',
```

renku/models/provenance/activities.py
```diff
@@ -79,6 +80,10 @@ class Activity(CommitMixin):
             '@reverse': 'prov:activity',
         }, kw_only=True, hash=False
     )
+    influenced = jsonld.ib(
+        context='prov:influenced',
```
rokroskar: where is this used?

jirikuncar: It's needed to mark parent directories as changed.

PR #412, opened 2019-02-12
api: move metadata to .renku/datasets

### LINKED ISSUE #396: Error that output directory is not empty because of metadata
**Describe the bug**
In case an output-directory is defined and it is also a renku-dataset, renku give the following error:
`Error: The output directory "data/VOM_output/AdelaideRiver/best" is not empty. `

`Delete existing files before running the command:`
`  (use "git rm <file>..." to remove them first)`

`	data/VOM_output/AdelaideRiver/best/metadata.yml`

`Once you have removed files that should be used as outputs,`
`you can safely rerun the previous command.`

**To Reproduce**
Just to illustrate, I did this simple example on this repository:
https://renkulab.io/gitlab/remko.nijzink/vom_test 
commit 3106563633f1c7b274311859a2648f3996a61597

`renku dataset create results`
`renku run bash src_sh/copy2dir.sh data/input/AdelaideRiver/dailyweather.prn data/results/`

jirikuncar:
Moves location of dataset metadata file from `data/<name>/metadata.yml`  to `.renku/datasets/<name>/metadata.yml`

(closes #396)

## TODO

- [x] `renku doctor`
- [x] `renku migrate datasets`

### DIFF
--- a/renku/cli/doctor.py
+++ b/renku/cli/doctor.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2019 - Swiss Data Science Center (SDSC)
+# A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+# Eidgenössische Technische Hochschule Zürich (ETHZ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


rokroskar:
We should actually move away from the dataset functionality altogether - not sure if that should be done first?

renku/cli/doctor.py
```diff
@@ -0,0 +1,50 @@
...
+@click.command()
+@pass_local_client
+@click.pass_context
+def doctor(ctx, client):
+    """Check your system and repository for potential problems."""
+    click.secho(
+        '\n'.join(
+            textwrap.wrap(
+                'Please note that the diagnosis report is used to help Renku '
+                'maitainers with debugging if you file an issue. Use all proposed '
```
rokroskar:
```suggestion
                'maintainers with debugging if you file an issue. Use all proposed '
```

jirikuncar:
> We should actually move away from the dataset functionality altogether ...

It is a bit bigger project and it can still use the same metadata files.

Issue #417, opened 2019-02-18
renku dataset list


nperraud:
**Describe the solution you'd like**
The command `renku dataset list` would be a nice addition to the renku CLI.

jirikuncar:
```console
$ renku dataset
foo
bar
baz
```

PR #419, opened 2019-02-18
feat(datasets): added listing of datasets

jsam:
# Description

This PR adds a feature for ```renku datasets``` command to list all created datasets within a repo.

Implements #417 

Example output which should be produced:

```
SHORT_ID    NAME         AUTHORS_CSV    CREATED
----------  -----------  -------------  -------------------
87c46dd9    somedataset  sam            2019-02-18 13:47:10
5aa65a26    dataset1234  sam            2019-02-18 13:47:39
```

### DIFF
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -81,18 +81,25 @@ will yield:
 import click
 from click import BadParameter
 
+from renku.models._tabulate import tabulate
 from renku.models.datasets import Author
 
 from ._client import pass_local_client
 from ._echo import progressbar
 
 
-@click.group()
+@click.group(invoke_without_command=True)
 @click.option('--datadir', default='data', type=click.Path(dir_okay=True))
+@pass_local_client(clean=True, commit=True)
 @click.pass_context
-def dataset(ctx, datadir):
+def dataset(ctx, client, datadir):
     """Handle datasets."""
     ctx.meta['renku.datasets.datadir'] = datadir
+    table = tabulate(
+        client.datasets.values(),
+        headers=["short_id", "name", "authors_csv", "created"],
+    )
+    click.echo(table)
 
 
 @dataset.command()
--- a/renku/models/datasets.py
+++ b/renku/models/datasets.py
@@ -185,3 +185,13 @@ class Dataset(object):
     def _now(self):
         """Define default value for datetime fields."""
         return datetime.datetime.utcnow()
+
+    @property
+    def short_id(self):
+        """Shorter version of identifier."""
+        return str(self.identifier)[:8]
+
+    @property
+    def authors_csv(self):
+        """Comma-separated list of authors associated with dataset."""
+        return ",".join(author.name for author in self.authors)
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -157,34 +157,34 @@
         },
         "lxml": {
             "hashes": [
-                "sha256:0dd6589fa75d369ba06d2b5f38dae107f76ea127f212f6a7bee134f6df2d1d21",
-                "sha256:1afbac344aa68c29e81ab56c1a9411c3663157b5aee5065b7fa030b398d4f7e0",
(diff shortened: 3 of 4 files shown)


renku/models/datasets.py
```diff
@@ -185,3 +185,13 @@ class Dataset(object):
...
+
+    @property
+    def short_id(self):
+        """Shorter version of identifier."""
+        return str(self.identifier)[:8]
+
+    @property
+    def authors_csv(self):
+        """Comma-separated list of authors associated with dataset."""
+        return ",".join(map(lambda a: a.name, self.authors))
```
jirikuncar:
```suggestion
        return ",".join(author.name for author in self.authors)
```

renku/cli/dataset.py
```diff
@@ -81,18 +81,25 @@
...
 @click.option('--datadir', default='data', type=click.Path(dir_okay=True))
 @click.pass_context
-def dataset(ctx, datadir):
+@pass_local_client(clean=True, commit=True)
+def dataset(client, ctx, datadir):
     """Handle datasets."""
     ctx.meta['renku.datasets.datadir'] = datadir
+    table = tabulate(
+        client.datasets.values(),
+        headers=["short_id", "name", "authors_csv", "created"]
```
```suggestion
        headers=["short_id", "name", "authors_csv", "created"],
```

renku/cli/dataset.py
```diff
@@ -81,18 +81,25 @@
...
 from ._echo import progressbar
 
 
-@click.group()
+@click.group(invoke_without_command=True)
 @click.option('--datadir', default='data', type=click.Path(dir_okay=True))
 @click.pass_context
-def dataset(ctx, datadir):
+@pass_local_client(clean=True, commit=True)
+def dataset(client, ctx, datadir):
```
jirikuncar: please keep `ctx` argument before `client`

renku/cli/dataset.py
```diff
@@ -81,18 +81,25 @@
...
 @click.option('--datadir', default='data', type=click.Path(dir_okay=True))
+@pass_local_client(clean=True, commit=True)
 @click.pass_context
-def dataset(ctx, datadir):
+def dataset(ctx, client, datadir):
     """Handle datasets."""
     ctx.meta['renku.datasets.datadir'] = datadir
+    table = tabulate(
+        client.datasets.values(),
+        headers=["short_id", "name", "authors_csv", "created"],
```
rokroskar: I'm coming here a bit late, but couldn't `authors_csv` be just `authors`?

jsam: Yes, it can, but then the output is kinda ugly.

Example:
```SHORT_ID    NAME         AUTHORS                                                             CREATED
----------  -----------  ------------------------------------------------------------------  -------------------
87c46dd9    somedataset  [Author(name='sam', email='contact@justsam.io', affiliation=None)]  2019-02-18 13:47:10
5aa65a26    dataset1234  [Author(name='sam', email='contact@justsam.io', affiliation=None)]  2019-02-18 13:47:39
```

That's why I went for authors_csv property where I merge the names of all authors into CSV format.

rokroskar: I just meant the name of the column. The `csv` part isn't needed imo

jsam: Yea, I understand. I guess the same argument could be also applied for `short_id` header. Should we  change our internal wrapper of tabulate to achieve this renaming or we have already this functionality somewhere?

jirikuncar: We could define headers as mapping and use keys for getting the attributes and values as column names.

```python
def tabulate(..., headers, ...):
    ...
    if isinstance(headers, dict):
        attrs, names = headers.keys(), headers.values()
    else:
        attrs = names = headers
    table = [(
        format_cell(cell, datetime_fmt=datetime_fmt)
        for cell in attrgetter(*attrs)(c)
    ) for c in collection]
    return tblte(table, headers=[h.upper() for h in names], **kwargs)
```

PR #420, opened 2019-02-19
feat(datasets): use uuid for metadata path

### LINKED ISSUE #417: renku dataset list

jirikuncar:
Changes dataset metadata path to use UUID instead of its name.

--
closes #417

### DIFF
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -90,16 +90,18 @@ from ._echo import progressbar
 
 @click.group(invoke_without_command=True)
 @click.option('--datadir', default='data', type=click.Path(dir_okay=True))
-@pass_local_client(clean=True, commit=True)
+@pass_local_client(clean=False, commit=False)
 @click.pass_context
 def dataset(ctx, client, datadir):
     """Handle datasets."""
     ctx.meta['renku.datasets.datadir'] = datadir
-    table = tabulate(
-        client.datasets.values(),
-        headers=["short_id", "name", "authors_csv", "created"],
-    )
-    click.echo(table)
+
+    if ctx.invoked_subcommand is None:
+        table = tabulate(
+            client.datasets.values(),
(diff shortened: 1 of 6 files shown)


renku/cli/dataset.py
```diff
@@ -90,16 +90,18 @@
 
 @click.group(invoke_without_command=True)
 @click.option('--datadir', default='data', type=click.Path(dir_okay=True))
-@pass_local_client(clean=True, commit=True)
+@pass_local_client(clean=False, commit=False)
```
jsam: Does it maybe make more sense to put this as a default value to the pass_local_client definition or would that break to much stuff? Is there a reason why bool should have 3 possible states to handle in this case?

tests/test_cli.py
```diff
@@ -312,7 +312,8 @@ def test_datasets_import(data_file, data_repository, runner, project, client):
     # create a dataset
     result = runner.invoke(cli.cli, ['dataset', 'create', 'dataset'])
     assert result.exit_code == 0
-    assert (client.renku_datasets_path / 'dataset' / client.METADATA).exists()
+    with client.with_dataset('dataset') as dataset:
```
jsam: Are we running pytest in parallel on some environments?

renku/cli/dataset.py
jirikuncar: Eventually, we should unify arguments and defaults on `pass_local_client` and `client.transaction`.

tests/test_cli.py
jirikuncar: You can with `pytest -n=auto`.

jsam: Thanks, I know. :+1: 
I'm just asking generally for environments since I'm too lazy to check .travis.yml. 

Nevermind, I answered myself in meantime. :+1: :smile:

PR #422, opened 2019-02-19
fix(cli): lazy lfs pull before run, rerun and update

### LINKED ISSUE #332: Lazy LFS pull
commands like `run` `update` `rerun` should check if the known input files have been pulled from LFS -- if not, do it automatically. 

Simply pull the files automatically -- we may want to warn the user if a particularly large amount of data is being fetched (e.g. > 1GB)

Fetching data for inputs at clone time

**Additional context**
This is critical for allowing users to work with lots of data.

jirikuncar:
Simply pull all inputs from LFS.

closes #332

### DIFF
--- a/renku/api/storage.py
+++ b/renku/api/storage.py
@@ -42,7 +42,7 @@ class StorageApiMixin(RepositoryApiMixin):
 
     _CMD_STORAGE_CHECKOUT = ['git', 'lfs', 'checkout']
 
-    _CMD_STORAGE_PULL = ['git', 'lfs', 'pull', 'origin', '-I']
+    _CMD_STORAGE_PULL = ['git', 'lfs', 'pull', '-I']
 
     def init_external_storage(self, force=False):
         """Initialize the external storage for data."""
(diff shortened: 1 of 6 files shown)


renku/api/storage.py
```diff
@@ -88,14 +88,18 @@ def track_paths_in_storage(self, *paths):
...
+                self.repo.commit(), path
+            )
+            run(
+                self._CMD_STORAGE_PULL + [str(path)],
+                cwd=str(client.path.absolute()),
+                stdout=PIPE,
+                stderr=STDOUT,
+            )
+        elif self.use_external_storage:
+            raise errors.ExternalStorageNotInstalled(self.repo)
```
jsam: We better readability we should consider documenting in docstring if a function raises something or not.

renku/api/storage.py
```diff
@@ -88,14 +88,18 @@ def track_paths_in_storage(self, *paths):
...
         """Pull a path from LFS."""
-        client, commit, path = self.resolve_in_submodules(
-            self.repo.commit(), path
-        )
-        run(
-            self._CMD_STORAGE_PULL + [str(path)],
-            cwd=str(client.path.absolute()),
-            check=True
-        )
+        if self.use_external_storage and self.external_storage_installed:
```
jsam: Instead of branching we could just 

```
        if self.use_external_storage and self.external_storage_installed is False:
            raise errors.ExternalStorageNotInstalled(self.repo)
        ....

```

Easier to read for the future.

jirikuncar: we can refactor it to a decorator since the same pattern is in more methods.

PR #423, opened 2019-02-19
test(fixture): add fixture for old repository with corresponding tests

jsam:
# Description

* Added fixture which will handle snapshots of old repositories in tests/fixtures (rename?)
* Simple tests which checks if new renku cli can parse the old one

Fixes #416 

(missing small cleanups)

PR #427, opened 2019-02-22
fix: removes typo causing wrong path resolution

jirikuncar:
💣


rokroskar:
this didn't trigger a failed test?

jirikuncar:
@rokroskar the tests will come after #423.

PR #438, opened 2019-03-01
perf: report exceptions to Sentry when configured

jirikuncar:
Requires `SENTRY_DSN` variable to be set and `sentry-sdk` package to be installed.

### DIFF
--- a/renku/cli/move.py
+++ b/renku/cli/move.py
@@ -15,7 +15,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Move a file in the repository."""
+"""Move or rename a file, a directory, or a symlink.
+
+Moving a files that belongs to a dataset will update its metadata. It also
+will attempt to update tracking information for files stored in an external
+storage (using Git LFS). Finally it makes sure that all relative symlinks work
+after the move.
+"""
 
 import os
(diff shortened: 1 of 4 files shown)


rokroskar:
add docs?

renku/cli/move.py
```diff
@@ -15,7 +15,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Move a file in the repository."""
+"""Move or rename a file, a directory, or a symlink.
+
+Moving a files that belongs to a dataset will update its metadata. It also
```
rokroskar:
```suggestion
Moving a file that belongs to a dataset will update its metadata. It also
```

renku/cli/move.py
```diff
@@ -15,7 +15,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Move a file in the repository."""
+"""Move or rename a file, a directory, or a symlink.
```
rokroskar: this isn't relevant for a PR on sentry

jirikuncar: It is at least in a separate commit.

rokroskar: better than nothing I suppose

PR #441, opened 2019-03-06
feat(Makefile): generate Makefile and compare to make

jirikuncar:
- Generates `Makefile` using `renku log --format=Makefile`
- Adds docs comparing `renku` to `make`


docs/comparison.rst
```diff
@@ -0,0 +1,76 @@
...
+
+.. code-block:: Makefile
+
+   foo:
+     @echo test > foo
+
+   foo.wc: foo
+     @wc -c < foo > foo.wc
+
+Renku also makes sure that if any of the :term:`inputs` is modified only
```
rokroskar:
```suggestion
Renku also makes sure that if any of the :term:`inputs` are modified only
```

docs/comparison.rst
```diff
@@ -0,0 +1,76 @@
...
+the necessary "rules" are invoked. Make also does not run the rule if all
```
rokroskar:
```suggestion
the necessary "rules" are invoked. In addition, `make` does not run the rule if all
```

docs/comparison.rst
```diff
@@ -0,0 +1,76 @@
...
+dependencies are older then targets.
```
rokroskar:
```suggestion
dependencies are older then the targets.
```

PR #451, opened 2019-03-08
refactor(cli): move dataset formatting to _format

jirikuncar:
just a cleanup


tests/cli/test_datasets.py
```diff
@@ -22,8 +22,11 @@
 import os
 
 import git
+import pytest
 
 from renku import cli
+from renku._compat import Path
+from renku.cli._format.datasets import FORMATS
```
```suggestion
from renku.cli._format.datasets import FORMATS as DATASETS_FORMATS
```

tests/cli/test_datasets.py
```diff
@@ -69,33 +72,24 @@ def test_datasets_import(data_file, data_repository, runner, project, client):
     assert result.exit_code == 0
 
 
-def test_datasets_list_empty(runner, project):
+@pytest.mark.parametrize('output_format', FORMATS.keys())
```
```suggestion
@pytest.mark.parametrize('output_format', DATASETS_FORMATS.keys())
```

tests/cli/test_datasets.py
```diff
@@ -69,33 +72,24 @@ def test_datasets_import(data_file, data_repository, runner, project, client):
...
 
-    output = result.output.split('\n')
-    assert output.pop(0).split() == ['ID', 'NAME', 'CREATED', 'AUTHORS']
-    assert set(output.pop(0)) == {'-', ' '}
-    assert output.pop(0) == ''
-    assert not output
 
-
-def test_datasets_list_non_empty(runner, project):
+@pytest.mark.parametrize('output_format', FORMATS.keys())
```
```suggestion
@pytest.mark.parametrize('output_format', DATASETS_FORMATS.keys())
```

tests/cli/test_datasets.py
```diff
@@ -145,7 +139,7 @@ def test_repository_file_to_dataset(runner, project, client):
 
     with client.with_dataset('dataset') as dataset:
         assert dataset.name == 'dataset'
-        assert '../../../a' in dataset.files
+        assert Path('../../../a') in dataset.files
```
jsam: Maybe we should think of adding a check that the file/link actually exists in its location on the filesystem, and not only inside metadata.

tests/cli/test_datasets.py
jirikuncar: not necessary at the moment

tests/cli/test_datasets.py
jirikuncar: no in the scope of this PR

jsam: And where is scope defined? 😁

jirikuncar: title

jsam: 😂 That's not definition. But 👌

PR #455, opened 2019-03-11
fix(datasets): link flag on dataset add

### LINKED ISSUE #448: Dataset add command with --copy/--no-copy works in reversed order
**Describe the bug**
I think --copy/--no-copy flags on add command are reversed.

**To Reproduce**
```
>> renku dataset add --no-copy my-dataset ../a 
>> ls -i data/my-dataset/
5904072 a
>> ls -i ../
4064143 a 
```

I would expect if I use --no-copy flag that no copy has been made. From the example above we can see that the copy has been made, even though I used --no-copy flag.

The opposite case `--copy` which I would expect to make a copy, creates a hard link with expected/same inodes.

Proposal 1: Fix the flag to be the opposite and add the test to validate this.
Proposal 2: Remove this flag altogether?

related #418

jsam:
This PR removes --copy/--no-copy flags on add command and introduces --link flag which will create a hardlink of the file into dataset.

Fixes #448

### DIFF
--- a/renku/api/datasets.py
+++ b/renku/api/datasets.py
@@ -157,7 +157,7 @@ class DatasetsApiMixin(object):
 
         dataset.files.update(files)
 
-    def _add_from_url(self, dataset, path, url, nocopy=False, **kwargs):
+    def _add_from_url(self, dataset, path, url, link=False, **kwargs):
         """Process an add from url and return the location on disk."""
         u = parse.urlparse(url)
 
@@ -190,7 +190,7 @@ class DatasetsApiMixin(object):
                             dataset,
                             dst,
                             f.absolute().as_posix(),
-                            nocopy=nocopy
+                            link=link,
                         )
                     )
                 return files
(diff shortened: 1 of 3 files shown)


renku/api/datasets.py
```diff
@@ -190,21 +190,21 @@ def _add_from_url(self, dataset, path, url, nocopy=False, **kwargs):
                             dataset,
                             dst,
                             f.absolute().as_posix(),
-                            nocopy=nocopy
+                            is_link=is_link
```
```suggestion
                            is_link=is_link,
```

renku/api/datasets.py
```diff
@@ -157,7 +157,7 @@ def add_data_to_dataset(
 
         dataset.files.update(files)
 
-    def _add_from_url(self, dataset, path, url, nocopy=False, **kwargs):
+    def _add_from_url(self, dataset, path, url, is_link=False, **kwargs):
```
jirikuncar: what is the purpose of `is_link`? maybe it should be called `link`.

jsam: is_ or at_ and similar prefixes/suffixes I'm using to denote certain types (like booleans or datetime). It's a matter of styled I used before, some habits die hard. I can rename if it bothers you.

PR #466, opened 2019-03-15
test: fix assert order to "constant cmp value"

jirikuncar:
Use `assert CONST OP VALUE` consistently.


jirikuncar:
It is not a general rule. It's mostly recommended for languages where `=` is used for assignment.

PR #469, opened 2019-03-15
feat(api): list datasets from a commit

jirikuncar:
Load datasets from a specific commit.

### DIFF
--- a/renku/api/datasets.py
+++ b/renku/api/datasets.py
@@ -51,6 +51,26 @@ class DatasetsApiMixin(object):
         """Return a ``Path`` instance of Renku dataset metadata folder."""
         return self.renku_path.joinpath(self.DATASETS)
 
+    def datasets_from_commit(self, commit=None):
+        """Return datasets defined in a commit."""
+        commit = commit or self.repo.head.commit
+
+        try:
+            datasets = commit.tree / self.renku_home / self.DATASETS
+        except KeyError:
+            return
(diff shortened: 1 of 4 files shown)


renku/api/datasets.py
```diff
@@ -51,6 +51,26 @@ def renku_datasets_path(self):
         """Return a ``Path`` instance of Renku dataset metadata folder."""
         return self.renku_path.joinpath(self.DATASETS)
 
+    def datasets_from_commit(self, commit=None):
```
jsam: Can we add a comment about which type is `commit`?

jirikuncar: Do you want to start including type references everywhere?

jsam: It helps me while reading the code and since I'm not super familiar with those types within git library, it would save me some time.

Not necessary everywhere.

jirikuncar: We can do a separate documentation sprint.

In this case, I am not a big fan of adding parameters docs just for a sake of having them if they are self explanatory:

```python
    """Return datasets defined in a commit.

    :param commit: ???
    :type commit: git.Commit
    """
```

I would rather see:

```python
    def datasets_from_commit(self, commit: Optional[git.Commit] = None) -> Generator[Dataset]:
```

jsam: :+1:

PR #475, opened 2019-03-18
refactor: filter dataset files by names

jirikuncar:
* Uses arguments to define datasets filter.


renku/cli/dataset.py
```diff
@@ -222,9 +216,7 @@ def _include_exclude(file_path, include=None, exclude=None):
     return True
 
 
-def _filter(
-    client, dataset_names=None, authors=None, include=None, exclude=None
-):
+def _filter(client, names=None, authors=None, include=None, exclude=None):
     """Filter dataset files by specified filters.
 
     :param dataset_names: Filter by specified dataset names.
```
jsam:
```suggestion
    :param names: Filter by specified dataset names.
```

PR #481, opened 2019-03-19
docs: group models and add missing commands

jirikuncar:
Improves docs 📖


docs/cli.rst
```diff
@@ -111,6 +111,22 @@ Renku Command Line
...
+
+.. _cli-migrate:
+
+``renku migrate``
+-----------------
+
+.. automodule:: renku.cli.migrate
+
+.. _cli-githooks:
+
```
jsam: Should we also add ls-files?

jirikuncar: It should be part of dataset documentation (`renku.cli.dataset.__doc__`)

PR #485, opened 2019-03-20
ci: include AppVeyor configuration

jirikuncar:
Run tests on Windows.


rokroskar:
@jirikuncar is this still being worked on?

jirikuncar:
@rokroskar the AppVeyron is working - however renku tests do not run on Windows.

rokroskar:
This PR then seems like it might be open indefinitely - or is there a clear way forward?

jirikuncar:
@rokroskar it mostly about paths - `\` vs `/`

rokroskar:
Is it that the tests fail on windows or that renku doesn't run on windows? Do we need to support windows atm?

jsam:
Progress bars with multiprocessing act differently on Windows shell as well so it would be nice to have this (if we have any windows users and promised support for windows platform already). If there is no promised Windows compatibility, we should close this until we do.

rokroskar:
I don't recall us having any requests for windows compatibility - am I wrong about this @jirikuncar?

PR #493, opened 2019-03-25
feat(cli): add dataset rm command

### LINKED ISSUE #418: Feature request: renku dataset delete
Removing manually a dataset now is pain. Woud be great to have an helping command for that.

`renku dataset delete d` should delete the dataset `d` and rebase if `d` has not beed used

jsam:
Adds dataset rm command.

closes #418

### DIFF
--- a/renku/api/datasets.py
+++ b/renku/api/datasets.py
@@ -78,20 +78,23 @@ class DatasetsApiMixin(object):
             result[path] = Dataset.from_yaml(path)
         return result
 
+    def dataset_path(self, name):
+        """Get dataset path from name."""
+        from renku.models.refs import LinkReference
+        path = self.renku_datasets_path / name / self.METADATA
+        if not path.exists():
+            path = LinkReference(
+                client=self, name='datasets/' + name
+            ).reference
+        return path
+
     def load_dataset(self, name=None):
         """Load dataset reference file."""
-        from renku.models.refs import LinkReference
         path = None
         dataset = None
 
         if name:
-            path = self.renku_datasets_path / name / self.METADATA
-
-            if not path.exists():
-                path = LinkReference(
-                    client=self, name='datasets/' + name
-                ).reference
-
+            path = self.dataset_path(name)
             if path.exists():
                 dataset = Dataset.from_yaml(path)
 
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -243,6 +243,21 @@ def unlink(client, name, include, exclude, yes):
         click.secho('OK', fg='green')
 
 
+@dataset.command('rm')
+@click.argument('names', nargs=-1)
+@pass_local_client(clean=True, commit=True)
+def remove(client, names):
+    """Delete a dataset."""
+    datasets = {client.dataset_path(name) for name in names}
+    with progressbar(
+        datasets, item_show_func=lambda item: str(item) if item else ''
+    ) as bar:
+        for dataset in bar:
+            if dataset and dataset.exists():
+                dataset.unlink()
+        click.secho('OK', fg='green')
+
+
 def _include_exclude(file_path, include=None, exclude=None):
     """Check if file matches one of include filters and not in exclude filter.
 
--- a/tests/cli/test_datasets.py
+++ b/tests/cli/test_datasets.py
@@ -322,11 +322,11 @@ def test_datasets_ls_files_tabular_empty(runner, project):
     """Test listing of data within empty dataset."""
     # create a dataset
     result = runner.invoke(cli.cli, ['dataset', 'create', 'my-dataset'])
-    assert result.exit_code == 0
+    assert 0 == result.exit_code
 
     # list all files in dataset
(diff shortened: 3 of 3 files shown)


renku/api/datasets.py
```diff
@@ -105,6 +105,15 @@ def store_dataset(self, dataset):
         with dataset.__reference__.open('w') as f:
             yaml.dump(dataset.asjsonld(), f, default_flow_style=False)
 
+    def remove_dataset(self, dataset):
+        """Removes dataset."""
+        path = (
```
```suggestion
        path = dataset.__reference__
```

renku/cli/dataset.py
```diff
@@ -243,6 +243,18 @@ def unlink(client, name, include, exclude, yes):
...
+@dataset.command('rm')
+@click.argument('names', nargs=-1)
+@pass_local_client(clean=True, commit=True)
+def remove(client, names):
+    """Delete a dataset."""
+    for name in names:
+        dataset = client.load_dataset(name=name)
+        if dataset:
+            client.remove_dataset(dataset)
+            click.secho('OK', fg='green')
```
```suggestion
    click.secho('OK', fg='green')
```

jsam: I think we should not print anything if no dataset is found. On unlink, we are handling this by not outputting anything so we should handle it in the same way here or we do a proper message for such cases on both commands.

jirikuncar: Maybe we can start with removing a single dataset first and then refactor the name resolution to something like:

```
datasets = {
    datasets.__reference__: dataset
    for dataset in (client.load_dataset(name) for name in names)
    if dataset
}
with progressbar(datasets.values()) as bar:
    for dataset in bar:
        client.remove_dataset(dataset)
click.secho('OK', fg='green')

Comment (jirikuncar) [reply renku/cli/dataset.py]: @jsam 🙄 ⬆️ ❓

Comment (jsam) [reply renku/cli/dataset.py]: Ah, yes! This is another issue now. How do you see this solution of resolving by name? Since the paths are uuid based now we cannot simply query the filesystem and get that answer. Do we have some additional mapping somewhere with which this can be done?

Comment (jsam) [reply renku/cli/dataset.py]: Ah, I see we could do LinkReference. Ok!

File: renku/api/datasets.py [hunk 3]
```diff
@@ -105,6 +108,14 @@ def store_dataset(self, dataset):
         with dataset.__reference__.open('w') as f:
             yaml.dump(dataset.asjsonld(), f, default_flow_style=False)
 
+    def remove_dataset(self, path):
```
jirikuncar: how is it linked to a dataset? should it take `Dataset` instance?

renku/cli/dataset.py
```diff
@@ -243,6 +243,18 @@ def unlink(client, name, include, exclude, yes):
         click.secho('OK', fg='green')
 
 
+@dataset.command('rm')
+@click.argument('names', nargs=-1)
+@pass_local_client(clean=True, commit=True)
+def remove(client, names):
+    """Delete a dataset."""
+    datasets = [str(client.dataset_path(name)) for name in names]
```
```suggestion
    datasets = [client.dataset_path(name) for name in names]
```

renku/api/datasets.py
jsam: Why allocate additional object if you don't need it?

renku/cli/dataset.py
```diff
@@ -243,6 +243,18 @@ def unlink(client, name, include, exclude, yes):
...
+    with progressbar(datasets) as bar:
```
```suggestion
    with progressbar(datasets, item_show_func=lambda item: str(item) if item else '') as bar:
```

renku/cli/dataset.py
```diff
@@ -243,6 +243,18 @@ def unlink(client, name, include, exclude, yes):
...
+        for dataset in bar:
+            client.remove_dataset(dataset)
```
```suggestion
            if dataset and dataset.exists():
                dataset.unlink()
```

renku/cli/dataset.py
jsam: So allocating additional stack frame for every function call is better than doing it straight away?

jirikuncar: 1. function call <<< syscall for unlink
2. we can completely remove the progressbar

jsam: 1. Same sentinel is on remove_dataset. So that's really correct. It's the same thing.
2. Why are spinning in circle then?
3. As I said, due to sentinel it's the same thing in the end. :+1:

tests/cli/test_datasets.py
```diff
@@ -582,3 +582,36 @@ def test_dataset_unlink_file(tmpdir, runner, client):
         assert new_file.basename not in [
             file_.path.name for file_ in dataset.files.values()
         ]
+
+
+def test_dataset_rm(tmpdir, runner, project, client):
+    """Test removal of a dataset."""
+    # create a dataset
+    result = runner.invoke(cli.cli, ['dataset', 'create', 'my-dataset'])
+    assert result.exit_code == 0
```
jirikuncar: see #466

PR #496, opened 2019-03-26
fix(api): make methods lock free

jirikuncar:
* Moves locking to CLI from API.

### DIFF
--- a/renku/api/datasets.py
+++ b/renku/api/datasets.py
@@ -109,37 +109,37 @@ class DatasetsApiMixin(object):
     def with_dataset(self, name=None):
         """Yield an editable metadata object for a dataset."""
         from renku.models.refs import LinkReference
-        with self.lock:
-            dataset = self.load_dataset(name=name)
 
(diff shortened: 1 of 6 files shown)


renku/api/datasets.py
```diff
@@ -106,37 +106,37 @@ def store_dataset(self, dataset):
...
-                setattr(dataset, '__reference__', path)
+        if dataset is None:
+            dataset = Dataset(name=name)
+            setattr(dataset, '__source__', {})
 
-                if name:
-                    LinkReference.create(
-                        client=self, name='datasets/' + name
-                    ).set_reference(path)
+            path = (
```
jsam: I think we can update this with dataset_path()

jirikuncar: will do in another PR

jirikuncar: well we can't really use the `dataset_path()` here since we don't want to resolve the `LinkReference` in this case.

PR #497, opened 2019-03-26
fix(cli): add check of missing references

jirikuncar:
Additional checks for #493.

### DIFF
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -323,7 +323,23 @@ def unlink(client, name, include, exclude, yes):
 def remove(client, names):
     """Delete a dataset."""
     from renku.models.refs import LinkReference
-    datasets = {client.dataset_path(name) for name in names}
+    datasets = {name: client.dataset_path(name) for name in names}
+
+    if not datasets:
+        raise click.BadParameter(
+            'use dataset name or identifier', param_hint='names'
+        )
+
+    unknown = [
+        name
+        for name, path in datasets.items() if not path or not path.exists()
+    ]
+    if unknown:
+        raise click.BadParameter(
+            'unknown datasets ' + ', '.join(unknown), param_hint='names'
+        )
+
+    datasets = set(datasets.values())
 
     with progressbar(
         datasets,
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -322,14 +322,28 @@ def unlink(client, name, include, exclude, yes):
 @pass_local_client(clean=True, commit=True)
 def remove(client, names):
     """Delete a dataset."""
+    from renku.models.refs import LinkReference
     datasets = {client.dataset_path(name) for name in names}
+
     with progressbar(
-        datasets, item_show_func=lambda item: str(item) if item else ''
+        datasets,
+        label='Removing metadata files'.ljust(30),
+        item_show_func=lambda item: str(item) if item else '',
     ) as bar:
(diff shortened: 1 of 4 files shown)


rokroskar:
What is a missing reference? I don't understand the error message to the user.

jirikuncar:
references - aliases for workflows and datasets

rokroskar:
Right, but the message still doesn't make sense to me. If a reference is missing, why would I resolve this with `git rm`?

jirikuncar:
@rokroskar because we don't have any `renku` specific command to do it. These symlinks are created behind the scene and if one deleted the file then the symlink is broken.

rokroskar:
aha so the "reference" here is literally a symlink - got it. `renku doctor`?

jsam:
`renku doctor` does not check references as far as I checked. @jirikuncar can you confirm this?

We should definitely add this check on `renku doctor` since our commands are not transactional and if the certain command fails user if forced to resolve it with git. Executing `renku doctor` in this case should give user some message on what he should do.

renku/cli/dataset.py
```diff
@@ -248,14 +248,28 @@ def unlink(client, name, include, exclude, yes):
...
+    with progressbar(
+        list(LinkReference.iter_items(client, common_path='datasets')),
+        label='Removing aliases'.ljust(30),
+        item_show_func=lambda item: item.name if item else '',
+    ) as bar:
+        for ref in bar:
+            if ref.reference in datasets:
+                ref.delete()
+
+    click.secho('OK', fg='green')
```
jsam: Please don't output success if nothing happened. It's just confusing from the users perspective

jirikuncar: https://github.com/SwissDataScienceCenter/renku-python/pull/497/files#diff-5ee2736a38fb1ec55353bdd962e82066R258

jirikuncar:
> renku doctor does not check references as far as I checked. @jirikuncar can you confirm this?

it does - see https://travis-ci.org/SwissDataScienceCenter/renku-python/builds/512045470?utm_source=github_status&utm_medium=notification

renku/cli/dataset.py
jsam: Consider the case where you have multiple datasets. Example:

```renku dataset
ID        NAME           CREATED              AUTHORS
--------  -------------  -------------------  ---------
0ad1cb9a  some-dataset2  2019-03-19 16:39:46
9436e36c  my-dataset     2019-02-28 16:48:09  sam
d40e0ce6  some-dataset   2019-03-09 16:34:34  sam
c53b021e  somedataset    2019-03-06 14:43:03  sam
```

And now as a user I make a mistake where I want to delete multiple datasets:

```renku dataset rm some-dataset2 some-dataset22                                                                  2 ↵
OK
```

In this case, we are reporting to the user that operation was successful and that we achieved what he wanted. Which is not the case cause we failed to remove the second dataset since it does not exist. We should not report OK in this case.

Let's create a ticket for this case to handle or handle it here.

jirikuncar: I've added an additional check.

jsam:
I had some issues with broken references and `renku doctor` din't detected them. I'll try to reproduce it again and open and ticket.

jirikuncar:
> I had some issues with broken references and `renku doctor` din't detected them. I'll try to reproduce it again and open a ticket.

@jsam I think I have done something wrong during refactoring, but it should work again.

PR #513, opened 2019-04-11
feat(datasets): dataset export to zenodo 

jsam:


renku/api/secrets.py
```diff
@@ -0,0 +1,38 @@
...
+
+@attr.s
+class SecretsApiMixin:
+    """Client for handling secrets for external providers."""
+
+    SECRETS_DIR = 'providers'
+
+    @property
+    def renku_secrets_path(self):
+        return self.renku_path.joinpath(self.SECRETS_DIR)
```
rokroskar: do you want this in the repository?

tests/cli/test_datasets.py
```diff
@@ -622,3 +622,49 @@ def test_dataset_rm(tmpdir, runner, project, client):
 
     result = runner.invoke(cli.cli, ['doctor'], catch_exceptions=False)
     assert 0 == result.exit_code
+
+
+@pytest.mark.integration
+def test_dataset_import_real_doi(runner, project):
+    """Test dataset import for existing DOI."""
+    result = runner.invoke(
+        cli.cli, ['dataset', 'import', '10.5281/zenodo.597964'], input='y'
```
rokroskar: is it possible to mock these?

renku/api/secrets.py
jsam: No, planned to add them to gitignore.

tests/cli/test_datasets.py
jsam: Since their API can change I wouldn't mock. I marked them as integration and now on the runner, we need to split the unit against integration tests. We have a couple of integrations already which are not marked/separated.

renku/api/secrets.py
rokroskar: still, there's a reasonable risk that someone will check in their password

tests/cli/test_datasets.py
rokroskar: oh I see, I didn't realize that was an option :)

renku/api/secrets.py
jsam: I know. That's why it's encoded, so it's not super obvious.

If you have a better suggestion I'm all ears. :smile:

rokroskar: credentials are per provider, not per dataset or repo - so having them in a renku configuration outside the repo would maybe make more sense?

Issue #526, opened 2019-04-25
The same author yields a different node in dataset json-ld


rokroskar:
**Describe the bug**

Take this json-ld generated by renku after I created a dataset and added a file:

```
{
    "name": "test",
    "created": "2019-04-25T13:55:17.466072",
    "identifier": "ddaf1940bfaf41aab55a7d7f0d15f29c",
    "authors": [
      {
        "name": "Rok Roskar",
        "email": "roskarr@ethz.ch",
        "affiliation": null,
        "@type": "dcterms:creator"
      }
    ],
    "files": {
      "data/test/README.rst": {
        "path": "data/test/README.rst",
        "url": "https://raw.githubusercontent.com/SwissDataScienceCenter/renku/master/README.rst",
        "authors": [
          {
            "name": "Rok Roskar",
            "email": "roskarr@ethz.ch",
            "affiliation": null,
            "@type": "dcterms:creator"
          }
        ],
        "added": "2019-04-25T13:57:47.745516",
        "@type": "schema:DigitalDocument"
      }
    },
    "@context": {
      "dcterms": "http://purl.org/dc/terms/",
      "dctypes": "http://purl.org/dc/dcmitypes/",
      "foaf": "http://xmlns.com/foaf/0.1/",
      "prov": "http://www.w3.org/ns/prov#",
      "schema": "http://schema.org",
      "name": "dcterms:name",
      "created": "schema:dateCreated",
      "identifier": {
        "@id": "dctypes:Dataset",
        "@type": "@id"
      },
      "authors": {
        "@container": "@list",
        "@id": "dcterms:creator"
      },
      "email": "dcterms:email",
      "affiliation": "schema:affiliation",
      "files": {
        "@container": "@index",
        "@id": "schema:DigitalObject"
      },
      "url": "schema:url",
      "added": "schema:dateCreated"
    },
    "@type": "dctypes:Dataset"
  }
```

If we look at the triples that will be generated from this, we get e.g.

```

_:b1 | http://purl.org/dc/terms/email | roskarr@ethz.ch 
_:b1 | http://purl.org/dc/terms/name | Rok Roskar

_:b3 | http://purl.org/dc/terms/email | roskarr@ethz.ch
_:b3 | http://purl.org/dc/terms/name | Rok Roskar
```

This means that the same author is being mapped as two separate entitites. We should include the email in the `@id` field of the `Author` in order to avoid duplication.

rokroskar:
[...]

PR #527, opened 2019-04-25
[WIP] fix: use json-ld containers correctly in dataset

### LINKED ISSUE #525: Incorrect JSON-LD with container objects
**Describe the bug**
container JSON-LD objects need to have an `@id` field -- this is currently missing from e.g. out `Dataset` JSON-LD context. 

We should properly implement the context for `Authors` and `Files` containers found in `Datasets`.

rokroskar:
- simplify json-ld contexts for dataset and author
- set id of the dataset correctly
- use the author's email as an id for now to de-duplicate authors

closes #525
addresses #526

### DIFF
--- a/renku/models/datasets.py
+++ b/renku/models/datasets.py
@@ -42,11 +42,10 @@ _path_attr = partial(
 
 
 @jsonld.s(
-    type='dcterms:creator',
+    type='schema:person',
     context={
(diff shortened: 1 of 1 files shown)


jsam:
Maybe we should add a few tests which would assert expected json-ld output?

rokroskar:
superseded by  #534

PR #533, opened 2019-05-07
feat(config) added global config manager

### LINKED ISSUE #532: Configuration manager
Implement configuration management for system-wide renku installation.

Configuration management should handle:
```
~/.renku
```

Where `~` is a user directory path (e.g. `C:\Users\<username>\.renku` or `/home/<username>/.renku`)

jsam:
closes #532


renku/cli/_config.py
```diff
@@ -158,3 +159,48 @@ def find_project_config_path(path=None):
...
+    def store_config(self, config):
+        """Persists global configuration object."""
+        with open(self.config_path, 'w') as file:
+            config.write(file)
+
+    def load_secret(self, provider):
+        """Load a secret for a external provider."""
+        try:
+            config = self.load_config()
+            return config[provider]['secret']
```
ableuler: maybe `return config[provider].get('secret')` instead of try/except?

jsam: `config` is not a dict but a config section object. But still, point taken. I've looked it up how to do it there so it should be updated, also added a test for it while at it. Thanks.

rokroskar:
I don't see any CLI for this?

rokroskar:
Note that the current `config` CLI was implemented to allow the user to specify a custom image registry that is used when pulling an image locally. This isn't used very much atm, but we should decide if it's worth keeping. In general, we will be faced with the problem of some configs belonging in the repo (i.e. used by everyone who might be sharing a project) and other configs being strictly for a single user (i.e. secrets). How should we deal with this?

renku/api/client.py
```diff
@@ -26,10 +26,7 @@
 
 @attr.s
 class LocalClient(
-    PathMixin,
-    StorageApiMixin,
-    RepositoryApiMixin,
-    DatasetsApiMixin,
+    PathMixin, StorageApiMixin, RepositoryApiMixin, DatasetsApiMixin
```
rokroskar: are you using yapf with the style file in this repo?

jsam: Yes.

rokroskar:
As discussed IRL: 

- [ ] implement `renku config --global` / keep the git config code
- [ ] remove dangling code from `/cli/_config.py`
- [ ] remove `/cli/env.py` and `/cli/endpoint.py`
- [ ] implement `ConfigManager` as another mixin

renku/cli/_config.py
ableuler: 👍

rokroskar:
Is this ready for re-review @jsam?

renku/api/config.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def __exit__(self, type, value, traceback):
+        """Release lock file."""
+        self._lock.close()
+
+    @property
+    def config_path(self):
+        """Renku config path."""
+        config = Path(self.config_dir)
+        if config.exists() is False:
```
rokroskar:
```suggestion
        if not config.exists():
```

renku/api/config.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def get_config(client, write_op, is_global):
+    """Get configuration object."""
+    if is_global:
+        return client
+
+    if write_op:
+        return client.repo.config_writer()
+    else:
```
rokroskar: don't need this `else`

tests/test_cli.py
```diff
@@ -1093,3 +1091,44 @@ def test_input_directory(runner, project, run):
         str(p.relative_to(cwd))
         for p in inputs.rglob('*') if p.name != '.gitkeep'
     ) == set(result.output.strip().split('\n'))
+
+
+def test_config_manager_creation(client):
+    """Check creation of configuration file."""
+    path_ = client.config_path
+    assert str(path_).endswith('renku.ini')
```
rokroskar: this doesn't test that it was created but that the path contains `renku.ini` - do you want to assert that the file exists?

renku/api/config.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def load_config(self):
+        """Loads global configuration object."""
+        config = configparser.ConfigParser()
+        config.read(str(self.config_path))
+        return config
+
+    def store_config(self, config):
+        """Persists global configuration object."""
+        with open(str(self.config_path), 'w') as file:
```
rokroskar: should we be setting some restrictive access flags on the file and directory, e.g. 600?

jsam: So currently they are 644. Why is 644 a problem?

rokroskar: multi-tenant environments, e.g. HPC clusters

PR #534, opened 2019-05-09
fix: modify json-ld for datasets

### LINKED ISSUE #525: Incorrect JSON-LD with container objects

jsam:
- simplify json-ld contexts for dataset and author
- set id of the dataset correctly
- use the author's email as an id for now to de-duplicate authors

closes #525
addresses #526


renku/models/_jsonld.py
```diff
@@ -218,6 +220,9 @@ def asjsonld(
     basedir=None,
 ):
     """Dump a JSON-LD class to the JSON with generated ``@context`` field."""
+    # if not getattr(inst.__class__, '_jsonld_fields', None):
```
rokroskar: remove?

PR #538, opened 2019-05-21
WIP: make --isolation option suited for parallel workflows

ableuler:
- Prevent renku from merging back _one_ branch automatically after running a command in isolation.
- Add a merge strategy for .gitattributes to .gitattributes which which allows automatic merge / rebase of a set branches (checked out in worktrees) which were used to execute workflow steps in parallel. The following two commands can then be used to rebase back. 
```
`git branch | grep isolation | xargs -I{branch} git rebase {branch}`
`git branch | grep isolation | xargs -I{branch} git branch -D {branch}`
```


jsam:
@ableuler Please, open a ticket with the followups for this topic. It seems really old and potentially very stale. Let's reopen a PR once, we have a clear way of moving forward with it.

PR #556, opened 2019-06-21
fix(datasets): strip query string from data filenames

### LINKED ISSUE #544: cli: dataset add should ignore query parameters
**Describe the bug**
The filename created with `renku dataset add` includes query parameters

**To Reproduce**
```
renku dataset create flights
renku dataset add flights https://www.dropbox.com/s/99w7evit5y7jxb3/2019-01-flights.csv.zip?dl=0
```

Look at the name of the resulting file.

**Expected behavior**

The query parameters should be stripped. In the above example, the result should be just `2019-01-flights.csv.zip`.

It might be necessary to ask the user how she wants the URL converted into a filename.

ableuler:
When adding data to a dataset from a URL which includes a query string, this query string used to be present in the filename. This is now fixed.

Closes #544

### DIFF
--- a/tests/cli/test_datasets.py
+++ b/tests/cli/test_datasets.py
@@ -75,6 +75,15 @@ def test_datasets_import(data_file, data_repository, runner, project, client):
     )
     assert 0 == result.exit_code
 
+    # add data from any URL (not a git repo)
+    result = runner.invoke(
+        cli.cli,
+        ['dataset', 'add', 'dataset', 'http://example.com/file.ext?foo=bar'],
+        catch_exceptions=False,
+    )
+    assert 0 == result.exit_code
+    assert os.stat('data/dataset/file.ext')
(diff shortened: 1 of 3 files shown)


ableuler:
> Thanks for the quick fix - could you please add a test?

done!

tests/cli/test_datasets.py
```diff
@@ -75,6 +75,15 @@ def test_datasets_import(data_file, data_repository, runner, project, client):
     )
     assert 0 == result.exit_code
 
+    # add data from any URL (not a git repo)
+    result = runner.invoke(
+        cli.cli,
+        ['dataset', 'add', 'dataset', 'http://example.com/file.ext?foo=bar'],
```
jsam: Instead of just mocking, could we also add a real integration test for this?

PR #557, opened 2019-06-21
docs: using ptvsd

rokroskar:
Documentation update for enabling external debugging.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -126,6 +126,7 @@ The containerized version of the CLI can be launched using Docker command.
 It makes sure your current directory is mounted to the same place in the
 container.
 
+
 Usage
 =====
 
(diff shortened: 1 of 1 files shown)


Pipfile
```diff
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-renku = {editable = true, path = "."}
+renku = {editable = true,extras = ["all"],path = "."}
```
jsam: Can we upgrade dependencies on the release cycle?

README.rst
```diff
@@ -126,6 +126,33 @@ The containerized version of the CLI can be launched using Docker command.
...
+        --editable \
+        --spec <path-to-renku-python>[all] \
+        renku
+
+This will install all the extras for testing and debugging.
+
+Debugging
+---------
+
+If you use Visual Studio Code, it is very convenient to debug ``renku`` using
```
jsam: I would not add editor specific debugging information to project level. It's personal what you use and we should be neutral about that topic on the project level.

renku/__init__.py
```diff
@@ -19,10 +19,23 @@
...
 from .api import LocalClient
 from .version import __version__
 
 __all__ = (
     'LocalClient',
     '__version__',
 )
+
+# Wait for the VS Code debugger to attach if requested
+VSCODE_DEBUG = os.environ.get('VSCODE_DEBUG') == '1'
```
jsam: Editor specific stuff should be kept in the editor. You can easily configure your editor to do this automatically, without putting it in the codebase.

rokroskar:
In general I agree with you, but here the debugger needs to either start a server for vscode to connect to or you need to specify the python script for the debugger to execute.

jsam:
Can we extract this server starting into its own script which you could put into your lunch.json to automatically control the server? That way we don't affect the runtime.

PR #558, opened 2019-06-21
Add datasets to the KG

### LINKED ISSUE #443: models: integrate dataset metadata to KG
- Export dataset metadata to KG.

rokroskar:
This PR adds `Dataset`s to the graph generated by `renku log`

This is the slightly collapsed graph with this PR:

noname gv

There are a few outstanding issues with the json-ld contexts, but those can be addressed in a separate PR. 

----

closes #443


Pipfile.lock
```diff
@@ -1,7 +1,7 @@
 {
```
jsam: Let's update deps on release.

renku/api/datasets.py
```diff
@@ -65,18 +64,16 @@ def datasets_from_commit(self, commit=None):
...
+            dataset.commit = commit
+            yield dataset
 
     @property
     def datasets(self):
         """Return mapping from path to dataset."""
         result = {}
         for path in self.renku_datasets_path.rglob(self.METADATA):
-            result[path] = Dataset.from_yaml(path)
+            result[path] = Dataset.from_yaml(path, self)
```
jsam:
```suggestion
            result[path] = Dataset.from_yaml(path, client=self)
```

renku/api/datasets.py
```diff
@@ -98,7 +95,7 @@ def load_dataset(self, name=None):
         if name:
             path = self.dataset_path(name)
             if path.exists():
-                dataset = Dataset.from_yaml(path)
+                dataset = Dataset.from_yaml(path, self)
```
jsam:
```suggestion
                dataset = Dataset.from_yaml(path, client=self)
```

renku/models/_jsonld.py
```diff
@@ -318,7 +323,7 @@ def from_jsonld(cls, data, __reference__=None, __source__=None):
             ) != type_:
                 new_cls = cls.__type_registry__[type_]
                 if cls != new_cls:
-                    return new_cls.from_jsonld(data)
+                    return new_cls.from_jsonld(data, client)
```
jsam:
```suggestion
                    return new_cls.from_jsonld(data, client=client)
```

renku/models/datasets.py
```diff
@@ -50,6 +51,7 @@
 )
 class Creator(object):
     """Represent the creator of a resource."""
+    client = attr.ib(default=None, kw_only=True)
```
jsam:
```suggestion

    client = attr.ib(default=None, kw_only=True)
```

renku/models/datasets.py
```diff
@@ -32,6 +32,7 @@
 from renku._compat import Path
 from renku.api.repository import default_path
 from renku.utils.doi import is_doi
+from renku.models.provenance.entities import Entity
```
jsam: Is this sorted? :dancing_men:

renku/models/_jsonld.py
```diff
@@ -346,24 +351,25 @@ def from_jsonld(cls, data, __reference__=None, __source__=None):
 
         if __reference__:
             with with_reference(__reference__):
-                self = cls(**data_)
+                self = cls(client=client, **data_)
```
jsam: I think not all models need a client. Right? I think this will fail for everything which is not Dataset.

renku/models/datasets.py
rokroskar: clearly not :)

renku/api/datasets.py
```diff
@@ -150,32 +151,36 @@ def add_data_to_dataset(
                     dataset, dataset_path, url, target, **kwargs
                 )
             else:
-                files = {}
+                files = []
                 for t in target:
-                    files.update(
+                    files.extend(
```
jsam: Why not append? All return lists are of length 1. Maybe instead of allocating additional lists and then merging them, we could just append the object?

renku/api/repository.py
```diff
@@ -375,3 +375,13 @@ def init_repository(self, name=None, force=False):
             metadata.updated = datetime.datetime.utcnow()
 
         return str(path)
+
+    def is_untracked(self, path):
```
jsam: Let's make this a normal function and put it to _git.py. That way I can reuse it in my check.

tests/models/__init__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
```
jsam: Since there is no logic needed in this file, we can remove it since we are trying to respect PEP420.

tests/cli/test_gitignore.py
```diff
@@ -45,6 +45,6 @@ def test_dataset_add(tmpdir, runner, client):
     result = runner.invoke(
         cli.cli,
         ['dataset', 'add', 'testing', '--force', ignored_file.strpath],
-        catch_exceptions=False,
+        catch_exceptions=True,
```
jsam: This is a default value.

tests/cli/test_datasets.py
```diff
@@ -597,15 +595,17 @@ def test_dataset_unlink_file(tmpdir, runner, client):
...
+            for file_ in dataset.files
+        }
 
     result = runner.invoke(
         cli.cli, [
             'dataset', 'unlink', 'my-dataset', '--include', new_file.basename,
             '-y'
-        ]
+        ],
+        catch_exceptions=False
```
jsam: Why is this needed?

tests/cli/__init__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
```
jsam: PEP 420.

We can delete this?

renku/models/provenance/activities.py
```diff
@@ -102,8 +101,13 @@ class Activity(CommitMixin):
 
     @generated.default
     def default_generated(self):
-        """Calculate default values."""
+        """Entities generated by this Activity."""
```
jsam:
```suggestion
        """Entities generated by this ``Activity``."""
```

renku/api/datasets.py
rokroskar: if there are subdirectories it will not be length 1

renku/api/repository.py
rokroskar: 👍

tests/models/__init__.py
rokroskar: as discussed IRL: needed for VSCode test discovery to work

renku/api/repository.py
rokroskar: actually I realized I'm not using it anywhere anymore so I removed it.

tests/cli/test_datasets.py
rokroskar: removed.

tests/cli/test_gitignore.py
rokroskar: fixed

renku/models/provenance/activities.py
```diff
@@ -135,12 +136,18 @@ def default_generated(self):
             if (self.client.path / path).is_dir():
                 entity_cls = Collection
 
-            entity = entity_cls(
-                commit=commit,
-                client=client,
-                path=str(path),
-                parent=collection,
-            )
+            # TODO: use a factory method to generate the entity
```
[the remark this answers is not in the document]
rokroskar: done: https://github.com/SwissDataScienceCenter/renku-python/issues/594

tests/cli/test_datasets.py
```diff
@@ -188,6 +188,60 @@ def test_dataset_create_exception_refs(
     assert 'a' in result.output
 
 
+def test_datasets_import(data_file, data_repository, runner, project, client):
```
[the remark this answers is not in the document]
rokroskar: added a ticket: https://github.com/SwissDataScienceCenter/renku-python/issues/595

PR #564, opened 2019-06-26
fix(status): fix renku status output when not in root folder

### LINKED ISSUE #551: renku status does not report correct results outside the root of the repo
**Describe the bug**
If you cd to a subdirectory of a repo and run `renku status`, you do not get the correct results

**To Reproduce**
1. Take any renku project in which renku run has been executed with an input file and an output file. Modify the input file and commit it.  
2. Run `renku status` in the root directory of the project. You will see that that it detects that the output file is out of date because the input file has changed. 
3. `cd` to some directory other than the root and execute `renku status`. 
4. This will show that all outputs are up-to-date, which is incorrect.

**Expected behavior**
The results of `renku status` should be the same, regardless of where you execute it. If you run it in a subdirectory, only the the files in that subdirectory should be considered in the result list.

**Desktop (please complete the following information):**
 - OS: macOS
 - Version: 0.5.0

ableuler:
The renku status command gave a wrong result when the current directory was not the root directory of the repository. This is now fixed.
Closes #551.


renku/cli/_graph.py
```diff
@@ -233,7 +233,9 @@ def dependencies(self, revision='HEAD', paths=None):
         """Return dependencies from a revision or paths."""
         result = []
 
-        if not paths:
+        if paths:
+            paths = map(lambda path: self.normalize_path(path), paths)
```
jsam:
```suggestion
            paths = [self.normalize_path(p) for p in paths]
```

ableuler: Thanks for the suggestion @jsam - I agree that would be a bit nicer to read. I used `map` instead of a list comprehension because `paths` can be a generator and the list comprehension would force the paths to be evaluated at this point, which - as far as I know - is not the case when using `map`.

ableuler:
> Python 3.5 does not like this PR. 🐙
> 
> The added test seems to be flaky since it fails 1/10 times (on py35). Do you know maybe why?

No, but I'll try to find out...

renku/cli/_graph.py
rokroskar: You can use a generator expression

ableuler: > You can use a generator expression

ok, done

ableuler:
> > Python 3.5 does not like this PR. 🐙
> > The added test seems to be flaky since it fails 1/10 times (on py35). Do you know maybe why?
> 
> No, but I'll try to find out...

Test are passing reliably now, also for 3.5. 
I think it's ready to be merged (@jsam @rokroskar ).

rokroskar:
cc @ciyer - could you check that it fixes the problem in your project?

PR #571, opened 2019-06-28
fix(cli): allow renku run with many inputs

### LINKED ISSUE #552: renku run error - too many dependent files
Trying to pre-process an audio dataset with 30000 files and merge them into a single hdf5 makes the following error:

**Renku version:** 0.5.0
**OS:** Linux (#1 SMP Tue Nov 27 12:40:57 EST 2018)
**Python:** 3.6.8

### Traceback

```
  File "[...]/renku/cli/_exc.py", line 100, in main
    result = super().main(*args, **kwargs)
  File "[...]/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "[...]/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "[...]/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "[...]/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "[...]/renku/cli/_client.py", line 80, in new_func
    result = ctx.invoke(method, client, *args, **kwargs)
  File "[...]/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "[...]/renku/cli/run.py", line 187, in run
    for _, path in tool.iter_input_files(client.workflow_path)
  File "[...]/renku/api/storage.py", line 122, in pull_paths_from_storage
    stderr=STDOUT,
  File "[...]/subprocess.py", line 423, in run
    with Popen(*popenargs, **kwargs) as process:
  File "[...]/subprocess.py", line 729, in __init__
    restore_signals, start_new_session)
  File "[...]/subprocess.py", line 1364, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
OSError: [Errno 7] Argument list too long: 'git'
```

ableuler:
This PR in principle allows the execution of `renku run` commands with thousands of input arguments. 

Note: This commit resolves only the case where a renku run command implicitly relies on many arguments which are NOT already expanded on the command line. 
Closes #552.


renku/api/storage.py
```diff
@@ -31,6 +31,10 @@
 
 HAS_LFS = call(['git', 'lfs'], stdout=PIPE, stderr=STDOUT) == 0
 
+# Batch size for when renku is expanding a large list
+# of files into an argument string.
+ARGUMENT_BATCH_SIZE = 1000
```
jsam: Wouldn't the problem still be there if we pass 1000 arguments with each size being bigger than 150 bytes? Should we make default batch size smaller?

tests/test_cli.py
```diff
@@ -103,6 +103,20 @@ def test_run_simple(runner, project):
     assert '.renku/workflow/' in result.output
 
 
+def test_run_many_args(client, run):
+    """Test a renku run command which implicitly relies on many inputs."""
+
+    os.mkdir('files')
+    output = 'output.txt'
+    for i in range(5000):
```
jsam: I see a potential edge case here. Can we test with a number which is not dividable by 1000? 😄

ableuler: Sure - how about 5003 (Sophie Germain prime number) acceptable? 😃

jsam: Yes! Perfect!

renku/api/storage.py
ableuler: Changed it to 100.

Issue #562, opened 2019-06-25
renku init --no-external-storage should be more explicit


rokroskar:
**Describe the bug**
If git lfs is enabled in a global config, `renku init --no-external-storage` will not prevent files from being added to LFS. Currently it does not *install* LFS but it also does not prevent LFS from being used if it is enabled in a higher-level config. It should instead run `git lfs install --skip-smudge --local`

mohammad-sdsc:
I don't understand the proposed solution here: `git lfs install --skip-smudge --local`. This does not prevent *adding* of files to LFS and only avoid automatic downloading of LFS files when pulling or cloning.

rokroskar:
No, running `git lfs install --skip-smudge --local` results in this `.git/config`:

```
[filter "lfs"]
        required = true
        clean = git-lfs clean -- %f
        smudge = git-lfs smudge --skip -- %f
        process = git-lfs filter-process --skip
```

rokroskar:
Actually I think the `--no-local-storage` option is just broken for some reason - even if I don't have git-lfs installed anywhere and I initialize the repo with the `-S` option it still tries to put outputs in LFS.

mohammad-sdsc:
Exactly, I've tried with different LFS variation (i.e. --skip-smudge added to local, user, and system's .gitconfig) and it still add files to LFS.

mohammad-sdsc:
```
$ git lfs install --help

...
* --skip-smudge:
    Skips automatic downloading of objects on clone or pull. This requires a
    manual "git lfs pull" every time a new commit is checked out on your
    repository.
...
```

mohammad-sdsc:
The problem is that when at the project creation time we provide the `--no-external-storage`, it is not persisted in the renku project file (if such a thing exist). Therefore, when we call the next command, renku has no idea that it shouldn't use LFS.

jsam:
The way how you can know that repo is initialized with `--no-external-storage` flag is by looking at .git/config and searching for LFS section. If there is no such section there, you know that repo was initialized with `--no-external-storage`. 

There is `.renku/metadata.yml` file if you need to persist certain options, but I'm not sure if we should persist this one since it could lead to inconsistency with git.

rokroskar:
The thing is that renku-python should *already* be doing this check. 

[...]

PR #575, opened 2019-07-02
fix: only check local repo for lfs filter

mohammad-sdsc:

# Description

Checking for LFS usage in Renku project was looking into current repo git config file and all the higher level files (i.e. user and system). If LFS filter was enabled in any of these config file, Renku would try to use it even if it was told not to when initialising the project.

This fix only looks in repository's git config.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/562.

## Type of change

## Checklist

### DIFF
--- a/renku/api/storage.py
+++ b/renku/api/storage.py
@@ -65,99 +65,116 @@ class StorageApiMixin(RepositoryApiMixin):
     @property
     def external_storage_installed(self):
         """Check that Large File Storage is installed."""
-        return HAS_LFS and self.repo.config_reader(
-        ).has_section('filter "lfs"')
+        return HAS_LFS
 
     def track_paths_in_storage(self, *paths):
         """Track paths in the external storage."""
-        if self.use_external_storage and self.external_storage_installed:
-            track_paths = []
-            attrs = self.find_attr(*paths)
+        if not self._use_lfs():
+            return
 
-            for path in paths:
-                # Do not add files with filter=lfs in .gitattributes
-                if attrs.get(path, {}).get('filter') == 'lfs':
-                    continue
-
-                path = Path(path)
-                if path.is_dir():
-                    track_paths.append(str(path / '**'))
-                elif path.suffix != '.ipynb':
-                    # TODO create configurable filter and follow .gitattributes
-                    track_paths.append(str(path))
-
-            call(
-                self._CMD_STORAGE_TRACK + track_paths,
-                stdout=PIPE,
-                stderr=STDOUT,
-                cwd=str(self.path),
-            )
-        elif self.use_external_storage:
+        if not self.external_storage_installed:
             raise errors.ExternalStorageNotInstalled(self.repo)
+
+        track_paths = []
+        attrs = self.find_attr(*paths)
+
+        for path in paths:
+            # Do not add files with filter=lfs in .gitattributes
+            if attrs.get(path, {}).get('filter') == 'lfs':
+                continue
+
+            path = Path(path)
+            if path.is_dir():
+                track_paths.append(str(path / '**'))
+            elif path.suffix != '.ipynb':
+                # TODO create configurable filter and follow .gitattributes
+                track_paths.append(str(path))
+
+        call(
+            self._CMD_STORAGE_TRACK + track_paths,
+            stdout=PIPE,
+            stderr=STDOUT,
+            cwd=str(self.path),
+        )
 
     def untrack_paths_from_storage(self, *paths):
         """Untrack paths from the external storage."""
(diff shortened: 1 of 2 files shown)


renku/api/storage.py
```diff
@@ -61,12 +59,14 @@ def init_external_storage(self, force=False):
     @property
     def external_storage_installed(self):
         """Check that Large File Storage is installed."""
-        return HAS_LFS and self.repo.config_reader(
-        ).has_section('filter "lfs"')
+        return call(['git', 'lfs'], stdout=PIPE, stderr=STDOUT) == 0
```
jsam: Maybe a bit more lightweight option to do this would be better.

```
from shutil import which
return not which('git-lfs')
```

renku/api/storage.py
```diff
@@ -88,24 +88,30 @@ def track_paths_in_storage(self, *paths):
...
             )
-        elif self.use_external_storage:
+        else:
             raise errors.ExternalStorageNotInstalled(self.repo)
 
     def untrack_paths_from_storage(self, *paths):
         """Untrack paths from the external storage."""
-        if self.use_external_storage and self.external_storage_installed:
+        if not self._use_lfs():
+            return
```
jsam: Usage of this function (places like remove and move) is computing certain path indexes which are not needed in this case. We should somehow detect that it's not needed to compute those things and not invoke this function at all if there is no storage from which we need to track/untrack.

Then we don't need this check at all.

renku/api/storage.py
```diff
@@ -88,24 +88,30 @@ def track_paths_in_storage(self, *paths):
...
+
+        if self.external_storage_installed:
             call(
                 self._CMD_STORAGE_UNTRACK + list(paths),
                 stdout=PIPE,
                 stderr=STDOUT,
                 cwd=str(self.path),
             )
-        elif self.use_external_storage:
+        else:
```
jsam: Instead of adding else, we could check for the oppositte case.

```
if not self.external_storage_installed:
    raise errors.ExternalStorageNotInstalled(self.repo)
```

renku/api/storage.py
rokroskar: why remove `HAS_LFS`? That one is only called at import time of the module instead of on every invocation of the method.

mohammad-sdsc: I'll undo this change.

renku/api/storage.py
mohammad-sdsc: I rather keep the scope of this fix limited and would postpone that refactoring to a later time. Moreover, since this is a public function it's good to have this test in place just a safety-net.

renku/api/storage.py
```diff
@@ -61,88 +61,106 @@ def init_external_storage(self, force=False):
...
-
-            call(
-                self._CMD_STORAGE_TRACK + track_paths,
-                stdout=PIPE,
-                stderr=STDOUT,
-                cwd=str(self.path),
-            )
-        elif self.use_external_storage:
+        if not self._use_lfs():
+            return
```
jsam: Let's make use_lfs method as part of the public API, cause I think we will need it in #588 and here we should raise an exception and not silently fail.

renku/api/storage.py
```diff
@@ -61,88 +61,106 @@ def init_external_storage(self, force=False):
...
-            for client_path, paths in client_dict.items():
-                run(
-                    self._CMD_STORAGE_PULL + [shlex.quote(','.join(paths))],
-                    cwd=str(client_path.absolute()),
-                    stdout=PIPE,
-                    stderr=STDOUT,
-                )
-        elif self.use_external_storage:
+        if not self._use_lfs():
+            return
```
jsam: Raise exception.

renku/api/storage.py
```diff
@@ -61,88 +61,106 @@ def init_external_storage(self, force=False):
...
                 stdout=PIPE,
                 stderr=STDOUT,
-                check=True,
             )
-        elif self.use_external_storage:
+
+    def checkout_paths_from_storage(self, *paths):
+        """Checkout a paths from LFS."""
+        if not self._use_lfs():
+            return
```
jsam: Raise exception.

mohammad-sdsc:
The first change (raising an exception) will break functionality of the existing code until https://github.com/SwissDataScienceCenter/renku-python/issues/588 is implemented. Is this something that we want now?
To me, it makes more sense to do this as the first part of https://github.com/SwissDataScienceCenter/renku-python/issues/588.

jsam:
Yes, ofc. Let's rebase and merge.

mohammad-sdsc:
I've rebased and fixed the conflicts. Please approve.

PR #584, opened 2019-07-08
fix: escape whitespaces in notebook name

mohammad-sdsc:

# Description

When there are whitespaces in notebook's names, the dot build fails with a message like `Path "notebooks/Test" does not exist.`. This happens due to a call to `renku log` where its input lacks proper file name escaping. See https://renkulab.io/gitlab/clement.lefebvre/test_blank_space/-/jobs/6521 for more info.

Fixes # (issue)

## Type of change

## Checklist

- [X] I have **NOT** removed any existing tests.
- [X] I have run ``./run-tests.sh`` locally.


mohammad-sdsc:
When calling `git ls-files --no-empty-directory --recurse-submodules -z | xargs -0 renku log --format dot >graph.dot` on my local machine, I get a warning with the following message:

```
file:///blob/3fb50f43b05c200dbf8f9256550e987fb814089d/notebooks/01-Count Flights.ran.ipynb does not look like a valid URI, trying to serialize this will break.
file:///blob/1ebeebc566ebe22cfbe3a37790c728da5b380226/notebooks/01-Count Flights.ipynb does not look like a valid URI, trying to serialize this will break.
file:///blob/4dc85a5356bcd750bc9a1675c1224cfd397ab4e0/notebooks/00-Filter Flights.ran.ipynb does not look like a valid URI, trying to serialize this will break.
...
```

However, it does not seems to me that we are going to serialize the concerned URI. so we can ignore this.

PR #590, opened 2019-07-11
feat: add SoftwareAgent to Activity

### LINKED ISSUE #508: models: export Agent for Activities
Each activity (not only `renku *`) should have an agent.

rokroskar:
This PR links every activity to the version of `renku` used to generate the triples. This way we can check in the future which parts of the graph may need to be regenerated. 

---

closes #508

### DIFF
--- a/renku/models/provenance/activities.py
+++ b/renku/models/provenance/activities.py
@@ -29,6 +29,7 @@ from renku.models.cwl import WORKFLOW_STEP_RUN_TYPES
 from renku.models.cwl._ascwl import CWLClass
 from renku.models.cwl.types import PATH_OBJECTS
 
+from .agents import renku_agent
 from .entities import Collection, CommitMixin, Entity, Process, Workflow
 from .qualified import Association, Generation, Usage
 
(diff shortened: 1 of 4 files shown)


renku/models/provenance/activities.py
```diff
@@ -100,6 +102,16 @@ class Activity(CommitMixin):
         kw_only=True,
     )
 
+    agent = jsonld.ib(context='prov:agent', kw_only=True)
+
+    @agent.default
+    def default_agent(self):
+        """Set the default agent."""
+        return SoftwareAgent(
```
jsam: Allocating activity is already pretty expensive operation and this adds a bit more to the problem. I propose we create this instantitation as a singleton and then we associate the same singleton object to the newly created Activity, and not allocate new instance every time we create new Activity.

rokroskar: good point - I have no idea how to implement a singleton with `attr` though?

jachro:
I generated some triples locally and they seem to be looking good.
In terms of the agent resource, I think it would be nice to have it the way you proposed `https://github.com/SwissDataScienceCenter/renku-python/tree/<version-sha>` instead of `mailto:renku-<version>` but I don't know how much work would that be to change it.

rokroskar:
> I don't know how much work would that be to change it 

shouldn't be too much work - I'll see what I can do

Issue #588, opened 2019-07-10
Fix external storage handling within click commands


jsam:
The change introduced in #575 did not address changes needed within click commands which are using storage interface. In those commands, we are computing certain indexes which as of #575 is not needed.

We should refactor those commands in such a way that we are only computing what is needed.

PR #592, opened 2019-07-12
fix: ensure external storage is handled correctly

jsam:
* fixes accidental removal of --no-external-storage check on run command
* refactors storage API to be more secure and clean
* prevents unnecessary computations inside click commands
* fixes the tests back to the expected behavior

closes: #588

### DIFF
--- a/renku/api/storage.py
+++ b/renku/api/storage.py
@@ -16,12 +16,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Client for handling a data storage."""
-
+import functools
 import shlex
 from collections import defaultdict
+from shutil import which
 from subprocess import PIPE, STDOUT, call, run
 
 import attr
+from werkzeug.utils import cached_property
 
 from renku import errors
 from renku._compat import Path
@@ -29,11 +31,26 @@ from renku._compat import Path
 from ._git import _expand_directories
 from .repository import RepositoryApiMixin
 
-HAS_LFS = call(['git', 'lfs'], stdout=PIPE, stderr=STDOUT) == 0
-
 # Batch size for when renku is expanding a large list
 # of files into an argument string.
 ARGUMENT_BATCH_SIZE = 100
+
+
+def ensure_external_storage(fn):
+    """Ensure management of external storage on methods which depend on it.
+
+    :raises: ``errors.ExternalStorageNotInstalled``
+    :raises: ``errors.ExternalStorageDisabled``
+    """
+    # noqa
+    @functools.wraps(fn)
+    def wrapper(self, *args, **kwargs):
+        if not self.has_external_storage:
+            pass
+        else:
+            return fn(self, *args, **kwargs)
+
+    return wrapper
 
 
 @attr.s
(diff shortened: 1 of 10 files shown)


renku/errors.py
```diff
@@ -263,6 +264,25 @@ def __init__(self, repo):
         super(ExternalStorageNotInstalled, self).__init__(msg)
 
 
+class ExternalStorageDisabled(RenkuException, click.ClickException):
+    """Raise when disabled repository storage API is trying to be used."""
+
+    def __init__(self, repo):
+        """Build a custom message."""
+        msg = (
+            'Git-LFS is not configured, '
```
rokroskar:
```suggestion
            'External storage is not configured, '
```

renku/errors.py
```diff
@@ -263,6 +264,25 @@ def __init__(self, repo):
...
+            'but this action is trying to trigger storage action.\n'
```
rokroskar:
```suggestion
            'but this action is trying to use it.\n'
```

renku/errors.py
```diff
@@ -263,6 +264,25 @@ def __init__(self, repo):
...
+            'By running this command without LFS enabled '
```
rokroskar:
```suggestion
            'By running this command without storage enabled '
```

renku/errors.py
```diff
@@ -263,6 +264,25 @@ def __init__(self, repo):
...
+            'you could be committing\n'
+            'large files directly to the git repository.\n\n'
+            'If this is your intention, please repeat the command with '
+            'the -S flag (e.g. renku -S run <cmd>), \n'
+            'otherwise install LFS with "git lfs install --local".'
```
rokroskar:
```suggestion
            'otherwise install e.g. git-LFS with "git lfs install --local".'
```

renku/api/storage.py
```diff
@@ -53,6 +70,30 @@ class StorageApiMixin(RepositoryApiMixin):
...
+        """Check if repository has external storage enabled.
+
+        :raises: ``errors.ExternalStorageNotInstalled``
+        :raises: ``errors.ExternalStorageDisabled``
+        """
+        repo_config = self.repo.config_reader(config_level='repository')
+        lfs_enabled = repo_config.has_section('filter "lfs"')
+
+        storage_enabled = lfs_enabled and self.storage_installed
+        if self.use_external_storage and not storage_enabled:
```
mohammad-sdsc: With this the user has to specify `-S` flag with each command even when the project is initialised with `-S`. This looks a bit inconvenient to me.

renku/cli/move.py
```diff
@@ -98,20 +98,25 @@ def fmt_dst(path):
...
+            click.echo(WARNING + 'There are custom .gitattributes.\n')
+            if click.confirm(
+                'Do you want to edit ".gitattributes" now?', default=False
+            ):
+                click.edit(filename=str(client.path / '.gitattributes'))
+
+    if tracked and client.has_external_storage:
+        client.track_paths_in_storage(
+            *(destinations[path] for path in tracked)
+        )
```
mohammad-sdsc: This `if` block can go inside the previous block.

renku/api/storage.py
jsam: This is exactly the check we need, otherwise, user can by accident commit the large file and destroy the repo.

Issue #543, opened 2019-05-29
renku --ignore-dirty option


ciyer:
**Is your feature request related to a problem? Please describe.**
Sometimes I have a dirty working tree, and I want to run a renku command, which I know will not interact with the files that are dirty. 

**Describe the solution you'd like**
I would like a `--ignore-dirty` option that looks at what files are dirty and ignores them when running the command. Alternatively, it could move all the dirty files to some tmp location, execute the renku command, and then move them back into the working tree.

**Describe alternatives you've considered**
I can run `git stash` or `git clean`, or `git commit` before the command, but this is sometimes inconvenient.

erbou:
The renku command could also present a human-readable error if the repo is dirty: e.g. list the modified files and suggest the options, including the `--ignore-dirty`.

Note that there is a risk that an undeclared input (bad practice), or a dependency of the command is dirty. Also be prepared to deal with a `renku run` command that create/change files that could conflict with the `git stash pop` and do a merge that a non-expert may be left to to deal with.

rokroskar:
Another thing to consider is what will happen if you stash large data?

Panaetius:
Design how this works for renku run/rerun/update and what other commands have this issue currently

Panaetius:
First draft of design in https://sdsc.atlassian.net/wiki/spaces/RENKU/pages/810778631/Working+with+dirty+repositories+in+renku-python

PR #607, opened 2019-07-24
feat: enable all datasets command to operate on dirty repository

jsam:
This PR enables all dataset subcommands to operate on a dirty repository by default. It also cleanups tests and adds code needed to enable more commands to operate on the dirty repository.

addresses: #543


renku/api/_git.py
```diff
@@ -230,30 +234,54 @@ def commit(self, author_date=None, commit_only=None):
...
 
+        if commit_only == COMMIT_DIFF_STRATEGY:
+            staged = {item.a_path for item in self.repo.index.diff(None)}
+
+            modified = {item.a_path for item in self.repo.index.diff('HEAD')}
+
+            if staged or modified:
+                self.repo.git.reset()
+
+            # Exclude files created by pipes.
```
rokroskar: is this needed for dataset commands?

renku/api/datasets.py
```diff
@@ -452,3 +454,11 @@ def check_for_git_repo(url):
         except InvalidGitRepositoryError:
             is_git = False
     return is_git
+
+
+def get_dataset_metadata_paths():
```
rokroskar: why does this need to be a function? it could just be a list.

tests/cli/test_rerun.py
```diff
@@ -0,0 +1,239 @@
+# -*- coding: utf-8 -*-
```
rokroskar: why is this part of this PR?

tests/cli/test_run.py
```diff
@@ -0,0 +1,73 @@
+# -*- coding: utf-8 -*-
```
rokroskar: why is this part of this PR?

tests/cli/test_rerun.py
jsam: Cause we need to refactor it anyway and since I entered regression with this PR I decided to left it here. If it bothers you I can make a its own PR for this.

tests/cli/test_run.py
jsam: Same as test_rerun.py

renku/api/datasets.py
jsam: well, since same code is used on multiple places (it was more than 2 like it is now) I decided to extract it as a function  cause of DRY'ed approach

renku/api/_git.py
jsam: yes

renku/api/datasets.py
rokroskar: yes I understand... but it doesn't need to be a function. 

```
DATASET_METADATA_PATHS = [
        Path(RENKU_HOME) / Path(DatasetsApiMixin.DATASETS),
        Path(RENKU_HOME) / Path(LinkReference.REFS),
]
```

tests/cli/test_run.py
rokroskar: I don't follow - this is refactoring the run/rerun tests, so it seems to me that it should not be in this PR. What am I missing?

jsam: Looks like github or somebody deleted my comment. :(

Test refactoring is on the todo list for a while now, and since I worked on workflow components as well here and during that refactored all this tests when removing this functionality from workflow commands (run/rerun/update). I decided to keep the tests refactored and sneak that refactoring in. If it is a problem I can create a ticket for this and refactor it on its own PR. I've seen no reason for all this extra work so I kept it like this since both of them are mergeable straight away.

tests/cli/test_rerun.py
jsam: Looks like my comment got deleted. :(

PR #614, opened 2019-07-26
build: update and cleanup build

rokroskar:
This PR:

* removes the docker build (not needed anymore since the KG is hosted and dot files are not generated in CI)
* push dev tags to have consistency between pypi releases and repo tags
* updates changelog


jsam:
Thanks for looking into this!

PR #622, opened 2019-08-09
chore: add tests.yml to set up github action

rokroskar:
try out github actions on this repo


rokroskar:
It more or less works, looks like there are just some dependencies missing still on mac os.

rokroskar:
ok it passes now - but it's just the test, no build and deploy

PR #626, opened 2019-08-19
feat: Dataverse import

Panaetius:
Adds import support for arbitrary Dataverse datasets.

Works with the following URI formats:
- `https://<server>/dataset.xhtml?persistentId=doi:<doi>`
- `doi:<doi>`
- `<doi>`

with `<doi>` being a DOI like `10.18710/OYIH83`.

Fixes tqdm progressbars when there are more files than lines in the terminal. Now only shows as many progressbars as there are multiprocessing workers.

Fixes extraction of downloaded zip archives.

Adds error logging of download errors.

Fixes bug with downloads where small files would have their size rounded to 0.0 for the progress bar, causing a crash.


renku/cli/_providers/dataverse.py
```diff
@@ -0,0 +1,254 @@
...
+class DataverseFileSerializer:
+    """Dataverse record file."""
+
+    _id = attr.ib(default=None, kw_only=True)
+
+    identifier = attr.ib(default=None, kw_only=True)
+
+    name = attr.ib(default=None, kw_only=True)
+
+    fileFormat = attr.ib(default=None, kw_only=True)
```
jsam: I see it's convenient, but ideally we want to keep same notation through out the code base. Can we rename this to `file_format`?

renku/cli/_providers/__init__.py
```diff
@@ -31,20 +34,26 @@ class ProviderFactory:
...
+        if is_doi_ is None:
             url = urlparse(uri)
             if bool(url.scheme and url.netloc and url.params == '') is False:
                 return None, 'Cannot parse URL.'
 
+        uri_lower = uri.lower()
+
         provider = None
-        if 'zenodo' in uri:
+        if 'zenodo' in uri_lower:
```
jsam:
```suggestion
        if 'zenodo' in uri.lower():
```

Panaetius: Good catch! That was left over from an earlier version of the code where I reused `uri_lower` a couple of times. Thanks

renku/cli/_providers/dataverse.py
```diff
@@ -0,0 +1,254 @@
...
+
+DATAVERSE_API_PATH = 'api'
+
+DATAVERSE_VERSION_API = 'info/version'
+DATAVERSE_METADATA_API = 'datasets/export'
+DATAVERSE_FILE_API = 'access/datafile/:persistentId/'
+DATAVERSE_EXPORTER = 'schema.org'
+
+
+def check_dataverse_uri(url):
```
rokroskar: this should probably be refactored to be a generic classmethod of the `ProviderApi` to simplify the factory function logic - but maybe in a separate PR.

renku/cli/_providers/doi.py
```diff
@@ -95,6 +97,8 @@ def _query(self, doi):
 
     def find_record(self, uri):
         """Finds DOI record."""
+        if uri.startswith('doi:'):
+            uri = uri[4:]
```
rokroskar: this is a bit hacky, could probably be better `urlparse` from `urllib`?

tests/cli/test_datasets.py
```diff
@@ -831,3 +833,80 @@ def test_dataset_edit(runner, client, project):
         catch_exceptions=False
     )
     assert 0 == result.exit_code
+
+
+@pytest.mark.parametrize(
+    'uri', (
+        '10.5281/zenodo.3363060', 'doi:10.5281/zenodo.3363060',
+        'hhttps://zenodo.org/record/3363060'
```
jsam: `https`

tests/cli/test_datasets.py
```diff
@@ -831,3 +833,80 @@ def test_dataset_edit(runner, client, project):
...
+    )
+)
+def test_dataset_provider_resolution_zenodo(dataverse_responses, uri):
+    """Check that zenodo uris resolve to ZenodoProvider."""
+    provider, _ = ProviderFactory.from_uri(uri)
```
jsam: If tests generally depend on having internet connection mark them as integration `@pytest.mark.integration`

renku/cli/dataset.py
```diff
@@ -517,7 +518,7 @@ def export_(client, id, provider, publish):
 def import_(ctx, client, uri, name, extract):
     """Import data from a 3rd party provider.
 
-    Supported providers: [Zenodo, ]
+    Supported providers: [Zenodo, Dataverse]
     """
     provider, err = ProviderFactory.from_uri(uri)
```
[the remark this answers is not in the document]
Panaetius: I added some errorhandling and now if a provider fails, a warning gets shown if it works with another provider/an error otherwise.

conftest.py
```diff
@@ -263,3 +263,188 @@ def zenodo_sandbox(client):
...
+
+    with responses.RequestsMock(assert_all_requests_are_fired=False) as rsps:
+
+        def doi_callback(request):
+            return (
+                200, {
+                    'Content-Type': 'application/json'
+                },
+                json.dumps({
+                    'type': 'dataset',
```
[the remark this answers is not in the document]
Panaetius: Added several new integration tests and removed the big mocked tests

renku/cli/_providers/dataverse.py
```diff
@@ -0,0 +1,277 @@
...
+    """Check if an URL points to a dataverse instance."""
+    url_parts = list(urlparse.urlparse(url))
```
rokroskar: sorry, I didn't notice this before - why turn the parsed URL into a list? It should be much cleaner to just refer to the parts of the URL by name.

Panaetius: `ParseResult` properties can't be set directl:
```
>>> from urllib.parse import urlparse
>>> r = urlparse("http://www.google.com/search?q=wer")
>>> r
ParseResult(scheme='http', netloc='www.google.com', path='/search', params='', query='q=wer', fragment='')
>>> r.path = "bla"
  File "<stdin>", line 1, in <module>
AttributeError: can't set attribute
```

It can be side by using the private method `_replace` as discussed here: https://stackoverflow.com/questions/21628852/changing-hostname-in-a-url though I'm not sure if using a private method is any cleaner. It is a bit of a weird design decision by the urllib developers.

rokroskar: Ah I didn't realize - that's pretty weird indeed. Also not possible to turn it into a dictionary...

tests/cli/test_integration_datasets.py
```diff
@@ -24,49 +24,43 @@
...
     """Test dataset import for existing DOI."""
     result = runner.invoke(
-        cli.cli, ['dataset', 'import', '10.5281/zenodo.597964'], input='y'
+        cli.cli, ['dataset', 'import', doi[0]], input=doi[1]
     )
     assert 0 == result.exit_code
     assert 'OK' in result.output
 
     result = runner.invoke(cli.cli, ['dataset'])
+    print(result.output)
```
jsam: Cleanup?

tests/cli/test_integration_datasets.py
```diff
@@ -24,49 +24,43 @@
...
-        ('10.5281/zenodo.3239256', 'n'),
-        ('10.5281/zenodo.3237813', 'n'),
-        ('10.5281/zenodo.3239988', 'y'),
-        ('10.5281/zenodo.3239986', 'n'),
-        ('10.5281/zenodo.3239984', 'n'),
-        ('10.5281/zenodo.3239982', 'n'),
-        ('10.5281/zenodo.3239980', 'n'),
-        ('10.5281/zenodo.3188334', 'y'),
-    ]
+    'doi', [('10.5281/zenodo.597964', 'y'), ('10.5281/zenodo.3236928', 'n'),
```
jsam: Did yapf do this? I've wanted to keep it 1 per line on purpose (easier to insert new ones without touching old)

conftest.py
```diff
@@ -299,6 +299,88 @@ def zenodo_sandbox(client):
...
 
 
+@pytest.fixture
+def doi_responses():
+    """Responses for doi.org requests."""
+    from renku.cli._providers.doi import DOI_BASE_URL
+    from renku.cli._providers.dataverse import (
+        DATAVERSE_API_PATH, DATAVERSE_VERSION_API
+    )
+    import json
```
jsam: Stdlib imports can be done globally.

renku/cli/_providers/__init__.py
```diff
@@ -19,35 +19,56 @@
...
-            provider = ZenodoProvider(is_doi=is_doi_)
+        warning = ''
+
+        for _, potential_provider in ProviderFactory.PROVIDERS.items():
+            try:
+                if potential_provider.supports(uri):
+                    provider = potential_provider
+                    break
+            except Exception as e:
+                print(e)
```
jsam: Why print?

renku/cli/_providers/__init__.py
```diff
@@ -19,35 +19,56 @@
...
+                print(provider)
```
jsam: Cleanup?

renku/cli/dataset.py
```diff
@@ -618,45 +644,63 @@ def import_(ctx, client, uri, name, extract):
...
         patoolib.extract_archive(filepath, outdir=data_folder_)
         filepath.unlink()
 
     def stream_to_file(request):
         """Stream bytes to file."""
         with open(download_to, 'wb') as f_:
+            scaling_factor = 1e-6
+            unit = 'MB'
+
+            if file.filesize < 1e5:
```
jsam: Can we add a comment on what this math is referring to? Just as a reference for the future.

tests/cli/test_datasets.py
```diff
@@ -15,7 +15,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Test ``dataset`` command."""
+"""Test dataset command."""
```
```suggestion
"""Test ``dataset`` command."""
```

renku/cli/_providers/__init__.py
```diff
@@ -19,35 +19,56 @@
...
-        if 'zenodo' in uri:
-            provider = ZenodoProvider(is_doi=is_doi_)
+        warning = ''
+
+        for _, potential_provider in ProviderFactory.PROVIDERS.items():
+            try:
+                if potential_provider.supports(uri):
+                    provider = potential_provider
+                    break
+            except Exception as e:
```
jsam: I think we should generally avoid catching generic `Exceptions`, but if we must then let's catch both (Exception, BaseException)

tests/cli/test_integration_datasets.py
```diff
@@ -24,49 +24,43 @@
 from renku import cli
 
 
+@pytest.mark.parametrize(
+    'doi', [(
+        '10.5281/zenodo.597964', 'y', 'pyndl_naive_discriminat_v064',
```
jsam: To keep readability up, tuples with so many elements are kinda hard to read. Maybe we could use dict here instead?

tests/cli/test_datasets.py
```diff
@@ -905,3 +907,28 @@ def test_dataset_file_date_created_format(tmpdir, runner, client, project):
         assert dp.parse(data_yaml['created'])
         assert validate_iso8601(data_yaml['created'])
         assert validate_iso8601(data_yaml['files'][0]['added'])
+
+
+@pytest.mark.parametrize(
+    'uri', (
```
```suggestion
    'uri', [
```

tests/cli/test_datasets.py
```diff
@@ -905,3 +907,28 @@ def test_dataset_file_date_created_format(tmpdir, runner, client, project):
...
+        '10.5281/zenodo.3363060', 'doi:10.5281/zenodo.3363060',
+        'https://zenodo.org/record/3363060'
+    )
```
```suggestion
    ]
```

tests/cli/test_datasets.py
```diff
@@ -905,3 +907,28 @@ def test_dataset_file_date_created_format(tmpdir, runner, client, project):
...
+)
+def test_dataset_provider_resolution_zenodo(doi_responses, uri):
+    """Check that zenodo uris resolve to ZenodoProvider."""
+    provider, _ = ProviderFactory.from_uri(uri)
+    assert type(provider) is ZenodoProvider
+
+
+@pytest.mark.parametrize(
+    'uri', (
```
```suggestion
    'uri', [
```

tests/cli/test_datasets.py
```diff
@@ -905,3 +907,28 @@ def test_dataset_file_date_created_format(tmpdir, runner, client, project):
...
+        '10.7910/DVN/TJCLKP', 'doi:10.7910/DVN/TJCLKP',
+        'https://dataverse.harvard.edu/dataset.xhtml' +
+        '?persistentId=doi:10.7910/DVN/TJCLKP'
+    )
```
```suggestion
    ]
```

tests/cli/test_datasets.py
```diff
@@ -905,3 +907,28 @@ def test_dataset_file_date_created_format(tmpdir, runner, client, project):
...
+)
+def test_dataset_provider_resolution_dataverse(doi_responses, uri):
+    """Check that dataverse uris resolve to DataverseProvider."""
```
```suggestion
    """Check that dataverse URIs resolve to ``DataverseProvider``."""
```

tests/cli/test_integration_datasets.py
```diff
@@ -115,34 +109,45 @@ def test_dataset_import_real_doi_warnings(runner, project):
...
     assert 2 == result.exit_code
-    assert 'URI not found.' in result.output
+    assert 'URI not found.' in result.output \
+        or 'Provider {} not found'.format(doi[1]) in result.output
 
 
+@pytest.mark.parametrize(
+    'url', [
+        'https://zenodo.org/record/2621208',
+        'https://dataverse.harvard.edu/dataset.xhtml' +
```
jsam:
```suggestion
        ('https://dataverse.harvard.edu/dataset.xhtml' 
        '?persistentId=doi:10.7910/DVN/S8MSVF')
```

tests/cli/test_integration_datasets.py
```diff
@@ -115,34 +109,45 @@ def test_dataset_import_real_doi_warnings(runner, project):
...
-    )
+    result = runner.invoke(cli.cli, ['dataset', 'import', url], input='y')
     assert 0 == result.exit_code
     assert 'OK' in result.output
 
 
+@pytest.mark.parametrize(
+    'url', [
+        'https://zenodo.org/record/2621201248',
+        'https://dataverse.harvard.edu/dataset.xhtml' +
```
jsam:
```suggestion
        ('https://dataverse.harvard.edu/dataset.xhtml'
         '?persistentId=doi:10.7910/DVN/S8MSVFXXXX')
```

tests/cli/test_integration_datasets.py
Panaetius: I think it was yapf, I kept it at 1 per line. I'll see if it does it again

renku/cli/_providers/__init__.py
Panaetius: The idea is that providers should handle expected exceptions themselves, so if there is any exception here, something went quite wrong. It isn't really necessary for the 2 providers we have now, but the more providers we have, the more difficult it'll be to account for all the possible interplay between them (Provider X might error when tested with an URI meant for Provider Y).

But it is relatively safe to assume that a provider will handle URIs (and exceptions) it was meant for, so if the provider is supposed to handle an URI, it should not throw some random exception for it.

In this case we show a warning, assume the provider does not support the URI, fall through to the next provider(s) and hopefully the URI can get handled successfully.

If no provider supports the URL, the warning is elevated to an exception.

This way, we should gracefully handle exceptions due to unforeseen edge-cases between providers without swallowing exceptions when we don't want to.

Good point on the `BaseException`.

PR #633, opened 2019-08-22
fix: security update

jsam:
Github did security scan on our repo and found a security problem in werkzeug package. This PR updates the package to suggested version.


jirikuncar:
@jsam you might want to run `pipenv update` too.

jsam:
Oh hey @jirikuncar, thanks! Completely forgot about that workflow. Usually we update that lock file on release.

PR #635, opened 2019-08-22
add human commiter to KG and clean up dataset creator ids

### LINKED ISSUE #632: `schema:creator` pointing to `mailto:None` for imported data-set
It looks like in cases when a data-set is imported, the `schema:creator` triple points to `<mailto:None>` resource. As a consequence, I imagine all the triples from all the projects and imported data-sets will be put into one `schema:Person` `<mailto:None>` bag during triples generation process.

Reproducing:
repo: `git@dev.renku.ch:virginiafriedrich/datasets-test.git`
commit: `46a77c18a31e42d356be7e6d4eac5ad35468e881`
command: `renku log --format rdf`

### LINKED ISSUE #634: No info about person on the commit triples
It looks that currently there's no information on the commit triples about a person who performed a commit. It would be great if that's added so we'd have something like follows:
```
  <rdf:Description rdf:about="file:///commit/<sha>">
    <prov:agent rdf:resource="file:///_<uuid>"/>
    ...
  </rdf:Description>
  ...
  <rdf:Description rdf:about="file:///_<uuid>">
    <rdf:type rdf:resource="http://schema.org/Person"/>
    <rdf:type rdf:resource="http://www.w3.org/ns/prov#Person"/>
    <schema:name>user name</schema:name>
    <schema:email><host@user></schema:email>
  </rdf:Description>
```

rokroskar:
This PR:

* prevents the creation of nodes with `mailto:None` ids - instead, if we don't have obvious identifying information like an email or an orcid-id we simply generate a blank node
* adds the `Person` `prov:agent` to the commit activity - this is the human agent that created the commit

This is a sample graph:

noname gv

Fixes #634 
Fixes #632

### DIFF
--- a/renku/models/datasets.py
+++ b/renku/models/datasets.py
@@ -69,7 +69,7 @@ class Creator(object):
         context='schema:name'
     )
 
-    _id = jsonld.ib(default=None, kw_only=True, context='@id')
+    _id = jsonld.ib(kw_only=True, context='@id')
 
     @property
     def short_name(self):
@@ -123,10 +123,18 @@ class Creator(object):
         """Create an instance from a Git commit."""
         return cls(name=commit.author.name, email=commit.author.email)
 
+    @_id.default
+    def default_id(self):
+        """Set the default id."""
+        if self.email:
+            return 'mailto:{email}'.format(email=self.email)
+        return '_' + str(uuid.uuid4())
+
     def __attrs_post_init__(self):
-        """Post-Init hook to set _id field."""
-        if not self._id:
-            self._id = 'mailto:{self.email}'.format(self=self)
+        """Finish object initialization."""
+        # handle the case where ids were improperly set
+        if self._id == 'mailto:None':
+            self._id = '_' + str(uuid.uuid4())
 
 
 @attr.s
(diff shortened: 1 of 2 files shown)


renku/models/datasets.py
```diff
@@ -123,10 +123,12 @@ def from_commit(cls, commit):
...
 
-    def __attrs_post_init__(self):
-        """Post-Init hook to set _id field."""
-        if not self._id:
-            self._id = 'mailto:{self.email}'.format(self=self)
+    @_id.default
+    def default_id(self):
+        """Set the default id."""
+        if self.email:
+            return 'mailto:{self.email}'.format(self=self)
```
jsam: I think we don't need to pass entire self context to format. I would just pass what's needed.

renku/models/datasets.py
```diff
@@ -123,10 +123,12 @@ def from_commit(cls, commit):
...
+        return '_'+str(uuid.uuid4())
```
jsam: Is this formatted? (yapf)

rokroskar: nope! thanks!

renku/models/datasets.py
rokroskar: true - fixed

PR #639, opened 2019-08-23
fix: migration of old datasets

### LINKED ISSUE #636: migrating data results in wrong paths in the dataset metadata
This was reported by @jen-thomas - 

after migrating the dataset metadata and running `renku dataset export zenodo`:

```
File "[...]/site-packages/git/util.py", line 333, in finalize_process
    proc.wait(**kwargs)
  File "[...]/site-packages/git/cmd.py", line 412, in wait
    raise GitCommandError(self.args, status, errstr)
git.exc.GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git rev-list HEAD -- ../../../data/glonass/gpsdata-20161227.log
  stderr: 'fatal: ../../../data/glonass/gpsdata-20161227.log: '../../../data/glonass/gpsdata-20161227.log' is outside repository
```

jsam:
closes #636


rokroskar:
can you make sure to check for any paths in a `Dataset` that are outside the repo root? Those should be migrated, see https://github.com/SwissDataScienceCenter/renku-python/issues/646 which will fail with 

```
stderr: 'fatal: ../../../data/zhbikes/2019_verkehrszaehlungen_werte_fussgaenger_velo.csv: '../../../data/zhbikes/2019_verkehrszaehlungen_werte_fussgaenger_velo.csv' is outside repository
```

even if the bug which causes the reported problem is fixed...

renku/cli/_providers/zenodo.py
```diff
@@ -214,9 +214,11 @@ def version(self):
     def is_last_version(self, uri):
         """Check if record is at last possible version."""
         if is_doi(uri):
+            breakpoint()
```
rokroskar: 😎

renku/cli/_providers/zenodo.py
```diff
@@ -214,9 +214,11 @@ def version(self):
...
             return uri == self.metadata.prereserve_doi['doi']
 
         record_id = self.metadata.prereserve_doi['recid']
+        breakpoint()
```
rokroskar: 🙈

renku/cli/migrate.py
```diff
@@ -45,33 +44,30 @@ def migrate():
...
-            relpath(str(old_path.parent / key), start=str(new_path.parent))
-        )
-
-        with new_path.open('w') as fp:
-            yaml.dump(asjsonld(dataset), fp, default_flow_style=False)
+        dataset.__reference__ = new_path
+        dataset.to_yaml()
 
-        old_path.unlink()
+        if isinstance(old_path, Path):
```
rokroskar: why not just

```
Path(old_path).unlink()
```

jsam: Cause I hate to allocate objects for readability purposes.

renku/cli/_providers/zenodo.py
jsam: Thank you for reviewing WIP. I push the last nite to save my work, was too tired to clean up properly. :clinking_glasses:

renku/cli/migrate.py
rokroskar: I think in this case it's not really performance-critical?

rokroskar: It's violating at least 4-5 of the commandments from https://www.python.org/dev/peps/pep-0020/ :)

jsam: `There should be one-- and preferably only one --obvious way to do it. Although that way may not be obvious at first unless you're Dutch.` 

Not sure how serious we should be about following this obvious meme. ;)

jsam:
> can you make sure to check for any paths in a `Dataset` that are outside the repo root? Those should be migrated, see #646 which will fail with
> 
> ```
> stderr: 'fatal: ../../../data/zhbikes/2019_verkehrszaehlungen_werte_fussgaenger_velo.csv: '../../../data/zhbikes/2019_verkehrszaehlungen_werte_fussgaenger_velo.csv' is outside repository
> ```
> 
> even if the bug which causes the reported problem is fixed...

I'm trying to reproduce it, but fail to understand how that happened. Let's merge this as is until I can look more in-depth into #646 issue.

rokroskar:
Using https://dev.renku.ch/projects/lorenzo.cavazzi.tech/zurich-bikes-tutorial I get:

```
renku doctor
...
stderr: 'fatal: ../../../data/zhbikes/2019_verkehrszaehlungen_werte_fussgaenger_velo.csv: '../../../data/zhbikes/2019_verkehrszaehlungen_werte_fussgaenger_velo.csv' is outside repository
```

renku/api/repository.py
```diff
@@ -99,11 +100,13 @@ def __attrs_post_init__(self):
...
-            ],
-                         cwd=str(self.path))
-        # TODO except
+            try:
+                check_output([
+                    'git', 'submodule', 'update', '--init', '--recursive'
+                ],
+                             cwd=str(self.path))
+            except subprocess.CalledProcessError:
+                pass
```
rokroskar: why fail silently? We should do/say _something_

renku/api/repository.py
```diff
@@ -251,37 +254,44 @@ def subclients(self, parent_commit):
...
-                    return (
-                        subclient,
-                        subclient.find_previous_commit(
-                            subpath, revision=submodule.hexsha
-                        ),
-                        subpath,
-                    )
-                except ValueError:
-                    pass
+                if (Path(submodule.path) / Path('.git')).exists():
```
rokroskar: doesn't a submodule by definition have a `.git` directory?

renku/api/repository.py
```diff
@@ -251,37 +254,44 @@ def subclients(self, parent_commit):
...
+        return subclients
 
     def resolve_in_submodules(self, commit, path):
         """Resolve filename in submodules."""
         original_path = self.path / path
-        if original_path.is_symlink() or str(path
-                                             ).startswith('.renku/vendors'):
+        in_vendor = str(path).startswith('.renku/vendors')
+
+        if original_path.is_symlink() or in_vendor:
```
rokroskar: Why not resolve the path before and do the `in_vendor` check on the resolved path?

renku/cli/_checks/migrate_datasets.py
```diff
@@ -52,3 +76,22 @@ def check_missing_files(client):
...
 
     return False
+
+
+def check_dataset_resources(client):
+    """Find missing datasets or files listed in metadata."""
+    missing_datasets = defaultdict(list)
+    missing_files = defaultdict(list)
+
+    for path, dataset in client.datasets.items():
```
rokroskar: it might be better to load the `yaml` or `json` directly - then you can also check if the file paths make sense, e.g. are valid inside the repository.

renku/cli/migrate.py
```diff
@@ -45,33 +48,84 @@ def migrate():
...
+            dataset_path = (
+                client.renku_datasets_path / uuid.UUID(dataset.identifier).hex
+            )
+
+        if not expected_path.exists():
+            shutil.move(str(dataset_path), str(expected_path))
+            dataset.path = expected_path
+            dataset.__reference__ = expected_path / client.METADATA
+
+        for file_ in dataset.files:
```
rokroskar: check here whether files have paths outside the repository root - if so, renormalize the paths

rokroskar:
Running through the toy example from https://github.com/SwissDataScienceCenter/renku-python/issues/636#issue-484095429 but using the code from this PR instead of renku `0.5.2` I end up in this state:

```
 git status
On branch master
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        deleted:    data/mydata/metadata.yml
```

and I still have a submodule in the repo:

```
cat .gitmodules
[submodule "test-migration"]
        path = .renku/vendors/local/private/tmp/test-migration
        url = /private/tmp/test-migration
```

checked out in the `.renku/vendors/local` directory:

```
tree .renku
.renku
├── datasets
│   └── 164bc1a7-3919-4851-8975-230242852f0e
│       └── metadata.yml
├── metadata.yml
├── refs
│   └── datasets
│       └── mydata -> ../../datasets/164bc1a7-3919-4851-8975-230242852f0e/metadata.yml
└── vendors
    └── local
        └── private
            └── tmp
                └── test-migration
                    ├── Dockerfile
                    ├── data
                    │   └── mydata
                    │       └── metadata.yml
                    ├── datafile
                    └── requirements.txt
```

renku/api/repository.py
jsam: There is no guarantee that subclients will resolve to the git repo. For example, you can have a case where submodule is listed in `.gitmodules` and it's pointing to a raw directory. In that case, code explodes here instead of giving user a nice message.

renku/api/repository.py
jsam: This is just a rewrite for easier reading the code. Code is the same as before, just nicer to read.

renku/api/repository.py
rokroskar: Then we should maybe try to do a better check - is this submodule actually listed in `.gitmodules`, if yes, can we re-initialize it

renku/api/repository.py
rokroskar: yep I know, I was just wondering about the order of operations

renku/cli/_checks/migrate_datasets.py
jsam: What do you mean? `client.datasets` will load it directly and since this is validation and it will do just that `check if paths make sense`.

renku/api/repository.py
rokroskar: I suppose we can leave that for another time and just leave this code as-is for now

renku/api/repository.py
jsam: We are not failing silently. Git output will be shown.

rokroskar:
argh... I'm finding now another issue where the file paths are absolute - so they can't be resolved when the repo is cloned somewhere else. Do you want to deal with that in the migration as well?

renku/cli/migrate.py
jsam: This check is right here, 2 line below. Not sure what are you referring to.

jsam:
How did you test this? I tested it again this morning and it works for me and also I have a test which reproduces the report from #636 which is passing.

rokroskar:
I don't see any different behavior - running these commands:

```
renku init test-migration
cd test-migration
pipx run --spec renku==0.3.3 renku dataset create mydata
touch datafile
git add datafile
git commit -m 'added datafile'
pipx run --spec renku==0.3.3 renku dataset add mydata datafile
pipx run --spec renku==0.5.2 renku migrate datasets
git show
```

I still have the local directory in the `.gitmodules` and the file paths point outside the repo:

```
'@context':
  _id: '@id'
  _label: rdfs:label
  _project: schema:isPartOf
  added: schema:dateCreated
  affiliation: schema:affiliation
  alternate_name: schema:alternateName
  created: schema:dateCreated
  creator: schema:creator
  date_published: schema:datePublished
  description: schema:description
  email: schema:email
  files: schema:hasPart
  identifier: schema:identifier
  in_language: schema:inLanguage
  keywords: schema:keywords
  license: schema:license
  name: schema:name
  path: prov:atLocation
  schema: http://schema.org/
  url: schema:url
  version: schema:version
  wfprov: http://purl.org/wf4ever/wfprov#
'@type':
- prov:Entity
- schema:Dataset
- wfprov:Artifact
_id: 6a05c063-b852-40dc-899d-8a460454ec7e
_label: 6a05c063-b852-40dc-899d-8a460454ec7e
_project:
  '@context':
    _id: '@id'
    foaf: http://xmlns.com/foaf/0.1/
  '@type':
  - foaf:Project
  - prov:Location
  _id: file:///Users/rok/Projects/renku-test-projects/test-migration
created: '2019-09-04 08:24:23.232835'
creator:
- '@type': schema:Person
  _id: mailto:rok.roskar@sdsc.ethz.ch
  affiliation: null
  alternate_name: null
  email: rok.roskar@sdsc.ethz.ch
  name: Rok Roskar
date_published: null
description: null
files:
- '@type':
  - prov:Entity
  - schema:DigitalDocument
  - wfprov:Artifact
  _id: blob/UNCOMMITTED/datafile
  _label: datafile@UNCOMMITTED
  _project: null
  added: '2019-09-04 08:24:25.298332'
  creator:
  - '@type': schema:Person
    _id: mailto:rok.roskar@sdsc.ethz.ch
    affiliation: null
    alternate_name: null
    email: rok.roskar@sdsc.ethz.ch
    name: Rok Roskar
  dataset: null
  path: ../../../data/mydata/datafile
  url: null
identifier: 6a05c063-b852-40dc-899d-8a460454ec7e
in_language: null
keywords: []
license: null
name: mydata
path: /Users/rok/Projects/renku-test-projects/test-migration/.renku/datasets/6a05c063-b852-40dc-899d-8a460454ec7e
url: null
version: null
```

rokroskar:
Trying to migrate the dataset from https://renkulab.io/gitlab/ACE-ASAID/meteorology (after deleting the new one) I get 

```
renku dataset ls-files
Error: Broken resource of type `datasets` found.
Migration is required.
Please check `renku migrate --help` command to fix the issue.
Hint: `renku migrate datasets`
```

which I guess is because of absolute paths:

```
cat .renku/datasets/ce2e0f25-dff1-49f5-8d37-19aaa12c0e22/metadata.yml
'@context':
  _id: '@id'
  _label: rdfs:label
  _project: schema:isPartOf
  created: schema:dateCreated
  creator: schema:creator
  date_published: schema:datePublished
  description: schema:description
  files: schema:hasPart
  identifier: schema:identifier
  in_language: schema:inLanguage
  keywords: schema:keywords
  license: schema:license
  name: schema:name
  path: prov:atLocation
  schema: http://schema.org/
  url: schema:url
  version: schema:version
  wfprov: http://purl.org/wf4ever/wfprov#
'@type':
- prov:Entity
- schema:Dataset
- wfprov:Artifact
_id: ce2e0f25-dff1-49f5-8d37-19aaa12c0e22
_label: ce2e0f25-dff1-49f5-8d37-19aaa12c0e22
_project:
  '@context':
    _id: '@id'
    foaf: http://xmlns.com/foaf/0.1/
  '@type':
  - foaf:Project
  - prov:Location
  _id: https://renkulab.io/ACE-ASAID/meteorology
authors:
- '@type': dcterms:creator
  affiliation: null
  email: jenny_t152@yahoo.co.uk
  name: Jen Thomas
created: '2018-11-23T14:42:17.985570'
creator: []
date_published: null
description: null
files:
- '@type':
  - prov:Entity
  - schema:DigitalDocument
  - wfprov:Artifact
  _id: blob/UNCOMMITTED/MAWS__SMSAWS__20161123.txt
  _label: MAWS__SMSAWS__20161123.txt@UNCOMMITTED
  _project: null
  added: '2018-11-23T15:12:42.232297'
  creator: []
  dataset: null
  name: MAWS__SMSAWS__20161123.txt
  path: MAWS__SMSAWS__20161123.txt
  url: /home/jen/projects/ace_data_management/data_staging_for_renku/meteorology/MAWS__SMSAWS__20161123.txt
- '@type':
  - prov:Entity
  - schema:DigitalDocument
  - wfprov:Artifact
  _id: blob/UNCOMMITTED/MAWS__SMSAWS__20161124.txt
  _label: MAWS__SMSAWS__20161124.txt@UNCOMMITTED
  _project: null
  added: '2018-11-23T15:12:42.257562'
  creator: []
  dataset: null
  name: MAWS__SMSAWS__20161124.txt
  path: MAWS__SMSAWS__20161124.txt
  url: /home/jen/projects/ace_data_management/data_staging_for_renku/meteorology/MAWS__SMSAWS__20161124.txt
...
```

rokroskar:
Another thing I noticed - `authors` is no longer migrated to `creator` - it looks like the model migrations are now skipped?

rokroskar:
apologies, my repo found itself in a state from which the only easy way out was a force-push...

renku/cli/_checks/migrate_datasets.py
```diff
@@ -0,0 +1,226 @@
...
+
+    return missing_datasets, missing_files
+
+
+def ensure_clean_lock(client):
+    """Make sure Renku lock file is not part of repository."""
+    lock_file = Path('.renku.lock')
+    if lock_file.exists():
+        lock_file.unlink()
+
```
rokroskar: If the lock file is deleted, it doesn't get properly comitted without: 
```suggestion
client.repo.git.add(str(lock_file))

```

renku/models/migrations/__init__.py
```diff
@@ -15,8 +15,14 @@
...
 
-from renku.models.migrations.dataset import migrate_dataset
+from renku.models.migrations.dataset import (
+    migrate_dataset_schema,
+    migrate_absolute_paths,
+)
 
-MIGRATIONS = [migrate_dataset]
+JSONLD_MIGRATIONS = {
+    'dctypes:Dataset': [migrate_dataset_schema, migrate_absolute_paths],
```
rokroskar: this will fail if the expanded form of the URI is used, e.g. this could be just `http://purl.org/dc/dcmitypes/Dataset` if it came out of a sparql query. We should address this in another PR.

renku/cli/_checks/migrate_datasets.py
```diff
@@ -0,0 +1,226 @@
...
+                relative = path.parent.relative_to(client.renku_datasets_path)
+                missing_files[str(relative)].append(str(filepath))
+
+    return missing_datasets, missing_files
+
+
+def ensure_clean_lock(client):
+    """Make sure Renku lock file is not part of repository."""
+    lock_file = Path('.renku.lock')
+    if lock_file.exists():
```
rokroskar: This should check if the lock file is in the git tree, not just on the filesystem. If it's been checked in and deleted then the current code doesn't remove it from git. If you use my suggestion below then the code fails when the file is present on the filesystem but it's not checked in git. A better approach might be:

```
    lock_file = Path('.renku.lock')
    try:
        client.repo.tree() / '.renku.lock'
    except KeyError:  # the lock file is not checked in
        return
    lock_file.unlink()
    client.repo.git.add(str(lock_file))
```

jsam: Well since you cannot run the structural migration without a clean repository, this check will actually check if its part of the git tree, since you will always run it from the HEAD clean state.

rokroskar: It isn't because it doesn't actually remove the file from git if it's checked in.

jsam: Hmm, I'm failing to reproduce what you're reporting. I've added the test which validates what I said [above](https://github.com/SwissDataScienceCenter/renku-python/pull/639/files#diff-7e3313f0ee1c3b9530e15596b9b67537R122). Also I've added to that migration that we ensure that lock file is part of the gitignore.

renku/cli/_checks/migrate_datasets.py
jsam: Could you please describe the steps on how I can reproduce the issue?

renku/models/migrations/__init__.py
jsam: True, except @type attribute seems to always be in human-readable format due to compaction. Could you provide us with a concrete example where that would be the case so we can add it as a test?

renku/cli/_checks/migrate_datasets.py
```diff
@@ -0,0 +1,232 @@
...
+        lock_file.unlink()
+
+    # Add lock file to .gitignore
+    with open('.gitignore', 'a+') as f:
```
rokroskar: This needs to be wrt the root of the project - if the command is run in a subdirectory it breaks

renku/cli/_checks/migrate_datasets.py
```diff
@@ -0,0 +1,232 @@
...
+            if not filepath.exists():
+                relative = path.parent.relative_to(client.renku_datasets_path)
+                missing_files[str(relative)].append(str(filepath))
+
+    return missing_datasets, missing_files
+
+
+def ensure_clean_lock(client):
+    """Make sure Renku lock file is not part of repository."""
+    lock_file = Path('.renku.lock')
```
rokroskar: Make sure you make this path relative to the root of the project

renku/cli/_checks/migrate_datasets.py
```diff
@@ -0,0 +1,232 @@
...
+    if lock_file.exists():
+        lock_file.unlink()
+
+    # Add lock file to .gitignore
+    with open('.gitignore', 'a+') as f:
+        ignored = f.read()
+        if str(lock_file) not in ignored:
```
rokroskar: Hmm - not sure why, but running `renku migrate datasets` several times results in:

```
renku migrate datasets
 rok   master → rok ↑2 ↓1  ~/Projects/renku-test-projects/meteorology  git show
commit 36cb6fb62e0c11124410aaf89f9e26100dbe302f (HEAD -> master)
Author: Rok Roskar <rok.roskar@sdsc.ethz.ch>
Date:   Wed Sep 11 18:29:34 2019 +0200

    renku migrate datasets

diff --git a/.gitignore b/.gitignore
index 85f6388..0827d95 100644
--- a/.gitignore
+++ b/.gitignore
@@ -257,4 +257,4 @@ tags
 # End of https://www.gitignore.io/api/macos,python,R,linux,vim,emacs

 .renku.lock
-.renku.lock
\ No newline at end of file
+.renku.lock.renku.lock
\ No newline at end of file
 rok   master → rok ↑2 ↓1  ~/Projects/renku-test-projects/meteorology  renku migrate datasets
 rok   master → rok ↑3 ↓1  ~/Projects/renku-test-projects/meteorology  git show
commit c62aad18d64fe08750f355b9ab54f35f0e4d9bc1 (HEAD -> master)
Author: Rok Roskar <rok.roskar@sdsc.ethz.ch>
Date:   Wed Sep 11 18:29:50 2019 +0200

    renku migrate datasets

diff --git a/.gitignore b/.gitignore
index 0827d95..115d318 100644
--- a/.gitignore
+++ b/.gitignore
@@ -257,4 +257,4 @@ tags
 # End of https://www.gitignore.io/api/macos,python,R,linux,vim,emacs

 .renku.lock
-.renku.lock.renku.lock
\ No newline at end of file
+.renku.lock.renku.lock.renku.lock
\ No newline at end of file
```

In the line below, make sure you add a newline.

renku/models/migrations/__init__.py
rokroskar: If the context doesn't include the prefix, or if it includes the prefix by a different name, e.g. I would be free to define `schemaorg: http://schema.org` then this breaks. The full solution is to use the jsonld library here instead of a simple string match IMO. That's why I suggested we handle it in a follow-up issue.

renku/cli/_checks/migrate_datasets.py
rokroskar: you're right, with the current changes it seems to work 🤞

Issue #646, opened 2019-08-26
cli: renku log figs/grid_plot.png


lorenzo-cavazzi:
## Describe the bug
Upgrading from renku-python 0.5.0 to 0.5.2 breaks `renku log <specific_path>`

## Details
I created the project "https://dev.renku.ch/gitlab/lorenzo.cavazzi.tech/zurich-bikes-tutorial" on https://dev.renku.ch when the version 0.5.0 of the CLI was used in the Docker image. I am now using the latest version of the cli (locally) but I got an error with the command `renku log <path>` for a few specific files, like `renku log figs/grid_plot.png`.
I verified that the problem is not my local environment. I tried the following:
- use 0.5.0 locally --> no error anymore
- update the Docker image to 0.5.2 and start it from https://dev.renku.ch --> get the error -- the report is from this setup

**Renku version:** 0.5.2
**OS:** Linux (#1 SMP Mon Oct 22 10:40:32 EDT 2018)
**Python:** 3.7.3

### Traceback

```
Traceback (most recent call last):
  File "[...]/site-packages/renku/cli/_exc.py", line 100, in main
    result = super().main(*args, **kwargs)
  File "[...]/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "[...]/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "[...]/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "[...]/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "[...]/site-packages/renku/cli/_client.py", line 83, in new_func
    result = ctx.invoke(method, client, *args, **kwargs)
  File "[...]/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "[...]/site-packages/renku/cli/log.py", line 111, in log
    FORMATS[format](graph)
  File "[...]/site-packages/renku/cli/_format/graph.py", line 30, in ascii
    echo_via_pager(str(DAG(graph)))
  File "[...]/site-packages/renku/cli/_ascii.py", line 81, in __str__
    return '\n'.join(self)
  File "[...]/site-packages/renku/cli/_ascii.py", line 88, in __iter__
    node_symbol, self.node_text(node), column_info
[...]

PR #648, opened 2019-08-27
fix: always add commit to dataset if possible

rokroskar:
Always add a `commit` to `Dataset` if possible - in the case where a dataset is first being created with `with_dataset`, it won't have been committed yet.

This addresses #646 but needs the resolution of #639 for a full solution. 

### DIFF
--- a/renku/models/_jsonld.py
+++ b/renku/models/_jsonld.py
@@ -315,7 +315,7 @@ class JSONLDMixin(ReferenceMixin):
         data,
         __reference__=None,
         __source__=None,
-        **kwargs,
+        **kwargs
     ):
         """Instantiate a JSON-LD class from data."""
         if isinstance(data, cls):
--- a/renku/models/_jsonld.py
+++ b/renku/models/_jsonld.py
@@ -313,9 +313,9 @@ class JSONLDMixin(ReferenceMixin):
     def from_jsonld(
         cls,
         data,
-        client=None,
         __reference__=None,
         __source__=None,
+        **kwargs,
     ):
         """Instantiate a JSON-LD class from data."""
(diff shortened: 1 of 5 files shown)


renku/models/_jsonld.py
```diff
@@ -313,9 +313,9 @@ class JSONLDMixin(ReferenceMixin):
     def from_jsonld(
         cls,
         data,
-        client=None,
         __reference__=None,
         __source__=None,
+        **kwargs,
```
jsam:
```suggestion
        **kwargs
```

renku/models/_jsonld.py
```diff
@@ -372,17 +370,17 @@ def from_jsonld(
         return self
 
     @classmethod
-    def from_yaml(cls, path, client=None):
+    def from_yaml(cls, path, **kwargs):
```
jsam: I generally think kwargs should only be used as a last resort. Can we be explicit with our params or at least document what can come in kwargs?

rokroskar: Sure, it was just starting to look wordy but you're probably right

renku/models/_jsonld.py
```diff
@@ -331,7 +332,7 @@ def from_jsonld(
             ) != type_:
                 new_cls = cls.__type_registry__[type_]
                 if cls != new_cls:
-                    return new_cls.from_jsonld(data, client=client)
+                    return new_cls.from_jsonld(data, client=client, commit=commit)
```
jsam:
```suggestion
                    return new_cls.from_jsonld(
                        data, 
                        client=client, 
                        commit=commit
                    )
```

Issue #628, opened 2019-08-20
KG: lazy load DatasetFile metadata


rokroskar:
We have several places where setting of various attributes for JSON-LD classes is not self-consistent. For example, doing `renku dataset import <doi>` will result in a list of files all with an `UNCOMMITTED` label. We should review the metadata in `Dataset` and `DatasetFile` classes and come up with a way that we get the same metadata regardless of how the instance of the class is made. 

For instance, the `_label` in `DatasetFile` is inherited from `CommitMixin` and is set using the `commit` which may not always be available. However, this is also internal to renku and should/could be set at the time of instantiation - we don't necessarily need to save this in the JSON-LD on disk. OTOH it probably *is* useful to save the sha1 of the files in the metadata as this could be used to verify that what is being read from disk is actually the same data that was initially part of the dataset.

PR #651, opened 2019-08-27
fix: adds _label and commit data to imported dataset files, single commit for imports

Panaetius:
closes: #628 

This fixes the bug mentioned in #628 though it does not make DatasetFile properties lazy loading, as that was not the cause of this problem.

When importing a `DatasetFile` via the `renku import` command, the `add_to_dataset` method in `renku/cli/dataset.py` first calls `client.add_data_to_dataset(...)` to add (and commit the files, but then calls `dataset.update_metadata(with_metadata)` to update the metadata. But since `files` are considered metadata by `update_metadata(...)`, the original, uncommitted file entries overwrite the newly created entries.

As this will be reworked soon with the big CLI changes, this is just a quick&dirty hotfix that copies over the labels/commits of the files before they get overwritten (as discussed with @jsam ).

`client.add_data_to_dataset(...)` now takes a list of urls as input and commits them in a single commit, getting rid of the commit spam when importing a dataset with many files.

`dataset.update_files(...)` now actually updates files commits if they get added again instead of just working for new files, though this is not necessary for fixing the issue of this PR. I Just left it in as it makes sense and doesn't break anything. Maybe updating all metadata of the files would be even better. Let me know if this makes sense or not and I can adjust the PR.

### DIFF
--- a/renku/api/datasets.py
+++ b/renku/api/datasets.py
@@ -145,29 +145,36 @@ class DatasetsApiMixin(object):
         dataset.to_yaml()
 
     def add_data_to_dataset(
-        self, dataset, url, git=False, force=False, **kwargs
+        self, dataset, urls, git=False, force=False, **kwargs
     ):
         """Import the data into the data directory."""
         dataset_path = self.path / self.datadir / dataset.name
-        git = git or check_for_git_repo(url)
 
-        target = kwargs.pop('target', None)
+        files = []
 
-        if git:
-            if isinstance(target, (str, NoneType)):
-                files = self._add_from_git(
-                    dataset, dataset_path, url, target, **kwargs
-                )
-            else:
-                files = []
-                for t in target:
+        for url in urls:
+            git = git or check_for_git_repo(url)
+
+            target = kwargs.pop('target', None)
+
+            if git:
+                if isinstance(target, (str, NoneType)):
                     files.extend(
                         self._add_from_git(
-                            dataset, dataset_path, url, t, **kwargs
+                            dataset, dataset_path, url, target, **kwargs
                         )
                     )
-        else:
-            files = self._add_from_url(dataset, dataset_path, url, **kwargs)
+                else:
+                    for t in target:
+                        files.extend(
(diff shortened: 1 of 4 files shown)


renku/api/datasets.py
```diff
@@ -145,29 +145,36 @@ def with_dataset(self, name=None):
...
+        files = []
 
-        if git:
-            if isinstance(target, (str, NoneType)):
-                files = self._add_from_git(
-                    dataset, dataset_path, url, target, **kwargs
-                )
-            else:
-                files = []
-                for t in target:
```
rokroskar: `target` can have multiple elements - but this should probably be handled in the `_add_from_*` methods anyway

Panaetius: I didn't change this part. github just has issues with following the changes, the diff is a bit of a mess (I indented the whole block and added `for url in urls:` at the top). I.e. this "removed" part is "added" again further down.

renku/cli/dataset.py
```diff
@@ -309,11 +308,13 @@ def add_to_dataset(
...
                             file_.path = added_.path
                             file_.creator = with_metadata.creator
+                            file_._label = added_._label
+                            file_.commit = added_.commit
 
                 dataset.update_metadata(with_metadata)
 
     except FileNotFoundError:
-        raise BadParameter('Could not process {0}'.format(url))
+        raise BadParameter('Could not process {0}'.format(', '.join(urls)))
```
jsam:
```suggestion
        raise BadParameter('Could not process \n{0}'.format('\n'.join(urls)))
```

renku/api/datasets.py
rokroskar: ah yep I see it, thanks!

rokroskar:
> Can we maybe add explicit assertion/test for _label and commit?

you mean in the code? or as a part of the tests? In the code, `commit` is allowed to be empty by `CommitMixin` - I would imagine things would break in various places if you change that, but perhaps not.

PR #653, opened 2019-08-28
fix: cleanup needed for integration tests on py35

jsam:


renku/cli/dataset.py
```diff
@@ -624,8 +624,8 @@ def _init(lock, id_queue):
 
         try:
             for p in processing:
-                p.wait()
-                p.get()
+                p.get(timeout=int(os.getenv('RENKU_REQUEST_TIMEOUT', 300)))
```
Panaetius: What's the reasoning behind this change? It'll raise a `TimeoutError` if a download takes more than 300s

renku/utils/doi.py
```diff
@@ -34,7 +34,5 @@ def extract_doi(uri):
     """Return the DOI in a string if there is one."""
     match = doi_regexp.match(uri)
 
-    if match is None:
-        return None
-
-    return match.group(2)
+    if match:
+        return match.group(2)
```
Panaetius: I prefer explicit return of None instead of implicit, not sure if we have any conventions regarding this.

renku/cli/dataset.py
jsam: Get internally invokes for the wait, so we don't need both. And I was having some trouble where I tried to import a dataset but entire TTY was just hanging for a very long time, so I quick patched it with this. I can remove the timeout until we get the ticket about it or extend it to 24 hours or something like that.

renku/utils/doi.py
jsam: Honestly, we have both messes in the repo. I would never say that this is explicit but just plain verbose. We should pick one style and stick with it for future reviews, and since I'm not in favor of adding `return None` to all unit functions I would vote for non-verbose option.

renku/cli/dataset.py
jsam: I'll remove it until we find a nice solution.

PR #658, opened 2019-08-30
fix: Removes unneccesary call to git lfs with no paths

Panaetius:
`git lfs track -- <...>` now only gets called when there actually are paths to call. This should fix the issue for the user, though I couldn't reproduce the underlying git lfs issue.

### DIFF
--- a/renku/api/storage.py
+++ b/renku/api/storage.py
@@ -132,12 +132,13 @@ class StorageApiMixin(RepositoryApiMixin):
                 # TODO create configurable filter and follow .gitattributes
                 track_paths.append(str(path))
 
-        call(
-            self._CMD_STORAGE_TRACK + track_paths,
-            stdout=PIPE,
-            stderr=STDOUT,
-            cwd=str(self.path),
-        )
+        if any(track_paths):
+            call(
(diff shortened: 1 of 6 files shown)


renku/api/storage.py
```diff
@@ -132,12 +132,13 @@ def track_paths_in_storage(self, *paths):
                 # TODO create configurable filter and follow .gitattributes
                 track_paths.append(str(path))
 
-        call(
-            self._CMD_STORAGE_TRACK + track_paths,
-            stdout=PIPE,
-            stderr=STDOUT,
-            cwd=str(self.path),
-        )
+        if any(track_paths):
```
rokroskar: why do you need the `any`?

jsam: Can we clean it up if we don't need it? I don't think we need it since it will always contain Path objects or an exception will be raised instead of None being put into a list.

renku/api/storage.py
```diff
@@ -132,12 +132,13 @@ def track_paths_in_storage(self, *paths):
...
+            call(
```
jsam: I think this can raise KeyboardInterrupt. Maybe we should handle it and display a nice message to the user instead of a stack trace?

Issue #99, opened 2018-02-09
Import datasets from renku-aware repos


rokroskar:
Importing from a git repository that contains a `.renku` directory should automatically reuse the included metadata about authors/creators of various entities. 

- [x] remove the local filesystem path (privacy issues)
- [ ] reference the original dataset metadata file: `$ref: ...`
- [ ] use submodule index to iterate over files when importing from a Git repo

addresses https://github.com/SwissDataScienceCenter/renga/issues/135

PR #666, opened 2019-09-10
99 import dataset from renku aware repo

mohammad-sdsc:

# Description

This PR enables adding of other Renku repositories to datasets. It also copies Creators of the original dataset over if that information exists. Moreover, it refactors the code to use submodule index to get list of files in a dataset (instead of recursive calls). Some minor bugs are fixed as well.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/99

## Type of change

## Checklist


renku/api/datasets.py
```diff
@@ -295,8 +298,9 @@ def _add_from_git(self, dataset, path, url, target, **kwargs):
         u = parse.urlparse(url)
         submodule_path = self.renku_path / 'vendors' / (u.netloc or 'local')
 
-        # Respect the directory struture inside the source path.
+        # Respect the directory structure inside the source path.
         relative_to = kwargs.get('relative_to', None)
```
Panaetius: You can change the default value for `kwargs.get(...)` instead of doing `relative_to = relative_to or ''`

```suggestion
         relative_to = kwargs.get('relative_to', '')
```

renku/api/datasets.py
```diff
@@ -352,69 +356,103 @@ def _add_from_git(self, dataset, path, url, target, **kwargs):
                 submodule_path.relative_to(self.path).as_posix()
             ])
 
-        src = submodule_path / (target or '')
+        target = target or ''
+        relative_to = relative_to or ''
```
Panaetius: Is this line needed, since you're doing it already above?

renku/api/datasets.py
```diff
@@ -352,69 +356,103 @@ def _add_from_git(self, dataset, path, url, target, **kwargs):
...
+            files.extend({e.path for e in src_root.traverse()})
+
+        # create a map of all submodule's dataset files to be used to
+        # get creators. Use Dataset's creator if a file does not have
+        # a creator
+        submodule_client = LocalClient(submodule_path)
+        submodule_dataset_creators = {}
+        for sb_dataset in submodule_client.datasets.values():
+            for f in sb_dataset.files:
+                creators = f.creator if f.creator else sb_dataset.creator
```
Panaetius: To be consistent with the other default value statements.
```suggestion
                creators = f.creator or sb_dataset.creator
```

renku/api/datasets.py
```diff
@@ -352,69 +356,103 @@ def _add_from_git(self, dataset, path, url, target, **kwargs):
...
+                file_creator = submodule_dataset_creators.get(entry)
+                if file_creator:
+                    creators = file_creator
+                else:
+                    # grab all the creators from the commit history
+                    for commit in submodule_repo.iter_commits(
+                        paths=dst_target
+                    ):
+                        creator = Creator.from_commit(commit)
+                        if creator not in creators:
```
Panaetius: if you use a `set()` you don't need to do this check.

renku/api/datasets.py
mohammad-sdsc: `kwargs` has a `relative_to` filed but its value can be `None` so `kwargs.get('relative_to', '')` returns `None` and not `''`.

renku/api/datasets.py
mohammad-sdsc: I prefer to keep it as it is because then I have to convert from set to dict.

renku/api/datasets.py
Panaetius: After a second look, do we actually need to use kwargs here? The only kwargs argument that's accessed is `relative_to` and they aren't passed anywhere (only recursively), so I see no reason why `relative_to` can't be a proper argument and `**kwargs` could be removed.

renku/api/datasets.py
```diff
@@ -76,7 +76,9 @@ def datasets_from_commit(self, commit=None):
     def datasets(self):
         """Return mapping from path to dataset."""
         result = {}
-        for path in self.renku_datasets_path.rglob(self.METADATA):
+        for path in (self.path / self.renku_datasets_path).rglob(
+            self.METADATA
+        ):
```
jsam:
```suggestion
        paths = (self.path / self.renku_datasets_path).rglob(self.METADATA)
        for path in paths:
```

renku/api/datasets.py
```diff
@@ -295,8 +306,8 @@ def _add_from_git(self, dataset, path, url, target, **kwargs):
         u = parse.urlparse(url)
         submodule_path = self.renku_path / 'vendors' / (u.netloc or 'local')
 
-        # Respect the directory struture inside the source path.
-        relative_to = kwargs.get('relative_to', None)
+        # Respect the directory structure inside the source path.
+        relative_to = relative_to or ''
```
jsam: We should set `''` as the default parameter in method signature.

renku/api/datasets.py
```diff
@@ -352,69 +363,102 @@ def _add_from_git(self, dataset, path, url, target, **kwargs):
                 submodule_path.relative_to(self.path).as_posix()
             ])
 
-        src = submodule_path / (target or '')
+        target = target or ''
```
jsam: Put into method signature?

renku/api/datasets.py
```diff
@@ -352,69 +363,102 @@ def _add_from_git(self, dataset, path, url, target, **kwargs):
...
+        if relative_to:
             relative_to = Path(relative_to)
             if relative_to.is_absolute():
                 assert u.scheme in {
                     '', 'file'
                 }, 'Only relative paths can be used with URLs.'
-                target = (Path(url).resolve().absolute() / target).relative_to(
-                    relative_to.resolve()
+                relative_to = relative_to.resolve().relative_to(
+                    Path(url).resolve().absolute()
```
jsam:
```suggestion
                    Path(url).resolve()
```
Will already make the path absolute.

renku/api/datasets.py
```diff
@@ -352,69 +363,102 @@ def _add_from_git(self, dataset, path, url, target, **kwargs):
...
                 )
+
+            if target:
+                if str(target) != str(relative_to):
```
jsam: Merge them together

```
if target and str(target) != str(relative_to)
```

or set the target default value first and then make the second check. We don't need 2 ifs here.

Also, comparing strings here is probably not the best way of going about this since they can be symbolic paths.

rokroskar:
I tried the following:

```
renku init test-renku-dataset
cd test-renku-dataset
renku dataset create mydata
renku dataset add mydata https://renkulab.io/gitlab/cramakri/renku-tutorial-flights.git --target data/flights/2019-01-flights.csv.zip
```

This works, but the metadata I get is this:

```
files:
- '@type':
  - prov:Entity
  - schema:DigitalDocument
  - wfprov:Artifact
  _id: blob/aa030cc96e04d1b450b24911c675b4b2bac0d868/data/flights/2019-01-flights.csv.zip
  _label: data/flights/2019-01-flights.csv.zip@aa030cc96e04d1b450b24911c675b4b2bac0d868
  _project:
    '@context':
      _id: '@id'
      foaf: http://xmlns.com/foaf/0.1/
    '@type':
    - foaf:Project
    - prov:Location
    _id: https://renkulab.io/cramakri/renku-tutorial-flights
  added: '2019-09-17T21:30:55.642386+00:00'
  creator:
  - '@type': schema:Person
    _id: mailto:cramakri@ethz.ch
    affiliation: null
    alternate_name: null
    email: cramakri@ethz.ch
    name: Chandrasekhar Ramakrishnan
```

Two issues here:

1. the `_id` and _label` fields of the file metadata are copied from the original dataset. These (for the time being) only make sense in the context of the current project since they include the local project path - hence they should not be copied over but regenerated in the current context instead. 

2. the `_project` is linking to the source project. This is not incorrect, but it should _also_ link to the current project. The only possible solution here might be to use a list of projects (like we do for `files` and `creator`.

rokroskar:
@jsam:

> A third concern which I have is related to submodules. Currently, if we import only 1 dataset (and others are not needed) from the project we put the whole git repository as a submodule (which is not needed, since we are interested in 1 dataset from that project). Since we only need to import metadata correctly (including the unique ID's) and copy the data, I'm not sure if we need to add it as a submodule at all.

This is not specific to this PR but how data import works from git repositories - this was done to allow us to use git mechanisms to update the contents of imported entities.

tests/cli/test_datasets.py
```diff
@@ -284,16 +342,16 @@ def test_relative_import_to_dataset(
...
 
     first_level = tmpdir.mkdir('first')
     second_level = first_level.mkdir('second')
 
-    first_data = first_level.join('data.txt')
+    first_data = first_level.join('first.txt')
     first_data.write('first')
 
-    second_data = second_level.join('data.txt')
+    second_data = second_level.join('second.txt')
```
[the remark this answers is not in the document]
mohammad-sdsc: There were 3 files with the same name, `data.txt`, at three different levels; a test was passing incorrectly due to this.

PR #687, opened 2019-09-18
feat: include creator in project metadata

### LINKED ISSUE #659: Project metadata to contain info about project name and creator
As a Renku owner, I should be able to find `schema:name` and `schema:creator` among the project triples like follows:
```
<rdf:Description rdf:about="https://dev.renku.ch/virginiafriedrich/datasets-test">
  <rdf:type rdf:resource="http://www.w3.org/ns/prov#Location"/>
  <rdf:type rdf:resource="http://xmlns.com/foaf/0.1/Project"/>
  <schema:name>namespace/project-name</ns1:name>
  <schema:creator rdf:resource="https://orcid.org/0000-0003-4498-2870"/>
</rdf:Description>
```

rokroskar:
closes #659


renku/api/repository.py
```diff
@@ -132,10 +133,9 @@ def cwl_prefix(self):
         return str(self.workflow_path.resolve().relative_to(self.path))
 
     @cached_property
```
jsam: Since we are not creating new attrs object (no additional initializers), I think there is no need to have it cached. In fact caching it might cause problems if I manually change the attributes on the project (e.g. path).

renku/models/_jsonld.py
```diff
@@ -331,7 +331,10 @@ def from_jsonld(
             raise ValueError(data)
 
         if '@type' in data:
-            type_ = tuple(sorted(data['@type']))
+            type_ = data['@type']
```
jsam: I find this `_jsonld` wrapper quite difficult to read. Could we add a commend what this change does? Just to make it a little bit easier for future to read/maintain this code.

renku/models/_jsonld.py
```diff
@@ -375,8 +381,10 @@ def from_jsonld(
 
         data_ = {}
         if client:
-            data_['client'] = client
-            data_['commit'] = commit
+            if client:
```
jsam: Why `if client` twice?

renku/models/_jsonld.py
```diff
@@ -375,8 +381,10 @@ def from_jsonld(
...
+                data_['client'] = client
+            if commit:
+                data_['commit'] = commit
```
jsam: Maybe we should add a comment in which cases commit could be None? Not sure if we have that comment anywhere, could be useful for a future reference.

renku/models/_jsonld.py
```diff
@@ -399,7 +407,7 @@ def from_yaml(cls, path, client=None, commit=None):
         import yaml
 
         with path.open(mode='r') as fp:
-            source = yaml.safe_load(fp) or {}
+            source = yaml.load(fp, Loader=yaml.BaseLoader) or {}
```
jsam: Why remove safe load? This might be a security issue.

tests/test_projects.py
```diff
@@ -62,17 +95,18 @@ def test_project_serialization():
     assert 'created' in context
 
 
-def test_project_metadata_compatibility():
+@pytest.mark.parametrize("project_meta,version", [(PROJECT_V1, '1'), (PROJECT_V2, '2')])
+def test_project_metadata_compatibility(project_meta, version):
     """Test loading of the initial version."""
-    project = Project.from_jsonld(yaml.safe_load(PROJECT_V1))
-
+    project = Project.from_jsonld(yaml.load(project_meta, Loader=yaml.BaseLoader))
```
jsam:
```suggestion
    project = Project.from_jsonld(yaml.safe_load(project_meta, Loader=yaml.BaseLoader))
```

tests/test_projects.py
```diff
@@ -62,17 +95,18 @@ def test_project_serialization():
     assert 'created' in context
 
 
-def test_project_metadata_compatibility():
+@pytest.mark.parametrize("project_meta,version", [(PROJECT_V1, '1'), (PROJECT_V2, '2')])
+def test_project_metadata_compatibility(project_meta, version):
```
jsam: Maybe we should add a test where we assert for translation of schema.

renku/api/repository.py
rokroskar: ah true, I removed the instantiation of the object here

renku/models/_jsonld.py
rokroskar: done

renku/models/_jsonld.py
rokroskar: because you have to subclass the loader to prevent it from loading datetime strings as datetime objects - using `BaseLoader` loads everything as strings. 

from the docs: 

> BaseLoader(stream) does not resolve or support any tags and construct only basic Python objects: lists, dictionaries and Unicode strings.

tests/test_projects.py
rokroskar: I started to do that but then realized that if the translation doesn't happen, `name` is set to `None` so this test fails.

rokroskar: I agree though that maybe an explicit test would be better - do you have a suggestion?

rokroskar:
@jsam thanks for the review - I tried to add some descriptive comments, please see if they make more sense now.

rokroskar:
ps: I added a commit that modifies the `Pipfile` and `Pipfile.lock`  -- this is to enable e.g. `pipenv run pytest`. I can move it do a different PR if anyone thinks it doesn't belong here.

renku/models/_jsonld.py
```diff
@@ -50,6 +50,35 @@
...
 
+# Shamelessly copy/pasting from SO:
+# https://stackoverflow.com/questions/34667108/ignore-dates-and-times-while-parsing-yaml
+# This is needed to allow us to load from yaml and use json down the line.
+class NoDatesSafeLoader(yaml.SafeLoader):
+    """Used to safely load basic python objects but ignore datetime strings."""
+    @classmethod
+    def remove_implicit_resolver(cls, tag_to_remove):
+        """
+        Remove implicit resolvers for a particular tag
```
jsam: Format is off, I think.

tests/test_projects.py
jsam: https://gist.github.com/jsam/3329f961d8062d6d917eddbb5dc45cd2

Pipfile
```diff
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-renku = {editable = true, path = "."}
+renku = {editable = true,extras = ["all"],path = "."}
```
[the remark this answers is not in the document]
rokroskar: why would it break existing PRs?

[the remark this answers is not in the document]
jsam: Sorry, I couldn't make the commend on the lock file, I'm referring to updating package in the lock file.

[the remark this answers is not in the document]
rokroskar: right - but even that won't matter - tests don't run with `pipenv`.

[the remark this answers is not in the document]
jsam: I use locally pipenv to setup different python environments to check how things work on different python versions.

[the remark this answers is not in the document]
rokroskar: 👍 I removed the commit

PR #693, opened 2019-09-19
feat: dataset tagging

### LINKED ISSUE #566: Dataset tagging
Currently, there is no support for automated versioning of datasets created and managed within renku. A field for versioning is there, but it's only used for import/export features. We should come up with renku way of versioning of datasets so we could automate and build support around versioning of datasets.

This feature would be needed also for building additional support for versioning datasets during export to Zenodo (and other providers).

Panaetius:
adds 3 new commands:
- `renku dataset tag <dataset> <tag> [-d <description>] [--force]` creates a tag
- `renku dataset ls-tags <dataset>` List all tags in a dataset
- `renku dataset rm-tags <dataset> <tags...>` Remove one or more tags from a dataset

Additionally, importing a dataset tags it with the remote version.
Exporting a dataset with tags allows selecting what tag to export and temporarily checks out the repo at that tag for exporting.

The export command does not currently work for dirty directories (as that'd require stashing local changes and popping them later, which I wasn't sure was necessary), but I can add that if wanted.

Closes #566


renku/api/datasets.py
```diff
@@ -449,6 +451,65 @@ def get_relative_url(self, url):
             url = str(url / submodule_url.name)
         return url
 
+    def get_newest_commit_for_dataset(self, dataset):
```
jsam: To make this useful in the future (and possible to track lineage). We should probably build an ordered index which contains all commits in which given dataset was referenced. Ensure that commits are ordered and then simply query for head or tail.

renku/api/datasets.py
```diff
@@ -449,6 +451,65 @@ def get_relative_url(self, url):
...
+        return commit
+
+    def add_dataset_tag(self, dataset, tag, description='', force=False):
+        """Adds a new tag to a dataset.
+
+        Validates if the tag already exists and that the tag follows
+        the same rules as docker tags.
+        See https://docs.docker.com/engine/reference/commandline/tag/
+        for a documentation of docker tag syntax.
+        """
```
jsam: Can we also add ``:raises: ValueError ``? Adding those in docstrings is a good practice we should enforce since it's much easier to reason about the function by just reading the docstring. (Same goes for other places in the PR)

renku/cli/_format/dataset_tags.py
```diff
@@ -0,0 +1,66 @@
...
+
+def tabular(client, tags):
+    """Format dataset tags with a tabular output.
+
+    :param client: LocalClient instance.
+    :param tags: Dataset tags.
+    """
+    from renku.models._tabulate import tabulate
+
+    echo_via_pager(
```
jsam: Since we are splitting things apart (library/click commands). We should split actual output to stdout from string building. If you look the way I refactored on the restructure PR you can see an example. echo_via_pager would be used in click command while formatting function can be used as a library - returning only string.

renku/cli/_format/dataset_tags.py
```diff
@@ -0,0 +1,66 @@
...
+    :param tags: Dataset tags.
+    """
+    from renku.models._json import dumps
+    from renku.models._jsonld import asjsonld
+
+    data = [asjsonld(tag) for tag in tags]
+    echo_via_pager(dumps(data, indent=2))
+
+
+FORMATS = {
```
jsam: Another change which makes sense to alleviate import hell is that we don't use generic FORMATS name across all formatting modules but more concrete approach. So, in the new structure, this would become DATASET_TAGS_FORMATS.

renku/cli/_format/datasets.py
```diff
@@ -37,6 +37,8 @@ def tabular(client, datasets):
                 ('created', None),
                 ('creators_csv', 'creators'),
             )),
+            # workaround for tabulate issue 181
```
jsam: What do you mean by issue #181?

renku/cli/dataset.py
```diff
@@ -490,6 +522,84 @@ def remove(client, names):
...
+@click.option('--force', is_flag=True, help='Allow overwriting existing tags.')
+@pass_local_client(
+    clean=False,
+    commit=True,
+    commit_only=COMMIT_DIFF_STRATEGY,
+)
+@click.pass_context
+def tag(ctx, client, name, tag, description, force):
+    """Create a tag for a dataset."""
+    tag_dataset(client, name, tag, description, force)
```
jsam: Thank you.

renku/cli/dataset.py
```diff
@@ -490,6 +522,84 @@ def remove(client, names):
...
+@click.argument('name')
+@click.argument('tags', nargs=-1)
+@pass_local_client(
+    clean=False,
+    commit=True,
+    commit_only=COMMIT_DIFF_STRATEGY,
+)
+def remove_tags(client, name, tags):
+    """Remove tags from a dataset."""
+    dataset = client.load_dataset(name)
```
jsam: Can we also do this as we did the creation of a tag?

renku/cli/dataset.py
```diff
@@ -490,6 +522,84 @@ def remove(client, names):
...
+    '--format',
+    type=click.Choice(DATASET_TAGS_FORMATS),
+    default='tabular',
+    help='Choose an output format.'
+)
+@pass_local_client(clean=False, commit=False)
+@click.pass_context
+def ls_tags(ctx, client, name, format):
+    """List all tags of a dataset."""
+    dataset_ = client.load_dataset(name)
```
jsam: Let's also split this as you did in creation.

renku/models/datasets.py
```diff
@@ -151,6 +151,56 @@ def creators_csv(self):
         return ','.join(creator.name for creator in self.creator)
 
 
+@jsonld.s(
+    type='schema:PublicationEvent',
+    context={'schema': 'http://schema.org/'},
+    slots=True,
```
jsam: I guess we could freeze it as well then.

renku/cli/dataset.py
```diff
@@ -490,6 +522,84 @@ def remove(client, names):
     click.secho('OK', fg='green')
 
 
+def tag_dataset(client, name, tag, description, force=False):
```
jsam: To fit the new structure better, you can decorate this function with `pass_local_client`, since its a library part. That way click commands are only decorated with click stuff and we keep library contained.

renku/cli/_format/datasets.py
Panaetius: This issue: https://bitbucket.org/astanin/python-tabulate/issues/181/disable_numparse-fails-on-empty-input

I found that yesterday. I originally saw that version was `1` instead of `1.0` for dataverse, and thought tabluate formatting numbers was responsible for it, so I added this option, but the issue was somewhere else.

I left the option in as it makes sense to not treat certain things as number, as this can lead to weird behaviour (e.g. hashes ending in e1 being treated as numbers, which is a case mentioned in the tabulate docs for this feature)

jsam: AHA! Ok, cool tnx! Put the link in the comment as well, so we can keep track of it and make a ticket here on our tracker.

renku/api/datasets.py
```diff
@@ -451,15 +451,23 @@ def get_relative_url(self, url):
...
+
+        if max_results:
+            kwargs['max_count'] = max_results
+
         paths = [(Path(dataset.path) / self.METADATA).resolve()]
 
         paths.extend(f.full_path for f in dataset.files)
 
-        commit = list(self.repo.iter_commits(max_count=1, paths=paths))[0]
+        commits = self.repo.iter_commits(paths=paths, **kwargs)
```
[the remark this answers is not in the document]
Panaetius: Originally I wanted to pass kwargs from `dataset_commits` to `iter_commits` to make it more versatile, but decided against it since kwargs = :nauseated_face: 

I removed it now

PR #698, opened 2019-09-20
[WIP] Adds basic QA checklist

### LINKED ISSUE #697: Define share QA protocol
After making a release it's often a good practice to test the release out. Since we don't have QA team dedicated to do this, it would be good to define a docs which describe the steps needed to test the release. That way we can share responsibility of QA for each release we make.

Panaetius:
Closes #697 

Basic QA checklist to test a new release

### DIFF
--- a/QA_PROTOCOL.rst
+++ b/QA_PROTOCOL.rst
@@ -1,0 +1,54 @@
+..
+    Copyright 2017-2019 - Swiss Data Science Center (SDSC)
+    A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+    Eidgenössische Technische Hochschule Zürich (ETHZ).
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+==============================
+ Renku Python QA Protocol
+==============================
+
+This document describes smoke tests for checking that a new ``renku-python``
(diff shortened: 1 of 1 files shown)


QA_PROTOCOL.rst
```diff
@@ -0,0 +1,54 @@
...
+
+Test new repository
+-------------------
+- [ ] Init renku in a bare repo
+- [ ] Create a dataset and add a file
+- [ ] List datasets and files
+- [ ] Import a dataset from Zenodo and/or Dataverse
+- [ ] List datasets and files
+- [ ] Remove a dataset
+- [ ] ``renku run`` a simple command, e.g. ``cp``
```
rokroskar: maybe add that this should have more than one component? I always do something like this

```
renku run echo "bla" > bla
renku run wc bla > bla.wc
renku run echo "bla2" > bla
renku status # should show that bla.wc is outdated
renku update # should update bla
```

rokroskar: however, isn't this what the tests are doing already anyway? Or in which cases would this manual check be needed even if the tests are passing?

jsam: Instead of a telling to QA explicit commands to execute (and like you pointed out, which tests are doing it anyway), tell them a use case they should replicate/achieve and let them figure out how to do it. That way you always have a human component in breaking things more efficiently. :smile: 

So, an example of a defined case would be: Using renku create a 3 step workflow process which counts the words in a file and writes them into another file. Create a new file and rerun the pipeline to see the new results.

Panaetius: I remember renku failing less than a week ago despite all the tests passing :wink: 

The point is that yes, this is what the tests are doing anyways, but in isolation. Doing everything in a single renku repo with different steps etc. might show issues that single tests don't catch but a human catches immediately.

rokroskar:
I like this - is there a reason it's still a `WIP`?

Panaetius:
> I like this - is there a reason it's still a `WIP`?

Not really any reason, it was basically a draft so I left [WIP] on there.

rokroskar:
I think we should put it in place and use it for the next release today or tomorrow.

PR #699, opened 2019-09-20
build: use "all" for pipenv install

rokroskar:
updates `Pipfile` and `Pipfile.lock` to include packages required for testing.


rokroskar:
probably not impossible, but I will rerun it with `--keep-outdated` - then it won't update any libraries, only add new ones.

rokroskar:
Hmm so apparently this doesn't quite work as expected - the packages still got updated

Panaetius:
> Hmm so apparently this doesn't quite work as expected - the packages still got updated

Apparently `--ignore-pipfile --keep-outdated` would work, just using Pipfile.lock for installing. Alternatively `pipenv sync --dev` should also work.


Have you thought about using [pyup](https://github.com/pyupio/pyup) ? I think it supports pipenv, though there might still be some issues with pipenv support as it's relatively new (Difficult to tell from the docs/issues how mature it is). Automated PRs for dependency updates would be nice, though.

rokroskar:
> Apparently --ignore-pipfile --keep-outdated would work, just using Pipfile.lock for installing. Alternatively pipenv sync --dev should also work.

not sure this is really ideal either -  `--ignore-pipfile` just ends up with an empty `Pipfile.lock` where the only dependency is `.[all]`. I guess I could do 

```
pipenv install --dev .[all]
```

which would leave the non-dev packages untouched.

rokroskar:
well apparently this results in a strange situation where none of the development dependencies actually appear in the `Pipfile.lock`, but they are installed. And it still modifies some of the non-dev packages. I give up.

PR #701, opened 2019-09-20
feat: use UI-resolved project path as project ID

### LINKED ISSUE #554: Standardize/complete the file URIs in the `isPartOf` triple
It looks the file URIs in the `isPartOf` triples do not point to the right resource on Renku. For instance, triples generated on `renku-kuba` environment have URIs pointing to a non-existing resource on Renku DEV and instead of `https://renku-kuba.dev.renku.ch/projects/rokroskar/vom_natt` they point to `https://dev.renku.ch/rokroskar/vom_natt`.

**Original issue description**
I found out that `renku log --format rdf` produces ~triples where the objects of the `isPartOf` triples are comprised of `<GitLabHost>/<projectPath>`. While I'm not aware of why it was done this way, I think it's unnecessary and having only `<projectPath>` should be totally sufficient. The reasons I'm flagging that up are:
* `<projectPath>` defines a project uniquely within a GitLab instance;
* currently, the `<GitLabHost>/<projectPath>` values used in the mentioned triples do not really point to the projects in GitLab, as for instance on the DEV environment, they look like that: `https://dev.renku.ch/rokroskar/vom_natt` so they're lacking the `gitlab` part; 
* there's a certain amount of the maintenance cost associated with keeping the host in there as once we change GitLab hostname, an update query would have to be run on all the triples;
* even if we'd like to do searches across different Renku instances (a federation of Renkus), I'd imagine we'd rather go through an external Renku API, not access the RDF store directly (exposing the store to the outside world does seem to be the best idea from the security point of view);
* because the graph-service needs to find triples for a project, the GitLab hostname has to be defined in the service's config. However, that hostname cannot be used for any REST calls to GitLab as it's simply not valid and would require yet another config value.

So could we drop the hostname from the `isPartOf` triples, please? I believe it will make our life easier.

rokroskar:
closes #554

also, updates the `id` of the serialized project if a remote exists and the previous `id` contains `file://`. 

The resulting json-ld looks like:

```
{
  "@context": {
    "schema": "http://schema.org/",
    "prov": "http://www.w3.org/ns/prov#"
  },
  "@id": "https://renkulab.io/projects/rok.roskar/test-init",
  "@type": [
    "prov:Location",
    "schema:Project"
  ],
  "schema:creator": {
    "@id": "mailto:rok.roskar@sdsc.ethz.ch",
    "@type": "schema:Person",
    "schema:email": "rok.roskar@sdsc.ethz.ch",
    "schema:name": "Rok Roskar"
  },
  "schema:dateCreated": "2019-09-20T11:30:38.639529+00:00",
  "schema:dateUpdated": "2019-09-20T11:30:38.641184+00:00",
  "schema:name": "test-init",
  "schema:schemaVersion": "2"
}
```

### DIFF
--- a/renku/core/models/projects.py
+++ b/renku/core/models/projects.py
@@ -18,6 +18,7 @@
 """Model objects representing projects."""
 
 import datetime
+import os
 
 import attr
 
@@ -25,6 +26,8 @@ from renku.core.models import jsonld
 from renku.core.models.datasets import Creator
 from renku.core.models.datastructures import Collection
 from renku.core.utils.datetime8601 import parse_date
+
+PROJECT_URL_PATH = 'projects'
 
 
 @jsonld.s(
@@ -65,7 +68,14 @@ class Project(object):
 
     client = attr.ib(default=None, kw_only=True)
 
-    creator = jsonld.ib(default=None, kw_only=True, context='schema:creator')
+    creator = jsonld.ib(
+        default=None,
+        kw_only=True,
+        context={
+            '@id': 'schema:creator',
+            '@type': 'schema:Person',
+        },
+    )
 
     _id = jsonld.ib(context='@id', kw_only=True, default=None)
 
@@ -88,8 +98,32 @@ class Project(object):
                 # this assumes the project is being newly created
                 self.creator = Creator.from_git(self.client.repo)
 
-        if not self._id and self.client:
-            self._id = self.client.project_id
+        self._id = self.project_id
+
+    @property
+    def project_id(self):
+        """Return the id for the project based on the repo origin remote."""
+        import pathlib
+        import urllib
+
+        # Determine the hostname for the resource URIs.
+        # If RENKU_DOMAIN is set, it overrides the host from remote.
+        # Default is localhost.
+        host = 'localhost'
+        owner = self.creator.email.split('@')[0] if self.creator else 'NULL'
+        name = self.name
+
+        if self.client:
+            remote = self.client.remote
+            host = self.client.remote.get('host') or host
+            owner = remote.get('owner') or owner
+            name = remote.get('name') or name
+        host = os.environ.get('RENKU_DOMAIN') or host
+        project_url = urllib.parse.urljoin(
+            'https://{host}'.format(host=host),
+            pathlib.posixpath.join(PROJECT_URL_PATH, owner, name or 'NULL')
+        )
+        return project_url
 
 
 class ProjectCollection(Collection):
--- a/renku/core/management/repository.py
+++ b/renku/core/management/repository.py
@@ -31,6 +31,7 @@ from werkzeug.utils import cached_property, secure_filename
 
 from renku.core.compat import Path
 from renku.core.management.config import RENKU_HOME
+from renku.core.models.git import GitURL
 from renku.core.models.projects import Project
 from renku.core.models.refs import LinkReference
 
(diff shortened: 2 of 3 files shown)


renku/core/models/projects.py
```diff
@@ -88,8 +98,52 @@ def __attrs_post_init__(self):
...
-            self._id = self.client.project_id
+        if not self._id or self._id.startswith('file://'):
+            self._id = self.project_id
+
+    @property
+    def project_id(self):
+        """Return the id for the project based on the repo origin remote."""
+        from renku.core.models.git import GitURL
+
+        # RENKU_DOMAIN environment variable is the host override
```
jsam: I think we should create and .env-example file as an example .env file where we keep all possible configuration variables. That way we have a simple and easier way on how renku can be tweaked. Of course, in the long run this should be all documented properly.

renku/core/models/projects.py
```diff
@@ -88,8 +98,52 @@ def __attrs_post_init__(self):
...
+            except IndexError:
+                url = None
+
+            if url:
+                # set the host if not set from RENKU_DOMAIN above
+                host = host or url.hostname or 'localhost'
+                owner = url.owner or owner
+                name = url.name or name
+
+        return 'https://{host}/{project_url_path}/{owner}/{name}'.format(
```
jsam: Can we build URL's in the same way as we do in [here](https://github.com/SwissDataScienceCenter/renku-python/blob/master/renku/core/commands/providers/zenodo.py#L49).

renku/core/models/projects.py
```diff
@@ -88,8 +98,52 @@ def __attrs_post_init__(self):
...
+                pass
+
+            try:
+                url = GitURL.parse(self.client.repo.remotes[remote_name].url)
+
+                # Remove gitlab. unless running on gitlab.com.
+                hostname_parts = url.hostname.split('.')
+                if len(hostname_parts) > 2 and hostname_parts[0] == 'gitlab':
+                    hostname_parts = hostname_parts[1:]
+                url = attr.evolve(url, hostname='.'.join(hostname_parts))
```
jsam: I think we can do a little bit better here. Improving a comment by explaining why and adding a test case would be a great addition to this. 

I would propose that we extract it to its own function and put proper docstring explaining what and why. That way swapping/editing this part for additional cases in the future becomes much easier.

jachro:
Hi Rok. I tested the change and it works beautifully :) However, could we put merging of this PR on hold until the corresponding issue on renku-graph (https://github.com/SwissDataScienceCenter/renku-graph/issues/134) is about to be merged? Otherwise, I'd have to add yet another change to renku-graph as part of the dataset PR.

renku/core/models/projects.py
jsam: Usually, .env is in package root so I'm okay with having .env-example in project root as well.

rokroskar:
@jachro sure, I think that should be possible.

renku/core/models/projects.py
rokroskar: ok this should then be done as part of the templating work and project-level settings - see also #730

renku/core/models/projects.py
rokroskar: done

renku/core/models/projects.py
rokroskar: actually I did that already :)

rokroskar: oh - did you mean just the gitlab part? I moved the whole piece to `core/management/repository.py`

rokroskar:
This should be merged after #742

PR #707, opened 2019-09-25
fix: force current project for entities

rokroskar:
@vfried and @jachro reported a problem that datasets were not showing up in queries for forked projects. This PR forces the project to be set to the current project even if this is not what is recorded in the metadata. This might seem strange, but:

- the information that the dataset/file is already a part of the parent project is already recorded in the KG
- the use is unlikely to scrutinize or use this part of the dataset metadata since it is internal to renku


jachro:
I've tested the change and it works like a charm!

rokroskar:
ahem... looks like they found some problems with this one still :/

Issue #681, opened 2019-09-18
Add data to a dataset without submodules


rokroskar:
Import data without adding the source repository as a submodule. 

If the source is a renku project, import the `creator`, and `createdAt` properties. Additionally create links to the source KG node by using a `sameAs` triple e.g.

```<current-node> rdf:sameAs <source-node>```

The `<source-node>` should be the `_id` of the file in the source project and if it doesn't already exist (if, for example, the file is not a part of a `Dataset`) then it can be generated as

```blob/<commit-sha>/<path-in-source-project>```

We should perhaps optionally also link to the source project by adding it to the list of projects that a file belongs to.

Issue #670, opened 2019-09-11
Improve path handling in `dataset add`


rokroskar:
Currently the `renku dataset add` command has two confusing flags: `--target` and `--relative-to`. These are used when importing from a git repository - `--target` specifies which path(s) inside the repository should be imported and `--relative-to` specifies how the paths should be trimmed. We should simplify this by providing e.g. 

```
renku dataset add mydataset <repo-URL> \
    --src /path/in/the/source/repo --dst /path/in/the/current/repo
```

if `--dst` is omitted, the default should be `data/<dataset-name>/filename`. If a directory is specified as `--src` then the default is `data/<dataset-name>/directory`. 

Question: should `--src` be required? If it's omitted, should the entire repository be imported? 

---

related to https://github.com/SwissDataScienceCenter/renku/issues/593

PR #710, opened 2019-09-26
681 add to dataset without submodules

### LINKED ISSUE #214: clean up import from git repos
There seem to be many problem with imports from git repositories. 

Some points to improve on: 

* how should a specific file from another repository be added?
* Including a specific file works with e.g. the `--target` option or `<git-repo-url>/<path>` but importing a directory does not work
* ~~--relative-to` is broken -- injects the project name where it shouldn't~~ (#308)

mohammad-sdsc:

# Description

This PR removes the need to use submodules for importing datasets from Git repos in Renku. It stores the source of each file added to a dataset by using a `schema:isBasedOn` (https://schema.org/isBasedOnUrl) field to point to the remote file.

It also changes the command line interface: The `--target` and `--relative-to` options are removed from `dataset add` commands. Now, one can use `--source` and `--destination` options to say which files are copied from a remote repo and to which location.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/681
Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/670
closes #214 

## Type of change

## Checklist

### DIFF
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -65,31 +65,34 @@ URL schemes. For example,
 
     $ renku dataset add my-dataset git+ssh://host.io/namespace/project.git
 
-Sometimes you want to import just a specific path within the parent project.
-In this case, use the ``--target`` flag:
+Sometimes you want to import just specific paths within the parent project.
+In this case, use the ``--source`` or ``-s`` flag:
 
 .. code-block:: console
 
-    $ renku dataset add my-dataset --target relative-path/datafile \
+    $ renku dataset add my-dataset --source path/within/repo/to/datafile \
         git+ssh://host.io/namespace/project.git
 
-To trim part of the path from the parent directory, use the ``--relative-to``
-option. For example, the command above will result in a structure like
+The command above will result in a structure like
 
 .. code-block:: console
 
     data/
       my-dataset/
-        relative-path/
-          datafile
+        datafile
 
-Using instead
+You can use ``--destination`` or ``-d`` flag to change the name of the target
+file or directory. The semantics here are similar to the POSIX copy command:
+if the destination does not exist or if it is a file then the source will be
+renamed; if the destination exists and is a directory the source will be copied
+to it. You will get an error message if you try to move a directory to a file
+or copy multiple files into one.
 
 .. code-block:: console
 
     $ renku dataset add my-dataset \
-        --target relative-path/datafile \
-        --relative-to relative-path \
+        --source path/within/repo/to/datafile \
+        --destination new-dir/new-filename \
         git+ssh://host.io/namespace/project.git
 
 will yield:
@@ -98,7 +101,8 @@ will yield:
 
     data/
       my-dataset/
-        datafile
+        new-dir/
+          new-filename
 
 Tagging a dataset:
 
@@ -380,18 +384,27 @@ def edit(dataset_id):
 @click.argument('name')
 @click.argument('urls', nargs=-1)
 @click.option('--link', is_flag=True, help='Creates a hard link.')
-@click.option('--relative-to', default=None)
-@click.option(
-    '-t',
-    '--target',
-    default=None,
-    multiple=True,
-    help='Target path in the git repo.'
-)
 @click.option(
     '--force', is_flag=True, help='Allow adding otherwise ignored files.'
 )
-def add(name, urls, link, relative_to, target, force):
+@click.option(
+    '-s',
+    '--src',
+    '--source',
+    'sources',
+    default=None,
+    multiple=True,
+    help='Path(s) within remote git repo to be added'
+)
+@click.option(
+    '-d',
+    '--dst',
+    '--destination',
+    'destination',
+    default='',
+    help='Destination file or directory within the dataset path'
+)
+def add(name, urls, link, force, sources, destination):
     """Add data to a dataset."""
     progress = partial(progressbar, label='Adding data to dataset')
     add_file(
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -408,7 +408,13 @@ def add(name, urls, link, force, sources, destination):
     """Add data to a dataset."""
     progress = partial(progressbar, label='Adding data to dataset')
     add_file(
-        urls, name, link, force, relative_to, target, urlscontext=progress
+        urls=urls,
+        name=name,
+        link=link,
+        force=force,
+        sources=sources,
+        destination=destination,
+        urlscontext=progress
     )
 
 
--- a/renku/core/commands/checks/migration.py
+++ b/renku/core/commands/checks/migration.py
@@ -31,7 +31,10 @@ from ..echo import WARNING
 
 def dataset_pre_0_3(client):
     """Return paths of dataset metadata for pre 0.3.4."""
-    return (client.path / 'data').rglob(client.METADATA)
+    project_is_pre_0_3 = int(client.project.version) < 2
+    if project_is_pre_0_3:
+        return (client.path / 'data').rglob(client.METADATA)
+    return []
 
 
 def check_dataset_metadata(client):
--- a/renku/core/management/datasets.py
+++ b/renku/core/management/datasets.py
@@ -21,11 +21,13 @@ import os
 import re
 import shutil
 import stat
+import tempfile
 import uuid
 import warnings
 from configparser import NoSectionError
 from contextlib import contextmanager
 from pathlib import Path
+from subprocess import PIPE, SubprocessError, run
 from urllib import error, parse
 
 import attr
@@ -34,7 +36,7 @@ import requests
 from renku.core import errors
 from renku.core.management.config import RENKU_HOME
 from renku.core.models.datasets import Creator, Dataset, DatasetFile, \
-    DatasetTag, NoneType
+    DatasetTag
 from renku.core.models.git import GitURL
 from renku.core.models.locals import with_reference
 from renku.core.models.refs import LinkReference
@@ -139,7 +141,14 @@ class DatasetsApiMixin(object):
         dataset.to_yaml()
 
     def add_data_to_dataset(
-        self, dataset, urls, git=False, force=False, **kwargs
+        self,
+        dataset,
+        urls,
+        git=False,
+        force=False,
+        sources=(),
+        destination='',
+        link=False
     ):
         """Import the data into the data directory."""
         dataset_path = self.path / self.datadir / dataset.name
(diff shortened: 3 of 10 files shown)


renku/api/datasets.py
```diff
@@ -282,141 +284,166 @@ def _add_from_url(self, dataset, path, url, link=False, **kwargs):
...
+                    raise errors.UsageError(
+                        'Cannot use --source within local repo subdirectories'
+                    )
+                source = Path(u.path).resolve().relative_to(repo_path)
+                sources = (source, )
+                url = repo_path.as_posix()
+        elif u.scheme in {'http', 'https', 'git+https', 'git+ssh'} or \
+                url.startswith('git@'):
+            repo_path = Path(tempfile.mkdtemp())
+            repo = Repo.clone_from(url, repo_path, recursive=True)
```
rokroskar: what happens with LFS objects here? We may want to use the `GIT_LFS_SKIP_SMUDGE=1` environment variable to prevent all the data being pulled and then only fetch the needed paths.

renku/cli/dataset.py
```diff
@@ -65,31 +65,34 @@
...
-          datafile
+        datafile
 
-Using instead
+You can use ``--destination`` or ``-d`` flag to change the name of the target
+file or directory. The semantics here are similar to the POSIX copy command:
+if the destination does not exist or if it is a file then the source will be
+renamed; if the destination exists and is a directory the source will be copied
+to it. You will get an error message if you try to move a directory to a file
+or copy multiple files into one.
```
rokroskar: what about specifying a directory that doesn't exist as a destination? Is it created or you get an error?

rokroskar:
I came across a funny bug when I tried out this PR. I tried this:

```
renku dataset add mydata https://renkulab.io/gitlab/rok.roskar/renku-env.git --destination env
Error: The following paths are ignored by one of your .gitignore files:

        data/mydata/env/bashrc
Please use the "--force" option if you really want to add them.
```

However:

```
tree
.
├── Dockerfile
├── data
│   └── mydata
│       └── env
│           └── bashrc
└── requirements.txt

3 directories, 3 files
```

So the files get copied first and then they are left there. Because they are ignored it's not easy to find them (or to get rid of them). You can use `git clean -dfX` but that has the downside of also deleting any other ignored files. If the command fails in this way, maybe we should warn the user and provide the relevant git command?

renku/core/commands/checks/migration.py
```diff
@@ -31,7 +31,10 @@
 
 def dataset_pre_0_3(client):
     """Return paths of dataset metadata for pre 0.3.4."""
-    return (client.path / 'data').rglob(client.METADATA)
+    project_is_pre_0_3 = int(client.project.version) < 2
```
jsam: @ciyer @mohammad-sdsc Small head's up here, just to avoid stepping on each other's toes. Since both of you are editing this, I think we should have both fixtures to ensure that everything works in the end.

renku/core/models/datasets.py
```diff
@@ -262,6 +262,8 @@ class DatasetFile(Entity, CreatorsMixin):
 
     url = jsonld.ib(default=None, context='schema:url', kw_only=True)
 
+    source = jsonld.ib(default=None, context='schema:isBasedOn', kw_only=True)
```
jsam: We should probably add a small comment to help differentiate and avoid possible future confusions with `__source__`.

rokroskar:
This is looking pretty good - I was just checking the metadata - for the case with my import from the comment above I get:

```
source:
    '@context':
      _id: '@id'
      _label: rdfs:label
      _project: schema:isPartOf
      added: schema:dateCreated
      creator: schema:creator
      dataset: schema:isPartOf
      name: schema:name
      path: prov:atLocation
      schema: http://schema.org/
      source: schema:isBasedOn
      url: schema:url
      wfprov: http://purl.org/wf4ever/wfprov#
    '@type':
    - prov:Entity
    - schema:DigitalDocument
    - wfprov:Artifact
    _id: blob/afc0b09b6474ef72a55058ceba626c482f1f4f6e/bashrc
    _label: bashrc@afc0b09b6474ef72a55058ceba626c482f1f4f6e
    _project: null
    added: '2019-10-02T12:19:56.601001+00:00'
    creator: []
    dataset: null
    name: bashrc
    path: bashrc
    source: null
    url: https://renkulab.io/gitlab/rok.roskar/renku-env.git
  url: https://renkulab.io/gitlab/rok.roskar/renku-env.git/bashrc
```

Two things here - first, `creator` is empty - we can/should get this from the original repo.  Second, what is the second `url` doing there?

renku/core/management/datasets.py
```diff
@@ -207,16 +212,15 @@ def _add_from_url(self, dataset, path, url, link=False, **kwargs):
...
-        else:
-            dst_path = os.path.basename(u.path)
+        if destination:
+            destination = self._resolve_paths(dataset_path,
+                                              [destination]).pop()
+
+        dst = self.path / dataset_path / destination
 
-        dst = path.joinpath(dst_path).absolute()
+        # FIXME raise if destination is a file and source is a directory
```
jsam: Can we open a ticket for this and put the ticket number in the comment?

renku/core/management/datasets.py
```diff
@@ -274,141 +278,195 @@ def _add_from_url(self, dataset, path, url, link=False, **kwargs):
...
+                    'path': dst.relative_to(self.path),
+                    'url': dst_url,
+                    'creator': creators,
+                    'dataset': dataset.name,
+                    'parent': self,
+                    'source': base
+                })
+
+                dst.parent.mkdir(parents=True, exist_ok=True)
+                shutil.copy2(str(src), str(dst))
```
jsam: :+1:  for using copy2 but that might be a problem due to copystat. To explain a bit why - I've seen people using renku on mounted filesystems and since copy2 will try to copy metadata through copystat accessing permission bits can cause this function to fail. In other places, for this reason, we just use copy instead of copy2.

renku/cli/dataset.py
mohammad-sdsc: It is created like `mkdir -p`

rokroskar: I tried with `--destination directory/` and if put the source into a file called `data/mydataset/directory` -- so I'm not sure how you can specify a directory as the destination?

rokroskar: a directory that doesn't exist, I should say

renku/core/management/datasets.py
mohammad-sdsc: Actually it was easier to fix this.

renku/core/models/datasets.py
mohammad-sdsc: I've changed this to `basedOn`

mohammad-sdsc:
> I came across a funny bug when I tried out this PR. I tried this:
> 
> ```
> renku dataset add mydata https://renkulab.io/gitlab/rok.roskar/renku-env.git --destination env
> Error: The following paths are ignored by one of your .gitignore files:
> 
>         data/mydata/env/bashrc
> Please use the "--force" option if you really want to add them.
> ```
> 
> However:
> 
> ```
> tree
> .
> ├── Dockerfile
> ├── data
> │   └── mydata
> │       └── env
> │           └── bashrc
> └── requirements.txt
> 
> 3 directories, 3 files
> ```
> 
> So the files get copied first and then they are left there. Because they are ignored it's not easy to find them (or to get rid of them). You can use `git clean -dfX` but that has the downside of also deleting any other ignored files. If the command fails in this way, maybe we should warn the user and provide the relevant git command?

Fixing this requires some larger refactoring in the dataset code and I didn't want to do that in this PR. Since files are ignored, the git repo should look clean and we can re-execute the command.
This will be addressed in the refactoring of this code that you and I talked about a while back.

mohammad-sdsc:
> This is looking pretty good - I was just checking the metadata - for the case with my import from the comment above I get:
> 
> ```yaml
> source:
>     '@context':
>       _id: '@id'
>       _label: rdfs:label
>       _project: schema:isPartOf
>       added: schema:dateCreated
>       creator: schema:creator
>       dataset: schema:isPartOf
>       name: schema:name
>       path: prov:atLocation
>       prov: http://www.w3.org/ns/prov#
>       schema: http://schema.org/
>       source: schema:isBasedOn
>       url: schema:url
>       wfprov: http://purl.org/wf4ever/wfprov#
>     '@type':
>     - prov:Entity
>     - schema:DigitalDocument
>     - wfprov:Artifact
>     _id: blob/afc0b09b6474ef72a55058ceba626c482f1f4f6e/bashrc
>     _label: bashrc@afc0b09b6474ef72a55058ceba626c482f1f4f6e
>     _project: null
>     added: '2019-10-02T12:19:56.601001+00:00'
>     creator: []
>     dataset: null
>     name: bashrc
>     path: bashrc
>     source: null
>     url: https://renkulab.io/gitlab/rok.roskar/renku-env.git
>   url: https://renkulab.io/gitlab/rok.roskar/renku-env.git/bashrc
> ```
> 
> Two things here - first, `creator` is empty - we can/should get this from the original repo. Second, what is the second `url` doing there?

The `creator` will be exactly the same as the creator of the added file. I did not want to repeat the same information, although it can be easily added if you think it's beneficial. 

The second url should be for the added file. I guess the indentation in your comment is not correct and `source` and the second `url` are in the same level and belong to the added file. I do not see a second url in my tests.

renku/cli/dataset.py
mohammad-sdsc: Well, actually if the destination is just one directory then a rename happens. If it is multi-level directory then all directories except the last one are created and then a rename to the last name happens:
`if the destination does not exist **OR** if it is a file then ...`.

It actually works like a copy command with the exception that in copy you cannot pass multi-level non-existing directories as destination bu here you can (and of course you don't have globs here).

renku/core/models/datasets.py
```diff
@@ -262,6 +262,10 @@ class DatasetFile(Entity, CreatorsMixin):
 
     url = jsonld.ib(default=None, context='schema:url', kw_only=True)
 
+    based_on = jsonld.ib(
+        default=None, context='schema:isBasedOn', kw_only=True
```
rokroskar: is there a reason you chose `isBasedOn` rather than `sameAs`? I guess you are assuming the user could modify the imported file?

rokroskar:
Specifying a non-existent path with `--source` should give an error? atm it passes silently and makes an empty commit.

renku/core/models/datasets.py
mohammad-sdsc: Yes. That was one of the reasons. The other reason is that `sameAs` can only be a URL while `isBasedOn` can also be a `CreativeWork`.

rokroskar: we will have stable URIs for all file and dataset objects soon - we might need to revisit this choice then.

PR #715, opened 2019-09-26
fix: use commit author for project creator

### LINKED ISSUE #713: `renku log` returns the wrong creator of Project
**Describe the bug**
`renku log` erroneously yields the current `git` user as the `Project` `Creator`  instead of yielding the user who initiated the project. 

Additionally this causes `renku log` to fail in cases where the git user is not configured.

rokroskar:
The PR corrects the wrong behavior that the user executing the git command was being set as the project creator. Instead, if the project creator does not exist, the git history of `.renku/metadata.yml` is searched for the first commit and its author is used as the project creator. 

closes #713

### DIFF
--- a/renku/core/management/repository.py
+++ b/renku/core/management/repository.py
@@ -217,8 +217,13 @@ class RepositoryApiMixin(GitCore):
         """Check if the path is a valid CWL file."""
         return path.startswith(self.cwl_prefix) and path.endswith('.cwl')
 
-    def find_previous_commit(self, paths, revision='HEAD'):
-        """Return a previous commit for a given path."""
+    def find_previous_commit(self, paths, revision='HEAD', return_first=False):
+        """Return a previous commit for a given path starting from ``revision``.
+
+        :param revision: revision to start from, defaults to ``HEAD``
+        :param return_first: show the first commit in the history
+        :raises KeyError: if path is not present in the given commit
+        """
         file_commits = list(self.repo.iter_commits(revision, paths=paths))
 
         if not file_commits:
@@ -228,7 +233,7 @@ class RepositoryApiMixin(GitCore):
                 )
             )
 
-        return file_commits[0]
+        return file_commits[-1 if return_first else 0]
 
     @cached_property
     def workflow_names(self):
--- a/renku/core/models/datasets.py
+++ b/renku/core/models/datasets.py
@@ -340,7 +340,6 @@ def _convert_keyword(keywords):
 @jsonld.s(
     type='schema:Dataset',
     context={
-        'added': 'schema:dateCreated',
         'affiliation': 'schema:affiliation',
         'alternate_name': 'schema:alternateName',
         'email': 'schema:email',
@@ -570,9 +569,10 @@ class Dataset(Entity, CreatorsMixin):
                     dataset_file.client = client
 
         try:
-            self.commit = self.client.find_previous_commit(
-                self.path, revision=self.commit or 'HEAD'
-            )
+            if self.client:
+                self.commit = self.client.find_previous_commit(
+                    self.path, revision=self.commit or 'HEAD'
+                )
         except KeyError:
             # if with_dataset is used, the dataset is not committed yet
             pass
--- a/renku/core/models/jsonld.py
+++ b/renku/core/models/jsonld.py
@@ -401,7 +401,8 @@ class JSONLDMixin(ReferenceMixin):
 
         for migration in set(migrations):
             data = migration(data)
-            __source__ = migration(__source__)
+            if __source__:
+                __source__ = migration(__source__)
 
         if data['@context'] != cls._jsonld_context:
             try:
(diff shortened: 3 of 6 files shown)


renku/models/datasets.py
```diff
@@ -340,7 +340,6 @@ def _convert_keyword(keywords):
 @jsonld.s(
     type='schema:Dataset',
     context={
-        'added': 'schema:dateCreated',
```
mohammad-sdsc: Why is this removed?

tests/test_dataset.py
```diff
@@ -25,7 +25,142 @@
...
+identifier: a4034ad6-dcf5-4e84-be4f-1a4a1440c800
+in_language: null
+keywords: []
+license: null
+name: newdata
+path: .renku/datasets/a4034ad6-dcf5-4e84-be4f-1a4a1440c800
+tags: []
+url: null
+version: null
+"""
```
mohammad-sdsc: I'd rather put this in a separate file.

tests/test_projects.py
```diff
@@ -153,3 +153,33 @@ def test_project_datetime_loading(project_meta):
...
+    project = client.project
+    project.creator = None
+    project.to_yaml()
+    client.repo.git.commit(
+        '-a', '--amend', '-C', 'HEAD', '--author',
+        'Johnny Doe <johndoe123456@example.com>', '--no-verify'
+    )
+    # now the creator should be the one from the commit
+    project = Project.from_yaml(client.renku_metadata_path, client=client)
+    assert project.creator.email == 'johndoe123456@example.com'
```
mohammad-sdsc: I'd suggest to use a different email address to make the change more clear.

renku/api/repository.py
```diff
@@ -216,8 +216,12 @@ def is_cwl(self, path):
         """Check if the path is a valid CWL file."""
         return path.startswith(self.cwl_prefix) and path.endswith('.cwl')
 
-    def find_previous_commit(self, paths, revision='HEAD'):
-        """Return a previous commit for a given path."""
+    def find_previous_commit(self, paths, revision='HEAD', first=False):
+        """Return a previous commit for a given path starting from ``revision``.
+
+        :param revision: revision to start from, defaults to ``HEAD``
+        :param first: instead of last, show the first commit in the history
```
jsam: If we are doing this change then let's add the most important part of it: exceptions being raised =))

renku/api/repository.py
```diff
@@ -216,8 +216,12 @@ def is_cwl(self, path):
         """Check if the path is a valid CWL file."""
         return path.startswith(self.cwl_prefix) and path.endswith('.cwl')
 
-    def find_previous_commit(self, paths, revision='HEAD'):
-        """Return a previous commit for a given path."""
+    def find_previous_commit(self, paths, revision='HEAD', first=False):
```
jsam: Can we rename this param to more explicit description? I propose `return_head` or `return_first`. That way its easier to reason while reading the usage of the function without going to function docs.

renku/api/repository.py
```diff
@@ -227,7 +231,7 @@ def find_previous_commit(self, paths, revision='HEAD'):
                 )
             )
 
-        return file_commits[0]
+        return file_commits[-1 if first else 0]
```
jsam: This will explode if file_commits is []. We should add a sentinel and raise proper renku exception in this case.

tests/test_dataset.py
jsam: We could create it as a fixture and put it into the fixture directory. If we decide to go this path I propose we create a list of all such testing fixtures (I remember there are project ones as well) and knock them out there.

renku/api/repository.py
rokroskar: it's raised a few lines above - the code only gets here if `file_commits` is not empty

renku/models/datasets.py
rokroskar: because having two mappings to `schema:dateCreated` is not possible and I kept the one coming from `Dataset`. We should find another solution for this value in `DatasetFile` but I don't know at the moment what the correct term would be.

tests/test_dataset.py
rokroskar: I actually decided to just generate the metadata - since the metadata is currently in flux I think this makes sense for the time being.

renku/api/repository.py
rokroskar: done

renku/models/datasets.py
mohammad-sdsc: Just keep in mind that if triples are not generated for `dateCreated` then it's probably the bug that I saw the other day about `isBasedOn`. You need to re-add this here in that case.

rokroskar: ah no, they were being created, that was the problem - the ones from the `Dataset` itself were not, however, because in the `context` you can apparently only map a term once - in this case `dateCreated` was mapped to `created` and `added.

rokroskar: but you're right - if at a later point we actually want to export the `added` information, it needs to be added above 👍

Issue #684, opened 2019-09-18
bug: when renku dataset create fails, metadata still created


emmjab:
**Describe the bug**
When `renku dataset create <name of dataset>` fails, there is still metadata created in `.renku` (but not a new directory under `data/`) which causes the subsequent attempt at creating a dataset with the same name to fail "mysteriously".

**To Reproduce**
1. `renku dataset create mydataset` <-- with some stuff that fails; you could try un-configuring your git config name and email, for instance.
2. `renku dataset create mydataset`

**Expected behavior**
Instead, it should be possible to create the dataset with the same name since the first time failed.

emmjab:
Just to add to this -- if a `renku dataset add` fails (can test by trying to add data from a fake path), and you accidentally wrote the wrong name for the dataset, a `data/<dataset name>` folder gets created and metadata get added for that new dataset name. I think the folder & metadata should be removed (or never added) if no data comes through.

mohammad-sdsc:
The issue with accidental creation of a dataset will be addressed in https://github.com/SwissDataScienceCenter/renku-python/issues/688 where we add an explicit command line argument for creation of non-existing datasets.

Issue #720, opened 2019-10-01
Decouple Renku from Click exceptions


jsam:
Currently, most of Renku exceptions are coupled with Click exceptions. To separate dependencies of renku core and click we need to first separate those exceptions.

Panaetius:
This applies to all `renku/core/commands/*` files

jsam:
The first step has been made in #690, the remaining work has been broken down into following tickets #703, #702, #720, #722.

Panaetius:
I already removed the respective exceptions in import/export in #702 so it'd be a good idea to wait for this until #702 is done

PR #729, opened 2019-10-02
fix: clean up dataset references when create or add fails

mohammad-sdsc:

# Description

When creating a dataset (either directly or via add command) fails, we cannot repeat the command by some weird error message. This is because symbolic references to dataset metadata is created and not removed upon an error. This PR fixes this issue by cleaning up after a failure.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/684

## Type of change

## Checklist

### DIFF
--- a/renku/core/management/datasets.py
+++ b/renku/core/management/datasets.py
@@ -110,11 +110,19 @@ class DatasetsApiMixin(object):
     def with_dataset(self, name=None):
         """Yield an editable metadata object for a dataset."""
         dataset = self.load_dataset(name=name)
+        clean_up_required = False
 
         if dataset is None:
+            clean_up_required = True
+            dataset_ref = None
             identifier = str(uuid.uuid4())
             path = (self.renku_datasets_path / identifier / self.METADATA)
-            path.parent.mkdir(parents=True, exist_ok=True)
+            try:
+                path.parent.mkdir(parents=True, exist_ok=False)
+            except FileExistsError:
+                raise errors.DatasetExistsError(
+                    'Dataset with reference {} exists'.format(path.parent)
+                )
 
             with with_reference(path):
                 dataset = Dataset(
@@ -122,13 +130,23 @@ class DatasetsApiMixin(object):
                 )
 
             if name:
-                LinkReference.create(client=self, name='datasets/' +
-                                     name).set_reference(path)
+                dataset_ref = LinkReference.create(
+                    client=self, name='datasets/' + name
+                )
+                dataset_ref.set_reference(path)
 
         dataset_path = self.path / self.datadir / dataset.name
         dataset_path.mkdir(parents=True, exist_ok=True)
 
-        yield dataset
+        try:
+            yield dataset
+        except Exception:
+            # TODO use a general clean-up strategy
+            # https://github.com/SwissDataScienceCenter/renku-python/issues/736
+            if clean_up_required and dataset_ref:
+                dataset_ref.delete()
+                shutil.rmtree(path.parent, ignore_errors=True)
+            raise
 
         # TODO
         # if path is None:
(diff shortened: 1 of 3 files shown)


tests/cli/test_datasets.py
```diff
@@ -1147,3 +1147,17 @@ def test_dataset_rm_tags_failure(tmpdir, runner, project, client):
...
+    """Test project is cleaned when dataset add fails for a new dataset."""
+    # add a non-existing path to a new dataset
+    result = runner.invoke(
+        cli,
+        ['dataset', 'add', 'new-dataset', 'non-existing-file'],
+        catch_exceptions=True,
+    )
+
+    assert result.exit_code == 1
+    assert not (client.renku_path / 'refs' / 'datasets' /
```
rokroskar: wouldn't `exists` be enough here?

mohammad-sdsc: The test wasn't failing with `exists`. It came out that for symlinks `exists` checks if the target exists or not:
```
Note If the path points to a symlink, exists() returns whether the symlink points to an existing file or directory.
```

rokroskar: but shouldn't the symlink be removed?

renku/core/management/datasets.py
```diff
@@ -110,25 +110,36 @@ def load_dataset(self, name=None):
     def with_dataset(self, name=None):
         """Yield an editable metadata object for a dataset."""
         dataset = self.load_dataset(name=name)
+        clean_up_required = False
```
jsam: Since there are several different possible cleanups that we could do in case of failure, I wouldn't do it here. Broken references are just one case in which we have to clean up, but I've seen others that could happen (e.g. Uncleaned lock files, partially created metadata to name a few). What I'm thinking is we should come up with a general enough solution which can handle those problems which can happen along the chain. Maybe something similar like migrations?

renku/core/management/datasets.py
```diff
@@ -110,25 +110,36 @@ def load_dataset(self, name=None):
...
 
         if dataset is None:
+            clean_up_required = True
+            dataset_ref = None
             identifier = str(uuid.uuid4())
             path = (self.renku_datasets_path / identifier / self.METADATA)
-            path.parent.mkdir(parents=True, exist_ok=True)
+            path.parent.mkdir(parents=True, exist_ok=False)
```
jsam: This will produce a non-click exception. Do we have a handler for it?

tests/cli/test_datasets.py
mohammad-sdsc: It is removed with this fix. I could not write a failing test for this using `exists`. When using `exists` the test will always pass either because the symlink is removed (this fix) or because the symlink is there but its target is not there (without this fix).

I can see that this might be a bit confusing, I'll change the logic to:
```
assert not ref.is_symlink() and not ref.exists()
```

renku/core/management/datasets.py
mohammad-sdsc: Let's discuss this tomorrow in Bern.

mohammad-sdsc: As we discussed, I've created https://github.com/SwissDataScienceCenter/renku-python/issues/736.

renku/core/management/datasets.py
mohammad-sdsc: Will be addressed in https://github.com/SwissDataScienceCenter/renku-python/issues/720.

renku/core/management/datasets.py
mohammad-sdsc: I'll add a TODO for it.

jsam: Let's just add the ticket number to code with a comment so we know in the future.

jsam: Let's add that ticket as a comment here as well so we know that we need to cleanup with that ticket.

PR #742, opened 2019-10-04
feat: use URIs for dataset ids

### LINKED ISSUE #661: `renku log` to generate dataset resource pointing the Renku dataset resource
As a Renku owner, I'd like `renku log` to generate project's resources triples pointing to the new `GET /knowledge-graph/datasets/:id` endpoint.

### LINKED ISSUE #719: renku log to generate single identifier for dataset imports of the same resource
As a Renku owner, I'd like to have a single dataset `resource` and `identifier` when importing the same dataset to different projects.

Acceptance criteria:
* When a dataset is manually created in Renku (a non-imported dataset case) then:
`@id` = `https://renku/datasets/<identifier>`
`schema: identifier` = `identifier` (a generated UUID);
* When a dataset is imported from Zenodo then:
`@id` = `https://renku/datasets/<identifier>` where `identifier` is url encoded
`schema: identifier` = the original dataset identifier as on the imported dataset (DOI)
* When a dataset is imported from Dataverse then:
`@id` = `https://renku/datasets/<identifier>` where `identifier` is url encoded
`schema: identifier` = the original dataset identifier as on the imported dataset (DOI)

Also, the `dateCreated` should not but put into triples generated by `renku log` command as it will be multiplied by a number of projects using a particular dataset.

rokroskar:
This PR tries to solve the problem of poorly-formed dataset `id`s. Specifically: 

* when a doi is available, use it as the identifier; we use just the doi itself e.g. `10.5281/zenodo.3251128` and not the full URI e.g. `https://doi.org/10.5281/zenodo.3251128`
* the id is always a URI that uses the `identifier`; if the `identifier` is a doi, it is URL-escaped
* older metadata (`id=doi` and `identifier=uuid`) is automatically migrated
* adds a `sameAs` triple pointing to the original doi

Example:

```json
{
  "@context": "http://schema.org",
  "id": "https://dev.renku.ch/datasets/10.5281%2Fzenodo.3247009",
  "type": [
    "http://www.w3.org/ns/prov#Entity",
    "Dataset",
    "http://purl.org/wf4ever/wfprov#Artifact"
  ],
  "creator": {
    "id": "https://orcid.org/0000-0002-3730-5348",
    "type": "Person",
    "name": "Kather, Jakob Nikolas"
  },
  "datePublished": "2019-06-16",
  "description": "",
  "hasPart": [
    {
      "id": "blob/f711cb1bd02b951f8f0ecc393471f31a26aa6a92/data/tissue_blocks_for_ebv_de_v01/falseNegative.zip",
      "type": [
        "http://www.w3.org/ns/prov#Entity",
        "DigitalDocument",
        "http://purl.org/wf4ever/wfprov#Artifact"
      ],
      "creator": {
        "id": "https://orcid.org/0000-0002-3730-5348",
        "type": "Person",
        "name": "Kather, Jakob Nikolas"
      },
      "name": "falseNegative.zip",
      "schema:url": "https://zenodo.org/api/files/7cf40388-2871-4a54-87ad-37572274961d/falseNegative.zip",
      "http://www.w3.org/ns/prov#atLocation": "data/tissue_blocks_for_ebv_de_v01/falseNegative.zip",
      "rdfs:label": "data/tissue_blocks_for_ebv_de_v01/falseNegative.zip@f711cb1bd02b951f8f0ecc393471f31a26aa6a92"
    },
    {
      "id": "blob/a26587d5eb9fcdf1f496efc01baadc1e84e7aec8/data/tissue_blocks_for_ebv_de_v01/falsePositive.zip",
      "type": [
        "http://www.w3.org/ns/prov#Entity",
        "DigitalDocument",
        "http://purl.org/wf4ever/wfprov#Artifact"
      ],
      "creator": {
        "id": "https://orcid.org/0000-0002-3730-5348",
        "type": "Person",
        "name": "Kather, Jakob Nikolas"
      },
      "name": "falsePositive.zip",
      "schema:url": "https://zenodo.org/api/files/7cf40388-2871-4a54-87ad-37572274961d/falsePositive.zip",
      "http://www.w3.org/ns/prov#atLocation": "data/tissue_blocks_for_ebv_de_v01/falsePositive.zip",
      "rdfs:label": "data/tissue_blocks_for_ebv_de_v01/falsePositive.zip@a26587d5eb9fcdf1f496efc01baadc1e84e7aec8"
    },
    {
      "id": "blob/d67d63f525061d2ae3ee0490f4551f7e43bfbbb9/data/tissue_blocks_for_ebv_de_v01/trueNegative.zip",
      "type": [
        "http://www.w3.org/ns/prov#Entity",
        "DigitalDocument",
        "http://purl.org/wf4ever/wfprov#Artifact"
      ],
      "creator": {
        "id": "https://orcid.org/0000-0002-3730-5348",
        "type": "Person",
        "name": "Kather, Jakob Nikolas"
      },
      "name": "trueNegative.zip",
      "schema:url": "https://zenodo.org/api/files/7cf40388-2871-4a54-87ad-37572274961d/trueNegative.zip",
      "http://www.w3.org/ns/prov#atLocation": "data/tissue_blocks_for_ebv_de_v01/trueNegative.zip",
      "rdfs:label": "data/tissue_blocks_for_ebv_de_v01/trueNegative.zip@d67d63f525061d2ae3ee0490f4551f7e43bfbbb9"
    },
    {
      "id": "blob/1d175c156820c92fdd34bcf65693d47b4cfbb969/data/tissue_blocks_for_ebv_de_v01/truePositive.zip",
      "type": [
        "http://www.w3.org/ns/prov#Entity",
        "DigitalDocument",
        "http://purl.org/wf4ever/wfprov#Artifact"
      ],
      "creator": {
        "id": "https://orcid.org/0000-0002-3730-5348",
        "type": "Person",
        "name": "Kather, Jakob Nikolas"
      },
      "name": "truePositive.zip",
      "schema:url": "https://zenodo.org/api/files/7cf40388-2871-4a54-87ad-37572274961d/truePositive.zip",
      "http://www.w3.org/ns/prov#atLocation": "data/tissue_blocks_for_ebv_de_v01/truePositive.zip",
      "rdfs:label": "data/tissue_blocks_for_ebv_de_v01/truePositive.zip@1d175c156820c92fdd34bcf65693d47b4cfbb969"
    }
  ],
  "identifier": "10.5281/zenodo.3247009",
  "keywords": [],
  "license": "http://creativecommons.org/licenses/by/4.0/legalcode",
  "name": "Tissue blocks for EBV detection in stomach cancer",
  "schema:sameAs": "https://doi.org/10.5281/zenodo.3247009",
  "subjectOf": [],
  "schema:url": "https://zenodo.org/record/3247009",
  "version": "v0.1",
  "http://www.w3.org/ns/prov#atLocation": ".renku/datasets/f2e6b3f3-5987-4800-9b7c-07d4821aa5c6",
  "rdfs:label": "10.5281/zenodo.3247009"
}
```

---

closes #719 
closes #661

### DIFF
--- a/renku/core/management/repository.py
+++ b/renku/core/management/repository.py
@@ -175,6 +175,37 @@ class RepositoryApiMixin(GitCore):
         if self.renku_metadata_path.exists():
             return Project.from_yaml(self.renku_metadata_path, client=self)
 
+    @property
+    def remote(self, remote_name='origin'):
+        """Return host, owner and name of the remote if it exists."""
+        from renku.core.models.git import GitURL
+
+        host = owner = name = None
+        try:
+            remote_branch = \
+                self.repo.head.reference.tracking_branch()
+            if remote_branch is not None:
+                remote_name = remote_branch.remote_name
+        except TypeError:
+            pass
+
+        try:
+            url = GitURL.parse(self.repo.remotes[remote_name].url)
+
+            # Remove gitlab. unless running on gitlab.com.
+            hostname_parts = url.hostname.split('.')
+            if len(hostname_parts) > 2 and hostname_parts[0] == 'gitlab':
+                hostname_parts = hostname_parts[1:]
+            url = attr.evolve(url, hostname='.'.join(hostname_parts))
+        except IndexError:
+            url = None
+
+        if url:
+            host = url.hostname
+            owner = url.owner
+            name = url.name
+        return {'host': host, 'owner': owner, 'name': name}
+
     def process_commit(self, commit=None, path=None):
         """Build an :class:`~renku.core.models.provenance.activities.Activity`.
 
--- a/renku/core/models/datasets.py
+++ b/renku/core/models/datasets.py
@@ -19,7 +19,10 @@
 
 import configparser
 import datetime
+import os
+import pathlib
 import re
+import urllib
 import uuid
 from functools import partial
 from pathlib import Path
@@ -30,7 +33,7 @@ from attr.validators import instance_of
 from renku.core import errors
 from renku.core.models.provenance.entities import Entity
 from renku.core.utils.datetime8601 import parse_date
-from renku.core.utils.doi import is_doi
+from renku.core.utils.doi import extract_doi, is_doi
 
 from . import jsonld as jsonld
 
@@ -40,6 +43,14 @@ _path_attr = partial(
     jsonld.ib,
     converter=Path,
 )
+
+
+def _extract_doi(value):
+    """Return either a string or the doi part of a URL."""
+    value = str(value)
+    if is_doi(value):
+        return extract_doi(value)
+    return value
 
 
 @jsonld.s(
@@ -382,7 +393,7 @@ class Dataset(Entity, CreatorsMixin):
         default=attr.Factory(uuid.uuid4),
         context='schema:identifier',
         kw_only=True,
-        converter=str
+        converter=_extract_doi
     )
 
     in_language = jsonld.ib(
@@ -434,6 +445,8 @@ class Dataset(Entity, CreatorsMixin):
         kw_only=True
     )
 
+    same_as = jsonld.ib(context='schema:sameAs', default=None, kw_only=True)
+
     @created.default
     def _now(self):
         """Define default value for datetime fields."""
@@ -462,13 +475,15 @@ class Dataset(Entity, CreatorsMixin):
     @property
     def uid(self):
         """UUID part of identifier."""
+        if is_doi(self.identifier):
+            return self.identifier
         return self.identifier.split('/')[-1]
 
     @property
     def short_id(self):
         """Shorter version of identifier."""
-        if is_doi(self._id):
-            return self._id
+        if is_doi(self.identifier):
+            return self.identifier
         return str(self.uid)[:8]
 
     @property
@@ -498,7 +513,9 @@ class Dataset(Entity, CreatorsMixin):
         :return: self
         """
         if is_doi(other_dataset.identifier):
-            self._id = other_dataset.identifier
+            self.same_as = urllib.parse.urljoin(
+                'https://doi.org', other_dataset.identifier
+            )
 
         for field_ in self.EDITABLE_FIELDS:
             val = getattr(other_dataset, field_)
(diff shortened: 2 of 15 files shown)


renku/core/models/datasets.py
```diff
@@ -42,6 +45,14 @@
 )
 
 
+def str_or_doi(value):
```
jsam: Isn't this gonna return string type in both cases? Maybe we should come up with a better name?

renku/core/management/repository.py
```diff
@@ -174,6 +174,38 @@ def project(self):
...
+        except TypeError:
+            pass
+
+        try:
+            url = GitURL.parse(self.repo.remotes[remote_name].url)
+
+            # Remove gitlab. unless running on gitlab.com.
+            hostname_parts = url.hostname.split('.')
+            if len(hostname_parts) > 2 and hostname_parts[0] == 'gitlab':
+                hostname_parts = hostname_parts[1:]
```
mohammad-sdsc: I don't get what this part is doing. Why `gitlab` should be removed?

renku/core/models/datasets.py
rokroskar: well it casts to string or returns the doi (which yes, is also a string). Originally the converter was just `str`.

renku/core/models/datasets.py
```diff
@@ -426,6 +437,8 @@ class Dataset(Entity, CreatorsMixin):
         kw_only=True
     )
 
+    source = jsonld.ib(context='schema:sameAs', default=None, kw_only=True)
```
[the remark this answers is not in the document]
rokroskar: good point - I renamed it to `same_as`

renku/core/models/datasets.py
jsam: Yes, hence the confusion. Maybe better name would be `origin_or_doi`?

renku/core/management/repository.py
rokroskar: this is a consequence of how gitlab is configured in renku deployments - it's not on a subdomain but instead on a path e.g. `renkulab.io/gitlab`. But the host should not include `gitlab`.

renku/core/models/datasets.py
rokroskar: I could rename it though I don't know that those names would make it clearer - a converter of just `str` is not confusing?

renku/core/management/repository.py
mohammad-sdsc: If `gitlab` is part of the path (and not part of the hostname) then this won't do anything. In this case, for example, `hostname` will be `renkulab.io`.

renku/core/models/datasets.py
jsam: Well no, cause we are only talking about typing with that converter. Here we are mixing semantics of logic with typing, let's go with only 1. If we would have DOI type then `str_or_doi` would make sense and thats what I would expect with this name, but since this is not the case this makes it harder to reason about it when reading it.

renku/core/management/repository.py
rokroskar: hmm you're right, I'm probably misremembering what the point was here, I have to look at the original commit that introduced this. It actually looks to be the opposite from what I described above, i.e. it's removing just the subdomain part. So in the case that the remote is `gitlab.renkuinstance.com` we don't want the IDs to have `gitlab` in them but just `renkuinstance.com`

renku/core/models/datasets.py
rokroskar: ok, renamed to `_extract_doi`

jachro:
👍Hurray! All seems to be working fine!

PR #744, opened 2019-10-07
refactor: remove click exceptions from Renku core

mohammad-sdsc:

# Description

Removes all uses of Click exception in `reknu.core`.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/720

## Type of change

- [x Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


tests/cli/test_datasets.py
```diff
@@ -417,7 +417,7 @@ def test_relative_git_import_to_dataset(tmpdir, runner, project, client):
         ],
         catch_exceptions=True,
     )
-    assert 2 == result.exit_code
+    assert 1 == result.exit_code
```
jsam: Exit code 2 represents user miss-use which is the guidance which we tried to follow. Why changing it here now? Do we want to migrate all user miss-use errors to general errors?

renku/core/errors.py
```diff
@@ -72,15 +76,30 @@ def return_or_raise(cls, response, expected_status_code):
         raise cls(response)
 
 
+class BadParameter(RenkuException):
```
jsam: There is a `click.BadParameter` exception as well which might lead to confusion. To enhance readability I would recommend that we come up with a better name? Maybe `InvalidParameter`?

renku/cli/exception_handler.py
```diff
@@ -59,6 +59,8 @@
 
 import click
 
+from renku.core.errors import RenkuException
+
 _BUG = click.style(
     'Ahhhhhhhh! You have found a bug. 🐞\n\n',
```
jsam: Should we also update the message to factor in the feedback we received in Bern? Some humble message which is not mocking @ciyer would do the trick (we can also open a ticket for that).

mohammad-sdsc: I think it deserves to have its own story.

renku/core/errors.py
mohammad-sdsc: I've changed to `ParameterError`.

tests/cli/test_datasets.py
mohammad-sdsc: Didn't know about this. Good point! I reverted back all these to return 2.

renku/core/errors.py
Panaetius: Do we even need our own exception for this? It seems the builtin `ValueError` has pretty much the same semantics.

mohammad-sdsc: It helps to give a better feedback to user, so, I would prefer to keep it.

PR #753, opened 2019-10-15
[WIP] add mls support

vigsterkr:
Add initial support for ML Schema

### DIFF
--- a/renku/core/models/mls.py
+++ b/renku/core/models/mls.py
@@ -1,0 +1,211 @@
+import configparser
+import datetime
+import re
+import uuid
+from functools import partial
+from pathlib import Path
+
+import attr
+from attr.validators import instance_of
+from dateutil.parser import parse as parse_date
+
+from renku.core import errors
+from renku.core.models.entities import Entity
+from renku.core.utils.doi import is_doi
+
+from renku.core.models import jsonld
+
+NoneType = type(None)
+
+_path_attr = partial(
+    jsonld.ib,
+    converter=Path,
+)
+
+def _convert_implementation(obj):
+    """Convert implementation object."""
+    if isinstance(obj, dict):
+        return Implementation.from_jsonld(obj)
+
+
+def _convert_algorithm(obj):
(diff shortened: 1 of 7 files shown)


rokroskar:
Hey @vigsterkr thanks for this effort, glad to see it coming to life! To make it easier for people to digest what's happening here, could you add a bit more info to the PR description?

renku/core/models/mls.py
```diff
@@ -0,0 +1,211 @@
...
+
+@jsonld.s(
+    type='mls:HyperParameterSetting',
+    slots=True,
+    context={
+        'mls': 'http://www.w3.org/ns/mls#',
+        'specified_by': 'mls:specifiedBy'
+    }
+)
+class HyperParameterSetting:
```
chrisbarber: Missing `_id`

renku/core/models/mls.py
```diff
@@ -0,0 +1,211 @@
...
+        if 'value' in v:
+            input_values.append(HyperParameterSetting.from_jsonld(v))
+    return input_values
+
+
+def _convert_parameter_value(obj):
+    """Convert parameter value."""
+    if isinstance(obj, dict):
+        # TODO: should be interpreted type value
+        return obj["@value"]
```
chrisbarber: I haven't been able to find anything that makes the jsonld we want from a python primitive e.g. `true` becomes `"value": {"@type": "xsd:boolean", "@value": "true"}`.

This is the closest I have found, which has all the mappings (https://github.com/RDFLib/rdflib/blob/f2d893145f6fba3513d5e169b964554451c5eee9/rdflib/term.py#L1441) but uses the full URL instead of `xsd:` prefix.

So I am just using:
```
from rdflib import Literal
```
...
```
    literal = Literal(obj)                                                                           
    return {"@type": "xsd:" + str(literal.datatype).split('#')[-1],                                  
            "@value": literal.value}
```

rokroskar:
I guess we can close this @vigsterkr ?

Issue #763, opened 2019-10-20
add `dataset update` to the docs


rokroskar:
We should add information about `renku dataset update` to the [renku-python docs](https://renku-python.readthedocs.io/en/latest/commands.html#module-renku.cli.dataset).

PR #766, opened 2019-10-22
doc: added docs for dataset update

mohammad-sdsc:

# Description

Document for `renku dataset update`

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/763

## Type of change

## Checklist

### DIFF
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -104,6 +104,26 @@ will yield:
         new-dir/
           new-filename
 
+Updating a dataset:
+
+After adding files from a remote Git repository, you can check for updates in
+those files by using ``renku dataset update`` command. This command checks all
+remote files and copies over new content if there are any. It also deletes
+files from the local dataset if they are deleted from the remote Git
+repository.
+
+You can limit the scope of updated files by specifying dataset names, using
+``--include`` and ``--exclude`` to filter based on file names, or using
+``--creators`` to filter based on creators. For example, the following command
+updates only CSV files from ``my-dataset``:
+
+.. code-block:: console
+
+    $ renku dataset update -I '*.csv' my-dataset
+
+Note that putting glob patterns in quotation is needed to tell Unix shell not
+to expand them.
+
 Tagging a dataset:
 
 A dataset can be tagged with an arbitrary tag to refer to the dataset at that


renku/cli/dataset.py
```diff
@@ -104,6 +104,26 @@
         new-dir/
           new-filename
 
+Updating a dataset:
+
+After adding files from a remote Git repository, you can check for updates in
+those files by using ``renku dataset update`` command. This command checks all
+remote files and copies over new content if there are any. It also deletes
+files from the local dataset if they are deleted from the remote Git
```
rokroskar: Hmm I hadn't thought of that before but I guess this will lead to some unexpected behavior. Is there an issue already for updating to a specific commit?

mohammad-sdsc: Not yet. It was supposed to be part of this one but I decided to do it later. I'll create on when we close this one. I don't see though how that would help; if a file is deleted from the latest update eventually we need a way to deal with it.

rokroskar: Yes sure, but right now I can imagine someone could do an update that deleted a file with no way to get back other than doing a git reset. If they were able to go to a previous commit it would be a non-issue.

rokroskar: anyway, this is not the right place for this discussion I realize :)  We should add support for updating to a specific commit reasonably soon to avoid having users that get stuck in some strange state.

renku/cli/dataset.py
```diff
@@ -104,6 +104,26 @@
...
+repository.
+
+You can limit the scope of updated files by specifying dataset names, using
+``--include`` and ``--exclude`` to filter based on file names, or using
+``--creators`` to filter based on creators. For example, the following command
```
rokroskar: creators - as determined by the email?

mohammad-sdsc: No, it is the full name of the creator. This is like some other dataset subcommands (e.g. `ls-files`)

renku/cli/dataset.py
mohammad-sdsc: As we discussed, I'll ask the user for deletion.

mohammad-sdsc:
I'll add changes in this PR to https://github.com/SwissDataScienceCenter/renku-python/pull/765 and close this one.

Issue #730, opened 2019-10-03
Enable local project configuration


rokroskar:
In certain cases we need to add project-specific configurations that get persisted in the repository (i.e. are not part of the git configuration). We already have a configuration handler but at the moment this combines the global and user-specific config (outside the repository) and the git config. We should extend this handler to allow also project-specific settings that supersede any global or user settings. 

Example use-case is setting a default path for the interactive session launch or graph-specific settings (see https://github.com/SwissDataScienceCenter/renku-python/pull/701#discussion_r331387367)

PR #777, opened 2019-10-29
730 local project configuration

mohammad-sdsc:

# Description

This PR adds local project configuration for Renku project. Local configurations are stored in `.renku/renku.ini` which is committed when there are some changes. When reading configuration, values in local configuration takes precedence over those in global configuration. To read only local or global configuration use `--local` or `--global` flags. When storing configuration, all changes are written into the local configuration unless `--global` flag is passed.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/730

## Type of change

## Checklist


renku/cli/config.py
```diff
@@ -46,13 +53,19 @@
 @click.command()
 @click.argument('key', required=True)
 @click.argument('value', required=False, default=None)
+@click.option(
+    '--local',
```
rokroskar: If local is the default, does it need a flag? Just to be explicit/more readable?

mohammad-sdsc: local is the default for storing. When reading, local and global are both read and keys in local override similar keys in global. `--local` in used only to limit the read to local config.

This is something I copied from Git. If this is confusing I can remove it.

renku/cli/config.py
```diff
@@ -46,13 +53,19 @@
...
+    is_flag=True,
+    help='Read from local configuration only.'
+)
 @click.option(
     '--global',
-    'is_global',
+    'global_only',
     is_flag=True,
-    help='Store to global configuration.'
+    help='Read/store from/to global configuration only.'
```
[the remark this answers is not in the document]
mohammad-sdsc: No. For the same reason as described below, local and global act differently.

renku/core/commands/client.py
```diff
@@ -45,6 +45,7 @@ def pass_local_client(
     commit_only=None,
     ignore_std_streams=True,
     allow_empty=True,
+    do_not_commit_if_empty=False,
```
jsam: Shouldn't this be defaulted to `True`? How do `allow_empty` and `do_not_commit_if_empty` differ?

renku/core/management/config.py
```diff
@@ -35,97 +35,127 @@ def print_app_config_path(ctx, param, value):
...
         return
-    click.echo(ConfigManagerMixin().config_path)
+    click.echo(ConfigManagerMixin().global_config_path)
     ctx.exit()
 
 
-def default_config_dir():
-    """Return default config directory."""
+def global_config_dir():
+    """Return User's config directory."""
```
jsam:
```suggestion
    """Return user's config directory."""
```

renku/core/management/config.py
```diff
@@ -35,97 +35,127 @@ def print_app_config_path(ctx, param, value):
...
 
             if not config[section].keys():
                 config.pop(section)
 
-        config = self.store_config(config)
-        return config
+            self.store_config(config, global_only=global_only)
+
+
+class _GlobalConfigLock:
```
jsam: The convention in the repository to achieve this is usually by using context managers.

mohammad-sdsc: This is a context manager but I moved it to a different class. Do you mean a context manager in the same class (i.e. LocalClient)?

renku/core/commands/client.py
mohammad-sdsc: `allow_empty` raises an exception when it is `False` (and. nothing to commit) while `do_not_commit_if_empty` just returns with no exception. Both of these are used so I kept them. I'll rename them to `commit_empty` and `raise_if_empty` (which is considered when `commit_empty` is false) to make it a bit more clear.

renku/cli/config.py
rokroskar: No it's fine, I realized what the point was after I started playing with it. But it's only useful if we also show the configuration, e.g. `renku config` without arguments would show you the combined configurations and `renku config --local` only the options you have set for the current project.

renku/core/management/config.py
jsam: This [way](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) is used in the repo. Also, we should write the handling of the lock files in the similar way which exists already in the [repo](https://github.com/SwissDataScienceCenter/renku-python/blob/master/renku/core/management/repository.py#L115).

mohammad-sdsc: That was way cleaner :sweat_smile:

renku/cli/config.py
mohammad-sdsc: Done.

mohammad-sdsc:
> Looks nice, thanks! I tested the functionality, seems to work as advertised.
> 
> I am wishing for two things:
> 
> 1. some way to remove a configuration setting. This would be especially nice if I set some local option that also has a global default and I then want to resort back to the default.
> 2. display/list the options
> 
> If these two requests seem too far out of scope of this PR, we can make new issues for them.

I've added both of these. With `--remove` option you can remove a value from either local or global config. `renku config` shows all the configuration and you limit its scope to local or global config.

mohammad-sdsc:
I've removed the `--config` flags from `renku` command-line. It was supposed to override the config file location (either itself or by reading `RENKU_CONFIG` env var) but in reality it won't do that. I'll create an issue to re-add this option in future.

rokroskar:
> `renku config` shows all the configuration and you limit its scope to local or global config.

Testing the latest changes and I was looking at what happens if I define the same option in the global and local configuration. Then `renku config` (without flags) uses the precedence of the local config and shows me only the local option. Is this intended? The functionality makes sense to me (i.e. `renku config` shows me what the configuration would be taken into account if I ran some command right now) but it's not what your comment says.

mohammad-sdsc:
> > `renku config` shows all the configuration and you limit its scope to local or global config.
> 
> Testing the latest changes and I was looking at what happens if I define the same option in the global and local configuration. Then `renku config` (without flags) uses the precedence of the local config and shows me only the local option. Is this intended? The functionality makes sense to me (i.e. `renku config` shows me what the configuration would be taken into account if I ran some command right now) but it's not what your comment says (or at least not how I understand it).

A `can` was missing from my comment: `... you can limit ...`. You can do that by using `--local` or `--global` (e.g. `renku config --global` shows the global config only).

rokroskar:
I was more referring to

> shows all the configuration 

do you mean that it shows the combined configuration?

mohammad-sdsc:
Yes. It shows the combined configuration.

renku/cli/config.py
```diff
@@ -27,32 +27,79 @@
...
 @click.command()
-@click.argument('key', required=True)
+@click.argument('key', required=False, default=None)
 @click.argument('value', required=False, default=None)
+@click.option('--remove', is_flag=True, help='Remove specified key.')
+@click.option(
+    '--local',
+    'local_only',
+    is_flag=True,
+    help='Read from local configuration only.'
```
rokroskar: just a nitpick here, but perhaps this should use the same wording as below for the global option? i.e. `Read/store from/to local configuration only.`

Issue #688, opened 2019-09-19
dataset create/add functionality


emmjab:
**Is your feature request related to a problem? Please describe.**
Currently `renku dataset add <my dataset name> <data path>` will silently first create the dataset if `<my dataset>` doesn't exist (but with no creator). Then, if the provided `<data path>` is malformed/no data is retrieved, there's irrelevant files and metadata around from the dataset creation step even though the data add fails.

**Describe the solution you'd like**
Instead, `renku dataset add <my dataset name> <data path>` should return an error: `Please first create the dataset with: renku dataset create <my dataset name>`.

**Additional context**
I think it's confusing to ask users to "create a dataset" that doesn't have any data in it. When you call the "create dataset" command, you're creating an empty folder to hold the data, and adding a file in preparation to be populated with metadata about where the data comes from, etc.

**Alternative based on additional context**
Maybe it makes sense to augment the `renku dataset create` command with `--add <path to data>` so that you can create and add data at the same time, and it's from the perspective of "creating a dataset":
ex:
`renku dataset create zhbikes --add <url to data.csv>`
`renku dataset create zhbikes --add <url to data.git> --rel-data path/to/mydata.csv`
`renku dataset create zhbikes --add <local/path/in/repo>`

rokroskar:
what about `renku dataset add --create zhbikes <url to data>`?

mohammad-sdsc:
I've picked @rokroskar suggestion when implementing this. Users must pass `--create` when adding data to a non-existing database or the operation fails with proper error message.
Moreover, if they provide `--create` with an existing database they also receive an error.

PR #779, opened 2019-10-29
feat: do not create dataset implicitly

mohammad-sdsc:

# Description

When adding file to a non-existing dataset, this PR prevent implicit creation of the dataset. User must pass `--create` for creation of a non-existing dataset.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/688

## Type of change

## Checklist

### DIFF
--- a/renku/core/commands/dataset.py
+++ b/renku/core/commands/dataset.py
@@ -106,8 +106,10 @@ def create_dataset(client, name, handle_duplicate_fn=None):
     :raises: ``renku.core.errors.ParameterError``
     """
     existing = client.load_dataset(name=name)
-    if (not existing or handle_duplicate_fn and handle_duplicate_fn(existing)):
-        with client.with_dataset(name=name, create=True) as dataset:
+    if not existing or (handle_duplicate_fn and handle_duplicate_fn(existing)):
+        with client.with_dataset(
+            name=name, create=True, fail_if_exists=False
+        ) as dataset:
             creator = Creator.from_git(client.repo)
             if creator not in dataset.creator:
                 dataset.creator.append(creator)
@@ -205,8 +207,10 @@ def add_to_dataset(
 
     except DatasetNotFound:
         raise DatasetNotFound(
-            'Dataset "{}" does not exists. Create the dataset or retry with '
-            '--create option for automatic dataset creation.'.format(name)
+            'Dataset "{0}" does not exists.\n'
+            'Use "renku dataset create {0}" to create the dataset or retry '
+            '"renku dataset add {0}" command with "--create" option for '
+            'automatic dataset creation.'.format(name)
         )
     except (FileNotFoundError, git.exc.NoSuchPathError):
         raise ParameterError('Could not process \n{0}'.format('\n'.join(urls)))
--- a/renku/core/commands/dataset.py
+++ b/renku/core/commands/dataset.py
@@ -100,19 +100,15 @@ def dataset_parent(client, revision, datadir, format, ctx=None):
 @pass_local_client(
     clean=False, commit=True, commit_only=DATASET_METADATA_PATHS
 )
-def create_dataset(client, name, handle_duplicate_fn=None):
+def create_dataset(client, name):
     """Create an empty dataset in the current repo.
 
     :raises: ``renku.core.errors.ParameterError``
     """
-    existing = client.load_dataset(name=name)
-    if not existing or (handle_duplicate_fn and handle_duplicate_fn(existing)):
-        with client.with_dataset(
-            name=name, create=True, fail_if_exists=False
-        ) as dataset:
-            creator = Creator.from_git(client.repo)
-            if creator not in dataset.creator:
-                dataset.creator.append(creator)
+    with client.with_dataset(name=name, create=True) as dataset:
+        creator = Creator.from_git(client.repo)
+        if creator not in dataset.creator:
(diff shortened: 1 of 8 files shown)


renku/core/management/datasets.py
```diff
@@ -109,12 +109,19 @@ def load_dataset(self, name=None):
...
+        self, name=None, identifier=None, create=False, fail_if_exists=False
+    ):
         """Yield an editable metadata object for a dataset."""
         dataset = self.load_dataset(name=name)
         clean_up_required = False
 
         if dataset is None:
+            if not create:
+                raise errors.DatasetNotFound(
+                    'Dataset does not exists. Create the dataset or retry '
```
jsam: I would prefer not to duplicate messages but keep them in 1 place.

renku/core/commands/dataset.py
```diff
@@ -158,18 +159,25 @@ def add_to_dataset(
     name,
     link=False,
     force=False,
+    create=False,
     sources=(),
     destination='',
     with_metadata=None,
-    urlscontext=contextlib.nullcontext
+    urlscontext=contextlib.nullcontext,
+    fail_if_exists=True
```
jsam: I think this should be the default behavior. Is there a reason for this flag? I see problems with `force` and `fail_if_exists` cause it can lead to undefined behavior since we don't have explicitly defined precedence anywhere.

renku/core/commands/dataset.py
```diff
@@ -107,7 +107,7 @@ def create_dataset(client, name, handle_duplicate_fn=None):
     """
     existing = client.load_dataset(name=name)
     if (not existing or handle_duplicate_fn and handle_duplicate_fn(existing)):
-        with client.with_dataset(name=name) as dataset:
+        with client.with_dataset(name=name, create=True) as dataset:
```
jsam: I think the above line contains unnecessary parentheses. Could we remove it? :smile:

renku/core/commands/dataset.py
```diff
@@ -195,6 +203,11 @@ def add_to_dataset(
 
                 dataset.update_metadata(with_metadata)
 
+    except DatasetNotFound:
+        raise DatasetNotFound(
+            'Dataset "{}" does not exists. Create the dataset or retry with '
+            '--create option for automatic dataset creation.'.format(name)
```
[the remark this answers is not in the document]
mohammad-sdsc: I've did it partially because `add` command takes more options which are tricky to correctly reproduce here.

renku/core/commands/dataset.py
mohammad-sdsc: In some other commands (e.g. `renku dataset create`) we need to be able to re-create a dataset (`create` prompts the user itself) so we need this flag. `force` is to tell git to force-add ignored files; dataset duplication tests is performed before and if it fails we don't reach the add part. Would you please explain how they can cause problem?

renku/core/commands/dataset.py
mohammad-sdsc: If you mean the `if` statement then I did not notice it (since I did not touch it) but I see your point. I think parenthesis is good to have there but the position is wrong; it should be like this:
`if not existing or (handle_duplicate_fn and handle_duplicate_fn(existing)):`

rokroskar:
@emmjab not sure you saw this - it's pretty close to getting merged so if you have a chance to have a look at the functionality it would be great!

renku/core/commands/dataset.py
rokroskar: should the files be removed from the dataset if it is recreated? This doesn't seem to be the case atm

mohammad-sdsc: Good catch! Should we remove this ability of recreating a dataset? We can simply instruct user to remove the dataset first if they want to recreate it. This would simplify the code in some places.

rokroskar: I think so - the behavior is also not consistent between using the `--create` flag and using `renku dataset create` - in the former you get an error, in the latter you are asked if you want to overwrite

mohammad-sdsc: I'll create an issue for it.

rokroskar: I think `renku dataset create` should also just create an error and it would be fine for the time being

emmjab:
Just tried it out! This part looks good to me (there's a typo but not important: "does not exist"):
```
renku dataset add two ../../testing_zenodo_import/requirements.txt 
Error: Dataset "two" does not exists.
Use "renku dataset create two" to create the dataset or retry "renku dataset add two" command with "--create" option for automatic dataset creation.
```

And I think this is good, too -- if you try to pass in the create flag but the dataset already exists, there's an error and the files aren't copied in.

```
renku dataset add two --create ../../../../Documents/UserStories.docx 
Error: Dataset with name "two" exists.
```

There's still a bug for when you locally try to add files from outside that exist within a git repo, and @mohammad-sdsc is opening a separate issue for this:

```
$ renku dataset add two --create ../../testing_zenodo_import/requirements.txt 
/Users/jablonsk/Documents/emma_dev_renku_ch/renku-python/renku/core/management/datasets.py:335: UserWarning: Importing local git repository, use HTTPS
  warnings.warn('Importing local git repository, use HTTPS')
Error: Cannot access remote Git repo: /Users/jablonsk/Documents/test/testing_zenodo_import
```

PR #783, opened 2019-10-30
fix: addressed CI problems with git submodules

jsam:


conftest.py
```diff
@@ -159,7 +159,7 @@ def project(repository):
     """Create a test project."""
     from git import Repo
 
-    repo = Repo(repository)
+    repo = Repo(repository, search_parent_directories=True)
```
rokroskar: 🙈

Issue #781, opened 2019-10-30
cli: renku migrate datasets fails


rokroskar:
## Describe the bug
`renku migrate datasets` fails catastrophically, but maybe the better question is why there is a `DatasetFile` with a blank `path`.

## Details
Git repo URL: `git@dev.renku.ch:virginiafriedrich/zurich-bikes-2.git`

**Renku version:** 0.7.0
**OS:** Darwin (Darwin Kernel Version 18.7.0: Thu Jun 20 18:42:21 PDT 2019; root:xnu-4903.270.47~4/RELEASE_X86_64)
**Python:** 3.7.4

### Traceback

```
Traceback (most recent call last):
  File "<attrs generated init renku.core.models.datasets.DatasetFile>", line 30, in __init__
    self.filename = __attr_converter_filename(__attr_factory_filename(self))
  File "[...]/renku/core/models/datasets.py", line 266, in <lambda>
    filename = attr.ib(kw_only=True, converter=lambda x: Path(x).name)
  File "[...]/pathlib.py", line 1010, in __new__
    self = cls._from_parts(args, init=False)
  File "[...]/pathlib.py", line 665, in _from_parts
    drv, root, parts = self._parse_args(args)
  File "[...]/pathlib.py", line 649, in _parse_args
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

jsam:
It seems that other commands fail with the same errors, and not just migrate datasets. After a quick patch and successful migration that file without filename is still there. I would consider this a broken state and handle this validation at the import step. As a followup, it would be extremely helpful if I can have DOI of the imported dataset which caused this problem so we can add it to the testing suite.

PR #786, opened 2019-10-31
fix: adds simple check on empty filename

jsam:
closes: #781

### DIFF
--- a/renku/core/models/datasets.py
+++ b/renku/core/models/datasets.py
@@ -128,6 +128,12 @@ def _convert_dataset_files_creators(value):
         return [Creator.from_jsonld(c) for c in coll]
 
 
+def convert_filename_path(p):
+    """Return name of the file."""
+    if p:
+        return Path(p).name
+
+
 @jsonld.s(
     type='schema:DigitalDocument',
     slots=True,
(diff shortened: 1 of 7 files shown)


renku/core/models/datasets.py
```diff
@@ -128,6 +128,12 @@ def _convert_dataset_files_creators(value):
         return [Creator.from_jsonld(c) for c in coll]
 
 
+def convert_filename_path(p):
+    """Return name of the file."""
+    if p and len(p) >= 1:
+        return Path(p).name
+
+
```
mohammad-sdsc: Can you please explain how this fix works? If `filename` is `None` then no `DatasetFile` object will be created!?

jsam:
> > I would add a test which caused this problem here as well, but I have no idea which DOI caused it, so any help with searching of that DOI is more than welcome. hear_no_evil
> 
> `git grep path .renku/datasets` shows that the empty path is in `.renku/datasets/dd38afca-0e96-4034-a2e5-dc7505c0af30/metadata.yml` so I guess the DOI is `https://doi.org/10.5281/zenodo.3348115` (`https://zenodo.org/record/3348115`).

Hmm, this is interesting. Seems like migration dumped that dataset since I don't even have it. :thinking:  Thanks! I'll look into this and add the test.

renku/core/models/datasets.py
jsam: I think it should be created, but `filename` field will remain None within the instantiated object.

The problem of why raise was happening is because this `Path(None)`  results in `TypeError`.

renku/core/models/datasets.py
```diff
@@ -128,6 +128,12 @@ def _convert_dataset_files_creators(value):
         return [Creator.from_jsonld(c) for c in coll]
 
 
+def convert_filename_path(p):
+    """Return name of the file."""
+    if p and len(p) >= 1:
```
rokroskar: is this different than just

```
if p:
  ...
```

rokroskar:
I still see the same issue with the `url` field of the files with a blank path.

Issue #770, opened 2019-10-24
RaaS: Add uploaded file to project


jsam:
Add a feature on renku service to add uploaded files to a cloned project.

rokroskar:
you mean outside of a dataset?

jsam:
There are 2 possible ways to add a data file to the project. 


**By upload**

In this scenario, we start by uploading a data file to the cache and receive a reference ID. This reference ID could be then used in the following endpoint:

```
/projects/add
```

where payload would be
```
{
    "type: 1,
    "file_id": <reference_id>,
    "project_id": <project_id>,
    "relative_path": <path from root of the repository where that files should be placed>,
    "commit_message": <commit_message>
}
```

This action could be performed in sync way and could return following response:
```
{
    "result": {
        "commit": <created commit>
    }
}
```

or error if one occurred:
```
{
    "error": {...}
}
```


**By external crawl**

In this scenario, we submit a remote URL containing the data we want to add to the project. We reuse the endpoint from 1) but with the following payload:

```
{
    "type: 2,
    "data_url": <url to be crawled>,
    "project_id": <project_id>,
    "relative_path": <path from root of the repository where that files should be placed>,
    "commit_message": <commit_message>
}
```

Since this action should be performed in an async way due to the potentially large size of `<data_url>`, we would submit the requested action as a job to the job queue and return the `<job_id>` in the response.

Response example:
```
{
    "result": {
        "job_id": <job_id>
    }
}
```

or error if one occurred,
```
{
    "error": {...}
}
```

Issue #732, opened 2019-10-03
Renku as a Service: Datasets


jsam:
# Description

The renku service should offer functionality to allow the following operations on data sets:

- create datasets
- add files to datasets
- remove files from datasets
- delete datasets

# Implementation

To provide this functionality, the renku service needs to do several additional things:

- clone the project repo
- push the project repo back to server
- receive uploaded files
- receive uploaded directories

# Design

To develop a PoC quickly, some design choices have been made:

- The service will respond to requests synchronously

# Original

With #690 being merged we have enabled the ability to implement all dataset management endpoints within a renku-python as a service.

Work which has been done in https://github.com/SwissDataScienceCenter/renku-core/issues should be ported into renku-python and implemented as a service.

related: #731

ciyer:
From today's discussion

**File/folder hierarchies**

If the user just uploads an individual file, then current handling is fine. But, users may want to upload folder/file hierarchies. The easiest solution is to let the user structure things the way they want on their file system and then select the root of what they want to upload. The UI will create an archive and upload that and include a flag as part of the request stating that the contents need to be unarchived.

ciyer:
From today's discussion

**Endpoints**

_dataset_

-/dataset/list
-/dataset/create
-/dataset/files-add

_projects/files_

- /cache/projects-list
- /cache/project-clone
- /cache/files-list
- /cache/files-upload

ciyer:
From today's discussion

*/cache/files-list*

The return should be an object with at least
```
{
  "id": [service-file-id],
  "path": [path of the file],
  "size": [size of the file]
}
```

rokroskar:
As discussed yesterday, we should immediately have a plan for enabling the service to handle requests asynchronously. To me it looks like [Celery](https://docs.celeryproject.org/) would be a good option because it would still allow us to use Flask as the web framework and not have to resort to too many exotic python libraries/dependencies/restrictions. As also discussed with @ciyer, using coroutines would be of limited value in cases where the long-running tasks were executing within python (e.g. git tasks in renku-python) anyway.

rokroskar:
This epic has been completed - new, finer-grained epics should be opened as needed. 🚀

PR #788, opened 2019-11-01
feat: added renku service with cache and datasets

jsam:
addresses: #732

closes: #770


renku/service/entrypoint.py
```diff
@@ -0,0 +1,129 @@
...
+    app.config['UPLOAD_FOLDER'] = UPLOAD_FOLDER.name
+    app.config['MAX_CONTENT_LENGTH'] = 16 * 1024 * 1024
+
+    make_cache()
+    build_routes(app)
+
+    return app
+
+
+def build_routes(app):
```
rokroskar: These should all be blueprints?

renku/service/views/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+    """Check if filename is allowed."""
+    has_extension = '.' in filename
+    is_allowed = filename.rsplit('.', 1)[1].lower() in ALLOWED_EXTENSIONS
+    return has_extension and is_allowed
+
+
+def find_file(user, file_id):
+    """Search for cached file."""
+    for filename in os.listdir(CACHE_UPLOADS_PATH):
+        metadata = jwt.decode(filename, JWT_KEY, algorithms=[JWT_ALGORITHM])
```
Panaetius: I don't like the idea of using filenames to store metadata. Why not use a file-based database like SQLite to store file metadata? That'd improve performance as well.

renku/service/management/openapi.py
```diff
@@ -0,0 +1,59 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Generate OpenAPI spec for Renku service."""
+import json
+from pathlib import Path
+
+from apispec import APISpec
+from apispec_webframeworks.flask import FlaskPlugin
```
rokroskar: This does not seem to have support for blueprints (see also https://github.com/marshmallow-code/apispec-webframeworks/issues/11) - could we use [flask-apispec](https://github.com/jmcarp/flask-apispec) instead?

renku/service/static/spec.json
```diff
@@ -0,0 +1,339 @@
+{
```
rokroskar: do you mean to save this file? isn't it always automatically generated?

renku/service/views/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+
+import jwt
+from flask import request, jsonify
+from git import Repo
+
+from renku.service.config import ALLOWED_EXTENSIONS, CACHE_PROJECTS_PATH, \
+    INVALID_PARAMS_ERROR_CODE, CACHE_UPLOADS_PATH, JWT_KEY, JWT_ALGORITHM
+
+is_git_url = re.compile(
+    r'((git|ssh|http(s)?)|(git@[\w\.]+))(:(//)?)([\w\.@\:/\-~]+)(\.git)(/)?',
```
rokroskar: you could use https://github.com/SwissDataScienceCenter/renku-python/blob/master/renku/core/models/git.py#L132

renku/core/management/datasets.py
```diff
@@ -296,7 +296,8 @@ def _add_from_url(self, dataset, dataset_path, url, link, destination):
         mode = dst.stat().st_mode & 0o777
         dst.chmod(mode & ~(stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH))
 
-        self.track_paths_in_storage(str(dst.relative_to(self.path)))
+        if self.has_external_storage:
```
rokroskar: Doesn't the `ensure_external_storage` decorator already do this?

rokroskar: there should probably instead be a check whether the external storage was requested...

renku/service/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM python:3.6
```
rokroskar: why two dockerfiles for the service?

renku/service/config.py
```diff
@@ -0,0 +1,41 @@
...
+API_URL = os.getenv(
+    'RENKU_SVC_SWAGGER_URL', '/api/{0}/spec'.format(API_VERSION)
+)
+
+UPLOAD_FOLDER = tempfile.TemporaryDirectory()
+
+CACHE_UPLOADS_PATH = Path(UPLOAD_FOLDER.name) / Path('uploads')
+CACHE_PROJECTS_PATH = Path(UPLOAD_FOLDER.name) / Path('projects')
+
+ALLOWED_EXTENSIONS = {'txt', 'pdf', 'csv'}
```
rokroskar: ? why would we want to impose this?

renku/service/views/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+        if metadata['user'] == user and file_id == filename.split('.')[1]:
+            return CACHE_UPLOADS_PATH / filename
+
+
+def validate_git_url(git_url):
+    """Parse Git URL."""
+    if not git_url or not is_git_url.match(git_url):
+        return None
+
+    name = git_url.split('/')[-1].split('.')[0]
```
rokroskar: use https://github.com/SwissDataScienceCenter/renku-python/blob/master/renku/core/models/git.py#L132 instead

renku/service/views/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+def find_file(user, file_id):
+    """Search for cached file."""
+    for filename in os.listdir(CACHE_UPLOADS_PATH):
+        metadata = jwt.decode(filename, JWT_KEY, algorithms=[JWT_ALGORITHM])
+        if metadata['user'] == user and file_id == filename.split('.')[1]:
+            return CACHE_UPLOADS_PATH / filename
+
+
+def validate_git_url(git_url):
+    """Parse Git URL."""
```
rokroskar: What is this function doing? The name says validate, but the docstring says parse.

renku/service/views/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+    return name
+
+
+def validate_project_id(user, project_id):
+    """Validate if given project id is in cache."""
+    if project_id and (CACHE_PROJECTS_PATH / user / project_id).exists():
+        return CACHE_PROJECTS_PATH / user / project_id
+
+
+def repo_sync(repo_path, remote_name):
```
rokroskar: shouldn't this go under `cache`?

renku/service/views/cache.py
```diff
@@ -0,0 +1,335 @@
...
+                  list of all files for a given user.
+
+            error:
+                description: |
+                  Contains `code` and `message` describing the error.
+
+    """
+    if request.method == 'GET':
+        files = []
+        for file_name in os.listdir(CACHE_UPLOADS_PATH):
```
rokroskar: use `os.walk` instead

renku/service/views/__init__.py
rokroskar: In a deployment you would then save the DB on the persistent volume? Why not just use postgres then?

renku/service/views/datasets.py
```diff
@@ -0,0 +1,325 @@
...
+
+    """
+    if request.method == 'GET':
+        dataset_name = request.args.get('dataset_name')
+        project_id = request.args.get('project_id')
+        file_id = request.args.get('file_id')
+        remote_name = request.args.get('remote_name', 'origin')
+
+        local_path = find_file(user, file_id)
+        if not local_path or not local_path.exists():
```
rokroskar: in which case would `local_path.exists()` return `False`?

renku/service/views/datasets.py
```diff
@@ -0,0 +1,325 @@
...
+            return jsonify(
+                error={
+                    'code': INVALID_PARAMS_ERROR_CODE,
+                    'message': 'invalid file_id argument'
+                }
+            )
+
+        pwd = validate_project_id(user, project_id)
```
rokroskar:
```suggestion
        project_dir = validate_project_id(user, project_id)
```

renku/service/views/datasets.py
```diff
@@ -0,0 +1,325 @@
...
+        if not pwd:
+            return jsonify(
+                error={
+                    'code': INVALID_PARAMS_ERROR_CODE,
+                    'message': 'invalid project_id argument',
+                }
+            )
+
+        cwd = os.getcwd()
+        os.chdir(str(pwd))
```
rokroskar: use `LocalClient`

renku/service/views/datasets.py
```diff
@@ -0,0 +1,325 @@
...
+        try:
+            add_file([str(local_path)],
+                     dataset_name,
+                     use_external_storage=False)
+            if not repo_sync(pwd, remote_name):
+                return jsonify(
+                    error={
+                        'code': INTERNAL_FAILURE_ERROR_CODE,
+                        'message': 'remote name not found'
```
rokroskar: The push could fail for other reasons - probably best to forward the git error?

renku/service/views/datasets.py
```diff
@@ -0,0 +1,325 @@
...
+        return jsonify(result={
+            'dataset': dataset_name,
+            'file_id': file_id,
+        })
+
+
+@requires_identity
+def create_dataset_view(user):
+    """
+    Create a new dataset in cloned repository.
```
rokroskar:
```suggestion
    Create a new dataset in a project.
```

renku/service/views/datasets.py
```diff
@@ -0,0 +1,325 @@
...
+
+    ---
+    get:
+        tags:
+        - "datasets"
+
+        summary: Create a new dataset with cached projects.
```
rokroskar:
```suggestion
        summary: Create a new dataset in a project.
```

renku/service/views/datasets.py
```diff
@@ -0,0 +1,325 @@
...
+
+            - name: dataset_name
+              in: query
+              description: Dataset name.
+              schema:
+                type: string
+              required: true
+
+            - name: remote_name
+              in: query
```
rokroskar: add `default` value

renku/service/views/datasets.py
```diff
@@ -0,0 +1,325 @@
...
+        if not pwd:
+            return jsonify(
+                error={
+                    'code': INVALID_PARAMS_ERROR_CODE,
+                    'message': 'invalid project_id argument',
+                }
+            )
+
+        cwd = os.getcwd()
+        os.chdir(str(pwd))
```
rokroskar: use `LocalClient`

renku/service/views/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+    return is_pushed
+
+
+def requires_identity(f):
+    """Wrapper which indicates that route requires user identification."""
+    # noqa
+    @wraps(f)
+    def decorated_function(*args, **kws):
+        """Represents decorated function."""
+        user = request.headers.get('Authorization', '').split(' ')[-1]
```
rokroskar: We should require a JWT from keycloak, validate it, and use the `sub` claim as the user-id.

rokroskar:
```suggestion
        user = request.headers.get('Authorization', '').split('Bearer')[-1]
```

renku/service/config.py
jsam: Because we want security?

renku/service/views/__init__.py
jsam: Why? It has nothing to do with cache but with project management.

renku/service/views/cache.py
```diff
@@ -0,0 +1,335 @@
...
+        file = request.files['file']
+        if file.filename == '':
+            return jsonify(
+                error={
+                    'code': INVALID_PARAMS_ERROR_CODE,
+                    'message': 'no file selected for uploading'
+                }
+            )
+
+        if file and allowed_file(file.filename):
```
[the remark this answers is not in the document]
jsam: I don't understand. Why don't you want security and control? :(

renku/service/views/cache.py
```diff
@@ -0,0 +1,335 @@
...
+        force = 'force' in request.args
+
+        name = validate_git_url(git_url)
+        git_url = urlparse(git_url)
+        url_with_auth = '{0}:{1}@{2}'.format(
+            git_username, git_access_token, git_url.netloc
+        )
+        git_url = git_url._replace(netloc=url_with_auth).geturl()
+
+        if not name:
```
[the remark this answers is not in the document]
jsam: Sure, let's discuss this.

renku/service/views/datasets.py
```diff
@@ -0,0 +1,325 @@
...
+        project_id = request.args.get('project_id')
+        pwd = validate_project_id(project_id)
+        if not pwd:
+            return jsonify(status=2, message='invalid project_id argument')
+
+        cwd = os.getcwd()
+        os.chdir(str(pwd))
+
+        datasets = [{
+            'id': ds['_label'],
```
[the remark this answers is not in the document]
jsam: Actually, the label is ID, cause I don't want to use URL structure as an ID but a UUID. We can encode and hash `_id` to ensure that it's only one data type. Using DOI as `_id` leads to all sorts of problems, this should have it's own field and keep `_id` as 1 data type and not multiple.

renku/core/commands/client.py
```diff
@@ -63,8 +65,17 @@ def pass_local_client(
         )
 
     def new_func(*args, **kwargs):
-        ctx = click.get_current_context()
-        client = ctx.ensure_object(LocalClient)
+        ctx = click.get_current_context(silent=True)
+        if not ctx:
```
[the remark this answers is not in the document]
jsam: Yes, it's a patch ... We should probably come up with a better solution for this.

renku/service/views/datasets.py
```diff
@@ -0,0 +1,325 @@
...
+
+    """
+    if request.method == 'GET':
+        project_id = request.args.get('project_id')
+        pwd = validate_project_id(project_id)
+        if not pwd:
+            return jsonify(status=2, message='invalid project_id argument')
+
+        cwd = os.getcwd()
+        os.chdir(str(pwd))
```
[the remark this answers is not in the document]
jsam: Because `LocalClient` is not publicly exposed to the interface. That abstraction is heavily broken and should not be used outside of the internals of the core. The core module exposes the interface to work with which are designed for this purpose. We could think of adding cd as an optional flag to those methods.

renku/service/views/datasets.py
```diff
@@ -0,0 +1,325 @@
...
+
+        datasets = [{
+            'id': ds['_label'],
+            'name': ds['name'],
+            'version': ds['version'],
+            'tags': ds['tags']
+        } for ds in json.loads(dataset_parent(None, 'data', 'json-ld'))]
```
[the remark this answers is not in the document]
jsam: I agree that interfaces that core module expose should be updated for readability and ease of use.

I disagree that it's hacky, to me your proposal seems extremely hacky cause it's using a broken abstraction from the core module which is not even exposed for external use. `LocalClient` should not be used outside of the core module. It's a broken abstraction and exposing it for external module use will lead to a lot of hacky solutions and problems.

On top of that, we should not deviate the usages from CLI and SVC, they need to stay the same to ensure that we can maintain things easier.

renku/service/views/datasets.py
```diff
@@ -0,0 +1,325 @@
...
+        pwd = validate_project_id(project_id)
+        if not pwd:
+            return jsonify(status=2, message='invalid project_id argument')
+
+        cwd = os.getcwd()
+        os.chdir(str(pwd))
+
+        files = [
+            ds for ds in
+            json.loads(list_files(dataset_name, None, None, None, 'json-ld'))
```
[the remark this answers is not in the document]
jsam: No.

renku/service/views/datasets.py
```diff
@@ -0,0 +1,325 @@
...
+    ---
+    get:
+        tags:
+        - "datasets"
+
+        summary: Add a file from cache to cloned repository.
+        description: Add a file from cache to cloned repository
+
+        parameters:
+            - name: dataset_name
```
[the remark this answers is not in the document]
jsam: Because this is not what the interface supports. When you add a file, you specify a dataset name and not dataset id (which we can argue that is broken due to DOI's in there).

renku/service/views/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+def allowed_file(filename):
+    """Check if filename is allowed."""
+    has_extension = '.' in filename
+    is_allowed = filename.rsplit('.', 1)[1].lower() in ALLOWED_EXTENSIONS
+    return has_extension and is_allowed
+
+
+def find_file(user, file_id):
+    """Search for cached file."""
+    for filename in os.listdir(CACHE_UPLOADS_PATH):
```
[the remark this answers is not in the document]
jsam: Thanks, but I need strings not `PosixPaths` for iterations. There is no point in using `iterdir` and then cast every object we get into a string, it's unnecessary. So if there's no other reason for your proposal than readability, I respectfully disagree with this - runtime should never be sacrificed just because somebody finds one way of writing nicer than the other way.

renku/service/views/cache.py
jsam: 1) No. See below.

2) Let's talk about this, there is a tradeoff.

renku/service/views/datasets.py
jsam: This is just a defensive pattern. The potential case would be: A local path is cached object, a function can return you a path of a file which in that time was in a process of cleanup by a TTL cleanup strategy. During `find_file` processing invocation and found value, cleanup strategy executor can easily delete a file that he found. By the time function returns it can happen that path no longer exists, that's why I added this additional check.

renku/service/views/datasets.py
jsam: No, we don't want to deviate implementations and how we use the core module. See above.

renku/service/entrypoint.py
jsam: As discussed there is a reason for not opting to blueprints immediately.

renku/service/entrypoint.py
```diff
@@ -0,0 +1,129 @@
...
+    """Register routes to given app instance."""
+    swaggerui_blueprint = get_swaggerui_blueprint(
+        SWAGGER_URL, API_URL, config={'app_name': 'RenkuSvc'}
+    )
+
+    app.register_blueprint(swaggerui_blueprint, url_prefix=SWAGGER_URL)
+
+    app.add_url_rule(
```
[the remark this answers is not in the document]
jsam: See above.

renku/service/views/__init__.py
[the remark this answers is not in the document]
jsam: Thanks, but no. See below.

renku/service/views/__init__.py
jsam: I tried to stay as thin as possible before committing to external processes. There was one important design constraint under which I understood we have to stay and that is that the service should be lightweight as possible. Adding the database and whole distributed processing stack is something I would not consider as "staying lightweight". 

Let's discussed this with next steps.

renku/service/views/__init__.py
jsam: This proposal is quite questionable since it would break the single responsibility principle. I don't think that every service in the system should validate a token which it receives. Infrastructure should be set up in a way where the only gateway would validate a request and ensure that incoming request into the microservice is already valid and proper. Authorization header in this service is here to ensure identity and not authentication or authorization. The goal is to identify not authorize.

jsam: What do we get with this change? It would cause a problem cause all user bounded directories would start with space, which would make a whole structure just plain ugly.

renku/service/views/datasets.py
jsam: No. See above.

renku/service/config.py
rokroskar: it's not very easy to know ahead of time all the possible extensions people will want to upload. Besides, I don't think enforcing something as easily changed as a file extension is a particularly robust security practice.

renku/service/views/cache.py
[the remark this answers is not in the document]
rokroskar: This doesn't actually add any security, but it does make it so you have to change the source code every time someone wants to upload a different sort of file.

renku/service/views/datasets.py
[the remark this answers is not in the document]
rokroskar: Then don't call it `id`, call it `label`. Alternatively, you can use `identifier`. I would prefer that we don't mix names here if not absolutely necessary.

renku/service/views/datasets.py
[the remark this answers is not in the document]
rokroskar: It's an unnecessary json roundtrip - you are deserializing from disk, serializing in `dataset_parent`, then deserializing in this function and serializing again to return json. You also have the potential to get a raised `MigrationRequired` exception which afaict is not handled? I'm not sure why `LocalClient` should not be used outside of core?

renku/service/views/__init__.py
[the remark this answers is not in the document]
rokroskar: Sorry if it wasn't clear - my point was not really about readability but that `os.listdir` returns a list while `Path().iterdir()` returns a generator.

renku/service/views/__init__.py
rokroskar: Right, but `clone_repository` is also under `cache`?

renku/service/views/__init__.py
rokroskar: The space can be stripped, my point was more that the header will include something other than the token in the value and in most APIs I've seen using JWTs this is `Bearer`.

rokroskar: Sure that's fine. But you are now using the user's token as the user-id, and I don't understand how this is expected to work? What happens when the token changes? I agree that validating it here is perhaps not the right place - but we could still require that it is a JWT with a `sub` claim which is then used for the user id.

renku/service/views/cache.py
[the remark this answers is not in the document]
rokroskar: Now there is :) 

https://github.com/SwissDataScienceCenter/renku-python/issues/825

renku/service/views/cache.py
```diff
@@ -0,0 +1,335 @@
...
+              required: true
+
+            - in: query
+              name: git_username
+              description: Username to access remote git repository.
+              schema:
+                type: string
+              required: true
+
+            - in: query
```
rokroskar: I don't think you want tokens in query parameters? It's going to be all over the logs. You could request the access token from the gateway?

renku/service/serializers/cache.py
```diff
@@ -0,0 +1,128 @@
...
+
+    override_existing = fields.Boolean(missing=False)
+    unpack_archive = fields.Boolean(missing=False)
+
+
+class FileUploadResponse(Schema):
+    """Response schema for file upload."""
+
+    file_id = fields.String(missing=lambda: uuid.uuid4().hex)
+    timestamp = fields.Integer(missing=time.time() * 1e+3)
```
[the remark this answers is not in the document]
jsam: `timestamp` is in milliseconds and `file_size` is bytes.

I added the comment next to the fields.

renku/service/views/cache.py
```diff
@@ -0,0 +1,281 @@
...
+              name: unpack_archive
+              description: Determines if renku should work with archive.
+              schema:
+                type: bool
+              required: false
+
+        responses:
+            result:
+                description: Contains `file_id` reference.
```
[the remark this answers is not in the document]
jsam: This is also deleted now since it's auto-generated via `flask-apispec`

renku/core/commands/dataset.py
```diff
@@ -143,9 +143,12 @@ def add_file(
     destination='',
     ref=None,
     with_metadata=None,
-    urlscontext=contextlib.nullcontext
+    urlscontext=contextlib.nullcontext,
+    use_external_storage=False
 ):
     """Add data file to a dataset."""
+    client.use_external_storage = use_external_storage
```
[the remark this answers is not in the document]
jsam: I think it doesn't change defaults, it just gives the ability to control that behavior which is needed if the repo does not have enabled external storage and we are trying to add a file to it. Without this, we can't add a files to projects created from the UI (which defaults to non-LFS enabled projects).

I guess we should streamline and define this behavior a little bit better. I'm not super happy with this solution either, it was a quick fix to so the project created from the UI would work.

renku/service/entrypoint.py
```diff
@@ -0,0 +1,102 @@
...
+                openapi_version=OPENAPI_VERSION,
+                version=API_VERSION,
+                plugins=[MarshmallowPlugin()],
+            ),
+        'APISPEC_SWAGGER_URL': API_SPEC_URL,
+    })
+    app.register_blueprint(cache_blueprint)
+    app.register_blueprint(dataset_blueprint)
+
+    swaggerui_blueprint = get_swaggerui_blueprint(
```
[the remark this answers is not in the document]
jsam: I have no idea, this is from their documentation. To what do you want me to change it to?

renku/core/commands/dataset.py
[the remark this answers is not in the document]
rokroskar: how does the UI default to non-LFS projects? Whether LFS is enabled or not is defined in the git config, either local or global. If you use `renku clone` LFS should be enabled by default.

renku/service/entrypoint.py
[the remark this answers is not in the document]
rokroskar: aha! I thought swagger -> openAPI but I could be wrong?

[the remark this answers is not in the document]
jsam: OpenAPI is the spec, swagger and redoc are just frontends for the spec afaik.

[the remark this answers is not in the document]
rokroskar: Ah ok so they changed the spec name but not the tooling. 👍

Dockerfile.svc
```diff
@@ -0,0 +1,18 @@
+FROM python:3.7-alpine
+
+RUN apk add --update --no-cache alpine-sdk g++ gcc linux-headers libxslt-dev python3-dev build-base openssl-dev libffi-dev git && \
```
rokroskar: needs `bash` as well - I think for whichever `gitpython` commands need the shell

rokroskar:
I tested this together with https://github.com/SwissDataScienceCenter/renku-python/pull/843 and https://github.com/SwissDataScienceCenter/renku-gateway/pull/181 and it seems to work as advertised. I would remove the `WIP` and merge it after review. The one outstanding issue is whether we want to keep this as a part of the `renku-python` repo for now or if we want to go ahead and move it to its own repo right away. One possibility would be to move it to a stand-alone project but add `renku-python` as a submodule for the time being to make co-development easier. @Panaetius @jsam @mohammad-sdsc what do you think?

renku/core/management/clone.py
```diff
@@ -33,23 +34,46 @@ def clone(
...
+            ) from e
+
+        raise e
+
+    if config:
+        config_writer = repo.config_writer()
+
+        for key, value in config.items():
+            key_path = key.split('.')
+            if len(key_path) != 2:
```
Panaetius: What's the reasoning behind this? There's quite a few git config settings that would have `len(key_path) > 2`, like `filter.lfs.clean=` or `mergetool.meld.cmd`

renku/service/views/cache.py
```diff
@@ -0,0 +1,247 @@
...
+            file_path.unlink()
+        else:
+            return jsonify(
+                error={
+                    'code': INVALID_PARAMS_ERROR_CODE,
+                    'reason': 'file exists',
+                }
+            )
+
+    file.save(str(file_path))
```
Panaetius: if there's any exception past this point (like extraction failing), the file will be on disk but not in the cache, leaving the disk in a dirty state. Would be cleaner to delete the file if there's a failure further down from here.

jsam: Right, the other way around is also the same problem so it's not really much cleaner since both solutions would use some deferring mechanism. Reason why I din't opt into complicating this here is that there is a more general solutions to this which (one of) is described in #851.

renku/core/management/clone.py
jsam: Right, but those are git extensions configuration options which we don't want to have configurable here (cause often we are influencing them via ENV variables - convention I'm happy to break if we decide to change it everywhere). Passed key is in format `<section>.<key>` so in this simple check I'm just making sure that you are setting git only configuration.

rokroskar: The reason the LFS flags need to be set via environment variables is because they are used on checkout - afaik it's not possible to set them in the repo config before the repo is cloned/instantiated. For other options that might not be the case.

Issue #789, opened 2019-11-01
feat(datasets): should not allow to created nested datasets


rokroskar:
## Describe the bug
`renku migrate datasets` crashes for project https://dev.renku.ch/projects/rokroskar/vom_natt:

```
Traceback (most recent call last):
  File "[...]/renku/core/commands/migrate.py", line 32, in migrate_datasets
    migration(client) is not False for migration in STRUCTURE_MIGRATIONS
  File "[...]/renku/core/commands/migrate.py", line 32, in <listcomp>
    migration(client) is not False for migration in STRUCTURE_MIGRATIONS
  File "[...]/renku/core/commands/checks/migration.py", line 148, in migrate_datasets_pre_v0_3
    ref.set_reference(new_path)
  File "[...]/renku/core/models/refs.py", line 83, in set_reference
    self.path.parent.mkdir(parents=True, exist_ok=True)
  File "[...]/pathlib.py", line 1258, in mkdir
    self._accessor.mkdir(self, mode)
FileExistsError: [Errno 17] File exists: '/Users/rok/Projects/renku-test-projects/vom_natt/.renku/refs/datasets/VOM_output/AdelaideRiver'
```

The reason is that there are nested datasets in the repository and migration tries to create a symlink with the same name as an existing directory in `refs/datasets/`. We should not allow the user to create nested datasets and fail with a useful error message if they try to do this.

mohammad-sdsc:
I could not reproduce this error using the latest commit in your branch. This error happens if I checkout the commit on June 13th, 2019 (`e4c52edd269eb740a1783d5d0cb82fbc3d334323`). Is this where this bug happens?

rokroskar:
ah yes, in the meantime I've fixed the project. :) If you check out 0a73a21 or earlier it should give the error.

mohammad-sdsc:
I also noticed another weird issue with this project: if I run `renku migrate datasets` on the last commit, `c3d365224a7d5b36098d7d117bd242ea378785a6`, the migrations succeeds but the metadata file `.renku/datasets/c7fe0b80-6064-46a3-a175-71d47d1d2a86/metadata.yml` contains some invalid entries:
```
...
  - VOM_output/AdelaideRiver
...
```

rokroskar:
yes I tried to fix it but didn't do it fully - I moved the sub-datasets into the top level `data/` directory but I didn't change the metadata. We should probably have this done as a part of `renku mv` command.

[...]

PR #796, opened 2019-11-05
feat: avoid creation of nested datasets

mohammad-sdsc:
# Description

In previous Renku version, we could create nested datasets (e.g. `new/2018`, `new/2019`). In recent, Renku versions this is not possible, however, the error message printed to the user is not useful. This PR prints a useful error message to users.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/789

### DIFF
--- a/renku/core/management/datasets.py
+++ b/renku/core/management/datasets.py
@@ -129,7 +129,7 @@ class DatasetsApiMixin(object):
                 path.parent.mkdir(parents=True, exist_ok=False)
             except FileExistsError:
                 raise errors.DatasetExistsError(
-                    'Dataset with UUID {} exists'.format(identifier)
+                    'Dataset with reference {} exists'.format(path.parent)
                 )
 
             with with_reference(path):
--- a/renku/core/management/datasets.py
+++ b/renku/core/management/datasets.py
@@ -115,6 +115,12 @@ class DatasetsApiMixin(object):
         clean_up_required = False
 
         if dataset is None:
+            # Avoid nested datasets: name mustn't have '/' in it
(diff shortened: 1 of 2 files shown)


renku/core/management/datasets.py
```diff
@@ -123,7 +129,7 @@ def with_dataset(self, name=None, identifier=None):
                 path.parent.mkdir(parents=True, exist_ok=False)
             except FileExistsError:
                 raise errors.DatasetExistsError(
-                    'Dataset with reference {} exists'.format(path.parent)
+                    'Dataset with UUID {} exists'.format(identifier)
```
jsam: I think datasets imported from 3rd party and which are stamped with DOI will have identifiers equal to DOI structure and not UUID. So let's keep the wording as it was?

tests/cli/test_datasets.py
```diff
@@ -182,6 +182,14 @@ def test_dataset_create_exception_refs(runner, project, client):
     assert 'a' in result.output
 
 
+@pytest.mark.parametrize('name', ['dataset/new', '/dataset', 'dataset/'])
+def test_dataset_name_is_valid(client, runner, project, name):
+    """Test dataset name has no '/' character to avoid nested datasets."""
+    result = runner.invoke(cli, ['dataset', 'create', name])
+    assert result.exit_code == 2
```
```suggestion
    assert 2 == result.exit_code
```

renku/core/management/datasets.py
jsam: This might be a bug though, which we should revert. Let's sync tomorrow.

PR #807, opened 2019-11-08
make Creator a subclass of Person

### LINKED ISSUE #793: missing schema:name for `prov:agent -> schema:Person`
**Describe the bug**

`models.datasets.Creator` includes the property `schema:name` but for `models.provenance.Person` this is missing resulting in inconsistencies from sparql queries.

### LINKED ISSUE #806: renku log fails on migrated datasets
**Describe the bug**
If `renku log` is called on a dataset that has been migrated and its metadata location changed, the log command fails because the original (now deleted/removed) file is gone. 

**To Reproduce**
Clone https://dev.renku.ch/gitlab/virginiafriedrich/datasets-test.git and run `renku migrate datasets`. Then run `renku log --format nt`

rokroskar:
This PR refactors `Creator` to be a subclass of `Person`. (closes #793)

### DIFF
--- a/renku/core/models/provenance/agents.py
+++ b/renku/core/models/provenance/agents.py
@@ -19,6 +19,8 @@
 
 import re
 
+from attr.validators import instance_of
+
 from renku.core.models import jsonld as jsonld
 from renku.version import __version__, version_url
 
@@ -31,6 +33,7 @@ from renku.version import __version__, version_url
     context={
         'schema': 'http://schema.org/',
         'prov': 'http://www.w3.org/ns/prov#',
+        'rdfs': 'http://www.w3.org/2000/01/rdf-schema#'
     },
     frozen=True,
     slots=True,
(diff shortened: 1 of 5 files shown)


renku/core/models/provenance/agents.py
```diff
@@ -31,28 +31,39 @@
...
     },
     frozen=True,
     slots=True,
 )
 class Person:
     """Represent a person."""
 
-    name = jsonld.ib(context='rdfs:label')
-    email = jsonld.ib(context='schema:email')
+    name = jsonld.ib(context='schema:name')
```
mohammad-sdsc: Should we also add `validator` and `kw_only` here, like it was in `Creator`?
```
...
kw_only=True,
validator=instance_of(str),
...
```

rokroskar: good point, thanks!

Issue #709, opened 2019-09-25
`renku log` to generate relevant triples on dataset removal


jachro:
As a Renku user, when I remove a dataset from my project, I'd like this dataset not to be visible among its datasets. However, it should still be visible on all the other projects which either forked from my project earlier when the dataset was present or imported it.

**Assumptions:**
* the issue is played after the [dataset immutability issue](https://github.com/SwissDataScienceCenter/renku-python/pull/1142).

**Acceptance criteria:**
* Renku CLI to create a new version of the removed dataset with additional `prov:invalidatedBy` link to the commit activity where the deletion happened;
* the new version of the dataset should be done in a way described in the dataset immutability issue.

rokroskar:
What we need to specify is what this should include. The easiest seems to be something like [`invalidated`](https://www.w3.org/TR/prov-o/#invalidated) from prov-o. From that definition:

> Invalidation is the start of the destruction, cessation, or expiry of an existing entity by an activity. The entity is no longer available for use (or further invalidation) after invalidation. Any generation or usage of an entity precedes its invalidation.

There is also the inverse `invalidatedBy` which would make an outgoing edge from the entity to the commit, which may be easier in your graph query. But if you just filter on the absence of this predicate I guess it doesn't matter.

jachro:
I quite like the idea of `invalidatedBy` as it would leave a trace of the removal operation. However, if its direction is from an entity (a dataset in this case) to a commit, the problem I can see here is if a dataset would be linked to another project (which I believe would be the case when a dataset was imported from another Renku project or whatever else). So having this kind of trace would be nice but from the other side, a more general question is appearing whether our KG should work as an audit log or rather simply be a source of truth about what's available now. I can imagine the second option could be easier to maintain, smaller in size and ... maybe good enough and what we would expect from it? Going this direction, the only thing we'd have to do for the dataset removal is just deleting the `isPartOf` triple from the dataset and eventually do a complete removal of a dataset with its parts, creators and so on, if there's no other `isPartOf` on it.

Ideas, thoughts?

rokroskar:
Right I guess that would be similar to the git/github behavior where it eventually garbage collects objects when there are no more references to them when history gets overwritten. But in this case there was no force-push, the object was just deleted from the working tree. So if the user wanted to see the state of the project before the delete, they couldn't. Therefore I think we can only remove on a force-push.

jachro:
[...]

PR #808, opened 2019-11-08
fix: ignore deleted paths in generated entities

### LINKED ISSUE #806: renku log fails on migrated datasets

rokroskar:
This PR fixes a problem with running `renku log` on deleted entities by enforcing a git diff filter to exclude deleted files from the paths listed in the paths affected by a commit. We should change this behavior eventually to add `isInvalidatedBy` triple to point to the invalidated entities. 

closes #806
related to https://github.com/SwissDataScienceCenter/renku-python/issues/709


renku/core/models/provenance/activities.py
```diff
@@ -192,6 +192,11 @@ def paths(self):
         index = set()
 
         for file_ in self.commit.diff(self.commit.parents or NULL_TREE):
+            # ignore deleted files (note they appear as ADDED)
+            # in this backwards diff
+            # TODO: set `deprecatedBy` for deleted paths
+            if file_.change_type == 'A':
```
jsam: Why do they appear as 'A'? According to [this](https://github.com/gitpython-developers/GitPython/blob/master/git/diff.py#L170) it says it's 'D'.

rokroskar: yeah it's confusing - it's because of the way the diff is being generated here. Since the file is deleted in the current commit and the diff direction is going backwards in history it then appears as added.

PR #809, opened 2019-11-11
feat: add project initialization from template

### LINKED ISSUE #546: Use the same project templates as the UI

We need to standardize project creation and use the same templates as the UI. 

We should by default use the same repository of templates as the UI does. Optionally, the user could provide a URL to a renku template. The user would have options like:

```bash
renku init --template-name [template name: e.g., basic-r, basic-python] renku-project
renku init --template https://github.com/<path-to-repo-with-template> renku-project
```

If multiple directories with templates exist, the user would be presented with a choice.

The templates in the https://github.com/SwissDataScienceCenter/renku-project-template project should be shipped with the CLI and can be used even if there is no network connection.

lorenzo-cavazzi:
# Description

This PR adds the possibility to initialize a Renku project starting from a remote [Jijna2](https://jinja.palletsprojects.com/en/2.10.x/) template.

It's still missing:
* [required] `--force` parameter
* [required] `git lfs`
* Embed basic templates
* Ask for the template name after fetching the remote repository
* [separate issue?] Modify templates to properly handle the description field
* [separate issue?] Allow users to provide custom variables for jinja template files

I am not sure how I should add the `Creator` data, I guess modifying the templates is the best option

Closes #546


renku/core/commands/init.py
```diff
@@ -0,0 +1,100 @@
...
+):
+    """Fetch the remote template and checkout the relevant folder.
+
+    Returns:
+        The path with template files.
+
+    """
+    # clone the repo locally without checking out files
+    template_repo = git.Repo.clone_from(
+        url, tempdir, no_checkout=True, depth=1
```
rokroskar: 👍

renku/core/commands/init.py
```diff
@@ -0,0 +1,100 @@
...
+from datetime import datetime, timezone
+from pathlib import Path
+
+import git
+
+from renku.core.management.config import RENKU_HOME
+
+
+def fetch_remote_template(
+    url, folder, branch='master', tempdir=Path(tempfile.mkdtemp())
```
rokroskar: This will set `tempdir` on import and it will never change. Is that intended? If not, `tempdir` should be `None` and you should set it in the body of the function if needed.

renku/core/commands/init.py
```diff
@@ -0,0 +1,100 @@
...
+        The path with template files.
+
+    """
+    # clone the repo locally without checking out files
+    template_repo = git.Repo.clone_from(
+        url, tempdir, no_checkout=True, depth=1
+    )
+
+    # check branch
+    template_branches = [branch.name for branch in template_repo.branches]
```
rokroskar: you could use a set comprehension here instead for faster lookup

renku/core/commands/init.py
```diff
@@ -0,0 +1,100 @@
...
+    if branch not in template_branches:
+        raise ValueError(
+            f'Branch "{branch}" doesn\'t exist in template "{url}"'
+        )
+
+    # check folder
+    # TODO: update logic: look at `manifest.yaml` for templates metadata
```
rokroskar: 👍 on this TODO :)

renku/core/commands/init.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+    # clone the repo locally without checking out files
+    template_repo = git.Repo.clone_from(
+        url, tempdir, no_checkout=True, depth=1
+    )
+
+    # check branch
+    template_branches = [branch.name for branch in template_repo.branches]
+    if branch not in template_branches:
+        raise ValueError(
```
rokroskar: I'm pretty sure these will result in `Aaaaaaaaaah` style messages in the user's console. We can probably make a new renku error e.g. `TemplateError` to handle this better (cc @mohammad-sdsc @Panaetius - can you advise?)

renku/core/management/repository.py
```diff
@@ -417,3 +419,42 @@ def init_repository(self, name=None, force=False):
             metadata.updated = datetime.now(timezone.utc)
 
         return str(path)
+
+    def init_empty_repository(self, force=False):
```
rokroskar: Why not use/reuse `init_repository`? There is now some code duplication here.

renku/core/management/repository.py
```diff
@@ -417,3 +419,42 @@ def init_repository(self, name=None, force=False):
...
+            raise FileExistsError(self.repo.git_dir)
+
+        path = self.path.absolute()
+        self.repo = Repo.init(str(path))
+
+        return str(path)
+
+    def import_from_template(self, template_path, metadata, force=False):
+        """Render template files from a template directory."""
+        # adding author info? Creator.from_git(self.repo)
```
rokroskar: That line in `init_repository` is just a hacky way to force a check on the git configuration. The "creator" is just the person making the initial commit so it's not necessary to add it anywhere manually.

renku/core/commands/init.py
```diff
@@ -0,0 +1,100 @@
...
+            )
+    for file in required_files:
+        if not Path(template_path, file).is_file():
+            raise ValueError(
+                f'File {file} is required for the template to be valid'
+            )
+    return True
+
+
+def create_from_template(
```
rokroskar: since this needs `client` it should probably be a part of the `RepositoryApiMixin`

renku/cli/init.py
```diff
@@ -189,3 +193,101 @@ def init(ctx, client, directory, name, force, use_external_storage):
...
+    '--template-branch',
+    default='master',
+    help='Change templates target branch.',
+)
+@click.option('--force', is_flag=True, help='Override target path.')
+@click.option('--description', help='Describe your project.')
+@option_use_external_storage
+@pass_local_client
+@click.pass_context
+def template_init(
```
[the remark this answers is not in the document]
lorenzo-cavazzi: right, the idea is to embed the current templates with the package so we can have a single `init` command. I temporarily split into 2 commands until I complete this step

conftest.py
```diff
@@ -337,6 +337,15 @@ def add_client(doctest_namespace):
     doctest_namespace['client'] = LocalClient(path=tempfile.mkdtemp())
 
 
+@pytest.fixture
+def local_client():
```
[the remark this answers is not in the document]
lorenzo-cavazzi: that one already invokes `renku init`, so it's not good for testing repository methods like `init_empty_repository`

renku/core/commands/init.py
lorenzo-cavazzi: I think the code in `renku/cli/init.py@290` prevents the `Aaaaaaaaaah` from displaying and shows a nicer output instead

renku/core/management/repository.py
lorenzo-cavazzi: right, I changed the logic quite a bit, I'll replace the `init_repository` once this PR is ready

renku/core/commands/init.py
lorenzo-cavazzi: This is definitely not the expected behavior, thanks for pointing this out!

renku/core/commands/init.py
lorenzo-cavazzi: Changed it, I hope it's in line with what you had in mind. Please let me know if there is an obvious better way to do that

renku/core/commands/init.py
lorenzo-cavazzi: This is essentially a wrapper around 2 functions actually part of the `RepositoryApiMixin` (`import_from_template` and `init_empty_repository`) plus the definition of an extra metadata (creation date). I thought it makes sense to have this function here instead of `cli/init.py` since I want to invoke the same logic from both the cli and the service/ui. Does this make sense?

lorenzo-cavazzi:
Thanks for the extensive review.
The duplication is only temporary, I didn't want to disrupt the current init in the PR so that it's easier to test, but later I will leave only this as "the" init command.

I added some comments inline and addressed the basic points. Feel free to un-resolve them if the solution is not correct. I'll implement the bigger ones in the next days.

My original idea for the templates was to embed a specific version when running the setup. It wouldn't be that different from what we currently do in the ui https://github.com/SwissDataScienceCenter/renku-ui/blob/9d10a68c9f4c7bf9a009a35ee98c6147d3e1d821/helm-chart/renku-ui/values.yaml#L61. My point is that we may introduce changes in the templates (even moving from jinja to something else) and we don't want to break all the old installation of `renku-python` at once by automatically downloading new versions. Does this make sense?

rokroskar:
Yeah makes sense - we should be able to download the last tagged version of the template when `python setup.py install` is executed. Is that what you mean?

lorenzo-cavazzi:
Kind of, but instead of getting the latest we get the one specified in `setup.py` by some variable. In this way we bind the template version to the renku-python version.

lorenzo-cavazzi:
I have tried to address the main concerns:

* merge `init` and `template-init` to a unique command
* embed the templates  -- this happens in `setup.py` but something is still wrong according to the tests
* properly handle expected exceptions in the cli
* remove unnecessary `required=True` on options

I fear I introduced a few problems with this new commit because tests complain about the new setup and `--force` parameter on `init` doesn't always work as expected.

Other minor concerns
* The code is not very pythonic, I am sure it can be improved. Any suggestion/help is appreciated :slightly_smiling_face: 
* Should I rename `core/commands/init.py` as `template.py` since everything there is about templates? :thinking: 
* I know I have to remove the f-strings to keep 3.5 compatibility, I guess I can postpone this to the last step :man_shrugging: 
* Currently, I don't allow the user to pass custom metadata serving as Jinja variable for the template files. I guess we can leave this out for now and open a new issue :no_good_man: 


P.S. Sorry for forcing, the `master` branch diverged to much and I rebased at some point. The initial commit code has not changed.

renku/cli/init.py
```diff
@@ -95,97 +98,191 @@ def store_directory(ctx, param, value):
...
-    branch_name = None
-
-    stack = contextlib.ExitStack()
-
-    if force and client.repo:
-        msg = 'Initialized project in {path} (branch {branch_name})'
-        merge_args = ['--no-ff', '-s', 'recursive', '-X', 'ours']
-
+    def print_done():
+        click.echo(' DONE ', nl=False)
```
jsam:
```suggestion
        click.secho('OK', fg='green')
```

renku/cli/init.py
```diff
@@ -95,97 +98,191 @@ def store_directory(ctx, param, value):
...
+        click.echo(click.style('\u2713', fg='green'))
```
jsam:
```suggestion

```

renku/cli/init.py
```diff
@@ -95,97 +98,191 @@ def store_directory(ctx, param, value):
...
+
+    # verify path status
+    if len(list(client.path.glob('**/*'))) > 0:
```
jsam:
```suggestion
    if list(client.path.glob('**/*')):
```

renku/cli/init.py
```diff
@@ -95,97 +98,191 @@ def store_directory(ctx, param, value):
...
+        if not force:
+            error = FileExistsError(
+                'Folder "{0}" is not empty. Please '
+                'add --force flag to transform it into a Renku repository.'.
+                format(str(path))
+            )
+            raise click.UsageError(error)
+        else:
```
jsam:
```suggestion

```

renku/cli/init.py
```diff
@@ -95,97 +98,191 @@ def store_directory(ctx, param, value):
...
+            try:
+                commit = client.find_previous_commit(
+                    str(client.renku_metadata_path),
+                )
+                branch_name = 'pre_renku_init_' + str(commit)[:8]
```
jsam:
```suggestion
                branch_name = 'pre_renku_init_'.format(str(commit)[:7])
```

or 
```suggestion
                branch_name = 'pre_renku_init_'.format(commit.hexsha[:7])
```

renku/cli/init.py
```diff
@@ -95,97 +98,191 @@ def store_directory(ctx, param, value):
...
+                    commit=commit,
+                    merge_args=['--no-ff', '-s', 'recursive', '-X', 'ours']
+                ):
+                    click.echo(
+                        'Saving current data in branch {0}'.
+                        format(branch_name)
+                    )
+            except AttributeError:
+                click.echo('Warning! Overwriting non-empty folder.')
+            except GitCommandError as e:
```
jsam: Instead of catching it here, we should probably wrap it inside core module and then reraise renku exception.

renku/cli/init.py
```diff
@@ -95,97 +98,191 @@ def store_directory(ctx, param, value):
...
+                with client.worktree(
+                    branch_name=branch_name,
+                    commit=commit,
+                    merge_args=['--no-ff', '-s', 'recursive', '-X', 'ours']
+                ):
+                    click.echo(
+                        'Saving current data in branch {0}'.
+                        format(branch_name)
+                    )
+            except AttributeError:
```
jsam: Who's raising this?

renku/cli/init.py
```diff
@@ -95,97 +98,191 @@ def store_directory(ctx, param, value):
...
-            commit = NULL_TREE
-            branch_name = 'renku/init/root'
-            merge_args.append('--allow-unrelated-histories')
-
-        ctx.obj = client = stack.enter_context(
-            client.worktree(
-                branch_name=branch_name,
-                commit=commit,
-                merge_args=merge_args,
+        except (ValueError, GitCommandError) as e:
```
jsam: Handle this exceptions inside core module and then raise renku exceptions from them so core module raises mostly renku exceptions.

renku/cli/init.py
```diff
@@ -95,97 +98,191 @@ def store_directory(ctx, param, value):
...
+    if template:
+        template_filtered = [
+            template_elem for template_elem in template_manifest
+            if template_elem['name'] == template
+        ]
+        if len(template_filtered) == 1:
+            template_data = template_filtered[0]
+        else:
+            click.echo(
+                'The template with name "{0}" is not avilable.'.
```
jsam:
```suggestion
                'The template with name "{0}" is not available.'.
```

renku/core/commands/init.py
```diff
@@ -0,0 +1,148 @@
...
+    :param source: url or full path of the templates repository
+    :param ref: reference for git checkout - branch, commit or tag
+    :param tempdir: temporary work directory path
+    :return: template manifest Path
+    """
+    if tempdir is None:
+        tempdir = Path(tempfile.mkdtemp())
+
+    # clone the repo locally without checking out files
+    template_repo = git.Repo.clone_from(source, tempdir, no_checkout=True)
```
jsam: Handle exception and raise renku exception in case of failure.

renku/core/commands/init.py
```diff
@@ -0,0 +1,148 @@
...
+    template_repo.remotes.origin.fetch(ref)
+    try:
+        template_repo.head.reset(template_repo.commit(ref))
+    except git.exc.BadName:
+        ref = 'origin/{0}'.format(ref)
+        template_repo.head.reset(template_repo.commit(ref))
+    git_repo = git.Git(tempdir)
+
+    # checkout the manifest
+    git_repo.checkout(TEMPLATE_MANIFEST)
```
jsam: This can fail also. Should we wrap it?

renku/core/commands/init.py
```diff
@@ -0,0 +1,148 @@
...
+    """Validate a local template.
+
+    :param template_path: path of the template to validate
+    """
+    # TODO: implement a better check
+    required_folders = [RENKU_HOME]
+    required_files = ['{0}/metadata.yml'.format(RENKU_HOME), 'Dockerfile']
+    for folder in required_folders:
+        if not Path(template_path, folder).is_dir():
+            raise ValueError(
```
jsam: Create a specific exception for this which inherits from RenkuException.

renku/core/commands/init.py
```diff
@@ -0,0 +1,148 @@
...
+                'Folder {0} is required for the template to be valid'.
+                format(folder)
+            )
+    for file in required_files:
+        if not Path(template_path, file).is_file():
+            raise ValueError(
```
jsam: Should probably be also based on renku exception.

renku/core/commands/init.py
```diff
@@ -0,0 +1,148 @@
...
+    return True
+
+
+def validate_template_manifest(manifest):
+    """Validate manifet content.
+
+    :param manifest: manifest file content
+    """
+    if not isinstance(manifest, list):
+        raise ValueError((
```
jsam: Should probably be also based on renku exception.

renku/core/commands/init.py
```diff
@@ -0,0 +1,148 @@
...
+            'The repository doesn\'t contain a valid',
+            '"{0}" file'.format(TEMPLATE_MANIFEST)
+        ))
+    for template in manifest:
+        if not template['name']:
```
jsam:
```suggestion
        if not isinstance(template, dict) or 'name' not in template:
```

renku/core/commands/init.py
```diff
@@ -0,0 +1,148 @@
...
+            raise ValueError((
```
jsam: Should probably be also based on renku exception.

renku/core/commands/init.py
```diff
@@ -0,0 +1,148 @@
...
+                'Every template listed in "{0}"',
+                ' must have a name'.format(TEMPLATE_MANIFEST)
+            ))
+        for attribute in ['folder', 'description']:
+            if not template[attribute]:
+                raise ValueError((
```
jsam: Should probably be also based on renku exception.

renku/core/commands/init.py
```diff
@@ -0,0 +1,148 @@
...
+            '"{0}" file'.format(TEMPLATE_MANIFEST)
+        ))
+    for template in manifest:
+        if not template['name']:
+            raise ValueError((
+                'Every template listed in "{0}"',
+                ' must have a name'.format(TEMPLATE_MANIFEST)
+            ))
+        for attribute in ['folder', 'description']:
+            if not template[attribute]:
```
jsam:
```suggestion
            if attribute in template:
```

Otherwise, you will get `KeyError`.

renku/core/commands/init.py
```diff
@@ -0,0 +1,148 @@
...
+def read_template_manifest(folder, checkout=False):
+    """Extract template metadata from the manifest file.
+
+    :param folder: path where to find the template manifest file
+    :param checkout: checkout the template folder from local repo
+    """
+    manifest_path = folder / TEMPLATE_MANIFEST
+
+    with manifest_path.open('r') as fp:
+        manifest = yaml.safe_load(fp)
```
jsam: If you don't read and write transactionally then you probably don't need file handling via contextmanagers. You could achieve the same with 

```
manifest = yaml.safe_load(manifest_path.read_text())
```
```

File: renku/core/commands/init.py [hunk 30]
```diff
@@ -0,0 +1,148 @@
...
+        validate_template_manifest(manifest)
+
+        if checkout:
+            git_repo = git.Git(folder)
+            template_folders = [template['folder'] for template in manifest]
+            for template_folder in template_folders:
+                template_path = folder / template_folder
+                git_repo.checkout(template_path)
```
jsam: This can raise, maybe we should handle it?

renku/core/management/repository.py
```diff
@@ -375,45 +376,40 @@ def with_workflow_storage(self):
...
-            if self.repo is None:
-                self.repo = Repo.init(str(path))
-        else:
-            if self.repo is not None:
-                raise FileExistsError(self.repo.git_dir)
-
-            self.renku_path.mkdir(parents=True, exist_ok=force)
-            self.repo = Repo.init(str(path))
+        if self.repo is not None and not force:
+            raise FileExistsError(self.repo.git_dir)
```
jsam: Consider creating renku exception for it.

renku/core/management/repository.py
```diff
@@ -375,45 +376,40 @@ def with_workflow_storage(self):
...
+                # with file.open() as fr:
+                #     file_content = fr.read()
+                #     # look for the required keys
+                #     env = Environment()
+                #     parsed = env.parse(file_content)
+                #     variables = meta.find_undeclared_variables(parsed)
+
+                with open(file) as file_content:
+                    # open and parse the file
+                    template = Template(file_content.read())
```
jsam:
```suggestion
                    template = Template(file.read_text())
```

jsam: And then you can remove contextmanager.

renku/core/management/repository.py
```diff
@@ -375,45 +376,40 @@ def with_workflow_storage(self):
...
+                    rendered_content = template.render(metadata)
+                    # write content to the new location
+                    with open(destination, 'w') as output_file:
```
jsam:
```suggestion
                    destination.write_text(rendered_content)
```

renku/core/management/repository.py
```diff
@@ -375,45 +376,40 @@ def with_workflow_storage(self):
...
+                        output_file.write(rendered_content)
+            except Exception:
```
jsam: Avoid catching top-level ```Exception```, we can be more specific in this case.

renku/core/management/repository.py
```diff
@@ -375,45 +376,40 @@ def with_workflow_storage(self):
...
+                if file.is_dir():
+                    destination.mkdir(parents=True, exist_ok=True)
+                else:
+                    shutil.copy(file, destination)
```
jsam: Maybe this part needs to be wrapped in exception as well?

.gitignore
```diff
@@ -1,5 +1,8 @@
 # Automatically generated by setuptools_scm
 renku/version.py
+renku/templates
```
jsam:
```suggestion
renku/templates/
```

.gitignore
```diff
@@ -1,5 +1,8 @@
...
+renku_data
```
jsam:
```suggestion
renku_data/
```

.gitignore
```diff
@@ -1,5 +1,8 @@
...
+data
```
jsam:
```suggestion
data/
```

renku/core/commands/init.py
lorenzo-cavazzi: good catch :smile: 
I guess it's `if attribute not in template:`

renku/cli/init.py
lorenzo-cavazzi: Going for the second.
`branch_name = 'pre_renku_init_{0}'.format(commit.hexsha[:7])`

renku/cli/init.py
lorenzo-cavazzi: I ended up removing `def print_done` and using `click.secho('OK', fg='green')` straight in the code

lorenzo-cavazzi:
Thanks @jsam for the thorough review! I addressed the code suggestions (I marked them as completed since the `Outdated` label was there also on non-addressed suggestions), and I left the Exceptions suggestions undone -- I'll tackle them next week.
I tried to fix the tests, now they seem to run on Github actions (thanks @rokroskar for suggesting `--editable` mode on `pip install`) but they hang at some point. I am debugging that -- thanks to @Panaetius for pointing me in the right direction on this

rokroskar:
The current failing test is due to the call to `renku doctor` which tries to validate the metadata. This is the failure:

```
renku log --strict --format nt
Error: Validation Report
Conforms: False
Results (1):
Constraint Violation in MinCountConstraintComponent (http://www.w3.org/ns/shacl#MinCountConstraintComponent):
        Severity: sh:Violation
        Source Shape: [ sh:maxCount Literal("1", datatype=xsd:integer) ; sh:minCount Literal("1", datatype=xsd:integer) ; sh:nodeKind sh:IRI ; sh:path prov:wasInformedBy ]
        Focus Node: <https://localhost/activities/commit/36cf59e674a800930a740e33d345a3444d9c2a9b>
        Result Path: prov:wasInformedBy
```

This has nothing to do with this PR but more with either the SHACL shape or the project template...

Panaetius:
`renku doctor` doesn't test workflow information, only dataset and project structure. 

I just tested it with the `--pre` build/newest master:

```shell
$ renku --version
0.8.3.dev6
$ renku init
OK
Initialized empty project in /home/zenon/DEV/renku-test
(renku-test) [zenon@ZENON-PC renku-test]$ renku doctor
Please note that the diagnosis report is used to help Renku
maintainers with debugging if you file an issue. Use all proposed
solutions with maximal care and if in doubt ask an expert around or
file an issue. Thanks!

Everything seems to be ok.
$ renku log --strict --format nt
Error: Validation Report
Conforms: False
Results (1):
Constraint Violation in MinCountConstraintComponent (http://www.w3.org/ns/shacl#MinCountConstraintComponent):
        Severity: sh:Violation
        Source Shape: [ sh:maxCount Literal("1", datatype=xsd:integer) ; sh:minCount Literal("1", datatype=xsd:integer) ; sh:nodeKind sh:IRI ; sh:path prov:wasInformedBy ]
        Focus Node: <https://localhost/activities/commit/9696428d4954e2c062c837b63d544e7b35f70505>
        Result Path: prov:wasInformedBy

Couldn't get log: Invalid Knowledge Graph data
```

So what you found is a new bug (might be SHACL or graph generation), but `renku doctor`failing is independent of that.

The `renku log` error means it generates an Activity that doesn't have a `prov:wasInformedBy` set. This is set to the parent commit, but since it's a new project, there is only one commit (no parent). So the SHACL should be change to allow 0 `wasInformedBy`.

I looked at the `renku doctor` issue quickly with @lorenzo-cavazzi last week and then it looked like `creator` not being set when initializing from a template.

So it looks like 2 separate issues.

rokroskar:
The error above can be fixed by:

```
diff --git a/renku/data/shacl_shape.json b/renku/data/shacl_shape.json
index 31d53f2..b818851 100644
--- a/renku/data/shacl_shape.json
+++ b/renku/data/shacl_shape.json
@@ -587,7 +587,7 @@
             {
                "nodeKind": "sh:IRI",
                "path": "prov:wasInformedBy",
-               "minCount": 1,
+               "minCount": 0,
                "maxCount": 1
             },
             {
```

But `renku doctor` still fails with

```
Warning: Invalid structure of project metadata
        http://schema.org/creator: Type: http://www.w3.org/ns/shacl#MinCountConstraintComponent, Node ID: <Anonymous>
```

I can't find any blank nodes when I run `renku log`...

rokroskar:
Our messages crossed - yes it is indeed 2 separate issues. I don't know which node is missing `creator` though?

renku/cli/init.py
```diff
@@ -95,97 +99,185 @@ def store_directory(ctx, param, value):
...
 @click.argument(
-    'directory',
+    'path',
     default='.',
     callback=store_directory,
-    type=click.Path(writable=True, file_okay=False, resolve_path=True)
+    type=click.Path(writable=True, file_okay=False, resolve_path=True),
+)
+@click.option(
+    '--name',
```
rokroskar: Is this option needed? If `init` is called with an argument, e.g. `renku init my-renku-project` the name is `my-renku-project`. If it is called within an existing directory the name is the directory name. I suppose you could still give it a name explicitly, but the default _should not_ be "Renku project".  I didn't notice that this had been set before, I just realized when I was taking a closer look at the triples that are generated...

Panaetius:
it was the project node. It wasn't so much missing the creator as the creator being in an old format ( single string vs Creator object data with affiliation etc.). 

`renku doctor` is special in that it validates `*.yml` directly without any of the `jsonld.py` migrations. In the test this is to make sure that newly creaty projects/dataset with the current code conform with the current SHACL, as a command it's to help find issues with the local metadata of a user when debugging, to see if something might go wrong with migrations etc.

rokroskar:
Are you sure? Here's what `renku log --format dot-debug` produces:

noname gv


So it looks like `creator` is a proper node.

Panaetius:
At least that was the issue last week, I'm not up to date if that's still the case.

The templates have old metadata though (see context in e.g. https://github.com/SwissDataScienceCenter/renku-project-template/blob/master/python-minimal/.renku/metadata.yml might just be because the context is in the old format).

rokroskar:
Thanks, that should be easy enough!

lorenzo-cavazzi:
Ops sorry for not mentioning before that I have already spent time on that and (thanks to the SHACL validation) it was easy to assess that a few tests are indeed failing because the author data are missing. I made this quick commit https://github.com/lorenzo-cavazzi/renku-python/commit/0d2d142e7bda04428b200620070cdd5afe054f63 and the corresponding tests are succeeding https://github.com/lorenzo-cavazzi/renku-python/commit/0d2d142e7bda04428b200620070cdd5afe054f63/checks?check_suite_id=349167643 :tada: 

I would fix that in the template and adjust the cli code accordingly but I wasn't sure if that was enough -- I would like to quickly discuss this topic today.

renku/cli/init.py
lorenzo-cavazzi: right, I restored the previous functionality since I would keep the possibility to give a custom name to the project but falling back to the folder name when this is not provided

renku/cli/init.py
lorenzo-cavazzi: Right, I guess there are many better ways to do that. Is `list()` penalizing performance? Or is there a more efficient/pythonic way to check that?

renku/cli/init.py
lorenzo-cavazzi: `find_previous_commit()` in `renku/core/management/repository.py`. If that fails, the current folder wasn't a git repository before and I just print a warning.
We may think at a better solution to adopt in this case (creating a "pre renku-init" commit?)

renku/cli/init.py
rokroskar: This list could potentially have millions of entries. One function is `os.listdir()` - this still returns a list but not recursively. The generator for walking the directory tree if you want to check subdirectories is `os.walk()`. You could have a function like

```
def not_empty(path):
  for dirpath, dirname, files in os.walk(path):
    if files:
      return True
  return False
```

rokroskar: You could argue this is unneeded optimization - if the renku project has millions of files, we have other things to worry about :)

renku/core/commands/init.py
lorenzo-cavazzi: my previous comment is wrong, but that has been addressed in the latest commit by using renku exceptions

renku/cli/init.py
lorenzo-cavazzi: Thanks for the suggestion, I believe this is a good point! Plus, keeping an eye on using the most efficient code is always a good attitude.
I have implemented something based on generators but slightly different.

lorenzo-cavazzi:
Sorry for progressing slowly on this. I think a couple of features are still missing:

* Overhaul `validate_template()`: the fastest fix would be removing the check on `.renku/metadata.yml` 
and remove the file from the templates repo.
* Allow users to provide `metadata` to be used on jinja templates to substitute variables: an extra `--metadata` option on the cli is enough to feed metadata to `create_from_template(), but we need to somehow look for the list of available variables -- either going through all the files or providing a file in the template with the list of available variables.

Minor details for improving clarity:
* the code from `renku/cli/init.py@189 (# verify path status)` should probably be moved to `core/management/repository.py`
* `core/commands/init.py` should be renamed to `templates/commands/init.py` since everything there is about templates

tests/cli/test_init.py
```diff
@@ -79,7 +79,7 @@ def test_init(isolated_runner):
     assert 0 == result.exit_code
     assert new_project.exists()
     assert (new_project / '.renku').exists()
-    assert (new_project / '.renku' / 'metadata.yml').exists()
```
rokroskar: shouldn't the metadata file still get created?

lorenzo-cavazzi: Good point, it makes sense to still check `metadata` in the cli tests and change that only on template validation tests

lorenzo-cavazzi:
The tests on 3.5 are now passing. There is only a weird error happening on Travis with python 3.7 and `REQUIREMENTS=lowest` for `DataverseProvider` related tests. Maybe someone has a suggestion on how to address that?

renku/cli/init.py
```diff
@@ -64,128 +64,255 @@
...
+    """Check if path contains files.
+
+    :ref path: target path
+    """
+    gen = Path(path).glob('**/*')
+    try:
+        next(gen)
+        return False
+    except StopIteration:
+        return True
```
Panaetius:
```
return not any(gen)
```

Would accomplish the same without having to deal with exceptions

lorenzo-cavazzi: Thanks for the cleanup suggestion!

rokroskar: does `any(gen)` create a list first though?

Panaetius: No, it calls iternext(...) and checks if that's NULL or not (in cython C code). So basically it does the same, just without having to deal with exceptions.

Any returns true on the first element encountered that evaluates to true, so if paths was `['', '']` there'd be a difference (above code returning False and `not any(...)` returning True, but I don't think that matters here.

And I find it more readable with `any`

PR #814, opened 2019-11-14
Publish SHACL to repo

Panaetius:
# Description

Github action to copy the `shacl_shape.json` to a different repository on commit to master with version.

Currently configured to go to https://github.com/Panaetius/SHACL_test_repo for testing.

Access Token needs to be generated for this with rights to the final repository where the SHACL should end up in. Until then it's WIP and it pushes to the test repo on every commit (not just on commit to master)

Closes https://github.com/SwissDataScienceCenter/renku/issues/685


.github/workflows/publish_shacl.yml
```diff
@@ -0,0 +1,37 @@
...
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install .[all]
+    - name: Publish shape
+      run: |
+        git clone https://${{secrets.SHACL_ACCESS_TOKEN}}@github.com/Panaetius/SHACL_test_repo.git shacl_publish
```
rokroskar: which user does the `SHACL_ACCESS_TOKEN` belong to? We should use the renku bot user for this

Panaetius: Right now it's my access token as it deploys to a testrepo under my name. THis has to be changed before merging (as mentioned above), not sure I have the rights to do that though.

rokroskar:
You can use path filtering to avoid that: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-a-workflow#filtering-for-specific-branches-tags-and-paths

Panaetius:
> Looks good! Thanks.
> 
> I just have one quick question. Is this gonna make a commit even if new commit on master does not update SHACL definitions? If yes, could/should we find a way to update only when they get updated?

I thought about that and I think it makes sense to add it even if it didn't change, to make it easier for others to know what schema to validate output from what version of renku python they're using. This way you can just use whatever is in the respective folder without having to do any logic.

jsam:
I see. But that ability should probably be part of the schema versioning itself since we gonna most likely need it inside the core module for migrations. Also, having a new directory for every changed schema version (vs. for every commit) makes it easier/possible to track changes of schema through time (I think the same wouldn't be the case after 200 commits on master with current setup).

I guess there is a tradeoff for both.

Panaetius:
i changed it so it only gets executed when `renku/data/shacl_shape.json` is in a push

.github/workflows/publish_shacl.yml
```diff
@@ -0,0 +1,34 @@
...
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install .[all]
+    - name: Publish shape
+      run: |
+        git clone https://${{secrets.SHACL_ACCESS_TOKEN}}@github.com/SwissDataScienceCenter/renku-ontology.git shacl_publish
+        cd shacl_publish
+        mkdir $(renku --version)
+        cp ../renku/data/shacl_shape.json ./$(renku --version)/
+        git config --local user.email "action@github.com"
```
rokroskar: can we use `renku@datascience.ch` or similar insted?

PR #822, opened 2019-11-15
chore: Automated documentation label

Panaetius:
Adds a `documentation:pending` label when a PR is created and changes were made in the `renku/` subdirectory and adds an action that fails if that label is present.


rokroskar:
hmm this seems a bit buggy still - adding and removing labels has some unpredictable effects.

Panaetius:
I think it is due to how github handles actions and checks. Actions are regarded as being on a per-commit basis, but they can be triggered with events other than pushing (like the labeled/unlabeled) events. This executes the action but not `in place`, it runs as a new action, so the old run and the new one stick around. Also actions can only set the `successful` and `failed` states, but not for instance the `pending`state.

I think github apps get get around this (Otherwise the `WIP` app wouldn't work), but I didn't see an app that 1. supports checking labels and 2. get run when labels are changed (There are some apps for 1. but paradoxically non of them update on label change...). 

https://github.com/marketplace/pr-label-enforcer might work (I couldn't install it to SDSC as I lack the rights), together with the labeler action that's adds the `documentation:pending` label to a newly opened PR

rokroskar:
I tried configuring the PR label enforcer by adding the `.github/pr_labels.yml` as per their instructions - but doesn't seem to be showing up in the checks. Maybe it needs to be on master?

rokroskar:
@Panaetius are we giving up on this?

Panaetius:
> @Panaetius are we giving up on this?

I have no idea what the issue with the pr label enforcer, maybe it needs to be on master. I'll open an issue in their github.

Theoretically it should work already without the pr label enforcer, just with having to manually rerun all checks when changing the labels (Same as with the conventional commits check). So if we can't get the PR label enforcer to work, we could just go with the more manual version.

rokroskar:
should be all set: 
image

PR #824, opened 2019-11-15
fix: ensure all Person instances have valid ids

### LINKED ISSUE #812: regression: mailto:None as id of dataset creators
`mailto:None` appears as a node ID of dataset creators. 

This was fixed but the regression came about during the refactoring of `Creator` and `Person` classes.

rokroskar:
closes #812


rokroskar:
yep, fixing the lowercase issue :)

your tests should be there

renku/core/models/datasets.py
```diff
@@ -44,6 +45,32 @@
...
+    if isinstance(value, dict):  # compatibility with previous versions
+        return [Person.from_jsonld(value)]
+
+    if isinstance(value, list):
+        return [Person.from_jsonld(v) for v in value]
+
+
+@attr.s
+class PersonMixin:
+    """Mixin for handling creators container."""
```
Panaetius: Maybe this should still be named `CreatorsMixin` as it is for setting the creator on an object. Especially given the docstring there.

Panaetius:
Thanks a lot for this change! it makes a lot of sense to get rid of `Creator`.

renku/core/models/datasets.py
rokroskar: good point - I changed it back

Issue #825, opened 2019-11-18
Add `renku clone` command


rokroskar:
There are some repository initialization steps that a user might miss when they clone a renku project. This typically leads to issues with git-LFS and potentially other problems. We should create a `renku clone` command that does these operations for the user to ensure a seamless experience. 

The steps that should be done automatically:

- `git lfs install --local`
- `renku githooks install`
- others?

closes #213

PR #828, opened 2019-11-19
feat: adds renku clone command

mohammad-sdsc:
# Description

Adds a `renku clone` commands that clones a Git repository and does Renku-specific initialisations (git-hooks and LFS).
It also refactors `renku githooks` command to have proper file structure (i.e. `commands/githooks.py`, `management/githooks.py`).

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/825


renku/core/commands/clone.py
```diff
@@ -0,0 +1,43 @@
...
+
+from .client import pass_local_client
+
+
+@pass_local_client
+def renku_clone(
+    client, url, path=None, install_githooks=True, skip_smudge=True
+):
+    """Clone Renku project repo, install Git hooks and LFS."""
+    click.echo('Cloning {} ...'.format(url))
```
jsam: This should probably happen only in CLI command.

renku/core/commands/clone.py
```diff
@@ -0,0 +1,43 @@
...
+
+    install_lfs = client.use_external_storage
+    clone(
+        url=url,
+        path=path,
+        install_githooks=install_githooks,
+        install_lfs=install_lfs,
+        skip_smudge=skip_smudge,
+        progress=GitProgress()
```
jsam: If I use this from the service, wouldn't this progress bar pollute the stdout?

tests/cli/test_clone.py
```diff
@@ -0,0 +1,46 @@
...
+# limitations under the License.
+"""Renku clone tests."""
+
+from pathlib import Path
+
+from renku.cli import cli
+from renku.core.management.storage import StorageApiMixin
+
+
+def test_renku_clone(runner, remote_project, directory_tree, monkeypatch):
```
jsam: Can we add a doc string?

tests/cli/test_clone.py
```diff
@@ -0,0 +1,46 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Renku clone tests."""
+
+from pathlib import Path
+
+from renku.cli import cli
+from renku.core.management.storage import StorageApiMixin
+
+
```
jsam: Can we add an integration test as well?

renku/core/management/clone.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def clone(
+    url,
+    path=None,
+    install_githooks=True,
+    install_lfs=True,
+    skip_smudge=True,
+    progress=None
+):
```
jsam: Can we add an option for a shallow clone?

renku/core/management/clone.py
```diff
@@ -0,0 +1,64 @@
...
+        install(client=client, force=True)
+
+    if install_lfs:
+        try:
+            command = ['git', 'lfs', 'install', '--local', '--force']
+            if skip_smudge:
+                command += ['--skip-smudge']
+            repo.git.execute(command=command, with_exceptions=True)
+        except GitCommandError as e:
+            raise errors.GitError('Cannot install Git LFS') from e
```
jsam: Would be great to enhance readability and wrap `repo.git.execute` into `try-except` since other lines can't raise exceptions that are being handled.

renku/core/management/clone.py
```diff
@@ -0,0 +1,64 @@
...
+    path = path or '.'
+    # Clone the project
+    try:
+        if skip_smudge:
+            os.environ['GIT_LFS_SKIP_SMUDGE'] = '1'
+        repo = Repo.clone_from(url, path, recursive=True, progress=progress)
+    except GitCommandError as e:
+        raise errors.GitError(
+            'Cannot clone remote Renku project: {}'.format(url)
+        ) from e
```
jsam: Same as below, I would wrap only the lines which can actually raise the exception.

renku/core/management/clone.py
```diff
@@ -0,0 +1,64 @@
...
+):
+    """Clone Renku project repo, install Git hooks and LFS."""
+    from renku.core.management.client import LocalClient
+
+    path = path or '.'
+    # Clone the project
+    try:
+        if skip_smudge:
+            os.environ['GIT_LFS_SKIP_SMUDGE'] = '1'
+        repo = Repo.clone_from(url, path, recursive=True, progress=progress)
```
jsam: `recursive=True` should probably be one of the controllable parameters on the invocation of the `clone(...)`

renku/core/management/clone.py
mohammad-sdsc: Do you have a use-case for it? Here, I believe we always need the whole Renku project history and shallow won't be used. Moreover, shallow cloning has many flags (e.g. `--depth`, `--shallow-since`, `--shallow-exclude`, ...) and I don't want to include them all here if not needed. However, if you have a use for it probably we can include `--depth`.

tests/cli/test_clone.py
mohammad-sdsc: Well, actually we just need the integration test and not this one!

renku/core/management/clone.py
jsam: Yes, most likely we will have to expose this on the service, cause for a simple use case like adding a file from UI we probably don't want full clone (cause full clone on big repositories could be quite slow). I think adding a `--depth` would be sufficient for that use case. We can also make a ticket for that if you feel it's out of scope here.

mohammad-sdsc: No, it's OK and already done 👍

Issue #816, opened 2019-11-14
 feat: add named dataset from other renku project


emmjab:
**Is your feature request related to a problem? Please describe.**
Labs want to be able to set up data repos where datasets are created and metadata is added, and then in the analysis repos import those datasets, like "fully formed datasets" in zenodo or dataverse.

**Describe the solution you'd like**
`renku dataset import renku:dataset-provenance/dataset-name`

**Additional context**
Based on the way users want to organize their projects (data somewhere that many different projects use), it makes more conceptual sense to be able to import these "renku" datasets rather than see the datasets in those repos as just random files.

PR #838, opened 2019-11-26
feat: import renku datasets

mohammad-sdsc:
# Description

A Renku dataset can be imported like Zenodo or Dataverse datasets. All files in a dataset among with their metadata are copied into a Renku project. The URI for a Renku dataset is the link to the dataset in the Renku UI:
```
https://dev.renku.ch/projects/virginiafriedrich/datasets-test/datasets/10.5281%2Fzenodo.3251128
https://dev.renku.ch/projects/virginiafriedrich/datasets-test/datasets/c82dcc4c-9923-4b57-b79a-4b2550a8c59b
```
Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/816


rokroskar:
How should I import from a repo using ssh? I tried:

```
renku dataset import git@dev.renku.ch:virginiafriedrich/datasets-test.git:parametrized_extreme_rai_210
Error: Invalid parameter value: Could not process git@dev.renku.ch:virginiafriedrich/datasets-test.git:parametrized_extreme_rai_210.
```

and

```
renku dataset import git+ssh://dev.renku.ch/virginiafriedrich/datasets-test:parametrized_extreme_rai_210
...
renku.core.errors.GitError: Cannot clone remote Renku project: git+ssh://dev.renku.ch/virginiafriedrich/datasets-test.git
```

renku/core/commands/providers/renku.py
```diff
@@ -0,0 +1,105 @@
...
+
+    is_doi = attr.ib(default=False)
+    _accept = attr.ib(default='application/json')
+
+    @staticmethod
+    def supports(uri):
+        """Whether or not this provider supports a given uri."""
+        return 'renku' in uri.lower()
+
+    def find_record(self, uri, client):
```
jsam:
```suggestion
    def find_record(self, uri, client=None):
```

My linter complains that the signature is not the same. :smile_cat:

jsam: Or instead update the `ProviderApi` interface

renku/core/commands/providers/renku.py
```diff
@@ -0,0 +1,105 @@
...
+                'Cannot find dataset with name "{}"'.format(dataset_name)
+            )
+
+        return RenkuRecordSerializer(dataset)
+
+    def get_exporter(self, dataset, access_token):
+        """Create export manager for given dataset."""
+        raise NotImplementedError()
+
+    def is_git_based(self):
```
jsam:
```suggestion
    @property
    def is_git_based(self):
```

rokroskar:
@jsam I think the URI has been changed - @mohammad-sdsc could you update the PR description?

renku/core/commands/providers/renku.py
```diff
@@ -0,0 +1,105 @@
...
+        """True if provider is git-based."""
+        return True
+
+    @staticmethod
+    def record_id(uri):
+        """Extract dataset id from uri."""
+        return uri.split(':')[-1]
```
rokroskar: we need something more elaborate here to support git URLs like `git@<renku>:<user>/<project>`. Maybe we need to specify the dataset name with a flag? E.g. 

`renku dataset import --name mydata git@renkulab.io/<name>/<project>`

mohammad-sdsc: The problem with git@ is not because of this. The template method where it checks if the URL is valid before handing it over to the providers and since it is not a valid URL that check fails. I'll fix it in the template method and keep the syntax as it is.
As we discussed last week in the design call, we are going to change this to be a valid URL (when UI can return JSON metadata that points to the location of the gitlab project).

mohammad-sdsc:
> Thank you for looking into this! I have one small question.
> 
> I'm trying to test it out and I'm using the following datasets generated from the integration tests here: ` https://dev.renku.ch/projects/contact/integration-tests/datasets/fbd66f4154f84df3a35d70de88f3866a/`
> 
> and I'm getting the following:
> 
> ```
> renku dataset import https://dev.renku.ch/projects/contact/integration-tests/datasets/fbd66f4154f84df3a35d70de88f3866a/
> Error: Invalid parameter value: Invalid Renku dataset URI: https://dev.renku.ch/projects/contact/integration-tests/datasets/fbd66f4154f84df3a35d70de88f3866a/
> ```
> 
> Shouldn't this be supported?

This is not supported for now. The problem is that we need Renku project's path in gitlab to be able to clone the repo and look for the requested dataset. 
The format that works at the moment is: <URL of the gitlab project>:<dataset name>.

renku/core/commands/providers/renku.py
rokroskar: right - I just think it's a bit dangerous to add a regex that is not actually a valid git URL to the function that checks if git URLs are valid... so I would prefer a different solution even if it is "ugly" for the time being.

renku/core/commands/providers/renku.py
```diff
@@ -0,0 +1,198 @@
...
+    """Renku API provider."""
+
+    is_doi = attr.ib(default=False)
+    _accept = attr.ib(default='application/json')
+
+    @staticmethod
+    def supports(uri):
+        """Whether or not this provider supports a given uri."""
+        u = urllib.parse.urlparse(uri)
+        return 'renku' in u.netloc and (
```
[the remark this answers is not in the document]
mohammad-sdsc: I've talked to Pamela and Kuba. Although current deployments have 'renku' somewhere in the URL, it is not mandatory. Moreover, we don't have a similar page like Dataverse. So, for now, I'll remove this check and only rely on the URL's path. An error is raise somewhere down the road if we cannot find a git repo or a Renku dataset.

mohammad-sdsc:
We need to re-generate graph on `dev.renku.ch` for the tests to pass. ATM, datasets URL on this deployment use DOI instead of UUID (we changed this in https://github.com/SwissDataScienceCenter/renku-python/issues/889) which makes the tests fail.

rokroskar:
Can @jachro help?

mohammad-sdsc:
Yes, I've already talked to him and he will make a deployment on him own env so that I can test this. Later, we will update `dev.renku.ch` too.

rokroskar:
Great thanks!

mohammad-sdsc:
Thanks @jsam for the review!

Regarding the first two links that don't work: If you go to the Gitlab page of the project that contains them (https://renkulab.io/gitlab/rok.roskar/flights-tutorial), there is not dataset in there. I did not dig where is the source of this problem in the KG, but as far as the functionality of this PR is concerned, it is working correctly as it can find anything when it clones that repo.

For the migration, I applied the current migration scheme that we have when forking the project. There is not check to see if migration is required or not because that's what I'm currently implementing in the migration story and I did not want to redo it here. This should not have any negative effect AFAIK.

The size and checksum of files when importing a dataset comes from the dataset's metadata, however, we don't store that information for Renku datasets. That's why there is nothing there to be displayed. I added some code to calculate this information OTF when importing the dataset. It should be enough for the sake of this story.

I don't have any control on the output of the Git LFS commands. I've just put an empty line there to avoid the green OK colliding with it.

rokroskar:
> The size and checksum of files when importing a dataset comes from the dataset's metadata, however, we don't store that information for Renku datasets.

Do we have an issue for adding this? I think it would certainly be useful

jachro:
I don't think we have one as yet

mohammad-sdsc:
> > The size and checksum of files when importing a dataset comes from the dataset's metadata, however, we don't store that information for Renku datasets.
> 
> Do we have an issue for adding this? I think it would certainly be useful

It might be a bit tricky to implement. We need to keep this information updated when for example we run a renku update; or when multiple dataset share the same file and we need to keep all of them updated.
I believe the best solution would be to query git for this information whenever needed. We can create the issue though.

rokroskar:
@mohammad-sdsc I'm seeing `basedOn` in the dataset json-ld context but afaict it's not actually used? Can we remove it?

mohammad-sdsc:
> @mohammad-sdsc I'm seeing `basedOn` in the dataset json-ld context but afaict it's not actually used? Can we remove it?

AFAICR if we remove this then triple generation won't work for dataset files. I've just checked this and triples are generated correctly without it. Maybe there was a bug in the code which was fixed. I'll remove this.

rokroskar:
@mohammad-sdsc I think `basedOn` was used only for renku datasets that were imported using `renku dataset import` - but if I see correctly, you've changed that to use `sameAs`. Right?

mohammad-sdsc:
> @mohammad-sdsc I think `basedOn` was used only for renku datasets that were imported using `renku dataset import` - but if I see correctly, you've changed that to use `sameAs`. Right?

No, it was used for adding data from git; for each file, we keep track of the remote file in the `isBasedOn` property. A `isBasedOn` is not required for the dataset (and we added it just for the reason I've mentioned earlier). `sameAs` was always used for imports though.

rokroskar:
Ah right - but then it shouldn't be removed, right??

mohammad-sdsc:
I've remove it in https://github.com/SwissDataScienceCenter/renku-python/pull/838/commits/e2fa92041082556552f5998275c07fbf1a303699 after your comment.

rokroskar:
Yep, i saw that - but my comment was based on incomplete information :) I think this means that the `basedOn` triple will be compacted away and never generated (though the `renku log` won't fail)

rokroskar:
Or am I maybe still misunderstanding what you mean :/

rokroskar:
Ok mystery(ies) solved - thanks for the patience @mohammad-sdsc!

This is looking really good to me - @emmjab could you give it a once-over to see if we missed something obvious from a usability point of view?

emmjab:
User hat on: tbh it was hard to figure out what the syntax is for this, even after reading this thread... I think there can be a number of improvements to the UX -- 

1. For instance, the helptext for `renku dataset` refers to "importing" for 3rd party datasets, which doesn't indicate that you can also import other renku datasets:

```
$ ▶ renku dataset --help
Usage: renku dataset [OPTIONS] COMMAND [ARGS]...

  Handle datasets.
...
...
  export    Export data to 3rd party provider.
  import    Import data from a 3rd party provider.
...
```
2. Once you drill down into help for `import`, you see "Renku" in the list, but I guess I would change import help in all places to: `Import data from a 3rd party provider or another renku project.`

```
$ ▶ renku dataset import --help
Usage: renku dataset import [OPTIONS] URI

  Import data from a 3rd party provider.

  Supported providers: [Dataverse, Renku, Zenodo]

Options:
  --short-name TEXT  A convenient name for dataset.
  -x, --extract      Extract files before importing to 
```
---
3. Additionally, it's not abundantly clear what the URI should be. On renkulab we have these links to the dataset that look like: `https://renkulab.io/datasets/f69a4955-3778-4e63-858d-863c435ad336`

Project on dev, tried to import a dataset from renkulab:
```
$ ▶ renku dataset import https://renkulab.io/datasets/f69a4955-3778-4e63-858d-863c435ad336
Error: Cannot access knowledge graph: https://renkulab.io/knowledge-graph/projects/cchoirat/hostmicrobe
Response code: 500
```

Project on dev, tried to import a dataset from dev:
```
$ ▶ renku dataset import https://dev.renku.ch/datasets/3f626670-49d7-4ae2-9a44-8611ae26d5b3
Error: Cannot access knowledge graph: https://dev.renku.ch/knowledge-graph/projects/renku-qa/sbb-data-qa
Response code: 500
```

Finally realized the URL on the dataset's page is different from those links:

```
$ ▶ renku dataset import https://dev.renku.ch/projects/renku-qa/flights-tutorial-playground/datasets/15f506ce-d62a-4763-9022-892499f01d82/
CHECKSUM                                  NAME                       SIZE (MB)  TYPE
----------------------------------------  -----------------------  -----------  ------
7ebdc31b1c05d3720896d1d968080bec69c3c213  2019-01-flights.csv.zip         7.90  zip
Do you wish to download this version? [y/N]: y
OK                                                      
```

Why are these urls different?

mohammad-sdsc:
> See comment in thread -- boils down to:
> 
> 1. update helptext to add renku to string involving "3rd parties" --> `Import data from a 3rd party provider or another renku project.`

Thanks for noticing this! Done!

> 2. update helptext to add syntax for URI

This makes the help text too long and it's not just for renku. We should add examples for zenodo and dataverse as well which all are in the docs.

> 3. add an example to the documentation

It's already there.

> 4. why doesn't it work for the datasets I tried?

As the error message shows, the knowledge graph is no accessible for this project: `https://dev.renku.ch/knowledge-graph/projects/renku-qa/sbb-data-qa`. It's probably an issue with the KG but this PR cannot do anything about it. Did you find the error message useful? If not, suggestions are welcome!

This one works for example: `renku dataset import https://dev.renku.ch/datasets/48299db3-8870-4cbe-a480-f75985c42a62`

rokroskar:
@mohammad-sdsc - when the URL is the generic dataset URL, how do you decide which project to download it from?

mohammad-sdsc:
> @mohammad-sdsc - when the URL is the generic dataset URL, how do you decide which project to download it from?

It gets the list of projects that contain the dataset from the KG and notes that the dataset is `sameAs` which dataset. Then it goes over the projects and queries the KG to get project's info (e.g. urls to project's gitlab page) and tries to clone the project. It continues until it succeeds with one project. Then it queries KG again to get the list of datasets for that project and copies the one which is `sameAs` the one it noted earlier.

rokroskar:
Ok so basically you traverse backwards until you get to the source project?

mohammad-sdsc:
> Ok so basically you traverse backwards until you get to the source project?

 No, it's traversing upwards. The loop is NOT for checking which project has the dataset; all of them have it or otherwise they won't show up in the first KG query's result. We know which dataset in that project is the same as the one request using `sameAs` property (they are both same as some dataset which can also be either of them).

The looping is to see which one we can clone from because there might be private projects which we don't have access to and ATM they are not filtered.

rokroskar:
I see - in the current implementation this will lead to problems since not all of those projects will necessarily have the dataset in the same state. Once we do the work of ensuring dataset versions it will no longer be a problem. For now we should instruct users to go to the specific project they want to import the dataset from.

mohammad-sdsc:
> I see - in the current implementation this will lead to problems since not all of those projects will necessarily have the dataset in the same state. Once we do the work of ensuring dataset versions it will no longer be a problem. For now we should instruct users to go to the specific project they want to import the dataset from.

Yes, it would be the safest to use the link with specific project. It's also faster since we query KG just once.

emmjab:
Thanks for the helptext changes @mohammad-sdsc !

Just to clarify, how will a user know why the dataset is not available? How does it appear in the UI for dataset search but not however you're accessing it?

For: https://renkulab.io/projects/cchoirat/hostmicrobe/datasets/f69a4955-3778-4e63-858d-863c435ad336/

On dev:
```
$ ▶ renku dataset import https://renkulab.io/projects/cchoirat/hostmicrobe/datasets/f69a4955-3778-4e63-858d-863c435ad336/
Error: Cannot access knowledge graph: https://renkulab.io/knowledge-graph/projects/cchoirat/hostmicrobe
Response code: 500
```

Just ran into another scenario; on dev:
```
$ ▶ renku dataset import https://dev.renku.ch/projects/renku-qa/corona-pandemic/datasets/06735ea2-7e27-44d3-a60d-28d67e54e333/
Username for 'https://dev.renku.ch': e.jablonski
Password for 'https://e.jablonski@dev.renku.ch': 
Error: Invalid parameter value - Cannot find any project for the dataset.
```

jachro:
@emmjab I'll take a look. There are some peculiarities about the mentioned projects and their Access Tokens available for Graph Services.

mohammad-sdsc:
> Just to clarify, how will a user know why the dataset is not available? How does it appear in the UI for dataset search but not however you're accessing it?

The reason is that I need to access the KG to get where is gitlab instance located to clone the repo; UI already has this information and does not need to do that.

> Just ran into another scenario; on dev:
> 
> ```
> $ ▶ renku dataset import https://dev.renku.ch/projects/renku-qa/corona-pandemic/datasets/06735ea2-7e27-44d3-a60d-28d67e54e333/
> Username for 'https://dev.renku.ch': e.jablonski
> Password for 'https://e.jablonski@dev.renku.ch': 
> Error: Invalid parameter value - Cannot find any project for the dataset.
> ```

This happens when the remote project cannot be cloned. As we discussed, I've changed the error message to reflect this:

```
$ renku dataset import https://dev.renku.ch/projects/renku-qa/corona-pandemic/datasets/06735ea2-7e27-44d3-a60d-28d67e54e333/
Username for 'https://dev.renku.ch':
Password for 'https://fdsfs@dev.renku.ch': 
Error: Invalid parameter value for https://dev.renku.ch/projects/renku-qa/corona-pandemic/datasets/06735ea2-7e27-44d3-a60d-28d67e54e333/: Cannot clone remote projects:
	git@dev.renku.ch:renku-qa/corona-pandemic.git
	https://dev.renku.ch/gitlab/renku-qa/corona-pandemic.git
```

Issue #476, opened 2019-03-18
cli: dataset add should strip username:password from source URL


jirikuncar:
Make sure that no user sensitive information included in the source URL are stored.

- [ ] username:password
- [ ] token
- [ ] absolute local paths
- [ ] ensure that usernames and tokens are not added to commit message

PR #841, opened 2019-11-26
feat: do not store url credentials anywhere

mohammad-sdsc:
# Description

Remove username and passwords from URLs from dataset metadata and commit message when adding data to a dataset. It also removes local paths from metadata when adding a file from local filesystem.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/476

### DIFF
--- a/renku/core/management/datasets.py
+++ b/renku/core/management/datasets.py
@@ -41,6 +41,7 @@ from renku.core.models.git import GitURL
 from renku.core.models.locals import with_reference
 from renku.core.models.provenance.agents import Person
 from renku.core.models.refs import LinkReference
+from renku.core.utils.urls import remove_credentials
 
 
 @attr.s
@@ -301,7 +302,7 @@ class DatasetsApiMixin(object):
             else:
                 return [{
                     'path': path_in_repo,
-                    'url': src.as_uri(),
+                    'url': path_in_repo,
                     'creator': dataset.creator,
                     'dataset': dataset.name,
                     'parent': self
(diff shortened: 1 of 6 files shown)


renku/core/utils/urls.py
```diff
@@ -39,3 +40,11 @@ def url_to_string(url):
         return url
 
     raise ValueError('url value not recognized')
+
+
+def remove_credentials(url):
+    """Remove username and password from a URL."""
+    if url is None:
+        return ''
```
rokroskar: why change type?

mohammad-sdsc: This is actually fixing the type. When url is `None` then `.geturl()` returns `bytes` (`b''`) instead of normal return value of `str`.

renku/core/management/datasets.py
```diff
@@ -297,7 +298,7 @@ def _add_from_local(self, dataset, path, link, destination):
             else:
                 return [{
                     'path': path_in_repo,
-                    'url': src.as_uri(),
+                    'url': None,
```
rokroskar: should we instead make this a relative path? It's only relevant for the user who creates the dataset but maybe that's better than `None`?

mohammad-sdsc: If the file is copied from an absolute path (e.g. a shared storage mounted in the same filesystem) then this would still expose the path. Besides, a relative path is not (technically) a URL. I am open to change this though if these are fine with you.

rokroskar: `file://../../some/path` is a valid URL. It's true that if you knew the location of the original project on disk you would then know the paths of the origin data, but we don't expose the absolute path of the project anywhere.

PR #846, opened 2019-11-27
Fix JSON-LD translation and related issues

### LINKED ISSUE #844: JSON-LD Project translation and deserialisation isn't applied to nested objects
Project nodes nested in the KG (in DatasetFile or Dataset) are not deserialized correctly, instead just being read in as string or dict and output as is, leading to outdated metadata (specifically, `foaf:Project` instead of `schema:Project`).

Writeup of what happens (on `datasets-test` repo with `renku log --format nt .renku/datasets/48299db3-8870-4cbe-a480-f75985c42a62/metadata.yml |grep foaf`:

- Attrs doesn't do anything with the `type` keyword, so if `_project` is a string (as it is on those dataset files), then `DatasetFile._project` would be a string. And if it's correctly set in the yaml, it'd be a dict (instead of `Project`). This only applies for `Project` on nested objects.
- Additionally, `CommitMixin._project` and `DatasetFile.dataset` are both `schema:isPartOf`, so actually in this case, `DatasetFile._project` is a list, of `[{...project data...}, "dataset_name"]`, making deserialization even more broken. This is due to the `pyld.jsonld.compact(...)` call around line 503 in `jsonld.py` (it gets expanded to `isPartOf` and all of those entries get compacted into one of the `isPartOf` properties as a list. This is important to consider for https://github.com/SwissDataScienceCenter/renku/issues/696 
- Since it's a list with a dictionary inside, with the dictionary containing the original data as loaded from `metadata.yml`, it gets serialized as-is back to jsonld when `asjsonld(...)` is called. 
- TL;DR: The CommitMixin._project data never gets beyond the dict stage and is never a Project instance. No transformation/logic is applied to it. And it gets exported as is.

On a side note, `Dataset._project` has `name` = None,  and created/updated = now() for zenodo imports (in the `datasets-test` project), which seems wrong (The data is there on disk). Though it gets overridden with `self.client.project` in the post init of `Dataset`, so this has no effect. But it's weird that it doesn't get deserialized correctly event though it's there in the yaml. The dataset with this issue is `10.5281%2Fzenodo.3266798`. This has to be verified after a fix for the above points, as it only happened when I forced deserializing of `_project` for `Dataset`.

TODO as I see it:
- Add a converter for `Project` that can handle dict and string, at least, then translation would work. Or just add a  generic converter for every property that has `type` set and where `type` is a deserializable object.
- Either never use the same ontology type for more than one property (might be difficult for generic things like `isPartOf` and `hasPart` ) or find a way that compaction doesn't mess with this.
- Figure out a migration for the above, if necessary.
- Check that `Project` metadata loading works even in cases where it is later overwritten with the `client.project`, just to make sure there's no other, additional issue.

Panaetius:
Fixes #844 

Removes `DatasetFile.dataset`.
Adds default converter for jsonld types so jsonld is always deserialized.
Propagates jsonld context so jsonld translation can work correctly
Fixes migrations
Fixes project id's for old projects
Removes type annotation from Person that was in the wrong place


renku/core/models/jsonld.py
```diff
@@ -523,6 +535,12 @@ def from_jsonld(
 
         for k, v in compacted.items():
             if k in fields:
+                if isinstance(v, dict) and '@context' not in v and\
```
rokroskar: 🙈

PR #848, opened 2019-12-03
chore: Adds semantic PR github action

Panaetius:
Adds a github action to enforce that PR titles conform to conventional commits syntax


Panaetius:
Currently, the action is triggered when the title is changed, but due to how Github handles actions, a previously failed check persists unless there's a new commit. In this case the failing action needs to be rerun to get it to pass. I don't see a way around this

Issue #840, opened 2019-11-26
Wrong file location when adding files to imported datasets


mohammad-sdsc:
**Describe the bug**
When adding a file to an imported datasets, the new file is added to a data directory with dataset's name instead of the dataset's directory.

**To Reproduce**
```
$ renku init
$ renku dataset import https://zenodo.org/record/3251128
$ ls data
contemporary_phenotypic_12
$ renku dataset add contemporary_phenotypic_12 https://example.com/index.htm
$ ls data
contemporary_phenotypic_12 'Contemporary phenotypic change in plant quantitative traits'
```

**Expected behavior**
Files must be added to the same folder.

rokroskar:
related to https://github.com/swissdatasciencecenter/renku-python/issues/791

Issue #791, opened 2019-11-04
Clean up use of dataset `display_name`


rokroskar:
There is some inconsistency with the use of dataset display name. Initially this was introduced to deal with really long dataset titles imported from external data repositories. However, the way it is used is inconsistent. For example, using `renku dataset` yields this:

```
renku dataset
ID                                    DISPLAY_NAME            VERSION    CREATED              CREATORS
------------------------------------  ----------------------  ---------  -------------------  ----------
2a2c6230-aeb2-41b0-89d5-f1cdd240d95a  maunaloa                           2019-01-14 12:29:45  R.Nijzink
```

But now to list the files I need to use a different name: 

```
renku dataset ls-files maunaloa
ADDED    CREATORS    DATASET    PATH
-------  ----------  ---------  ------

renku dataset ls-files MaunaLoa
ADDED                CREATORS       DATASET    PATH
-------------------  -------------  ---------  --------------------------------------------------------
2019-11-04 10:55:23  Remko Nijzink  MaunaLoa   /<dataset path>/data/MaunaLoa/weekly_in_situ_co2_mlo.csv
```

So I somehow need to remember that the dataset was named `MaunaLoa` when it was created even though it's not possible for me to retrieve this information unless i look at the name of the directory for the dataset. We should standardize this behavior and make it easier for the user to refer to a dataset. 

Perhaps a solution would be to ask the user what they want to name the dataset when they are importing it? That way we probably don't run the risk of absurdly long dataset names _and_ the user will actually know under which name they can find it.

Issue #515, opened 2019-04-12
cli: add metadata on dataset creation


jsam:
During creation of a dataset we should ask the user for some simple metadata information such as description, creator affiliations, display title, tags, etc.

rokroskar:
Maybe some of these, like affiliation, should be optionally persisted in (or read from) the user's renku configuration?

PR #850, opened 2019-12-04
feat: metadata on dataset creation

mohammad-sdsc:
# Description

When creating datasets, users can provide its description, creators, and display name using command line options. Creator is a string with "Name <email>" format and users can pass multiple creators. 
This also makes a consistent use of `short_name` in the code. `short_name` is used as the dataset's data directory name and dataset's reference name. `name` is not used by Renku other than storing it as a metadata. If users don't provide a short name when creating a dataset then one is automatically created.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/515
Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/791
Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/840

### DIFF
--- a/renku/core/management/datasets.py
+++ b/renku/core/management/datasets.py
@@ -133,7 +133,7 @@ class DatasetsApiMixin(object):
                 'Dataset exists: "{}".'.format(name)
             )
 
-        dataset_path = self.path / self.datadir / dataset.display_name
+        dataset_path = self.path / self.datadir / dataset.internal_name
         dataset_path.mkdir(parents=True, exist_ok=True)
 
         try:
@@ -155,27 +155,23 @@ class DatasetsApiMixin(object):
         dataset.to_yaml()
 
     def create_dataset(
-        self,
-        name,
-        display_name='',
-        description='',
-        creators=(),
-        is_import=False
+        self, name, internal_name='', description='', creators=()
     ):
         """Create a dataset."""
-        if self.load_dataset(
-            name=name
-        ) or (is_import and self.load_dataset(name=display_name)):
-            raise errors.DatasetExistsError(
-                'Dataset exists: "{}".'.format(name)
-            )
-
         if not name:
             raise errors.ParameterError('Dataset name must be provided.')
-        # Do not validate imported datasets' names
-        if not is_import and not is_dataset_name_valid(name):
+
+        if not internal_name:
+            internal_name = name
+
+        if not is_dataset_name_valid(internal_name):
             raise errors.ParameterError(
-                'Dataset name {} is not valid.'.format(name)
+                'Dataset name {} is not valid.'.format(internal_name)
+            )
+
+        if self.load_dataset(name=internal_name):
+            raise errors.DatasetExistsError(
+                'Dataset exists: "{}".'.format(internal_name)
             )
 
         identifier = str(uuid.uuid4())
@@ -192,15 +188,13 @@ class DatasetsApiMixin(object):
                 client=self,
                 identifier=identifier,
                 name=name,
-                display_name=display_name,
+                internal_name=internal_name,
                 description=description,
                 creator=creators
             )
 
-        reference_name = display_name or name
-
         dataset_ref = LinkReference.create(
-            client=self, name='datasets/' + reference_name
+            client=self, name='datasets/' + internal_name
         )
         dataset_ref.set_reference(path)
 
@@ -220,7 +214,7 @@ class DatasetsApiMixin(object):
     ):
         """Import the data into the data directory."""
         warning_message = ''
-        dataset_path = self.path / self.datadir / dataset.display_name
+        dataset_path = self.path / self.datadir / dataset.internal_name
 
         destination = destination or Path('.')
         destination = self._resolve_path(dataset_path, destination)
--- a/renku/core/management/datasets.py
+++ b/renku/core/management/datasets.py
@@ -36,7 +36,8 @@ from git import GitCommandError, GitError, Repo
 from renku.core import errors
 from renku.core.management.clone import clone
 from renku.core.management.config import RENKU_HOME
-from renku.core.models.datasets import Dataset, DatasetFile, DatasetTag
+from renku.core.models.datasets import Dataset, DatasetFile, DatasetTag, \
+    is_dataset_name_valid
 from renku.core.models.git import GitURL
 from renku.core.models.locals import with_reference
 from renku.core.models.provenance.agents import Person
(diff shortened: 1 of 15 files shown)


emmjab:
I tried the following:
```
$ renku dataset create new -d "this is a test dataset" -c "Emma <e.jablonski@gmail.com>" --display-name "mountain"
OK
$ ls
Dockerfile		requirements.txt
```

(1) didn't realize that the `data` directory and a `new` or `mountain` subdirectory wasn't going to be created

```
(env) IC-LM-208782:my_new_project jablonsk$ renku dataset 
ID                                    DISPLAY_NAME    VERSION    CREATED              CREATORS
------------------------------------  --------------  ---------  -------------------  ----------
3041b30a-df24-4dcb-a0b9-709340433f88  mountain                   2019-12-04 14:31:11  Emma
(env) IC-LM-208782:my_new_project jablonsk$ renku dataset add new /Users/jablonsk/Documents/permafrosthackathon/permafrost-refresh/data/annotations/annotations.zip 
Error: Dataset "new" does not exist.
Use "renku dataset create new" to create the dataset or retry "renku dataset add new" command with "--create" option for automatic dataset creation.
```

(2) Didn't realize `display-name` was then going to be how I add files to my dataset & the name of the directory inside data.

```
(env) IC-LM-208782:my_new_project jablonsk$ renku dataset add mountain /Users/jablonsk/Documents/permafrosthackathon/permafrost-refresh/data/annotations/annotations.zip 
Warning: Adding data from local Git repository. Use remote's Git URL instead to enable lineage information and updates.
(env) IC-LM-208782:my_new_project jablonsk$ ls
Dockerfile		data			requirements.txt
```

- Also checked that the name & email get validated (i.e. has to be of the format "Name <email>")
- that you can pass in any combination of the options

I don't quite understand `display_name` -- what makes something a valid renku name?

mohammad-sdsc:
1. Data directory is created when you add a file to the dataset.
2. `display_name` is used as a valid Renku name, however, its name is a bit misleading and we might change this.

emmjab:
> 1. Data directory is created when you add a file to the dataset.

Not gonna dig into this one now; can see it going either way (i.e. if the directory shows up, users will think they can add files to the directory without `renku dataset add`-ing them, but if it doesn't show up it's kind of opaque about what happens when you "create a dataset".

> 2. `display_name` is used as a valid Renku name, however, its name is a bit misleading and we might change this.

What's the definition of a "valid Renku name"?

> 3. But then ls-files still shows the dataset being called new

What about the inconsistency between whether it's named by `renku dataset create <firstname>` v. `renku dataset create --display_name <lastname>`?

mohammad-sdsc:
> Not gonna dig into this one now; can see it going either way (i.e. if the directory shows up, users will think they can add files to the directory without `renku dataset add`-ing them, but if it doesn't show up it's kind of opaque about what happens when you "create a dataset".

Please create an issue for this if you think we need to discuss it. 

> What's the definition of a "valid Renku name"?

A valid [Git reference](https://git-scm.com/docs/git-check-ref-format). Basically, one can use alphanumeric, ., -, and _. Some more characters are allowed by Git to use but we should (and will) disallow them.

> What about the inconsistency between whether it's named by `renku dataset create <firstname>` v. `renku dataset create --display_name <lastname>`?

The `--display-name` option is removed due the apparent confusion it causes. Now, there is only dataset's `name` (`<firstname>` in your example).

rokroskar:
Since we are now using `internal_name`, could we also specify `--name` or `--title` to give it a more human-readable name?

renku/core/management/datasets.py
```diff
@@ -174,6 +154,54 @@ def with_dataset(self, name=None, identifier=None, create=False):
 
         dataset.to_yaml()
 
+    def create_dataset(
+        self, name, internal_name='', description='', creators=()
```
rokroskar: should allow for specifying an identifier?

mohammad-sdsc:
> Since we are now using `internal_name`, could we also specify `--name` or `--title` to give it a more human-readable name?

We already have a `name` field for datasets which is the same as `internal_name` in locally-created datasets and is whatever-is-set-as-name for imported datasets. Probably, we should use a `title` field for human-readable names for both of these two types of datasets but in that case we are changing `name` field for imported datasets.

rokroskar:
Right - I'm wondering if we could follow the same convention for newly-created datasets - if the name given by the user matches the "allowed characters" regex, then we keep it as is and `name` is the same as `internal_name`. However, if a user specifies the name in this way:

```
renku dataset create "Exploratory dataset"
```

then this would become

```
name -> "Exploratory dataset"
internal_name -> exploratory.dataset
```

mohammad-sdsc:
I believe this will confuse some users as they expect to the name `"Exploratory dataset"` to refer to the dataset but they must actually use `exploratory.dataset`. This is the same situation as a previous commit of this PR that we had `--display-name` and a `name`.

rokroskar:
What if we followed up the dataset creation with a helpful message that said something like:

```
Creation successful! Use the name exploratory.dataset to refer to this dataset.
```

We also need something like this to follow the import command anyway.

mohammad-sdsc:
I had to force-push to resolve merge conflicts. Please review the last two commits.

rokroskar:
Cool! I think that works nicely:

```
renku dataset create "My awesome dataset"
Use the name "my_awesome_dataset" to refer to this dataset.
OK
 rok   master  ❲c❳ base  ~/Projects/renku-test-projects/test-850  renku dataset
ID                                    INTERNAL_NAME             VERSION    CREATED              CREATORS
------------------------------------  ------------------------  ---------  -------------------  ---------------------------------------
8bdee124-1a8a-408e-9ff1-b98f9ba007c4  my_awesome_dataset                   2019-12-16 14:59:50  R.Roskar
```

@emmjab what do you think?

emmjab:
Hah -- I think this is still confusing. Why does the name have to change? It has to be a valid git name? Why? it's not a submodule anymore, was that why?

rokroskar:
Because it's annoying to type something that is super long and (potentially) contains special characters?

rokroskar:
It's not obvious in this example, clearly, but if you import a dataset from zenodo it's likely that the name will be annoyingly long. You don't want to have to type that or copy/paste that every time you use it

emmjab:
hmm... autocomplete? 😅 

> if you import a dataset from zenodo it's likely that the name will be annoyingly long

Why is that? I can google this

rokroskar:
Take https://zenodo.org/record/3549866 for example. The name is "Synthetic dataset used in "The maximum weighted submatrix coverage problem: A CP approach"" - how do you suppose to use that on the command line?

emmjab:
why not use the DOI?

rokroskar:
Ok so @emmjab and I have been discussing this a bit offline. I find "internal_name" a bit confusing - what if we called it "short_name"? We could store this using `schema.org/alternateName` to give the user the option to set it or change it. The default would be what we are doing now, i.e. automatically generate it. 

So it would look like

```
dataset create --shortname dataset1 "This ia a dataset about stuff"
```

or 

```
dataset import --shortname dataset1 <DOI>
```

Issue #479, opened 2019-03-18
cli: renku doctor should check pre-commit hooks


jirikuncar:
Add check that the pre-commit hooks exist and were not modified (linked to current renku sources).

rokroskar:
We should most likely check these more often if not on every invocation of the `renku` command.

PR #854, opened 2019-12-10
feat: check for required git hooks

mohammad-sdsc:
# Description

`renku doctor` command also checks for the existence of the git hooks and checks that their content is up-to-date.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/479

### DIFF
--- a/renku/core/commands/checks/githooks.py
+++ b/renku/core/commands/checks/githooks.py
@@ -15,20 +15,23 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Define repository checks for :program:`renku doctor`."""
+"""Check for required Git hooks."""
+from pathlib import Path
 
-from .githooks import check_git_hooks_installed
-from .migration import check_dataset_metadata, check_missing_files
-from .references import check_missing_references
-from .validate_shacl import check_project_structure, check_datasets_structure
+from git.index.fun import hook_path as get_hook_path
 
-# Checks will be executed in the order as they are listed in __all__.
-# They are mostly used in ``doctor`` command to inspect broken things.
-__all__ = (
-    'check_git_hooks_installed',
-    'check_dataset_metadata',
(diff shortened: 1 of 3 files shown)


renku/core/commands/checks/githooks.py
```diff
@@ -0,0 +1,72 @@
...
+                'Use "renku githooks install" to install them. \n'
+            return False, message
+
+        with hook_path.open() as file_:
+            actual_hook = _extract_renku_hook(file_)
+        with StringIO(_read_resource(hook)) as file_:
+            expected_hook = _extract_renku_hook(file_)
+
+        if not expected_hook:
+            message = WARNING + 'Cannot check for existence of Git hooks.\n'
```
rokroskar: I guess this would happen if the renku installation is somehow corrupted? Maybe we should suggest to reinstall renku?

renku/core/commands/checks/githooks.py
```diff
@@ -0,0 +1,72 @@
...
+            return False, message
+
+        if actual_hook != expected_hook:
+            message = WARNING + 'Git hooks are outdated or not installed. ' \
+                'Use "renku githooks install --force" to update them. \n'
```
rokroskar: This works nicely!

renku/core/commands/checks/githooks.py
mohammad-sdsc: This was for myself to avoid comparing empty dicts in case the pattern was not parsed correctly :D. This is basically a safety-net for us in case we update the hooks and mess something up so that we see some tests fail. User should never see this and if something is wrong with their renku installation then they probably won't even reach this code.

mohammad-sdsc:
> One thing that comes to mind though - should we run `renku doctor` automatically every once in a while? The downside is that the the shacl check can be slow. Or maybe we could just suggest to the user periodically to run `renku doctor`.

I am not normally a fan of doing automatic things without users knowing about them. We probably can do this (or ask users to do it) only when there is an update in renku version to make sure that project is compatible with a newer version. I'll create a story for that.

Issue #768, opened 2019-10-24
RaaS: Commit messages should be specified as optional parameters


jsam:
When renku service creates commits, all commit messages are being reported as `app.py` since that is the source (or the command being executed). We need a facility in renku core commands in which we can specify commit messages coming from renku service.

PR #859, opened 2019-12-11
feat: add optional commit message

jsam:
closes: #768

### DIFF
--- a/renku/core/management/git.py
+++ b/renku/core/management/git.py
@@ -230,7 +230,11 @@ class GitCore:
 
     @contextmanager
     def commit(
-        self, commit_only=None, commit_empty=True, raise_if_empty=False
+        self,
+        commit_only=None,
+        commit_empty=True,
+        raise_if_empty=False,
+        commit_message=None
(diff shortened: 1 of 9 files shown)


renku/core/management/git.py
```diff
@@ -302,12 +306,15 @@ def commit(
                 raise errors.NothingToCommit()
             return
 
-        argv = [os.path.basename(sys.argv[0])
-                ] + [remove_credentials(arg) for arg in sys.argv[1:]]
+        if commit_message is None or not isinstance(commit_message, str):
```
mohammad-sdsc: It's good to check for the empty message as well. Also, for the case of non-string message, I'd rather raise an exception or just use `str(commit_message)` to make it visible instead of silently ignoring it. Since we are the user of this API, `str(commit_message)` would be enough.

jsam:
> This is going to be a nice improvement, but why limit it to the dataset commands?

They were not really limited, just implicit. I've added now explicit params to all other commands.

PR #861, opened 2019-12-12
feat: python 3.8 compatibility

lorenzo-cavazzi:
# Description

This PR aims to make renku compatible with python 3.8.

There is an issue with `cgi.escape()` that has been removed in favour of `html.escape()` https://docs.python.org/3.7/library/cgi.html#cgi.escape
I changed that in our codebase but it turns out it's used also in some other libraries that are sub-dependencies. This makes a test fail (specifically for `renku log --format dot-debug`). The second commit simply removes that from the format list -- I guess this workaround is not good since that format was introduced a few months ago

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -9,4 +9,4 @@ renku = {editable = true, extras = ["all"], path = "."}
 [dev-packages]
 
 [requires]
-python_version = "3.7"
+python_version = "3.8"
--- a/Pipfile
+++ b/Pipfile
@@ -9,4 +9,4 @@ renku = {editable = true, extras = ["all"], path = "."}
 [dev-packages]
 
 [requires]
-python_version = "3.8"
+python_version = "3.7"
--- a/renku/core/commands/format/graph.py
+++ b/renku/core/commands/format/graph.py
@@ -199,7 +199,7 @@ def _rdf2dot_reduced(g, stream):
     Adapted from original source:
     https://rdflib.readthedocs.io/en/stable/_modules/rdflib/tools/rdf2dot.html
     """
-    import cgi
+    import html
     import collections
 
     import rdflib
@@ -228,7 +228,7 @@ def _rdf2dot_reduced(g, stream):
 
     def formatliteral(l, g):
         """Format and escape literal."""
-        v = cgi.escape(l)
+        v = html.escape(l)
         if l.datatype:
             return '&quot;%s&quot;^^%s' % (v, qname(l.datatype, g))
         elif l.language:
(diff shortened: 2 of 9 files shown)


renku/core/commands/format/graph.py
```diff
@@ -97,7 +97,6 @@ def dot(graph, simple=True, debug=False, landscape=False, strict=False):
 dot_full = functools.partial(dot, simple=False, landscape=False)
 dot_landscape = functools.partial(dot, simple=True, landscape=True)
 dot_full_landscape = functools.partial(dot, simple=False, landscape=True)
-dot_debug = functools.partial(dot, debug=True)
```
rokroskar: why is this removed? I'd prefer if it stays

lorenzo-cavazzi: When `renku log --format dot-debug` is invoked, a library uses `cgi.escape()` which is not available in 3.8 anymore and therefore fails. I quickly checked and there is no new version compatible with 3.8 at the moment.
The second commit removing the `dot_debug` was intended to show a possible fix. If it makes sense, I can spend a bit more time trying to understand if `dot_debug` can be kept by using a different library.

jsam:
@lorenzo-cavazzi I pushed a patch with updated dependencies on pipenv.lock. Currently, patch is going to patch rdflib until they pushed a fix for 3.8. Last commit they had was ages ago so this might take some time. We can try to submit a patch PR to them with a fix and then update rdflib and remove the patch.

@rokroskar There is a design problem with that removed flag which I would like to remove (if it's only a debug feature) from end user facing interface or rename it if it's an actual feature which is used by someone. If it's the second one I would love to understand the use case (cause the same information is on both `dot-debug` and `dot-full`), if it's the first one I propose we rename it to `dot-extract-types` / `dot-show-types` / `dot-typed` cause that's reflecting on whats doing a little bit more.

rokroskar:
@jsam we can have that discussion separately, but it has nothing to do with this PR so please remove the commit.

jsam:
@rokroskar Why? I fixed the problem and put back the removed flag.

rokroskar:
Ah ok I didn't see that you put it back. That's fine then IMO.

Pipfile
```diff
@@ -9,4 +9,4 @@ renku = {editable = true, extras = ["all"], path = "."}
 [dev-packages]
 
 [requires]
-python_version = "3.7"
+python_version = "3.8"
```
rokroskar: is this intended? I don't think we want to force 3.8?

jsam: It's intended yes. I see no reasons why not, do you? Want me to roll it back?

rokroskar: In my view this should be the default supported version - I think we consider 3.8 experimental, no?

lorenzo-cavazzi: I also believe it makes sense to stay for a bit longer on `3.7` while just providing `3.8` compatibility.

PR #869, opened 2019-12-17
fix: enforce label on SoftwareAgent

### LINKED ISSUE #868: bug: missing label on SoftwareAgent
`SoftwareAgent` should have a label, but it does not due to missing `rdfs` definition in the context.

rokroskar:
# Description

label was missing from `SoftwareAgent`

Fixes #868


renku/core/models/provenance/agents.py
```diff
@@ -142,6 +142,7 @@ def __attrs_post_init__(self):
     context={
         'prov': 'http://www.w3.org/ns/prov#',
         'wfprov': 'http://purl.org/wf4ever/wfprov#',
+        'rdfs': 'http://www.w3.org/2000/01/rdf-schema#'
```
jsam:
```suggestion
        'rdfs': 'http://www.w3.org/2000/01/rdf-schema#',
```

Issue #866, opened 2019-12-16
Wrong metadata when importing and extracting datasets


mohammad-sdsc:
**Describe the bug**
If we use `--extract` to decompress an imported dataset then the extracted files are not added to the datasets metadata.

**To Reproduce**
```
$ renku dataset import 10.5281/zenodo.597964 --extract

CHECKSUM                          NAME                SIZE (MB)  TYPE
--------------------------------  ----------------  -----------  ------
15da253c243022e510c0ad4801cbda43  pyndl-v0.6.4.zip     0.498099  zip
Do you wish to download this version? [y/N]: y
patool: Extracting /tmp/tmpdjcza12u/pyndl-v0.6.4.zip ...                                                                                                                                              
patool: ... /tmp/tmpdjcza12u/pyndl-v0.6.4.zip extracted to `/tmp/tmpdjcza12u'.
OK

$ ls data/pyndl_naive_discriminat_v064/quantling-pyndl-c34259c/

CONTRIBUTING.rst  doc  LICENSE.txt  Makefile  MANIFEST.in  pyndl  README.rst  requirements.txt  setup.cfg  setup.py  stubs  tests  tox.ini

$ vi .renku/refs/datasets/pyndl_naive_discriminat_v064

# None of the above files are in the metadata
```

**Expected behavior**
Extracted files must be in the metadata.

Issue #858, opened 2019-12-11
Dataset import fails with --extract


mohammad-sdsc:
**Describe the bug**
Importing a dataset which contains non-compressed file with `--extract` causes Renku to fail.

**To Reproduce**
```
$ renku dataset import --extract https://zenodo.org/record/3251128
Ahhhhhhhh! You have found a bug. 
...
Traceback (most recent call last):
  File "/home/mohammad/.local/lib/python3.7/site-packages/patoolib/util.py", line 178, in __call__
    return self.cache[args]
KeyError: (PosixPath('/tmp/tmp4u4528dp/Gorne_Diaz_database_2019.csv'),)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/home/mohammad/sdsc/code/renku-python/renku/cli/dataset.py", line 362, in download_file_with_progress
    stream_to_file(r)
  File "/home/mohammad/sdsc/code/renku-python/renku/cli/dataset.py", line 358, in stream_to_file
    extract_dataset(data_folder, local_filename)
  File "/home/mohammad/sdsc/code/renku-python/renku/cli/dataset.py", line 311, in extract_dataset
    patoolib.extract_archive(filepath, outdir=data_folder_)
  File "/home/mohammad/.local/lib/python3.7/site-packages/patoolib/__init__.py", line 684, in extract_archive
    return _extract_archive(archive, verbosity=verbosity, interactive=interactive, outdir=outdir, program=program)
  File "/home/mohammad/.local/lib/python3.7/site-packages/patoolib/__init__.py", line 468, in _extract_archive
    format, compression = get_archive_format(archive)
  File "/home/mohammad/.local/lib/python3.7/site-packages/patoolib/__init__.py", line 291, in get_archive_format
    mime, compression = util.guess_mime(filename)
  File "/home/mohammad/.local/lib/python3.7/site-packages/patoolib/util.py", line 180, in __call__
    self.cache[args] = value = self.func(*args)
  File "/home/mohammad/.local/lib/python3.7/site-packages/patoolib/util.py", line 241, in guess_mime
    mime, encoding = guess_mime_mimedb(filename)
  File "/home/mohammad/.local/lib/python3.7/site-packages/patoolib/util.py", line 265, in guess_mime_mimedb
[...]

Issue #855, opened 2019-12-11
Cannot import datasets that contain ignored files


mohammad-sdsc:
**Describe the bug**
When importing a dataset that contains file names that are ignored by Git, an error message is printed to use `--force` option. However, there is no `--force` in dataset import command and as a result there is not way to import such datasets.

**To Reproduce**
```
$ renku dataset import --extract 10.5281/zenodo.3239980

Do you wish to download this version? [y/N]: y
patool: Extracting /tmp/tmpm0e8p8_8/pybossa-v2.12.2.zip ...                                                                                                                             
patool: ... /tmp/tmpm0e8p8_8/pybossa-v2.12.2.zip extracted to `/tmp/tmpm0e8p8_8'.
Error: The following paths are ignored by one of your .gitignore files:

        data/scifabricpybossa_v212_v2122/Scifabric-pybossa-3010c8c/pybossa/themes/README.md
        data/scifabricpybossa_v212_v2122/Scifabric-pybossa-3010c8c/pybossa/plugins/__init__.py
        data/scifabricpybossa_v212_v2122/Scifabric-pybossa-3010c8c/uploads/.gitignore
        data/scifabricpybossa_v212_v2122/Scifabric-pybossa-3010c8c/alembictest.ini
Please use the "--force" option if you really want to add them.

$ renku dataset import --extract --force 10.5281/zenodo.3239980

Usage: renku dataset import [OPTIONS] URI
Try "renku dataset import -h" for help.

Error: no such option: --force
```

**Additional context**
Add `--force` flag or just force-add them.

Issue #839, opened 2019-11-26
Imported datasets use UUID instead of DOI


mohammad-sdsc:
**Describe the bug**
When importing a dataset from Zenodo or Dataverse, the dataset ID is a UUID instead of being DOI of the imported dataset. Similarly, the dataset directory in `.renku/datasets` is also a UUID.

**To Reproduce**
```
renku init
renku dataset import https://zenodo.org/record/3251128
ls .renku/datasets/
```

**Expected behavior**
Dataset's DOI should be used for ID and directory name.

**Additional context**
Make sure that `renku migrate datasets` works after this fix.

mohammad-sdsc:
After discussion with @Panaetius, it came out that this is a feature and not a bug. If we use DOI as ID there will be an issue with dataverse, namely that dataverse has versions like 4.0 but doesn't change DOIs between versions, so a user might try to add the "same" dataverse dataset with different versions (with different short_names), which would fail as the DOI is the same.

Moreover, there's also the usecase of adding a dataset twice with different short_names so it ends up in different `data/` subfolders.

rokroskar:
> After discussion with @Panaetius, it came out that this is a feature and not a bug. If we use DOI as ID

Do you mean `id` or `identifier`? I guess you mean `identifier`.

mohammad-sdsc:
After discussion with @jachro and @rokroskar, we decided to give imported datasets unique IDs and not using DOI as ID. We rely on `sameAs` property to check if two datasets are identical or have the same origin.

PR #882, opened 2019-12-19
fix: various bugs in dataset import

mohammad-sdsc:
# Description

Various bug fixes regarding dataset imports (`--extract` won't add extracted files to metadata, ...). It also refactors the import logic. 

~~Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/839~~
Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/855
Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/858
Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/866

### DIFF
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -357,7 +357,7 @@ def download_file_with_progress(extract, data_folder, file, chunk_size=16384):
         if extract:
             extract_dataset(data_folder, local_filename)
 
-    with requests.get(file.url.geturl(), stream=True) as r:
+    with requests.get(file.url, stream=True) as r:
         r.raise_for_status()
         stream_to_file(r)
 
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -246,6 +246,7 @@ import click
 import editor
 import requests
 import yaml
+from tqdm import tqdm
 
 from renku.core.commands.dataset import add_file, create_dataset, \
     dataset_parent, dataset_remove, edit_dataset, export_dataset, \
@@ -256,6 +257,7 @@ from renku.core.commands.format.dataset_files import DATASET_FILES_FORMATS
 from renku.core.commands.format.dataset_tags import DATASET_TAGS_FORMATS
 from renku.core.commands.format.datasets import DATASETS_FORMATS
 from renku.core.errors import DatasetNotFound, InvalidAccessToken
+from renku.core.management.datasets import DownloadProgressCallback
 
 
 def prompt_access_token(exporter):
@@ -579,9 +581,33 @@ def import_(uri, short_name, extract):
         uri=uri,
         short_name=short_name,
         extract=extract,
-        interactive=True,
+        with_prompt=True,
+        progress=_DownloadProgressbar
     )
     click.secho('OK', fg='green')
+
+
+class _DownloadProgressbar(DownloadProgressCallback):
+    def __init__(self, description, total_size):
+        """Default initializer."""
+        self._progressbar = tqdm(
+            total=total_size,
+            unit='iB',
+            unit_scale=True,
+            desc=description,
+            leave=False,
+            bar_format='{desc:.32}: {percentage:3.0f}%|{bar}{r_bar}'
+        )
+
+    def update(self, size):
+        """Update the status."""
+        if self._progressbar:
+            self._progressbar.update(size)
+
+    def finalize(self):
+        """Called once when the download is finished."""
+        if self._progressbar:
+            self._progressbar.close()
 
 
 @dataset.command('update')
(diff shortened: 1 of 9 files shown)


renku/cli/dataset.py
```diff
@@ -297,71 +292,6 @@ def prompt_tag_selection(tags):
     return None
 
 
-def download_file_with_progress(extract, data_folder, file, chunk_size=16384):
```
Panaetius: This method was put here when we refactored renku into `cli` and `core`. The download method was originally part of the download code, but since it contains cli/ui stuff (showing the progress bar via `tqdm`) that doesn't make sense in e.g. the Service part, I moved the download method to `renku.cli`. Ideally, only the progress bar parts would have to be injected and the main download functionality would be in `renku.core` but due to the download code being a bit of a mess, I decided that would be too much effort to split up and it'd be easier to inject download as a single function (plus multiprocessing also doesn't make things any easier). See https://github.com/SwissDataScienceCenter/renku-python/issues/737 and https://github.com/SwissDataScienceCenter/renku-python/issues/737 

Moving tqdm back to `renku.core` is suboptimal, but i guess using the `disabled=` property works. The main issue is, if the UI ever implements polling for this operation or or there is some other client that needs feedback on the progress of the operation in some other form, this will be a mess. Now that we're actually implementing the renku service, we can make a more informed decision than we did back then, maybe @jsam can chime in on this.

Changing import to use dataset add is a neat idea, though.

renku/core/management/datasets.py
```diff
@@ -356,30 +388,56 @@ def _add_from_local(self, dataset, path, link, destination):
...
-    def _add_from_url(self, dataset, url, destination):
+    def _add_from_urls(self, dataset, urls, destination, extract, interactive):
+        destination.mkdir(parents=True, exist_ok=True)
+
+        files = []
+        max_workers = min(os.cpu_count() + 4, 8)
+        with concurrent.futures.ThreadPoolExecutor(max_workers) as executor:
+            futures = {
+                executor.submit(
+                    self._add_from_url, dataset, url, destination, extract,
```
Panaetius: This will lead to a regession where tqdm progress bars are broken (display bug) when importing datasets with many files (e.g. test with `doi:10.7910/DVN/TJCLKP`). the `id_queue` code higher up was added to fix that issue. The code used to create a fixed number of slots that progress bars can be shown in according to the number of concurrent processes and each worker would only write to the progress bar in its slow, preventing multiple parallel workers with progress bars from interfering with each other.

renku/cli/dataset.py
mohammad-sdsc: I've added a callback class and moved the progress bar out of the core.

renku/core/management/datasets.py
mohammad-sdsc: I've tested this before and I've tested it again with DOI you provided and the progress bars do not overlap and do not leave a trace. The difference is that this new implementation uses multiple threads instead of multiple processes. TQDM takes care of the overlaps in the multi-threaded mode and we don't need to do anything special. You can test it with something that takes a while to download and see the progress bars next to each other: `https://doi.org/10.5281/zenodo.3583477`

Moreover, the current implementation shows only one progress bar at any time; I did not check that if it is just the display or it does not download in parallel in the first place. If you checkout the master branch and use the above DOI you should be able to re-produce this issue. I've found this when I was doing the refactoring and since this PR solves that issue I did not create a bug report.

Panaetius: The problem comes when there's less lines in the terminal than files being downloaded, then it gets a weird scrolling behaviour/display bug. You can see it if you drag the terminal smaller than fullscreen. With the current master there's 4 bars at a time (depends in number of CPUs), which doesn't have this issue. So I don't see the issue you mention on master:
master_progress


Not sure if a cosmetic issue like this is worth our time, though, it's a bit complicated :)

renku/cli/dataset.py
Panaetius: Great, thank you!

renku/core/management/datasets.py
mohammad-sdsc: You are right in both cases. Sorry for my comment about the current implementation; for some reason I was excepting to see 5 parallel downloads per core ;).
Please let me know what do you think we should do regarding this corner case so that we can close this PR: keep the new code or revert to the old one? I'd prefer to keep the new code since it is more concise. Someone who is using less than 8 lines in console probably doesn't care much about polluted console output :).

Panaetius: I also think it's not worth fixing. The fix is uglier than the problem, it's purely cosmetic and only happens very rarely. With the old implementation it wasn't too big of a change to fix it, but now that it's combined with renku add, I think we should just leave it this way and just live with the display issue :).

PR #883, opened 2019-12-20
feat: Plugin support for renku-run

### LINKED ISSUE #867: feat: enable workflow graph extensions
As a Renku user, I would like to add custom metadata to the `ProcessRun` node upon completion of a `renku run` execution. The process could be something like this:

1. The 3rd party library needs to be made aware of renku and produce the required metadata. It should store it in the `.renku/<extension-subdir>`
2. There should be a factory method of `ProcessRun` that instantiates the relevant subclass if the current commit contains metadata in the designated place
3. `renku log` will automatically include the metadata attached to the subclassed `ProcessRun` instance. 

addresses #594 
helps with #753 

Outstanding issues not considered here: how to extend the SHACL shapes checked by `renku log --strict` to include the third-party extensions.

Panaetius:
Adds plugin functionality to renku run to record additional metadata

Closes #867


docs/plugins.rst
```diff
@@ -0,0 +1,52 @@
...
+    from setuptools import setup
+
+    setup(
+        ...
+        entry_points={"renku": ["name_of_plugin = myproject.pluginmodule"]},
+        ...
+    )
+
+
+where `myproject.pluginmodule` points to a renku `hookimpl` e.g.:
```
mohammad-sdsc: Perhaps we should use Renku (with a capital R) in the docs.

rokroskar:
Would be cool to see a demo of this functionality with the associated graph that is generated via the plugin. @Panaetius could you generate one and add it to the PR description?

renku/core/models/cwl/annotation.py
```diff
@@ -0,0 +1,39 @@
...
+    cmp=False,
+)
+class Annotation:
+    """Represents a custom annotation for a research object."""
+
+    _id = jsonld.ib(context='@id', kw_only=True)
+
+    body = jsonld.ib(default=None, context='oa:hasBody', kw_only=True)
+
+    source = jsonld.ib(default=None, context='dcterms:creator', kw_only=True)
```
rokroskar: why `dcterms` and not `schema`?

Panaetius: That's what the Web Annotation Vocabulary uses: https://www.w3.org/TR/annotation-vocab/#dcterms-creator

PR #893, opened 2020-01-10
fix: integration tests

mohammad-sdsc:
# Description

Fixes integration tests failure after https://github.com/SwissDataScienceCenter/renku-python/pull/882.


tests/cli/test_integration_datasets.py
```diff
@@ -59,10 +58,6 @@ def test_dataset_import_real_doi(runner, client, doi, sleep_after):
     assert doi['file'] in result.output
     assert doi['creator'] in result.output
 
-    doi_path = urllib.parse.quote(doi['doi'], safe='')
-    metadata_path = client.renku_path / 'datasets' / doi_path / 'metadata.yml'
-    assert metadata_path.exists()
```
rokroskar: why remove this?

mohammad-sdsc: We don't use the DOI anymore so there is no need to check for this specific path (it's a UUID now and we can't check it like this). We also check the proper creation of the dataset in other tests.

Issue #889, opened 2020-01-09
Do not use DOI as dataset ID when migrating projects


mohammad-sdsc:
**Describe the bug**
When migrating a project with imported datasets, DOI of the dataset is used as its `id` and `identifier`. This is not in par with the import logic which gives unique IDs to datasets. Like this, two datasets in different project are shown as one in the KG even if they are modified.

**Expected behavior**
Each dataset must have a unique `id` (e.g. a UUID) which parts of it is used as its `identifier`.

Related: https://github.com/SwissDataScienceCenter/renku-python/issues/839

PR #896, opened 2020-01-10
fix: do not use doi as id when migrating

mohammad-sdsc:
# Description

This PR avoids using DOI as `identifier` or `_id` for the imported datasets when migrating them. It also replaces DOI identifiers with UUID for existing datasets (if any).

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/889


mohammad-sdsc:
> The `renku/core/commands/checks/` checkes are usually performed by `renku doctor`, but the new check is run on any dataset command instead. But migrations are performed automatically in `jsonld.py` around line 500. (Including the `migrate_doi_identifier` in this PR). Why do we need to check for migrations on every dataset command if they're executed automatically on loading data?
> 
> Would it make sense to add a SHACL check to make sure identifier is a UUID?
> 
> And we really need a more sane way to deal with migrations instead of executing them on the fly every time the dataset is loaded. But I think there's already an issue for this.

I know. The only thing is that dataset metadata file will be in the wrong directory until with run the migration. This usually does not cause any problem. I thought it might be better to keep everything tidy all the time. I can happily revert the change and rely on `renku doctor` to require migration. 

I couldn't agree more with you regarding the migrations: We shouldn't rely on heuristics/magic for that.

I am not sure if SHACL check is OK to add because when we import a dataset, it temporarily has DOI as its ID until we create another dataset in Renku. We use this ID to set `sameAs` field of the created dataset. This is a bit ugly and we should fix it as well but for now let's not add a SHACL check.

Issue #876, opened 2019-12-18
Dataset create raises `Dataset exists` for similar names


jsam:
There is a bug on dataset create command for datasets with long names and high similarity score.

Example:
```
renku dataset create "my super awesome dataset research project with super long name 1" 
```
results in
```
Use the name "my_super_awesome_dataset" to refer to this dataset.
OK

```

Now we create another one:
```
renku dataset create "my super awesome dataset research project with super long name 2"
```
results in
```
Error: Dataset exists: "my_super_awesome_dataset".
```

Expected behavior is that both of those datasets should be created.

mohammad-sdsc:
We cannot avoid this fully because we create `short_name` from `name`. Even if we remove the length limitation we cannot do something meaningful regarding special character: the following two commands create the same `short_name` and the second one will fail:
```
renku dataset create 'some name'
renku dataset create `#some name#`
```

One solution will be to treat `name` as `short_name` (i.e. like before) in `renku dataset create` and pass a human-friendly name with `--name` or `--title`. @rokroskar what do you think?

rokroskar:
@mohammad-sdsc yes this (I think) was sort of along the lines of what I was advocating for before at some point. It has the added benefit of solving the issue that @emmjab has brought up, which is that the user creates a dataset with some name but then the name they need to use to refer to that dataset is somehow different.

PR #898, opened 2020-01-13
fix: dataset naming

mohammad-sdsc:
# Description

When creating a dataset, users now provide a `short_name` which is easy to work with in a command line shell and an optional `title` which is human-readable name for the dataset. Since all cli commands now work with `short_name`, this PR also uses `short_name` consistently in the command line.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/876


renku/core/models/datasets.py
```diff
@@ -540,20 +540,21 @@ def __attrs_post_init__(self):
             )
 
 
-def is_dataset_name_valid(name, safe=''):
-    """A valid name is a valid Git reference name with no /."""
-    # TODO make name an RFC 3986 compatible name and migrate old projects
+def is_dataset_short_name_valid(short_name, safe=''):
```
Panaetius: Can we remove `safe`? it's not actually used.

mohammad-sdsc: Done!

PR #901, opened 2020-01-14
fix: changes dataset description default value to None

### LINKED ISSUE #849: Dataset's description and person's affiliation not to occur when blank
It looks `renku log` produces Dataset's `description` and person's `affiliation` properties even if their values are blank. They should not be created in such cases.

Panaetius:
Changes default value of dataset.description to None, so it's not exported in `renku log` if it's not set.

I couldn't reproduce affiliation being empty.

fixes #849


mohammad-sdsc:
There is this `test_dataset_deserialization` in `tests/core/commands/test_serialization.py` which I believe should fail after this change. Can you please verify it? I guess you should also change the dataset model and `DatasetApiMixin`:
```
    description = jsonld.ib(
        default='', context='schema:description', kw_only=True
    )
```

```
    def create_dataset(
        self, name, short_name=None, description='', creators=None
    ):
```

renku/core/commands/providers/dataverse.py
```diff
@@ -211,6 +211,13 @@ def as_dataset(self, client):
         files = self.get_files()
         dataset = Dataset.from_jsonld(self._json, client=client)
 
+        if dataset.description == '':
```
jsam:
```suggestion
        if not dataset.description.strip()
```

PR #902, opened 2020-01-14
fix: integration tests

jsam:


renku/core/utils/requests.py
```diff
@@ -0,0 +1,42 @@
...
+from contextlib import contextmanager
+
+import requests
+from requests.adapters import HTTPAdapter
+from urllib3.util.retry import Retry
+
+
+@contextmanager
+def retry(
+    total_requests=5, backoff_factory=1, statuses=(500, 502, 503, 504, 429)
```
Panaetius: Typo, should be `backoff_factor`

Issue #908, opened 2020-01-15
Incorrect file names after import


mohammad-sdsc:
**Describe the bug**
After importing a dataset from Dataverse, dataset files have incorrect names.

**To Reproduce**
```
$ renku dataset import 10.7910/DVN/F4NUMR --short-name dataset
CHECKSUM    NAME                                 SIZE (MB)  TYPE
----------  ---------------------------------  -----------  -------------------------
            Candidate Web Coding Sheet            0.16896   application/msword
            Candidate Website Data, 2002-2006     0.583644  text/tab-separated-values
            Data Key 2002-2006                    0.128     application/msword
            Data Key 2008                         0.312832  application/msword
            Data Key 2010                         0.300544  application/msword
            Errata                                0.024064  application/msword
            Information about Web Data            0.023552  application/msword
Do you wish to download this version? [y/N]: y
OK                                                                                                                                                                         

$ ls data/dataset/                                                                                                                                       
2327548  2327549  2327550  2327551  2327552  2327553  2359758

```
**Expected behavior**
Files must retain their original names.

Issue #668, opened 2019-09-11
enable export to dataverse


rokroskar:
We should enable export of datasets to dataverse. See the data deposit API docs here: http://guides.dataverse.org/en/latest/api/getting-started.html#depositing-data

Also see: https://github.com/IQSS/dataverse-sample-data/blob/407aba72cfa140ae18e9ab80d2e504e0686252a9/create_dataset.py

cc @pdurbin

pdurbin:
@rokroskar I just held my screen up to @cchoirat 's screen to show you slide 11 from https://www.dataone.org/sites/default/files/sites/all/documents/dataone-wt-webinar-2019-10-08.pdf

Here's a screenshot:

<img width="1259" alt="Screen Shot 2019-11-05 at 11 05 14 AM" src="https://user-images.githubusercontent.com/21006/68224295-523a5580-ffbc-11e9-8237-3daa3dc9eeb1.png">

That's from the "Developing, Packing, and Sharing Reproducible Research Objects: The Whole Tale Approach" video at https://www.dataone.org/webinars/developing-packaging-and-sharing-reproducible-research-objects-whole-tale-approach where @craig-willis gave a very nice demo of exporting capsules from Whole Tale to DataONE using that BagIt-RO standard.

Here's where we're tracking Capsulization and Packaging of Replication Objects in Dataverse: https://github.com/IQSS/dataverse/issues/6085

PR #909, opened 2020-01-15
feat: dataverse export

mohammad-sdsc:
# Description

Enables exporting datasets to Dataverse. Fixes wrong file names after import.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/908
Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/668


conftest.py
```diff
@@ -363,6 +363,17 @@ def zenodo_sandbox(client):
     )
 
 
+@pytest.fixture
+def dataverse_demo(client):
+    """Configure environment to use Dataverse demo environment."""
+    client.set_value(
+        'dataverse', 'access_token', '4ca13597-cf43-4815-8763-b64994058c19'
```
jsam: Access token should probably be extracted into env variable and encrypted for the [CI](https://docs.travis-ci.com/user/encrypting-files/).

tests/cli/test_integration_datasets.py
```diff
@@ -269,8 +269,26 @@ def test_dataset_reimport_removed_dataset(runner, project, sleep_after):
 
 
 @pytest.mark.integration
+def test_dataset_import_preserve_names(runner, project, sleep_after):
+    """Test import keeps original file names."""
+    DOI = '10.7910/DVN/F4NUMR'
```
jsam: Let's not define constants in local scopes, otherwise pylint is crying. Just make it normal variable (lower case).

renku/core/commands/providers/dataverse.py
```diff
@@ -270,7 +278,180 @@ def remote_url(self):
...
+    def _make_url(self, api_path, **query_params):
+        """Create URL for creating a dataset."""
+        url_parts = urlparse.urlparse(self.server_url)
+        path = pathlib.posixpath.join(DATAVERSE_API_PATH, api_path)
+
+        query_params = urllib.parse.urlencode(query_params)
+        url_parts = url_parts._replace(path=path, query=query_params)
+        return urllib.parse.urlunparse(url_parts)
+
+    def _post(self, url, json=None, data=None, files=None):
```
jsam: We should make this more robust, with #902 merged there is a new utility for retrying.
With it you could:
```
with retry() as session:
    session.post(...)
```

tests/cli/test_integration_datasets.py
```diff
@@ -269,8 +269,26 @@ def test_dataset_reimport_removed_dataset(runner, project, sleep_after):
 
 
 @pytest.mark.integration
```
jsam: With #902, there is a new decorator with which you can specify integration tests for a rerun

```
@flaky(...)
...
```

This should make IT more robust in case of failure when running this in parallel.

conftest.py
mohammad-sdsc: This a free account on their demo server (same as the one above for Zenodo). I'd prefer to keep it like this for easier execution of tests in our local machines.

Issue #860, opened 2019-12-12
Dataset content preview table has wrong indentation


rokroskar:
**Describe the bug**

```
renku dataset import 10.5281/zenodo.3549866
CHECKSUM                          NAME              SIZE (MB)  TYPE
--------------------------------  --------------  -----------  ------
e00d2805076a00aa5515fb988fda6573  data_small.zip     4.69636   zip
8fdc30aa0546ed2b8a520c06b4b387b4  data.zip        1635.33      zip
4008ea48e3afae4d8466fdb9b9f2f65a  generation.py      0.010506  py
```

Indentation is inconsistent.

mohammad-sdsc:
This looks pretty consistent to me. Can you add more context?

rokroskar:
Ideally the size column would be either left or right justified, no?

mohammad-sdsc:
It is aligned around the decimal point which to me makes it easy to read and compare the sizes. Maybe we should limit the number of digits after the decimal point to some thing smaller than 6 (e.g. 2)?

Issue #847, opened 2019-11-30
feat: show loading bar for file download from url


emmjab:
**Is your feature request related to a problem? Please describe.**
When a user calls `renku dataset add <dataset> <url>` the user does not get any feedback about how much data is being downloaded.

**Describe the solution you'd like**
1. before continuing with the download, inform the user about how much data is coming so they have some idea of how long they'll have to wait, and inform the user it's too much for the pod to handle if possible (i.e. I found out the hard way that max 20GB can be handled in the pod, and then everything is lost because you can't push to gitlab from a frozen notebook).
2. in addition, show a loading bar while the data is downloading.

I think you can get the size of the data in the following way:
```
>>> import requests
>>> response = requests.head('https://storageaccountperma8980.blob.core.windows.net/hackathon-on-permafrost/timelapse_images_fast.zip')
>>> response.headers
{'Content-Length': '7223888389', 
'Content-Type': 'application/zip',
 'Content-MD5': '5ROuO4TmjSAx6iH6gBu7eA==', 
'Last-Modified': 'Thu, 28 Nov 2019 14:02:31 GMT',
 'ETag': '0x8D7740B9C8B23EB',
'Server': 'Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0', 
'x-ms-request-id': 'cc747c09-a01e-00de-7e73-a790c5000000', 
'x-ms-version': '2009-09-19',
'x-ms-lease-status': 'unlocked', 
'x-ms-blob-type': 'BlockBlob', 
'Date': 'Sat, 30 Nov 2019 11:44:59 GMT'}
```

**Additional context**
I think this is pretty high priority considering how many questions users have about how much data they can use.

**Resources**
Here's an example dataset for downloading; it's between 8 and 9 GB: renku dataset add images https://storageaccountperma8980.blob.core.windows.net/hackathon-on-permafrost/timelapse_images_fast.zip

rokroskar:
You certainly don't always know the amount of data ahead of time - but I suppose we could check the headers and present the information if it exists.

Panaetius:
This should be fixed already with the newest changes to datasets by @mohammad-sdsc

PR #911, opened 2020-01-16
feat: better UX when adding to a dataset

mohammad-sdsc:
# Description

Adds a progressbar when adding a file from a URL and warns about insufficient disk space.

Fixes https://github.com/swissdatasciencecenter/renku-python/issues/847
Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/860


renku/core/commands/dataset.py
```diff
@@ -182,6 +190,24 @@ def add_to_dataset(
...
+                        total_size += int(r.headers.get('content-length', 0))
+                except requests.exceptions.RequestException:
+                    pass
+        usage = shutil.disk_usage(client.path)
+
+        if total_size > usage.free:
+            GB = 2**30
+            message = 'There isn\'t enough disk space ' \
+                      '(required: {:.2f} GB/available: {:.2f} GB). ' \
+                      'Continue anyway?'.format(total_size/GB, usage.free/GB)
```
jsam: I'm just wondering. Is there a use case for continue in such cases when there is not enough disk space? Doesn't it make more sense just to abort any further computation in case on insufficient disk space?

Panaetius: I agree, abort or make it a loop that rechecks if the user freed up some space in the meantime

mohammad-sdsc: 👍 The operation is aborted if there is not enough disk space.

Issue #821, opened 2019-11-15
Improve terminal outputs on dataset commands


jsam:
Currently, we have some inconsistencies on dataset terminal output commands which we could improve. This ticket is about improving the naming and formating issue of those commands.

1. dataset parent subcommand (which servers as a list) and dataset ls-files format format list of creators differently
2. figure out how to deal with use case when there is 30+ creators in the display list
3. show a relative path for files
4. use display name (or better reference strategy method) as dataset name everywhere (#791)
5. renku created datasets should always resolve to same name

rokroskar:
Is 4. in the list above different from 5.?

PR #927, opened 2020-01-21
feat: improve dataset command output

mohammad-sdsc:
# Description

Minor fixes to dataset commands output. To address long output list, this PR adds a column selection option to `renku dataset` and `renku dataset ls-files` commands where user can use to show a selected list of properties. 

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/821

### DIFF
--- a/renku/core/commands/format/dataset_files.py
+++ b/renku/core/commands/format/dataset_files.py
@@ -16,11 +16,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Serializers for dataset list files."""
-
-from collections import OrderedDict
-
-from renku.core import errors
-from renku.core.models.tabulate import tabulate
+from .tabulate import tabulate
 
 
 def tabular(client, records, *, columns=None):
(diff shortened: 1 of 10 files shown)


renku/core/commands/format/dataset_files.py
```diff
@@ -52,8 +57,33 @@ def jsonld(client, records):
     return dumps(data, indent=2)
 
 
+def _make_headers(columns):
```
Panaetius: This method is identical to the `_make_headers` for `DatasetFile` except for `DATASET_COLUMNS` vs `DATASET_FILES_COLUMNS`. We can have a generic `make_headers` method that takes `DATASET_COLUMNS`/`DATASET_FILES_COLUMNS` as additional input parameter to reduce code duplication.

renku/core/commands/format/datasets.py
```diff
@@ -20,30 +20,46 @@
...
+    headers = _make_headers(columns)
+
+    # Sort based on the first requested field
+    attr = list(headers.keys())[0]
+    datasets = sorted(datasets, key=lambda d: getattr(d, attr))
+
+    return tabulate(datasets, headers=headers, disable_numparse=True)
+
+
+def _make_headers(columns):
```
Panaetius: See other comment

tests/cli/test_datasets.py
```diff
@@ -268,19 +268,47 @@ def test_datasets_list_empty(output_format, runner, project):
...
+            'dataset', 'create', 'my-dataset', '--title', 'Long Title', '-c',
+            'John Doe <john.doe@mail.ch>'
+        ]
+    )
+    assert 0 == result.exit_code
+
+    result = runner.invoke(cli, ['dataset', '--columns', columns])
+    assert 0 == result.exit_code
+    for value in values:
+        assert value in result.output
```
Panaetius: This test would pass if `--columns` is ignored and all columns are output, we should also test that the excluded columns are not present.

renku/core/commands/format/dataset_files.py
mohammad-sdsc: I've actually factored out `tabular` function which also has a similar pattern in both files.

tests/cli/test_datasets.py
mohammad-sdsc: Good point! I've modified this test and added new ones just for that purpose.

PR #928, opened 2020-01-22
fix: moved IT environment secret

jsam:


jsam:
Yeah,thanks for the patiance!

Issue #937, opened 2020-01-23
bug: zenodo dataset with embargo_date


emmjab:
**Describe the bug**
Tried to import a dataset that has an embargo date.

**To Reproduce**
```
$ renku dataset import  10.5281/zenodo.1175627
Ahhhhhhhh! You have found a bug. 🐞

1. Open an issue by typing "open";
2. Print human-readable information by typing "print";
3. See the full traceback without submitting details (default: "ignore").

Please select an action by typing its name (open, print, ignore) [ignore]: 
Traceback (most recent call last):
  File "/home/jovyan/.local/bin/renku", line 8, in <module>
    sys.exit(cli())
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/renku/cli/exception_handler.py", line 128, in main
    self._handle_github()
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/renku/cli/exception_handler.py", line 175, in _handle_github
    getattr(self, '_process_' + value)()
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/renku/cli/exception_handler.py", line 119, in main
    result = super().main(*args, **kwargs)
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/renku/cli/exception_handler.py", line 90, in main
    return super().main(*args, **kwargs)
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

PR #938, opened 2020-01-23
fix: added embargo date on zenodo import

jsam:
closes: #937


rokroskar:
It's great that this fixes the problem but the real issue is that we don't actually care about embargo date and we don't need the internal zenodo metadata. We should just be fetching the jsonld which has enough information for us. This would have the added benefit of standardizing import between zenodo and dataverse and enable importing anything from Google dataset search.

jsam:
I'm closing this PR in favor of #941.

PR #939, opened 2020-01-23
feat: Adds protection for renku relevant paths in dataset add

### LINKED ISSUE #873: Protect .renku directory from being used in commands
Currently it's possible for user to add a data file and place it in `.renku` directory. This behavior should be protected since it can lead to weird use cases. 

For example, the following command should raise a corresponding exception and warn user from such miss use:

```
renku dataset add ds1 .renku/metadata.yml .renku/mydatarepo/*
```

Panaetius:
Adds `.git/, .renku/, Dockerfile, environment.yml, .dockerignore, .gitignore, .gitlab-ci.yml, requirements.txt` as protected files that can't be added to a dataset. Adding to a dataset will store the file on LFS and the gitlab runner doesn't have LFS access, so I added more paths than the original issue asked for.

Closes #873

### DIFF
--- a/tests/cli/test_datasets.py
+++ b/tests/cli/test_datasets.py
@@ -1400,9 +1400,51 @@ def test_add_same_filename_multiple(runner, client, directory_tree):
     result = runner.invoke(
         cli, [
             'dataset', 'add', '--force', 'my-dataset1', directory_tree.strpath,
-            'Dockerfile'
+            'README.md'
         ]
     )
+    assert 0 == result.exit_code
+
+
+def test_add_protected_file(runner, client, tmpdir):
+    """Check adding same filename multiple times."""
+    result = runner.invoke(
+        cli, ['dataset', 'add', '-c', 'my-dataset1', '.renku']
+    )
+
+    assert 1 == result.exit_code
+    assert 'Error: The following paths are protected' in result.output
+
+    result = runner.invoke(
+        cli, ['dataset', 'add', '-c', 'my-dataset1', '.renku/']
+    )
+
+    assert 1 == result.exit_code
+    assert 'Error: The following paths are protected' in result.output
+
+    result = runner.invoke(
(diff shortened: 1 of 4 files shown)


tests/cli/test_datasets.py
```diff
@@ -1400,12 +1400,54 @@ def test_add_same_filename_multiple(runner, client, directory_tree):
     result = runner.invoke(
         cli, [
             'dataset', 'add', '--force', 'my-dataset1', directory_tree.strpath,
-            'Dockerfile'
+            'README.md'
```
jsam: Will github/gitlab show readme.md in case renku moves it to LFS?

jsam: In case it breaks it, maybe we should add this to the blacklist as well.

Panaetius: Probably it will just show a `The source could not be displayed because it is stored in LFS. You can download it instead. ` style message.

But this is not a file that's needed for any process/workflow/ci/what have you, so it doesn't affect renku. So if users want to put it on LFS, I don't see a reason to stop them.

jsam: Sure my point is more aiming at protecting users from themselves and crazy `accidents` people execute on the shell. I agree that there is no need to hardcode blacklisting of such files, maybe we should still think about putting such files to (future) `.renkuignore` which could be related to template important files. One of such files there maybe it could be `README.md`

tests/cli/test_datasets.py
```diff
@@ -1400,12 +1400,55 @@ def test_add_same_filename_multiple(runner, client, directory_tree):
...
         ]
     )
     assert 0 == result.exit_code
 
 
+def test_add_protected_file(runner, client, tmpdir):
```
jsam: Maybe we should future proof this and add it as a parametrized test.

PR #945, opened 2020-01-24
chore: update Pipfile.lock to match latest Pipfile

ableuler:
Concourse keeps failing because of the `Pipfile.lock` being outdated. This PR should fix the problem.


jsam:
Thank you for the PR. This problem keeps biting us now for a while. Maybe we should set up dependabot to fix it properly this time? 

https://dependabot.com/ would be great long term solution for this.

PR #962, opened 2020-01-28
chore: add action to update Pipfile

rokroskar:
# Description

Adds an action to automatically update `Pipfile.lock` whenever `setup.py` changes.

### DIFF
--- a/.github/workflows/pipfile-update.yml
+++ b/.github/workflows/pipfile-update.yml
@@ -1,0 +1,31 @@
+name: check and update Pipfile.lock
+
+on:
+  pull_request:
+    paths:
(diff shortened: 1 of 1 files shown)


jsam:
I think instead of hot gluing this, we should move the dependencies from setup.py to Pipfile and make setup.py read from there. That way dependabot can easily take care of this, and we get a nicer way of managing deps through first-class Pipfile support.

rokroskar:
I was hoping we could do most of our development through github actions... no? :)

rokroskar:
Seems to work: https://github.com/SwissDataScienceCenter/renku-python/runs/413239366?check_suite_focus=true

PR #963, opened 2020-01-28
fix: changes lfs pull paths to be relative to parent

### LINKED ISSUE #903: renku run does not pull dataset data
**Describe the bug**

I expect `renku run` to handle LFS data for me - if I need a file that is not pulled from LFS, it should be pulled automatically. This functionality was implemented, but perhaps it only works files that are created as outputs of a `renku run` command. We should check all paths that we can reasonably interpret as being needed for the command that is being executed and pull the data first.

Panaetius:
LFS does not work with absolute paths, for whatever reason.  So with cwd = `/work/flights-tutorial-data/`.
`git lfs pull -I /work/flights-tutorial-data/data/201901_us_flights_1/2019-01-flights.csv.zip` does not download the file from lfs, but `git lfs pull -I data/201901_us_flights_1/2019-01-flights.csv.zip` does.

I couldn't find any info on this on the net, but this PR fixes it.

Closes #903

### DIFF
--- a/renku/core/management/storage.py
+++ b/renku/core/management/storage.py
@@ -175,7 +175,8 @@ class StorageApiMixin(RepositoryApiMixin):
             client, commit, path = self.resolve_in_submodules(
                 self.repo.commit(), path
             )
-            client_dict[client.path].append(str(path))
+            rel_path = Path(path).relative_to(client.path)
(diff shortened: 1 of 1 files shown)


rokroskar:
Argh, good catch! That's strange (and frustrating). Maybe worth opening up an issue with git-lfs?

Issue #936, opened 2020-01-23
fix: improve service IT


jsam:
Current, service integration tests are executed against 1 repo and work only on master branch. This is a problem cause executing this tests often leads to a certain stress which renku cannot handle. 

Currently, we executed them for a month which lead to 6k commits at which point repository got so big that renku does not finish any of the commands being executed on it. I've tried to execute renku command on it and left it there hanging for over 3 hours and still it did not finish.

The idea of improvement is to eliminate this slowness after each run. To achieve this we should run all the integration tests on it's own branch. This would help significantly with the speed and stability of the IT.

emmjab:
if I'm reading this correctly, this sounds like a useful QA case for improvement of renku performance for repositories with 6k commits. Where does the repo live?

jsam:
It's [here]( https://dev.renku.ch/gitlab/contact/integration-tests).

Issue #862, opened 2019-12-12
feat: service endpoint for project creation


rokroskar:
With #809 we can create projects from any template URL. We should then be able to offer a `/renku/projects.create` endpoint.

The payload would be:

```
{
  "project_name": "myproj",
  "template_url": https://template.url,
  "template_options": 
  {
    "option1": 1,
    "option2": 2
  }
}
```

The template options could be obtained from an endpoint like `/renku/projects.template.options?template_url=https://template.url`

lorenzo-cavazzi:
I am a bit confused by `template_options`. Should they be the variables provided to feed the jinja interpreter? e.g., my template contains `{{ name }}`, and I will provide `{ "name": "Myname" }`?
I would rename that to `template_metadata` or `template_variables`. We are also missing the template name since a template repository may contain more than one, and we should allow the user to chose a tag/branch/commit

* /renku/projects.create
```
{
  "project_name": "myproj",
  "template_url": https://template.url,
  "template_ref": "master",
  "template_name": "Python basic",
  "template_variables": 
  {
    "var1": 1,
    "var2": "abc"
  }
}
```

Should we change the other endpoint name as well and make it a bit more `rpc` style?
* /renku/template.read_variables
```
{
  "template_url": https://template.url,
  "template_ref": "master",
  "template_name": "Python basic",
}
```

* /renku/template.read_manifest
We probably need another endpoint to read the manifest file to present the user with the available template names and descriptions. 
```
{
  "template_url": https://template.url,
  "template_ref": "master"
}
```
I am not sure about the url naming conventions, so the urls are just an example

jsam:
I think we could tackle this topic this sprint. After I knockout all the DS issues I could address this. @lorenzo-cavazzi Is that okay with you?

lorenzo-cavazzi:
If you want to do that firsthand it's ok, otherwise I can also do that in the current sprint :slightly_smiling_face:

PR #964, opened 2020-01-29
fix: added branch isolation in service tests

jsam:
closes: #936


rokroskar:
I'm wondering if it would be better to 

a) create a bot account on dev for this
b) make a fixture that creates a fresh repo for every build

jsam:
I don't think it's necessary mutually exclusive, I would keep both just to have better QA process. I was thinking about adding such test once we implement #862 so we can add this as acceptance criteria there?

For service bot I created a ticket here: #965, so let's add this as a followup.

PR #970, opened 2020-01-30
chore: add automerge action

rokroskar:
Adds the automerge action (see https://github.com/pascalgn/automerge-action).

It will automatically merge approved PRs if you add the `automerge` label. The default merge mode is squash using the automatic github message.


rokroskar:
This doesn't seem to do what we want - it doesn't automatically update the branch for some reason. Not clear why yet...

Issue #815, opened 2019-11-14
create a PoC using a pointer file to link to arbitrary external data


rokroskar:
As a Renku user, I would like to be able to create links to external data from my project and have those links lead to unambiguously identifying the data I am using in the Knowledge Graph. "External data" might be data on a shared filesystem or data that is located off-site. If the data is on a local, shared filesystem, the goal is to use the data without making copies.

One option is to use a "pointer file" that is saved to disk and which contains the information required to replace the pointer file with either a link to the data (if the data is accessible on a local filesystem) or the original raw data. This could be a very generic solution that could be adapted to a variety of backends (local/network storage, cloud storage/object store etc.)

If a pointer file is to be used, a decision must be made how the contents of the pointer file are to be swapped in both directions. One option would be to create custom git filters and rely on the same underlying functionality as git-lfs and similar systems. A demo of using a custom git filter with an arbitrary data server can be found here: https://gist.github.com/rokroskar/9fd57f2a6428c1ae8444972394fc10f9

The PoC should, however consider the simpler case of only linking to data on a local (shared) filesystem. There are a further two possible cases here: 1. the data is in a plain directory on a posix filesystem, and 2. the data is in a directory that is also a git repository. In the case of 1. we should compute the hash and use it as a part of the ID in the graph. In the case of 2. the hash and potentially other (renku-related) metadata can be obtained from the git repository.

mohammad-sdsc:
Pointer file should work fine when we can download/copy the actual data to the repository. Git smudge filter can be used to retrieve the actual content of the file. Since the file is in the Git repo, everything works as before.

However, in the case of local/network storage, we don't want to copy the actual file to the repo and we want to just keep a symlink to the file location.

I've tried having a pointer file and replace it with a symlink upon checkout using a smudge filter (like the provided demo). The issue with this solution is that filtering is not supposed to change the file type and should only change its content; I am changing a normal file to a symlink which shows up as a change in Git and repo will be in a dirty state right after the checkout.

Here, `file.yml` is a pointer file containing a URI which is changed to a symlink by the filter (let's ignore the file exists error which does not cause any harm normally):
```
[~/playground/815]$ git status
HEAD detached at f4d2860
nothing to commit, working tree clean
[~/playground/815]$ git co master
running filter on file.yml
content: "b'uri: /home/mohammad/metadata.yml\n'"
error: unable to create file file.yml: File exists
T       file.yml
[...]

PR #974, opened 2020-01-31
feat(dataset): adding data from external files without copy

mohammad-sdsc:
# Description

To add an external file to a dataset pass `--external` option to the `dataset add` command. To check for modification in external files run `renku dataset update --external`.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/815

### DIFF
--- a/renku/core/commands/checks/external.py
+++ b/renku/core/commands/checks/external.py
@@ -1,0 +1,48 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2020 - Swiss Data Science Center (SDSC)
+# A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+# Eidgenössische Technische Hochschule Zürich (ETHZ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Checks for external files."""
+from pathlib import Path
+
+import click
+
+from renku.core.commands.echo import WARNING
+
+
+def check_missing_external_files(client):
+    """Find external files that are missing."""
+    missing = []
+
+    for path, dataset in client.datasets.items():
+        for file_ in dataset.files:
+            if file_.external:
+                target = Path(file_.path).resolve()
+                if not target.exists():
+                    missing.append((file_.path, str(target)))
+
+    if not missing:
+        return True, None
+
+    problems = (
+        '\n' + WARNING + 'There are missing external files.\n'
+        '  (make sure that external volume is mounted and accessible)' +
+        '\n\n\t' + '\n\t'.join(
+            click.style(path, fg='yellow') + ' -> ' +
+            click.style(target, fg='red') for path, target in missing
+        ) + '\n'
+    )
+    return False, problems
--- a/renku/core/commands/checks/external.py
+++ b/renku/core/commands/checks/external.py
@@ -39,8 +39,8 @@ def check_missing_external_files(client):
 
     problems = (
         '\n' + WARNING + 'There are missing external files.\n'
-        '  (make sure that external volume is mounted and accessible)' +
-        '\n\n\t' + '\n\t'.join(
+        '  (make sure that external paths are accessible)' + '\n\n\t' +
+        '\n\t'.join(
             click.style(path, fg='yellow') + ' -> ' +
             click.style(target, fg='red') for path, target in missing
         ) + '\n'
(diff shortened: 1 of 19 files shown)


rokroskar:
This is starting to look nice! I just tried it and one thing I notice is that in the resulting graph there is no link between the pointer file and the file in the dataset (see below). 

noname gv

Another thought I had was that if we are adding a pointer reference in `.renku` maybe it should be generic? That way I could do something like `renku run --external-input <path-outside-of-repo> myscript.sh`

emmjab:
I tried it out in the following way:

1. created a simple text file in a temporary location
2. added to the dataset with the new "external" flag
3. ran a renku run a script to take that file as an input; write the contents into a new file in the repo
4. pushed the repo to dev.renku.ch

I was able to view the lineage from the output file (but not the data file or the pointer file) [1]
The pointer files were effectively not there in the repo, but the metadata was preserved.

```
renku init
renku dataset create test
renku dataset add test /tmp/testing_renku/data.txt --external
renku run python scripts/modify_test.py data/test/data.txt out.txt
```

[1] https://dev.renku.ch/projects/e.jablonski/test_external/files/lineage/out.txt 
[2] https://dev.renku.ch/projects/e.jablonski/test_external/files/blob/.renku/datasets/f6826b9c-f9a3-49ad-92ed-5464780874be/pointers/a1e20fa1-8ace-4fd8-8797-472113f6785a-e8e09898d91cc3a412ba36751fdba4ac931acb5c

Seems good to me! I didn't try changing data in the input file or code and running a renku update yet

rokroskar:
Awesome, thanks for making the project - This looks like a good starting point!

mohammad-sdsc:
@rokroskar I'd say the pointer file should not show up in the graph or lineage at all. It's just an implementation detail that users shouldn't care about. I'm not sure though if it is possible to exclude it.

emmjab:
Also just tested the `renku dataset update --external` and called `renku update`; also works as I would expect & generates a linage: https://dev.renku.ch/projects/e.jablonski/test_external/files/lineage/out.txt

rokroskar:
> I'd say the pointer file should not show up in the graph or lineage at all. It's just an implementation detail that users shouldn't care about. I'm not sure though if it is possible to exclude it.

I see - I suppose at the moment there isn't a way to exclude it. But if it's there, maybe a link between the two entities would make sense? 

Two other points that come to mind at the moment:

1. naming the external file nodes: should we have some stable naming convention? As it stands, I could use the same data from two different projects but this would not be obvious from the KG (apart from maybe the last part of the path being the same). I guess eventually they could be linked via their sha256 or similar. Have you thought about that?

2. Could there be a mechanism to easily update the link in case the location of the external data changes? 

Actually, I have a third :) 

3. Could this be extended to allow something like:

```
pwd
  <a renku project root>
ln -s ../../../data/path/file.csv
renku external add file.csv
```

So now in my renku project, I have a symbolic link `file.csv` pointing to the external data and the command `renku external add` registered this file and created the a file in `.renku/external` (?) that holds the metadata (checksum, etc.) Alternatively, the registration could be implicit, i.e. only when the link is used in a renku command. So I could create the symbolic link and commit it, but only when I did 

```
renku run --input file.csv script.py 
```

we would see that it's a symbolic link and create the requisite metadata. This is very similar to the dataset functionality but it allows me to do it without having to explicitly create a dataset.

ciyer:
This is very nice, and works pretty well! 

I tested a case where I have code in one repo and a file that is processed just in a regular directory. It is here. The process and results are documented here:

  https://dev.renku.ch/projects/renku-external-data-one-repo/renku-external-data-code/

I found the same thing that Emma found:

* The pointer file is shown as an upstream input to the [result](https://dev.renku.ch/projects/renku-external-data-one-repo/renku-external-data-code/files/lineage/data/out.txt), but it does not have [lineage itself](https://dev.renku.ch/projects/renku-external-data-one-repo/renku-external-data-code/files/lineage/data/some-data/data.txt), which is probably a bug

**Question**
Is there a reason the pointer files are stored in git-lfs? I would think they can/should be stored in plain git.

mohammad-sdsc:
Thanks all for your comments!
The  issue with lineage of external file is solved and the lineage is available now.

> 1. naming the external file nodes: should we have some stable naming convention? As it stands, I could use the same data from two different projects but this would not be obvious from the KG (apart from maybe the last part of the path being the same). I guess eventually they could be linked via their sha256 or similar. Have you thought about that?

As we discussed this will be implemented later.

> 2. Could there be a mechanism to easily update the link in case the location of the external data changes?

You can re-add the files with `--force` flag from the new location.

renku/core/commands/checks/external.py
```diff
@@ -0,0 +1,48 @@
...
+                target = Path(file_.path).resolve()
+                if not target.exists():
+                    missing.append((file_.path, str(target)))
+
+    if not missing:
+        return True, None
+
+    problems = (
+        '\n' + WARNING + 'There are missing external files.\n'
+        '  (make sure that external volume is mounted and accessible)' +
```
rokroskar: This works really nicely! Just a minor wording suggestion - in most cases it might not be a mounted volume so the user might wonder what that's all about. 

```suggestion
        '  (make sure that the external path is accessible)' +
```

Issue #932, opened 2020-01-23
feat: Improved handling of non-renku repositories


jsam:
Currently, if I execute any of the commands on the repository which is not initialized as renku repository (no .renku directory), I get the following:

```
Ahhhhhhhh! You have found a bug. 🐞

1. Open an issue by typing "open";
2. Print human-readable information by typing "print";
3. See the full traceback without submitting details (default: "ignore").

Please select an action by typing its name (open, print, ignore) [ignore]: 
```

Instead, of this message I would expect a more user friendly message, such as:

```
WARNING: This is not a renku initialized repository.

Would you like renku to manage this repository? [y/N]
```

ciyer:
This happens also happens on commands that do not require access to the repo. Try running `renku dataset add --help` outside a repo.

Commands that access the repo should prompt to initialize a repo as above. Help commands should should just print help.

ciyer:
Regarding the behavior in this situation, I think the behavior should be different in two situations:

## 1. In a git repository
### Option 1
```
WARNING: This is not a renku initialized repository.

Would you like renku to manage this repository? [y/N]
```

### Option 2
```
This is not a renku initialized repository. Run 'renku init .' to make it one.
```

## 2. Not in a git repository

```
fatal: not a renku repository.
```

rokroskar:
+1 for Option 2. 

Note, however, case 2. could also be covered by Option 2. Could it not?

ciyer:
Yes, you are right. Though in that case, the message should mention that it is not even a git repository. Reasoning: sometimes I type a renku command in a plain git repository, thinking I'm in a renku repository, but if I type a renku command in a regular directory, I'm usually not in the directory I think I'm in.

PR #989, opened 2020-02-04
fix: handle missing renku directory

jsam:
closes: #932


renku/cli/__init__.py
```diff
@@ -173,6 +186,18 @@ def print_global_config_path(ctx, param, value):
 @click.pass_context
 def cli(ctx, path, renku_home, use_external_storage):
     """Check common Renku commands used in various situations."""
+    renku_path = Path(path) / renku_home
+
+    if not renku_path.exists() and not is_allowed_command(ctx):
+        click.echo(
+            WARNING +
+            '`{0}` is not a renku initialized repository.'.format(path)
```
ciyer: With this wording, you should write `renku-initialized` and `renku-managed` below. But I think it's ok to just say `renku repository`, e.g., 

```WARNING + '`{0}` is not a renku repository.'.format(path)```

File: renku/cli/__init__.py [hunk 2]
```diff
@@ -173,6 +186,18 @@ def print_global_config_path(ctx, param, value):
...
+        )
+        raise UsageError((
```
ciyer: If you want to word it like this, you should write `renku-managed`, but I am ok with just calling it a renku repository
```'To initialize this as a renku repository, use: `renku init`'```

The use of warning and error looks a bit strange:

```
Warning: `...` is not a renku initialized repository.
Error: To initialize this as a renku managed repository use: `renku init`
```

Could this not be made to 

```
Error: `...` is not a renku repository.
To initialize this as a renku repository, use: `renku init`
```

File: tests/cli/test_errors.py [hunk 3]
```diff
@@ -0,0 +1,48 @@
...
+            assert expected_output == result.output
+
+
+def test_cli_initialization_no_err_help(runner):
+    """Check correct user warning within non-renku repository."""
+    from renku.core.utils.contexts import chdir
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        with chdir(tmpdir):
+            result = runner.invoke(cli, ['--help'])
```
mohammad-sdsc: Please consider to test other cases here like the safe commands (i.e. clone, init, help) and also when asking for a help on a subcommand (e.g. renku dataset -h/--help) by using pytest parametrize.

Issue #984, opened 2020-02-04
feat(svc): add endpoint to check jobs requested by user


jsam:
Add the following endpoint:

```
GET /jobs
```

Which should return the list of all requested jobs for a user. Same auth-based user identification as on other endpoints should be used.

Example response:

```
{
  "result": {
      "jobs": [
         { 
           "id": "...", 
           "state": [ENQUEUED|IN PROGRESS|COMPLETED], 
           "created_at": "", 
           "updated_at": "", 
           "extras": { [ job related metadata ] } }
       ]
   }
}
```

PR #1001, opened 2020-02-11
feat: added list user jobs endpoint

jsam:
closes: #984


jsam:
Yes, I'll try to make a small PoC and bring the discussion to the table. Thank you.

PR #1002, opened 2020-02-11
fix: flush old keys for user projects and files

jsam:


renku/service/jobs/cleanup.py
```diff
@@ -39,7 +39,11 @@ def cache_files_cleanup():
         user = {'user_id': user}
         for file_meta in cache.hash_table(file_key).values():
             data = json.loads(file_meta)
-            old = ((time.time() * 1e+3) - data['timestamp']) / 1e+3
+
+            # If record does not contain timestamp,
+            # it means its an old record, and we should mark it for deletion.
+            created_at = data.get('timestamp', (time.time() - ttl) * 1e+3)
```
rokroskar: I guess this will work but it seems a bit ad-hoc. If you want to expire the cache, shouldn't this be done at startup/deploy time?

Issue #922, opened 2020-01-20
Add migration mechanism


Panaetius:
- Throw exception if not migrated. Prevent execution unless migrated
- new migration mechanism. Incremental numbers, no backwards migrations (git reset)
- Save current migration in project metadata
- Automated revert if migration breaks, alert user to downgrade renku version

rokroskar:
Should step 0 be to compile a list of needed migrations to-date?

Panaetius:
See https://github.com/swissdatasciencecenter/renku-python/issues/920

Issue #920, opened 2020-01-20
Create Initial Migration


Panaetius:
Apply all on the fly/adhoc migrations to disk to get an initial state for migrations as first migration.

mohammad-sdsc:
Consider to convert all submodule-based dataset imports to non-submodule: https://github.com/SwissDataScienceCenter/renku-python/issues/681.

PR #1003, opened 2020-02-12
feat: new migration mechanism

mohammad-sdsc:
# Description

An extensible migration mechanism is provided. Renku projects have a version number which is checked against the Renku installation version. If project version is older, user are prompted to migrate their project to the latest version.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/922
Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/920


renku/core/management/migrate.py
```diff
@@ -0,0 +1,104 @@
...
+    if not _is_renku_project(client):
+        return
+
+    project_version = _get_project_version(client)
+    migration_executed = False
+
+    for version, path in get_migrations():
+        if version > project_version:
+            module = importlib.import_module(path)
+            for name in module.__all__:
```
Panaetius: I think it would be cleaner to define a single entrypoint for each migration file that gets executed (e.g. `def migrate(client):`) instead of relying on `__all__`. With `__all__` you still need to manually add all methods yourself, so it's not less work than putting them inside a function, so I'd go with the simpler interface.

renku/core/management/migrations/m_0003__1_jsonld.py
```diff
@@ -15,13 +15,50 @@
...
+        ],
+    }
+
+    old_metadata_paths = _dataset_pre_0_3(client)
+    new_metadata_paths = (client.path / client.renku_datasets_path).rglob(
+        client.METADATA
+    )
+
+    for file_ in itertools.chain(old_metadata_paths, new_metadata_paths):
+        dataset = Dataset.from_yaml(
```
Panaetius: I'm not sure we can rely on `Dataset.from_yaml` here, because the object might have change from the state of the dataset on disk. 

I think the safest way is to load the yaml to dict, then pass the dict to the migrations, and save the dict to yaml again. Basically what `from_jsonld` does internally, but without having to go through the `Dataset` class.

Migrations, especially the initial one, should depend on the yaml itself as much as possible, instead of on the classes of code, so we can change those classes without having to worry that it might break migrations (otherwise we end up with a kind of chicken and egg problem). 

Let me know if I can help, translating all our current migrations to a form that doesn't depend on Dataset code is probably quite an annoying chore.

renku/core/management/migrations/m_0003__1_jsonld.py
```diff
@@ -15,13 +15,50 @@
...
 
-def migrate_dataset_schema(data):
+__all__ = [
+    'migrate_datasets_metadata',
+]
+
+
+def migrate_datasets_metadata(client):
+    """Apply all initial JSON-LD migrations to datasets."""
+    jsonld_migrations = {
```
Panaetius: The initial migration is missing the `translate` on-the-fly migrations we have at the moment, see https://github.com/SwissDataScienceCenter/renku-python/blob/master/renku/core/models/projects.py#L42 for an example.

renku/core/commands/migrate.py
```diff
@@ -15,23 +15,24 @@
...
+    if is_migration_required(client):
+        raise MigrationRequired
+    elif is_project_unsupported(client):
+        raise ProjectNotSupported
 
-    if all(results) and client.repo.index.diff(None):
-        return results
+
+@pass_local_client(clean=True, commit=True, commit_empty=False)
+def migrate_project(client, commit_message=None):
```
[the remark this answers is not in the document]
mohammad-sdsc: I believe this is required by the `service`. Some other commands have it (e.g. dataset, config)

renku/core/management/datasets.py
```diff
@@ -623,6 +624,9 @@ def _fetch_lfs_files(self, repo_path, paths):
     def _fetch_files_metadata(client, paths):
         """Return metadata for files from paths."""
         files = {}
+
+        migrate(client)
```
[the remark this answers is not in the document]
mohammad-sdsc: Good point! However, we should not block users from adding data from a newer version. I'll add this check and return immediately instead of failing. Preferably, we should print a warning message but that requires lots of changes.

mohammad-sdsc:
Thanks @Panaetius for the review! I've made bunch of changes and rebased on `master`. You can review commit-by-commit if you wish. Everything starting from `c6d4400` commit is new.

Panaetius:
I tried it out locally, comparing new and old output, and it looks really good. The output differs in a qualified generation being added for the `metadata.yaml` due to the migration commit, but I think that's a feature, i.e. the output of renku log is not identical, but having the information that the dataset was changed is good.

Really good work, I know it wasn't easy!

One suggestion would be to slightly improve UX. Right now when running renku migrate, at the end it only prints `OK`. 

Something along the lines of:
```
Applying migration m_0003__1_jsonld ...
Applying migration m_0003__2_initial.py...
Successfully applied 2 migrations, project is up to date
```
would look very nice. If this can be done without too much effort?

Issue #1011, opened 2020-02-13
bug: dataset commands fail in subdirectories


rokroskar:
Running `renku dataset` commands fails if the user is not in the repository root.

PR #1012, opened 2020-02-14
fix: enable dataset cmd in sub directories

jsam:
closes: #1011


tests/cli/test_datasets.py
```diff
@@ -1481,3 +1482,20 @@ def test_pull_data_from_lfs(runner, client, tmpdir):
 
     result = runner.invoke(cli, ['storage', 'pull', str(relative_path)])
     assert 0 == result.exit_code
+
+
+def test_dataset_cmd_subdirectory(runner, project):
```
rokroskar: Thanks!

PR #1021, opened 2020-02-17
docs(svc): doc strings typos fixed

jsam:


renku/service/serializers/datasets.py
```diff
@@ -122,7 +122,7 @@ class DatasetListRequest(Schema):
 
 
 class DatasetDetails(Schema):
-    """Serialize dataset to response object."""
+    """Serialize a dataset to response object."""
```
rokroskar:
```suggestion
    """Serialize a dataset to a response object."""
```

renku/service/serializers/datasets.py
```diff
@@ -150,7 +150,7 @@ class DatasetFilesListRequest(Schema):
 
 
 class DatasetFileDetails(Schema):
-    """Serialzie dataset files to response object."""
+    """Serialize dataset files to response object."""
```
rokroskar:
```suggestion
    """Serialize dataset files to a response object."""
```

renku/service/views/datasets.py
```diff
@@ -129,7 +129,7 @@ def list_dataset_files_view(user, cache):
 @use_kwargs(DatasetAddRequest)
 @marshal_with(DatasetAddResponseRPC)
 @header_doc(
-    'Add uploaded file to cloned repository.', tags=(DATASET_BLUEPRINT_TAG, )
+    'Add an uploaded file to cloned repository.', tags=(DATASET_BLUEPRINT_TAG, )
```
rokroskar:
```suggestion
    'Add an uploaded file to a cloned repository.', tags=(DATASET_BLUEPRINT_TAG, )
```

renku/service/serializers/datasets.py
jsam: My linter says both forms are correct. :thinking:

rokroskar: run the grammar check in word to break the tie :)

jsam: I don't have to. I trust you and your years of experience of being grammar police! I applied the requested changes!

Issue #1025, opened 2020-02-18
Error in exporting an imported Zenodo dataset to Dataverse


mohammad-sdsc:
**Describe the bug**
A dataset which was previously imported from Zenodo cannot be exported to Dataverse.

**To Reproduce**
1. Import a dataset
```
$ renku dataset import https://zenodo.org/record/3606669 --short-name my-data
CHECKSUM                          NAME                SIZE (MB)  TYPE
--------------------------------  ----------------  -----------  ------
9eb4075f9abc17c5f76a2babdb475a1b  pyndl-v0.7.0.zip         0.50  zip
Do you wish to download this version? [y/N]: y
OK
```

2. Export dataset and select Tag [1] when prompted

```
$ renku dataset export my-data dataverse --dataverse-name SDSC-Test
Tag to export: 

<HEAD>  [1]
v0.7.0  [2]

Tag [1]: 1
Usage: renku dataset export [OPTIONS] SHORT_NAME PROVIDER

Error: Invalid value: Invalid control character at: line 29 column 51 (char 3028)
```

3. Export dataset and select Tag [2] when prompted

```
$ renku dataset export my-data dataverse --dataverse-name SDSC-Test
Tag to export: 

<HEAD>  [1]
v0.7.0  [2]

Tag [1]: 2
Usage: renku dataset export [OPTIONS] SHORT_NAME PROVIDER

Error: Invalid value: Dataset is not found.
```

PR #1028, opened 2020-02-19
fix: Dataverse export

mohammad-sdsc:
# Description
Fixes some issues with exporting datasets to Dataverse

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/1025


mohammad-sdsc:
Thanks for the review! 

As discussed, `HEAD` means the current version of the dataset (i.e. what is at the `HEAD` of the git). Since this is not exclusive to Dataverse and works the same for exporting to all providers, it does not belong to this PR (and will be addressed in other PRs if we decided to fix it).

PR #1030, opened 2020-02-19
build: use generic action for component update

### LINKED ISSUE #1029: use generic action to update renku-core version in main chart
To avoid duplication of common code, we should use the generic action from the main renku repo for updating the renku-core dependency.

rokroskar:
Replaces several steps of the publish action with a generic Docker action
from the main renku repo. 

Successful run here: https://github.com/SwissDataScienceCenter/renku-python/runs/456554251?check_suite_focus=true

Fixes #1029

### DIFF
--- a/.github/workflows/test_deploy.yml
+++ b/.github/workflows/test_deploy.yml
@@ -309,12 +309,14 @@ jobs:
     - uses: actions/checkout@v2
       with:
         fetch-depth: 0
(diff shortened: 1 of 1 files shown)


rokroskar:
This needs the corresponding PR in the renku repo.

rokroskar:
The PR in the main Renku repo has been merged so this should be safe to merge now.

PR #1033, opened 2020-02-20
chore: helm update for datasets workers

jsam:
Since dataset jobs should be independently be able to scale out, this PR adds additional worker container only for dataset jobs queue.


rokroskar:
Is there a test-deployment with this setup?

jsam:
No, I haven't tested it fully yet.

helm-chart/renku-core/values.yaml
```diff
@@ -15,9 +15,9 @@ cleanupProjectsTTL: 1800
 logLevel: WARNING
 
 image:
-  repository: renku/renku-core
+  repository: psam/renku-core
   tag: 'latest'
-  pullPolicy: IfNotPresent
+  pullPolicy: Always
```
rokroskar:
```suggestion
  pullPolicy: IfNotPresent
```

helm-chart/renku-core/values.yaml
```diff
@@ -15,9 +15,9 @@ cleanupProjectsTTL: 1800
 logLevel: WARNING
 
 image:
-  repository: renku/renku-core
+  repository: psam/renku-core
```
rokroskar:
```suggestion
  repository: renku/renku-core
```

helm-chart/renku-core/templates/deployment.yaml
```diff
@@ -40,18 +46,52 @@ spec:
...
             - name: RQ_WORKER_LOG_LEVEL
               value: {{ .Values.logLevel }}
+          volumeMounts:
+            - name: shared-volume
+              mountPath: {{ .Values.cacheDirectory }}
+
+        - name: {{ .Chart.Name }}-management-workers
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          command: ["python"]
```
rokroskar: you want to run this without gunicorn?

jsam: It's a worker, no need for it. We should think of daemonizing it internally, but for now since we are running it on k8s, it should be fine.

helm-chart/renku-core/values.yaml
jsam: I know - ty! I'm waiting for 1052 and 1048 to be merge so I can test this PR one last time before merging it.

helm-chart/renku-core/values.yaml
jsam: I noticed there is an issue with caching images in k8s. I'm not sure what exactly was going on, but workers image received super old image without new code. I agree we should switch back, but we should also inspect the underlying issue. Which command do you use to provision deployment? Do you set some additional flag to force redownloading of new image?

rokroskar: no if I'm pushing the same image while developing I do it like this - but otherwise you should not use `latest` to avoid this altogether. Some implementations always pull images tagged with `latest` even if `IfNotPresent` is set, but others don't - I'm not sure if this is a k8s versioning issue or a rancher issue or maybe both.

rokroskar: But, if you want your changes reflected in realtime on a live deployment, the way to go is to use telepresence - I think there are some docs in the UI repo about how they set it up, but @lorenzo-cavazzi and @vfried do this all the time

jsam: I see. I think our issue is a bit different. So basically what happened is that I deployed `latest` which was this `docker pull renku/renku-core:0.1.0-763a51d`, and renku-core service container was deployed with correct image tag, but other containers in same deployment had container without `worker.py` or `scheduler.py` implying that the image version there was prior we introduce cleanups, and the error message I was receiving all 3 types of containers was `stat: not found renku/service/worker.py`, `stat: not found renku/service/scheduler.py`. Forcing the deployment to always pull the image fixed this issue, so (to me) smells like it could be some caching problem.

rokroskar: I would put my money on there being an error in the tags used rather than a layer caching problem with docker running the containers - or maybe you hit a name collision with the 7-digit hash?? Unlikely but possible I suppose.  

Also `renku/renku-core:0.1.0-763a51d` is not `renku/renku-core:latest` so I'm not sure I understand?

jsam: At the moment of testing (at friday), `latest` and `0.1.0-763a51d` were pointing to the same image, and I haven't touch any deployment configuration (in regards to image config) - so I'm not sure that it was a mistake on my part. Let's see if I can reproduce the image after we merge remaining PR's.

PR #1044, opened 2020-02-21
build: pin gitdb version

rokroskar:
Pins `gitdb` to avoid incompatibility with gitpython.


jachro:
✔️ Checked on renku-kuba

PR #1045, opened 2020-02-21
docs: added github templates for proposal and improvement requests

jsam:
image

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -4,6 +4,8 @@ about: Create a report to help us improve
 
 ---
 
+**Note: for support questions, please use our discourse (https://renku.discourse.group/)**.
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -4,6 +4,8 @@ about: Create a report to help us improve
 
 ---
 
+**Note: for support questions, please use our discourse (https://renku.discourse.group/)**.
```
rokroskar: Could we use html comments for these headings? Otherwise they often end up in the issue text. Something like 

```
<!-- Note: for support questions, please use our discourse (https://renku.discourse.group/) -->

<!--  Describe the bug --> 
```

jsam: Done. Anything else?

PR #1048, opened 2020-02-21
build(deps): attrs version bump

jsam:


rokroskar:
I don't get it - attrs wasn't actually changed in `Pipfile.lock`

rokroskar:
Oh I see -- the requirement was `>=`

Issue #1047, opened 2020-02-21
feat: add service filter for specific jobs


jsam:
Currently jobs endpoint will return all user jobs. Would be nice to add a filter where we specify only `job_id` and receive updates only for this specific job.

PR #1050, opened 2020-02-21
feat(svc): adds job details endpoint

jsam:
closes: #1047

### DIFF
--- a/tests/service/views/test_utils.py
+++ b/tests/service/views/test_utils.py
@@ -43,3 +43,30 @@ def test_result_response(svc_client):
     assert {'result'} == set(response.keys())
     assert {'datasets'} == set(response['result'].keys())
     assert ctx == response['result']
+
+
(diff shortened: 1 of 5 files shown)


tests/service/views/test_utils.py
```diff
@@ -43,3 +43,30 @@ def test_result_response(svc_client):
...
     assert ctx == response['result']
+
+
+def test_result_response_with_none(svc_client):
+    """Test result response with None value."""
+    response = result_response(DatasetListResponseRPC(), None).json
+
+    assert response
+    assert {'result'} == set(response.keys())
+    assert not response['result']
```
mohammad-sdsc: Should we say `response['result'] is None`?

jsam: Done. We should probably set this as a standard on the repository. I agree 100% with it!

PR #1051, opened 2020-02-23
fix: cache instance cleanup

jsam:
Share same cache instance on service and workers.

### DIFF
--- a/renku/service/cache/__init__.py
+++ b/renku/service/cache/__init__.py
@@ -35,8 +35,7 @@ def make_cache():
     sub_dirs = [CACHE_UPLOADS_PATH, CACHE_PROJECTS_PATH]
 
     for subdir in sub_dirs:
-        if not subdir.exists():
-            subdir.mkdir()
+        subdir.mkdir(parents=True, exist_ok=True)
 
     return ServiceCache()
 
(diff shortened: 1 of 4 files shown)


renku/service/jobs/datasets.py
```diff
@@ -75,6 +75,7 @@ def dataset_import(
     dataset_uri,
     short_name=None,
     extract=False,
+    timeout=None,
```
mohammad-sdsc: This is not used anywhere. Is there a reason for that?

renku/service/cache/__init__.py
```diff
@@ -27,3 +28,19 @@ class ServiceCache(
...
     pass
+
+
+def make_cache():
+    """Create cache structure."""
+    sub_dirs = [CACHE_UPLOADS_PATH, CACHE_PROJECTS_PATH]
+
+    for subdir in sub_dirs:
+        if not subdir.exists():
+            subdir.mkdir()
```
mohammad-sdsc: I'd rather write `subdir.mkdir(parents=True, exist_ok=True)` and remove `if not subdir.exists():`.

renku/service/jobs/datasets.py
jsam: When you set on a RQ queue a timeout, this parameter gets injected into worker jobs and I would not like to handle this with kwargs since it's not self documenting. So I think this explicit approach is slightly better than kwargs.

renku/service/cache/__init__.py
jsam: Done. Ty!

PR #1058, opened 2020-02-24
fix: removes recreate strategy

jsam:
It seems that this bug has to be addressed in another way (https://github.com/rq/rq-scheduler/issues/167).


rokroskar:
I'd say lets leave it as is for now and close this - the last update on dev went through fine so it was probably only causing problems when the policy changed from one deploy to another. I know this doesn't solve the problem of potentially needing multiple schedulers but for now we don't have that need.

Issue #1035, opened 2020-02-20
fix: move IT to use staging


jsam:
Move tests to use staging environments of Zenodo and Dataverse!

mohammad-sdsc:
After talking to Dataverse people:
1. It's OK to create datasets and delete them before publishing
2. It's OK to have a few hundred published datasets
3. Published datasets cannot be deleted (it required superuser permission which are not given to us)

PR #1061, opened 2020-02-25
fix: Dataverse integration tests

mohammad-sdsc:
# Description

Deletes unpublished Dataverse datasets when integration tests are done. For testing published datasets, it creates a GH action that runs this tests on merges to master when one of the files in `renku/core/commands/providers` change.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/1035

### DIFF
--- a/.github/workflows/test_deploy.yml
+++ b/.github/workflows/test_deploy.yml
@@ -117,7 +117,9 @@ jobs:
         git config --global --add user.email "renku@datascience.ch"
     - name: Test with pytest
       env:
+        DATAVERSE_ACCESS_TOKEN: ${{ secrets.DATAVERSE_ACCESS_TOKEN }}
         IT_OAUTH_GIT_TOKEN: ${{ secrets.IT_OAUTH_GIT_TOKEN }}
+        ZENODO_ACCESS_TOKEN: ${{ secrets.ZENODO_ACCESS_TOKEN }}
       run: pytest -m integration -v
 
   test-macos-integration:
@@ -147,7 +149,9 @@ jobs:
         git config --global --add user.email "renku@datascience.ch"
     - name: Test with pytest
       env:
+        DATAVERSE_ACCESS_TOKEN: ${{ secrets.DATAVERSE_ACCESS_TOKEN }}
         IT_OAUTH_GIT_TOKEN: ${{ secrets.IT_OAUTH_GIT_TOKEN }}
+        ZENODO_ACCESS_TOKEN: ${{ secrets.ZENODO_ACCESS_TOKEN }}
       run: pytest -m integration -v
 
   # test-macos-integration-pr:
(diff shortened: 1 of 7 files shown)


.github/workflows/test_publish.yml
```diff
@@ -0,0 +1,70 @@
+name: Integration Test when Publishing to External Data Warehouses
+
+on:
+  pull_request:
+    types: 
+    - closed
```
rokroskar: why run on closed pull requests? This will run the job twice if the PR is merged to master.

.github/workflows/test_publish.yml
```diff
@@ -0,0 +1,70 @@
...
+    paths:
+    - renku/core/commands/providers/*.py
+
+jobs:
+  test-linux-integration:
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 4
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
```
rokroskar: Since this is mainly checking our API assumptions, do we need this matrix?

.github/workflows/test_publish.yml
mohammad-sdsc: There is also a condition for PR being on `master`. So, this runs when a PR on master is closed (which means the PR is merged to `master`). I don't get why it runs twice.

.github/workflows/test_publish.yml
mohammad-sdsc: To be clear, it will be `python-version: [3.6]`.

.github/workflows/test_publish.yml
rokroskar: oh I see, sorry I confused it with being run on push to master (not PR to master). I think we tend to use `push` instead of `pull_request` in some of our other configs, so something like this:

```yaml
on:
  push:
    branches: [master]
    paths: [renku/core/commands/providers/*.py]
```

mohammad-sdsc: Using `push` is simpler actually 🤦‍♂

setup.cfg
```diff
@@ -28,7 +28,7 @@ all_files = 1
 universal = 1
 
 [pydocstyle]
-add_ignore = D401
+add_ignore = D202,D401
```
jsam: This implies we need cleanup of `#noqa` in some places?

.github/workflows/test_deploy.yml
```diff
@@ -147,7 +149,9 @@ jobs:
         git config --global --add user.email "renku@datascience.ch"
     - name: Test with pytest
       env:
+        DATAVERSE_ACCESS_TOKEN: ${{ secrets.DATAVERSE_ACCESS_TOKEN }}
         IT_OAUTH_GIT_TOKEN: ${{ secrets.IT_OAUTH_GIT_TOKEN }}
+        ZENODO_ACCESS_TOKEN: ${{ secrets.ZENODO_ACCESS_TOKEN }}
       run: pytest -m integration -v
 
   # test-macos-integration-pr:
```
jsam: What's the reason to keep this commented out? Should we just delete it? Imho, It's part of git history anyway so no need to keep dead code around.

mohammad-sdsc: Agreed! Please approve!

setup.cfg
mohammad-sdsc: Good point! I missed one actually. Please re-approve.

PR #1062, opened 2020-02-25
chore: wait for chart before helm update

rokroskar:
Once the chart is pushed to the repository it takes a few minutes before it's available to `helm update` - a small delay between steps can avoid potential issues with chart retrieval.


.github/workflows/test_deploy.yml
```diff
@@ -320,6 +320,8 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.RENKU_CI_TOKEN }}
         DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
         DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+    - name: Wait for chart to be available
+      run: sleep 120
```
jsam: Should we make it a env variable in case number will need a change?

rokroskar: You'll just need to change the value of the env variable in this file anyway so I'm not sure it simplifies much?

jsam: It eliminates the need to solve operational problems with additional commits.

rokroskar: Yeah it's a good point though it seems like a secret for this is a bit heavy? Not sure. We had it set to two minutes in the Travis builds and I'm not sure it was ever a problem.

rokroskar: Of course the second I merge this it's bound to fail :)

PR #1065, opened 2020-02-25
build: fix render problem and add test

rokroskar:
An incorrect heading level caused PyPi to reject the distro - fix the heading and add a test using `twine` to prevent this in the future.


CHANGES.rst
```diff
@@ -19,7 +19,7 @@ Changes
 =======
 
 `0.9.1 <https://github.com/SwissDataScienceCenter/renku-python/compare/v0.9.0...v0.9.1>`__ (2020-02-24)
-=======================================================================================================
+-------------------------------------------------------------------------------------------------------
```
jsam: TY.

Issue #1027, opened 2020-02-18
feat(svc): add models for service resources


jsam:
Service has no concept of model which can be saved to cache. For it currently we are using plain dicts and this is quite limiting. We would need a support for better general life cycle management of such resources (projects, files, jobs).

For example, internal ID's could be generated within a model, state mutation should happen on the object itself, object cleanup and error management should be self contained.

I've found [walrus](https://walrus.readthedocs.io/en/latest/models.html#creating-updating-and-deleting) library to be potentially quite a good choice for it. I propose we add it to our resource management to have things in more cleaner and testable way.

rokroskar:
It can even function as a triple store! https://walrus.readthedocs.io/en/latest/graph.html 👍 

Walrus seems like it would be a reasonable choice for this - any cons?

PR #1087, opened 2020-03-04
chore(service): adds models for service resources

jsam:
closes: #1027


renku/service/jobs/cleanup.py
```diff
@@ -16,37 +16,30 @@
...
-        user = {'user_id': user}
-        for file_meta in cache.hash_table(file_key).values():
-            data = json.loads(file_meta)
-
+    for user, files in cache.user_files():
+        for file in files:
             # If record does not contain timestamp,
             # it means its an old record, and we should mark it for deletion.
-            created_at = data.get('timestamp', (time.time() - ttl) * 1e+3)
+            created_at = file.timestamp
```
Panaetius: `Job` has `created_at` and `updated_at` but `File` uses `timestamp` instead of `created_at`, we should keep naming consistent. `Project` also uses `timestamp`.

renku/service/jobs/datasets.py
```diff
@@ -22,59 +22,47 @@
...
-        self.user = user
         self.job = job
 
         super().__init__(None, None)
 
     def __call__(self, description, total_size):
         """Job progress call."""
-        self.job['state'] = USER_JOB_STATE_IN_PROGRESS
-        self.job['extras'] = {
+        self.job.state = USER_JOB_STATE_IN_PROGRESS
```
Panaetius: This could be put in a method on job, to not expose the internal of the job state to the outside world. In line with `Job.fail_job()`

renku/service/jobs/datasets.py
```diff
@@ -22,59 +22,47 @@
...
-        self.job['extras']['progress_size'] = size
-        self.cache.set_job(self.user, self.job)
+        self.job.extras['progress_size'] = size
+        self.job.save()
 
     def finalize(self):
         """Finalize job tracking."""
-        self.job['state'] = USER_JOB_STATE_COMPLETED
-        self.cache.set_job(self.user, self.job)
+        self.job.state = USER_JOB_STATE_COMPLETED
```
Panaetius: Same as above

PR #1098, opened 2020-03-11
chore: remove brew

### LINKED ISSUE #832: ci: issues with publishing to brew
There seem to be some problems with publishing to brew recently. See [here](https://travis-ci.org/SwissDataScienceCenter/renku-python/jobs/615276636?utm_medium=notification&utm_source=github_status) and [here](https://travis-ci.org/SwissDataScienceCenter/renku-python/jobs/615275432?utm_medium=notification&utm_source=github_status).

rokroskar:
* removes brew from build and deploy process as well as the instructions. 

* adds CLI docker image build to the deploy steps and fixes the broken docker build

closes #832


README.rst
```diff
@@ -56,63 +56,63 @@ analysis tasks.
...
 
-Use following installation steps based on your operating system and preferences
-if you would like to work with the command line interface and you do not need
-the Python library to be importable.
 
-Homebrew
---------
+``pipx`` installs renku into its own virtual environment, making sure that it
+does not pollute any other packages or versions that you may have already
+installed.
```
jsam: I really like this paragraph and that's excatly what I meant by "we should improve the docs first". 😄 

Maybe we should add a small `NOTE:` section here which would describe that if user has installation in project virtual environment that that installation will be different from the pipx one and that upgrading in this case needs to be in accordance to the way renku is being used?

rokroskar: I tried to add this using the `.. note:` directive but I'm not sure it's picked up by github...

README.rst
```diff
@@ -56,63 +56,70 @@ analysis tasks.
...
 
-The recommended way of installing Renku on MacOS and Linux is via
-`Homebrew <brew.sh>`_.
+.. note::
 
-::
+    If you install renku as a dependency in a virtual environment and the
```
jsam: Perfect! Thank you. :heart:

Issue #1074, opened 2020-03-02
feat: enable editing of metadata


rokroskar:
Renku users would like to edit dataset metadata after they are created. This needs to be done through the CLI as well as in the service so it can be used by the UI.

rokroskar:
Related to https://github.com/SwissDataScienceCenter/renku-python/issues/1075

PR #1111, opened 2020-03-17
feat(dataset): metadata edit

mohammad-sdsc:
# Description

Users can edit datasets' metadata by using command line to pass new values for title, description, and creators by using `--title`, `--description`, and `--creator` flags.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/1074


renku/cli/dataset.py
```diff
@@ -464,6 +464,10 @@ def edit(short_name, title, description, creator):
         click.echo('Nothing to update.')
     else:
         click.echo('Successfully updated: {}.'.format(', '.join(updated)))
+        if no_email_warnings:
+            click.echo(
+                WARNING + 'No email set for: ' + ', '.join(no_email_warnings)
```
rokroskar: maybe it would help here to print the expected syntax on the next line? But perhaps that's too verbose, not sure.

mohammad-sdsc: I think also it would too verbose; besides, we already print the format in `renku dataset edit -h`.

rokroskar: Maybe we can reword slightly then: "No email or wrong format for: "

Issue #1110, opened 2020-03-17
unexpected keyword argument 'access_conditions' when importing zenodo dataset in Jupyterlab terminal


jen-thomas:
**Describe the bug**
Unexpected keyword argument 'access_conditions' when importing zenodo dataset in Jupyterlab terminal

**Link to project**
https://renkulab.io/projects/ACE-ASAID/seawater_stable_oxygen_isotopes/environments
This is private but I can add a user if need be.

**To Reproduce**
- Start new environment in renkulab
- Open terminal in jupyterlab
- `renku dataset import 10.5281/zenodo.1494915`

**Expected behavior**

**Screenshots and/or execution output**
Please select an action by typing its name (open, print, ignore) [ignore]: ignore
Traceback (most recent call last):
  File "/home/jovyan/.local/bin/renku", line 8, in <module>
    sys.exit(cli())
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/renku/cli/exception_handler.py", line 128, in main
    self._handle_github()
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/renku/cli/exception_handler.py", line 175, in _handle_github
    getattr(self, '_process_' + value)()
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/renku/cli/exception_handler.py", line 119, in main
    result = super().main(*args, **kwargs)
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/renku/cli/exception_handler.py", line 90, in main
    return super().main(*args, **kwargs)
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
[...]

PR #1112, opened 2020-03-17
fix(datasets): raises correct error message on import of protected dataset

jsam:
closes: #1110

This issue leads to another feature which we currently don't support, but probably should. Jen in that ticket wanted to import a dataset which contains protected files. Now this is currently not supported within renku on any of the providers. We should probably think about building that soon since we have people requesting this.

I created a followup ticket for this: #1113


renku/core/commands/providers/zenodo.py
```diff
@@ -236,7 +236,7 @@ def get_jsonld(self):
 
     def get_files(self):
         """Get Zenodo files metadata as ``ZenodoFile``."""
-        if len(self.files) == 0:
+        if not self.files:
             raise LookupError('no files have been found')
```
mohammad-sdsc: It's better to mention that there might be protected/restricted files: `'no files have been found or there are files with restricted access'`

jsam: Sure. Maybe the following wording would work better?

`no files have been found - deposit empty or protected'

mohammad-sdsc: `no files have been found - deposit is empty or protected` sounds better to me but up to you.

Issue #1118, opened 2020-03-19
Too many requests error for dataset import from Zenodo


jen-thomas:
**Describe the bug**
HTTPError: 429 Client Error: TOO MANY REQUESTS for import of files from Zenodo (see error below)

**Link to project**
https://renkulab.io/projects/ACE-ASAID/cruise-track-legs0-4

**To Reproduce**
renku dataset import 10.5281/zenodo.3369130
(select y to download set of files)

**Expected behavior**
Files should be imported from Zenodo.

**Screenshots and/or execution output**
```
Traceback (most recent call last):
  File "[...]/renku/cli/exception_handler.py", line 119, in main
    result = super().main(*args, **kwargs)
  File "[...]/renku/cli/exception_handler.py", line 90, in main
    return super().main(*args, **kwargs)
  File "[...]/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "[...]/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "[...]/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "[...]/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "[...]/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "[...]/renku/cli/dataset.py", line 659, in import_
    progress=_DownloadProgressbar
  File "[...]/renku/core/commands/client.py", line 103, in new_func
    result = ctx.invoke(method, client, *args, **kwargs)
  File "[...]/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "[...]/renku/core/commands/dataset.py", line 584, in import_dataset
    total_size=total_size,
  File "[...]/renku/core/commands/dataset.py", line 240, in add_to_dataset
    progress=progress,
  File "[...]/renku/core/management/datasets.py", line 240, in add_data_to_dataset
    progress=progress
  File "[...]/renku/core/management/datasets.py", line 425, in _add_from_urls
    files.extend(future.result())
  File "[...]/concurrent/futures/_base.py", line 428, in result
    return self.__get_result()
  File "[...]/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
  File "[...]/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
[...]

Issue #1082, opened 2020-03-03
fix(cli): improve download progress bar


jsam:
Progress bar which is used on datasets import is known to be causing issues and unexpected behaviour on certain terminals. We should look into improving the UX around this. Some of the improvements which we could implement are:

  * don't hide progress bar of completed files (instead sequntially build and display them as we progress)
  * lower update refresh rate

PR #1119, opened 2020-03-19
fix(datasets): fixed importing large amount of small files

jsam:
closes: #1118

Even though this fixes the reported issue in an acceptable manner, UX experience remains unaddressed. Ticket for the UX part is still here #1082 and in this case its very clear that we should put more attention to the UX of this progress bar. On my machine progress bar disappears so fast that for a majority of the time I have a perception that the process is hanging even though it's working hard to import the data.

### DIFF
--- a/renku/core/management/datasets.py
+++ b/renku/core/management/datasets.py
@@ -23,6 +23,7 @@ import re
 import shutil
 import stat
 import tempfile
+import time
 import uuid
 import warnings
 from configparser import NoSectionError
@@ -362,7 +363,7 @@ class DatasetsApiMixin(object):
         return checksum, filesize
 
     def _check_protected_path(self, path):
-        """Checks if a path is protected by renku."""
+        """Checks if a path is a protected path."""
         try:
             path_in_repo = path.relative_to(self.path)
         except ValueError:
@@ -376,7 +377,7 @@ class DatasetsApiMixin(object):
         return False
 
     def _add_from_local(self, dataset, path, link, external, destination):
-        """Add a file or directory from local filesystem."""
+        """Add a file or directory from a local filesystem."""
         src = Path(path).resolve()
 
         if not src.exists():
@@ -453,7 +454,7 @@ class DatasetsApiMixin(object):
         destination.mkdir(parents=True, exist_ok=True)
 
         files = []
-        max_workers = min(os.cpu_count() + 4, 8)
+        max_workers = min(os.cpu_count() - 2, 4) or 1
         with concurrent.futures.ThreadPoolExecutor(max_workers) as executor:
             futures = {
                 executor.submit(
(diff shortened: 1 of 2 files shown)


renku/core/management/datasets.py
```diff
@@ -453,7 +454,7 @@ def _add_from_urls(
         destination.mkdir(parents=True, exist_ok=True)
 
         files = []
-        max_workers = min(os.cpu_count() + 4, 8)
+        max_workers = min(os.cpu_count() - 2, 4) or 1
```
mohammad-sdsc: Why do you decrease the number of workers? These are values recommended by Python and worked fine so far. Also, `os.cpu_count() - 2` can be negative when there is one core.

renku/core/management/datasets.py
```diff
@@ -469,21 +470,31 @@ def _add_from_urls(
...
+            start = time.time() * 1e+3
             paths = _download(
                 url=url,
                 download_to=destination,
                 extract=extract,
                 progress_class=progress
             )
-        except error.HTTPError as e:  # pragma nocover
+
+            exec_time = (time.time() * 1e+3 - start) // 1e+3
```
mohammad-sdsc: I am not sure these multiplication and division by 1000 is needed. We can use the output of `time.time()` and then check if `exec_time < 1` to have the same effect and it's easier to understand the logic.

renku/core/management/datasets.py
```diff
@@ -469,21 +470,31 @@ def _add_from_urls(
...
+            # If execution time was less or equal to zero seconds,
+            # block the thread a bit to avoid being rate limited.
+            if exec_time == 0:
+                time.sleep(max(os.cpu_count() - 2, 4))
```
mohammad-sdsc: This wait at least 4 seconds between consequent downloads which is too much. Since we have at most 8 threads, isn't it enough to just wait 1 second max (or 1 - exec_time if above comment is applied)?

renku/core/management/datasets.py
jsam: Where is this recommendation coming from? I reduce it, since the wait has to be equal to the thread pool size due to the throttling.

renku/core/management/datasets.py
jsam: I tried that logic and on my fiber I was still hitting the rate limit very easily (after few minutes of import). With this logic you normalize the requests rate across the thread pool, which results in not being throttled or rate limited. The additional problem with `1 - exec_time` is that once you hit a rate limit, and eventually continue, you hit the rate limit again and those subsequent rate limit hits are extremely punishing - I was waiting up to 45 minutes to be allowed to make a new request - that's why this normalization across thread pool is needed.

mohammad-sdsc: My guess is that there might be other rate limits there in place; if it is was only this 14 req/sec, then with 8 thread and maximum one request per second, we should be well below it. Anyways, keeping it like this is OK although not very user friendly.

renku/core/management/datasets.py
mohammad-sdsc: I got it from https://docs.python.org/3/library/concurrent.futures.html and then just reduced the max value. Basically, since downloading is an IO-bound operation, Python library used 5x more download thread up to 3.5 and they adjust it to be a bit lower from 3.6 upwards.
I believe UX will be a bit degraded with these values but we can adjust later if it is the case.

Issue #1105, opened 2020-03-13
Metadata is not committed when deleting a file from a dataset


mohammad-sdsc:
To reproduce:
1. Create a dataset and add some files to it
2. Unlink some files: `renku dataset unlink my-data -I 'some-file'`

Metadata is updated but it is not committed (an empty commit is created) and repository state is dirty.

PR #1120, opened 2020-03-19
fix(dataset): commit after unlinking files

mohammad-sdsc:
# Description

Changes after a `renku dataset unlink` commands weren't committed.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/1105


mohammad-sdsc:
> Could we also add to the test that the commit is created and not only not dirty?

Good point, done! Thanks for the review!

PR #1123, opened 2020-03-20
feat(dataset): show file size in ls-files

mohammad-sdsc:
# Description

Displays file sizes in `renku dataset ls-files`. It shows the current size of the file in the disk (which for files that are tracked in LFS but not pulled, it is equal to the pointer file size).


rokroskar:
> for files that are tracked in LFS but not pulled, it is equal to the pointer file size

you mean the size of the object as listed in the pointer file or really the size of the pointer file on disk (i.e. a few bytes)

mohammad-sdsc:
> > for files that are tracked in LFS but not pulled, it is equal to the pointer file size
> 
> you mean the size of the object as listed in the pointer file or really the size of the pointer file on disk (i.e. a few bytes)

The size of the pointer file itself. My line of thinking was that this can be confusing if we show actual file size which differs from its size on the disk. Moreover, it can be used to check which files are not pulled from LFS.

rokroskar:
Hmm I guess that's tricky. I would think the user would want to know the actual file sizes?

jsam:
This is great addition, only thing I'm wondering is if we could could use some other measure than bytes? I have some datasets which contain archives between 10 and 15 TB and when I list them the number is a bit hard to read (`17,492,679,888,515` :o ). I propose we default to MB or use relative measures.

renku/core/commands/format/dataset_files.py
```diff
@@ -29,13 +32,48 @@ def tabular(client, records, *, columns=None):
     if not columns:
         columns = 'added,creators,dataset,full_path'
 
+    if 'size' in columns.split(','):
+        _get_lfs_file_sizes(client, records)
```
rokroskar, line 36: This is assuming all dataset files are in LFS - what if that isn't the case?

mohammad-sdsc: It shows nothing for the size field when the file is not tracked by LFS.

rokroskar: Shouldn't it work regardless of how the file is stored?

mohammad-sdsc: I personally find it confusing to show size on the disk for some files and LFS size (which can be different) for others. But I see your point here. I'll include size on the disk when not in LFS.

mohammad-sdsc: What do you think about different units for file sizes (e.g. B, KB, MB, ...) which comes from LFS? Should I normalise those to be all MB?

rokroskar: Maybe use https://pypi.org/project/humanize/?

mohammad-sdsc: Now, this displays file size on the disk for those files that are not in LFS. To be compatible with LFS format I've used `humanize` to display humanized file sizes.

PR #1127, opened 2020-03-20
feat: renku init options refactor

### LINKED ISSUE #1125: make renku init options more user-friendly
1. [ ] Change the output of --print-manifest to have these columns:
    * INDEX: template position in the manifest.yaml
    * ID: folder
    * DESCRIPTION: combination of {name}: {description} from manifest.yaml
2. [ ] Remove the ambiguous `--template` and add `--template-index` and `--template-id` instead
3. [ ] `--print-manifest` is not intuitive. Change to `--list-templates` (other names are welcome!)

lorenzo-cavazzi:
# Description
Restyle `renku init` options to make them more intuitive and user friendly.
The list of changes is available here #1125

Fixes #1125

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -39,7 +39,6 @@ import yaml
 from _pytest.monkeypatch import MonkeyPatch
 from click.testing import CliRunner
 from git import Repo
-from tests.core.commands.test_init import TEMPLATE_ID
 from walrus import Database
 
 
@@ -156,7 +155,7 @@ def repository():
 
     with runner.isolated_filesystem() as project_path:
         result = runner.invoke(
-            cli, ['init', '.', '--template-id', TEMPLATE_ID],
+            cli, ['init', '.', '--template-id', 'python-minimal'],
             catch_exceptions=False
         )
         assert 0 == result.exit_code
@@ -562,7 +561,9 @@ def remote_project(data_repository, directory_tree):
     runner = CliRunner()
 
     with runner.isolated_filesystem() as project_path:
-        runner.invoke(cli, ['-S', 'init', '.', '--template-id', TEMPLATE_ID])
+        runner.invoke(
+            cli, ['-S', 'init', '.', '--template-id', 'python-minimal']
+        )
         result = runner.invoke(
             cli, ['-S', 'dataset', 'create', 'remote-dataset']
         )
--- a/conftest.py
+++ b/conftest.py
@@ -39,6 +39,7 @@ import yaml
 from _pytest.monkeypatch import MonkeyPatch
 from click.testing import CliRunner
 from git import Repo
+from tests.core.commands.test_init import TEMPLATE_ID
 from walrus import Database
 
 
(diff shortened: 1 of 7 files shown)


lorenzo-cavazzi:
What is the `documentation-pending` label? :thinking:

tests/cli/test_init.py
```diff
@@ -19,38 +19,56 @@
...
+    assert (
+        '1 folder_python Template Python: Description Python'
+    ) in stripped_sentence
+    assert '2 folder_R Template R: Description R' in stripped_sentence
     assert instructions not in stripped_sentence
     full_sentence = create_template_sentence(templates, True)
     assert instructions in full_sentence
 
 
+@pytest.mark.integration
```
mohammad-sdsc: This test does not need to be an integration test. We normally use this marker when the test needs to access the Internet.

tests/cli/test_init.py
```diff
@@ -19,38 +19,56 @@
...
+    """Test listing templates."""
+    new_project = Path('test-new-project')
+    assert not new_project.exists()
+    result = isolated_runner.invoke(cli, INIT + LIST_TEMPLATES)
+    assert 0 == result.exit_code
+    assert not new_project.exists()
+    assert TEMPLATE_ID in result.output
+
+
 @pytest.mark.integration
```
mohammad-sdsc: Same as above.

conftest.py
```diff
@@ -155,7 +155,7 @@ def repository():
 
     with runner.isolated_filesystem() as project_path:
         result = runner.invoke(
-            cli, ['init', '.', '--template', 'Basic Python Project'],
+            cli, ['init', '.', '--template-id', 'python-minimal'],
```
mohammad-sdsc: I know that this is decided somewhere else, but is there a reason to have the `-minimal` postfix here? It would be cleaner to have just `python` and `R`. Also, `--template` looks cleaner to me than `--template-id`.

rokroskar:
> What is the documentation-pending label?

It's to remind you to make sure that you need update the documentation if needed. In this case it seems like that is definitely the case...

tests/cli/test_init.py
lorenzo-cavazzi: ok good to know, then I should remove it also from other tests here, only 1 is accessing internet to download a repo

conftest.py
lorenzo-cavazzi: for `-minimal` I tend to agree with you but that should be changed here first https://github.com/SwissDataScienceCenter/renku-project-template
We may open an issue there.

We split `--template` in `--template-id` and `--template-index` in this PR, I think `--template` was confusing and we agreed on this type of change during the last design meeting. We can further change this if it turns out not to be clear enough, but I would rather avoid using `--template` since there are many options named `--template-xxx`

lorenzo-cavazzi:
@rokroskar I added docs for the template part that was not covered before

PR #1128, opened 2020-03-23
feat(dataset): wildcard support when adding data from git

mohammad-sdsc:
# Description

Now, you can use wildcards when selecting which files to be added from a git repo:
```
renku dataset add my-dataset -s notebooks/**/*py -s notebooks/*renku https://github.com/SwissDataScienceCenter/r10e-ds-py.git
```

### DIFF
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -118,12 +118,13 @@ The command above will result in a structure like
       my-dataset/
         datafile
 
-You can use shell-like waldcards (e.g. *, **, ?) when specifying paths to be
-added:
+You can use shell-like wildcards (e.g. *, **, ?) when specifying paths to be
+added. Put wildcard patterns in quotes to prevent your shell from expanding
+them.
 
 .. code-block:: console
 
-    $ renku dataset add my-dataset --source path/**/datafile \
+    $ renku dataset add my-dataset --source 'path/**/datafile' \
         git+ssh://host.io/namespace/project.git
 
 You can use ``--destination`` or ``-d`` flag to change the name of the target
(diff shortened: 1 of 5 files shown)


rokroskar:
Will this wildcard repeat the selection when you run `renku dataset update`?

mohammad-sdsc:
> Will this wildcard repeat the selection when you run `renku dataset update`?

No, it won't. This is evaluated at the time the command is executed and gets translated to a list of files.

rokroskar:
So should the workflow be then to just repeat the `add` command with the wildcard? will it fail if the files are already there?

mohammad-sdsc:
> So should the workflow be then to just repeat the `add` command with the wildcard? will it fail if the files are already there?

True. Repeating the workflow would work but we need to pass `--force` flag too. Adding of existing files fails unless this flag is passed in.

rokroskar:
would it make sense to have a second one, e.g. `--ignore-existing` or similar?

renku/cli/dataset.py
```diff
@@ -118,6 +118,14 @@
       my-dataset/
         datafile
 
+You can use shell-like waldcards (e.g. *, **, ?) when specifying paths to be
+added:
+
+.. code-block:: console
+
+    $ renku dataset add my-dataset --source path/**/datafile \
```
rokroskar: do the options with asterisks need to be in quotes?

mohammad-sdsc: It's better to do it. If the path does not exist in the CWD then shell won't expand it and it's safe to use it without quotes.
I'll add quotes in the example.

Issue #1129, opened 2020-03-23
flag to disable external storage does not work


rokroskar:
**Describe the bug**
`renku -S run <script>` still puts files in LFS.

**To Reproduce**

```
renku -S run echo woop > woop
git lfs ls-files
9d65e6fd03 * woop
```
**Expected behavior**
`woop` should not be listed under lfs files above.

PR #1130, opened 2020-03-23
fix(dataset): --no-external-storage flag not working

mohammad-sdsc:
# Description

Running renku with `--no-external-storage` was still tracking files in LFS which will be fixed by this PR.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/1129

### DIFF
--- a/renku/core/management/storage.py
+++ b/renku/core/management/storage.py
@@ -83,17 +83,19 @@ class StorageApiMixin(RepositoryApiMixin):
         :raises: ``errors.ExternalStorageNotInstalled``
         :raises: ``errors.ExternalStorageDisabled``
         """
+        if not self.use_external_storage:
+            return False
+
         repo_config = self.repo.config_reader(config_level='repository')
         lfs_enabled = repo_config.has_section('filter "lfs"')
 
-        storage_enabled = lfs_enabled and self.storage_installed
-        if self.use_external_storage and not storage_enabled:
+        if not lfs_enabled:
             raise errors.ExternalStorageDisabled(self.repo)
 
-        if lfs_enabled and not self.storage_installed:
+        if not self.storage_installed:
             raise errors.ExternalStorageNotInstalled(self.repo)
 
-        return lfs_enabled and self.storage_installed
+        return True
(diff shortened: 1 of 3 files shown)


renku/core/management/storage.py
```diff
@@ -83,17 +83,19 @@ def has_external_storage(self):
         :raises: ``errors.ExternalStorageNotInstalled``
         :raises: ``errors.ExternalStorageDisabled``
         """
+        if not self.use_external_storage:
```
rokroskar: Isn't this unrelated to whether the repository has external storage enabled? I think they should be kept as two separate checks.

tests/core/commands/test_cli.py
```diff
@@ -320,7 +320,7 @@ def test_configuration_of_no_external_storage(isolated_runner, monkeypatch):
         monkey.setattr(StorageApiMixin, 'storage_installed', False)
         # Missing --no-external-storage flag.
         result = runner.invoke(cli, ['run', 'touch', 'output'])
-        assert 'is not configured' in result.output
+        assert 'External storage is not configured' in result.output
```
rokroskar: configured or installed?

renku/core/management/storage.py
mohammad-sdsc: I am not sure if it was the intention of this property to make them separated because `use_external_storage` was checked here before:
```
if self.use_external_storage and not storage_enabled:
    raise errors.ExternalStorageDisabled(self.repo)
```
Moreover, it does not make sense to complain about LFS not installed when user explicitly asked to not use LFS.

tests/core/commands/test_cli.py
mohammad-sdsc: Configured. Installed is checked somewhere else.

renku/core/management/storage.py
rokroskar: it does if it's necessary to pull LFS objects in order to re-execute something

mohammad-sdsc: OK.

Issue #1126, opened 2020-03-20
Read variables list when using renku init


lorenzo-cavazzi:
**Is your feature request related to a problem? Please describe.**
Users need to know the variable list required by a template to initialize a new project

**Describe the solution you'd like**
Once SwissDataScienceCenter/renku-project-template#68 is merged, we can 
* [x] Print the variables list in `--print-manifest` (renamed as `--list-templates`)
* [x] Remove `--description` (currently not used) since there is no need to have it as a special variable
* [x] Rename `--template-variables` to `--variables`
* [ ] [bonus]: when some variables are required but not specified, interactively ask them to the user -- we may move this to a new issue

**Describe alternatives you've considered**
We could scan all the files in the template and try to find all the jinja variables, but that requires checking out all the files and it doesn't provide a description.

**Additional context**
Listing the variables is needed also for implementing #862

PR #1134, opened 2020-03-24
feat(init): read and display template variables

lorenzo-cavazzi:
This addresses the first 3 points of #1126 (I separated the fourth since I'm facing a few obstacles)

* Variables are read from the template and showed (with their description).
* `--description` is no longer a special variable with its own option but it may be provided through `--variables` wherever needed.
`--name` stays as the only special case. It can be provided by its own option or taken from the folder name. In any case, it's also forwarded as the variable `name` to the templates, making it unnecessary in the manifest (change it where needed https://github.com/SwissDataScienceCenter/renku-project-template/blob/eca8b97d22903650a45b1635b6624688f7bc0986/manifest.yaml#L5)


lorenzo-cavazzi:
ok this is now ready for review

renku/cli/init.py
```diff
@@ -75,15 +75,41 @@
...
+can check them by listing the templates `--list-templates`.
+
+To provide variables, use the ``--variables```option and list them as in a
+Python dictionary or a JSON object requiring either single quotes ``'`` or
+single quotes ``"`` for variables names.
+
+.. code-block:: console
+
+    $ renku init --template-id python-minimal --variables \
+    '{ "description": "my new shiny project" }'
```
rokroskar: could the ux around this be a bit nicer, maybe like this?

```
renku init --template-id python-minimal --parameter description="my new shiny project" --parameter another-parameter=value
```

This would follow the click convention for specifying multiple parameters - forcing users to write JSON might not go over so well :)

lorenzo-cavazzi: I definitely like this idea :+1:

renku/cli/init.py
```diff
@@ -75,15 +75,40 @@
...
-Updating an existing project
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Provide parameters
+~~~~~~~~~~~~~~~~~-
+
+Some templates require parameters to properly initialize a new project. You
+can check them by listing the templates ``--list-templates``.
+
+To provide parameters, use the ``--parameter`` option and provide each
+parameter using ``--parameter "param1"="value1"``.
```
mohammad-sdsc, line 92: Do we need to also put `param1` in the quotes? It's not necessary when there is no space and I guess parameter names have no space in them.

renku/cli/init.py
```diff
@@ -75,15 +75,40 @@
...
+    Initializing new Renku repository... OK
+
+If you don't provide the required parameters, the template will use an empty
+strings instead.
+
+.. note:: Every project requires a ``name`` that can either be provided using
+   ``--name`` or automatically taken from the target folder. This is
+   also considered as a special parameter, therefore it's automatically added
+   to the list of parameters forwarded to the ``init`` command. Providing a
+   different value for ``name`` through ``--parameter`` has no effect.
```
mohammad-sdsc, line 108: We should hide the `name` parameter then. There is no point in showing it when users cannot set its value. Is this possible?

renku/cli/init.py
```diff
@@ -133,29 +157,18 @@
...
-    )
+def parse_parameters(ctx, param, value):
+    """Parse parameters to dictionary."""
+    parameters = {}
+    for index, parameter in enumerate(value):
+        splitted = parameter.split('=', 1)
+        if len(splitted) < 2 or len(splitted[0]) < 1:
+            raise errors.ParameterError(
+                'Parameter format must be --parameter "param1"="value". ',
+                f'--parameter #{index + 1}'
```
mohammad-sdsc, line 168: It's better to show the parameter itself instead of its index: `f'"{parameter}"`

renku/cli/init.py
```diff
@@ -236,31 +256,39 @@ def check_git_user_config():
     callback=validate_name,
     help='Provide a custom project name.',
 )
-@click.option('--template-id', help='Provide the id of the template to use.')
 @click.option(
+    '-i', '--template-id', help='Provide the id of the template to use.'
+)
```
mohammad-sdsc, line 261: These short options are good but a bit inconsistent which makes them hard to remember and kinda defeat the purpose. May I suggest that we use `-t` for `--template-id` since it should be used most, and then use `-i` for `--template-index` (which will be consistent with the reset)?

tests/core/commands/test_init.py
```diff
@@ -96,6 +99,14 @@ def test_read_template_manifest():
...
         assert 'Description 2' == manifest[1]['description']
 
+        variables1 = manifest[0]['variables']
+        variables2 = manifest[1]['variables']
+        assert 0 == len(variables1)
+        assert 1 == len(variables2)
+        assert 1 == len(variables2.keys())
+        assert 'custom' in variables2.keys()
+        assert 'Custom Value' == variables2['custom']
+
```
mohammad-sdsc, line 109: What happens if user passes a non-existing parameter? We should have a test for that.

renku/cli/init.py
lorenzo-cavazzi: Not sure if quotes are really needed here for `param1`, I assumed there may be a (not so trivial) valid python variable name that, when passed through the cli, could be misinterpreted without the quotes -- if that is not the case, I can remove them.

renku/cli/init.py
lorenzo-cavazzi: The user can pass a custom `name`, otherwise `name = <current_folder_name>`.

This should clarify that passing a custom `--parameter name=something` has no effect because the parameter `name` is already automatically added from `--name`. It's a special case.

I fear it's not explained well :sweat_smile:  Please feel free to suggest something more understandable to users :slightly_smiling_face:

renku/cli/init.py
lorenzo-cavazzi: this makes sense :+1:

renku/cli/init.py
lorenzo-cavazzi: Good point! Thanks for considering the consistency with other commands here. I changed it as suggested

tests/core/commands/test_init.py
lorenzo-cavazzi: This is currently tested in `test_create_from_template` later in the file (line 170). We are passing parameters not listed in the expected parameters (`description`).

Anyway, that part is a bit confusing now but it will change soon since the last point of #1126 is about checking parameters validity.

renku/cli/init.py
mohammad-sdsc: It's OK, keeping them doesn't hurt.

renku/cli/init.py
mohammad-sdsc: It's just a bit confusing that `name` is shown in the parameters' list (`renku init -l`) but it cannot be set as a normal parameter. It would be ideal either to remove it from the parameters' list or remove `--name` and treat `name` as a normal parameter.

lorenzo-cavazzi: oh yes, you are totally right! I will open a PR to update that in the templates repository. The original idea was slightly different, this is why `name` was added as a parameter there

lorenzo-cavazzi: This should remove this confusing behavior SwissDataScienceCenter/renku-project-template#72

Once merged and tagged, no template will require `name` as a parameter anymore.

mohammad-sdsc: Great! We can remove this line then: `Providing a different value for ``name`` through ``--parameter`` has no effect.`

lorenzo-cavazzi: done :+1:

Issue #778, opened 2019-10-29
warn the user when they provide an unusable URL


rokroskar:
**Is your feature request related to a problem? Please describe.**
When a user provides a generic file dropbox link to the `renku dataset add` command, the result is that they get the HTML of the dropbox page instead of the contents of the file. This is easily remedied by adding `?dl=1` to the URL, but most users don't know this so we should tell them. A similar problem occurs when users copy/paste the URL of a file in gitlab/github instead of the "raw" URL. When this happens for URLs that we know are problematic, we should fail gracefully and warn the user with instructions on how to fix it. 

**Describe the solution you'd like**
```bash
renku dataset add https://dropbox.com/s/<path>

WARNING: you provided a generic Dropbox URL 
This will *not* download your data but the HTML of the Dropbox page instead. 
To retrieve the data, append `?dl=1` to the URL. 
```

**Describe alternatives you've considered**
Let the user figure it out for themselves; pretend we are `curl`

PR #1135, opened 2020-03-24
feat(datasets): add files from dropbox

jsam:
closes: #778

### DIFF
--- a/renku/core/management/datasets.py
+++ b/renku/core/management/datasets.py
@@ -30,6 +30,7 @@ from contextlib import contextmanager
 from pathlib import Path
 from subprocess import PIPE, SubprocessError, run
 from urllib import error, parse
+from urllib.parse import ParseResult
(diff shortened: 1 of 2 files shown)


renku/core/management/datasets.py
```diff
@@ -468,14 +468,35 @@ def _add_from_urls(
...
+            query = query.replace('dl=0', 'dl=1')
+        else:
+            query += 'dl=1'
+
+        url = parse.urlunparse((url._replace(query=query)))
+        return url
+
+    def provider_check(self, url):
+        """Check additional provider related operations."""
+        if 'dropbox.com' in url:
```
rokroskar: would it be safer here to parse the url and specifically look at the `host`?

rokroskar: `netloc` not `host` :) 👍

Issue #1063, opened 2020-02-25
feat(svc): enable adding remote files to a dataset


jsam:
We should modify existing endpoint to support importing remote files via pull strategy. To do this we could edit the existing endpoint:
```
POST /datasets.add
{
  "project_id": <project_id>,
  "dataset_name": <dataset_name>,
  "files" [
    {
       "file_url": "http://..."
    },
    ...
   ]
}
```

For each `file_url` we should schedule a user job which will pull the data and add it to a dataset.

PR #1139, opened 2020-03-25
feat(service): add remote files to dataset

jsam:
closes: #1063


tests/service/jobs/test_datasets.py
```diff
@@ -271,3 +272,55 @@ def test_dataset_import_twice_job(doi, svc_client_with_repo):
...
     assert 'Dataset exists' in extras['error']
+
+
+@pytest.mark.parametrize(
+    'url', [
+        'https://gist.github.com/jsam/d957f306ed0fe4ff018e902df6a1c8e3',
+    ]
+)
+@pytest.mark.integration
+@pytest.mark.service
```
mohammad-sdsc: Does this need flaky as well?

tests/service/views/test_dataset_views.py
```diff
@@ -807,3 +807,195 @@ def test_import_dataset_job_enqueue(
...
+            'dataset_name': uuid.uuid4().hex,
+            'create_dataset': True,
+            'files': [{
+                'file_url': url_gist,
+            }, {
+                'file_url': url_dbox,
+            }]
+        }),
+        headers=headers,
+    )
```
mohammad-sdsc: This code has a lot of redundancy with the above function. Please consider to factor it to a function.

tests/service/jobs/test_datasets.py
jsam: done

tests/service/views/test_dataset_views.py
jsam: done.

Issue #1064, opened 2020-02-25
chore(svc): make cleanup strategy aware of user jobs


jsam:
Currently, both cleanup strategies are not aware of user jobs. This could lead to an edge case where, a cleanup job could potentially delete a cached project (or a file) which is being used in some other en-queued user job. To prevent this, we can easily add a logic to strategy which would check en-queued user jobs, and if:
 1) there is a job which is not yet completed and 
 2) it contains a reference to a resource (project/files) affected by a TTL

It will simply skip cleanup of those resources.

PR #1140, opened 2020-03-25
chore(service): cleanup aware of user jobs

jsam:
closes: #1064


renku/service/jobs/cleanup.py
```diff
@@ -21,61 +21,107 @@
...
+    """Check if file should be deleted."""
+    ttl = ttl or int(os.getenv('RENKU_SVC_CLEANUP_TTL_FILES', 1800))
+
+    # If record does not contain created_at,
+    # it means its an old record, and we should mark it for deletion.
+    created_at = (file.created_at -
+                  datetime.utcfromtimestamp(0)).total_seconds() * 1e+3
+
+    if not file.created_at:
+        created_at = (time.time() - ttl) * 1e+3
```
Panaetius, line 48: Shouldn't you just return here directly? because `time.time() - (time.time() - ttl)` is guaranteed to be `> ttl` further down, so this code makes it look like some calculation is happening, even though it just returns.

renku/service/jobs/cleanup.py
```diff
@@ -21,61 +21,107 @@
...
+def project_cleanup_ttl_check(project, ttl=None):
+    """Check if file should be deleted."""
+    ttl = ttl or int(os.getenv('RENKU_SVC_CLEANUP_TTL_PROJECTS', 1800))
+    # If record does not contain created_at,
+    # it means its an old record, and we should mark it for deletion.
+    created_at = (project.created_at -
+                  datetime.utcfromtimestamp(0)).total_seconds() * 1e+3
+
+    if not project.created_at:
+        created_at = (time.time() - ttl) * 1e+3
```
Panaetius, line 94: Same as above.

renku/service/jobs/cleanup.py
```diff
@@ -21,61 +21,107 @@
...
 
-                if file_path.exists() and file_path.is_dir():
+                if file_path.is_dir():
                     shutil.rmtree(str(file_path))
 
                 file.delete()
 
 
+def project_cleanup_ttl_check(project, ttl=None):
+    """Check if file should be deleted."""
```
Panaetius, line 86:
```suggestion
    """Check if project should be deleted."""
```

renku/service/jobs/cleanup.py
```diff
@@ -21,61 +21,107 @@
...
+            job for job in cache.get_jobs(user) if job.state in
+            [USER_JOB_STATE_ENQUEUED, USER_JOB_STATE_IN_PROGRESS]
+        ]
 
-            if not project.created_at:
-                created_at = (time.time() - ttl) * 1e+3
-
-            old = ((time.time() * 1e+3) - created_at) / 1e+3
+        for project in projects:
+            if project_cleanup_lock_check(jobs, project):
```
Panaetius, line 121: Small nitpick but could we flip the logic in this method around?

We have
```
def project_cleanup_ttl_check(project, ttl=None):
    """Check if file should be deleted."""
```

where `True` means "delete the project"

and
```
def project_cleanup_lock_check(jobs, project):
    """Check if project should be deleted."""
```
where `True` means "Don't delete the project", which is a bit inconsistent.

renku/service/jobs/cleanup.py
jsam: Right, this look's like my rebase went bad. This is not up to date code, let me check if last commit is pushed. Thanks for the review btw!

jsam:
@Panaetius Thank you for the review one more time and I apologize for failed rebase. All the points have been actually addressed in a way in latest version, so let me just give you a short update. All project and file resources have helper methods which is used in cleanup jobs and they follow the identical interface. This interface specifies: `abs_path`, `exists`, `ttl_expired`, `purge` and `is_locked`. Adding new resources in the future should also follow this interface.

Things which are different from previous version which you saw are: 
    1) cleanup jobs follow similar implementation 
    2) common interface between resources (mentioned above) 
    3) `ttl_expired` is the new method in which we check if the resource has time to live and it does returns early in case of older non-migrated cached objects

Hope this helps with review!

Issue #926, opened 2020-01-21
Immutable Datasets


mohammad-sdsc:
As a Renku user, I would like to be able to differentiate an imported dataset that is modified by me from its origin.

**Additional context**
When we import a dataset, we set `schema:sameAs` property of the imported dataset to point to the DOI/URL of the original dataset. If we modify this dataset locally then it is not literally the same dataset as its origin anymore.

**Acceptance criteria**
* a new dataset `schema:identifier` should be created in case of a dataset modification;
* all the original dataset properties should be kept with few exceptions:
  * the `schema:sameAs` (and I guess the `schema:url`) should be dropped if existed;
  * the `schema:dateCreated` should be set to the date of the modification;
  * the person who performed the modification should be added to `schema:creator`;
  * all the other properties like `name`, `description`, `hasPart`, and so on should be changed only if that was the part of the modification to the dataset;
* a new `prov:wasDerivedFrom` pointing to the original dataset should be added to the newly created dataset entity;

rokroskar:
This sounds good to me - the only thing that is missing (maybe on accident?) is that `wasDerivedFrom` should point to the original dataset so the provenance path is clear. 👍

emmjab:
I'm not sure if this is the top level issue for this, but for instance on dataverse, the DOI seems to stay the same even if you "republish" with changes, and a "V2", "V3", etc. are updated. In that case we would want to keep (DOI/URL, Version) for sameAs somehow?

PR #1142, opened 2020-03-26
feat(dataset): immutability

mohammad-sdsc:
# Description

ATM, there is no way in Renku to distinguish between datasets in different projects that originate from the same source. We always assume that these datasets are `schema:sameAs` the original dataset. This assumption won't hold if one of those datasets is modified.

This PR changes dataset's identifier on each change to the dataset which means that all datasets are immutable. We link all different versions of a dataset using `prov:wasDerivedFrom`.

# Note:
* Although tags are stored in dataset's metadata, we don't change the ID when adding/removing tags
* When importing a Renku dataset, always the latest version is imported

Fixes: https://github.com/SwissDataScienceCenter/renku-python/issues/926


rokroskar:
Thanks for this change @mohammad-sdsc, it's going to open up a lot of nice possibilities! 

I've started looking at this and the graph that comes out looks pretty reasonable but there are a few issues. 

In the [test project](https://dev.renku.ch/projects/rokroskar/test-1142/) I have this sequence of commands:

```
renku dataset create mydata
renku dataset add mydata <path-to-local-file>
renku dataset add mydata <url>
```

So I would expect to see 

```
<dataset-rev1> <-- wasRevOf -- <dataset-rev2> <-- wasRevOf -- <dataset-rev3>
```

and `<dataset-rev1>` should have no files, `<dataset-rev2>` should contain the first file and `<dataset-rev3>` should contain both files. 

When I look at the graph that I get though I see that `<dataset-rev1>` contains the first file and `<dataset-rev2>` contains both files - so something must be going wrong during graph generation. This is the full dataset graph (you can see it as well in [the test project notebook](https://dev.renku.ch/projects/rokroskar/test-1142/files/blob/notebooks/graph-tests.ipynb):

noname gv

Another thing I'm not sure about - the metadata file is kept the same (under the original dataset uuid) even though the uuids of the datasets are actually changing. Would it make sense that these are kept consistent so that the name of the metadata directory is the same as the uuid of the dataset? Or is there a reason you kept them the same? On the other hand, it _might_ be nice to have the dataset metadata identified here by the dataset's short name - then all the revisions go under the same easily identifiable name and we don't need the whole "refs" machinery at all. We could also keep all the metadata around (each in its own uuid directory) - that might get a bit verbose/redundant, but it might be the state where we end up anyway eventually if the graph is stored as a flat list in the future. @Panaetius any thoughts on this?

mohammad-sdsc:
@rokroskar Thanks for the feedback!

> When I look at the graph that I get though I see that <dataset-rev1> contains the first file and <dataset-rev2> contains both files.

That's true! ATM, I am using `prov:wasRevisionOf` to generate the chain of changes in each commit to make it possible to generate this graph locally per commit (i.e. `renku log --format dot-full`). I've made a mistake that shows the current changes in the previous version.

I will remove this `wasRevisionOf` from the final PR. KG processes each commit separately and makes the link between them and it does not need this chain in each commit.

My line of thinking was to use `prov:wasRevisionOf` to make the chain of changes and use `prov:wasDerivedFrom` to keep track of the very first URL/id that we imported from. However, after discussing with Kuba, we agreed that keeping both is redundant and we only need to keep `prov:wasDerivedFrom` and use it also for chaining the changes.

> Another thing I'm not sure about - the metadata file is kept the same (under the original dataset uuid)

This was a known issue and I wrote that in the PR's notes. I'm going to change this to be `.renku/datasets/<short_name>.yml` or `.renku/datasets/<short_name>/metadata.yml`. I haven't decided yet but the latter is more like what we have now. I also don't think that we should keep all metadata for all revisions and agree that we can remove the `ref` mechanism.

> could you explain what the reasoning is behind the schema:Series relations?

The idea is to keep the chain of all `_id` changes along with their corresponding commit sha in the metadata. This is useful for example if we want to get a specific version of the dataset (i.e. specific `_id`) locally without querying KG. I am open to suggestions here.

Panaetius (approved):
I tested it with different commands and everything works great! thank you!

Issue #1013, opened 2020-02-14
fix(svc): additional logging


jsam:
We need more logging on service views (especially on cache views).

edit: would be nice to have common `logging.yml` which we can reuse in all our services.

edit2:

AC:
* Add reusable logging.yml
* Add request logging
* Add scheduler logging
* Add worker logging

PR #1152, opened 2020-03-30
chore(service): added additional logging

jsam:
closes: #1013

This PR needs to be merged after #1140 due to the need to add additional logging in those refactored jobs.

### DIFF
--- a/renku/service/logging.yaml
+++ b/renku/service/logging.yaml
@@ -1,0 +1,31 @@
+---
+formatters:
+  simple:
+    format: "[%(asctime)s] [%(levelname)s] [%(name)s] %(message)s"
+handlers:
+  console:
+    class: logging.StreamHandler
+    formatter: simple
+    level: DEBUG
+    stream: "ext://sys.stdout"
+loggers:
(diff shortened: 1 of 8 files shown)


renku/service/logging.yaml
```diff
@@ -0,0 +1,31 @@
...
+  console:
+    class: logging.StreamHandler
+    formatter: simple
+    level: DEBUG
+    stream: "ext://sys.stdout"
+loggers:
+  service:
+    handlers:
+      - console
+    level: DEBUG
```
rokroskar: can we make the level configurable through an environment variable so we can easily toggle it in a running service?

jsam: Right, I have to test this a bit more after rebase, but this should be already in place and this config should be only for default value.

rokroskar: are they exposed through the chart?

jsam: I just checked and one is and one is not, I'll add them so both have explicitly set default values there. :+1:

PR #1155, opened 2020-03-31
fix(core): directory input regression

mohammad-sdsc:
# Description

Fixes a regression in which absolute paths are generated in workflows.


rokroskar:
Thanks for the quick turn-around on this @mohammad-sdsc!

PR #1160, opened 2020-04-02
feat(service): create projects from templates

### LINKED ISSUE #862: feat: service endpoint for project creation

lorenzo-cavazzi:
# Description

**DRAFT -- ONLY MANIFEST READ**

Allow an authorized user to invoke the service to clone a template repository, get the information from the manifest and initialyze a new project from one of the available templates.

Fixes #862

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -772,6 +772,48 @@ def svc_client_with_repo(integration_lifecycle):
     yield svc_client, deepcopy(headers), project_id, url_components
 
 
+@pytest.fixture(scope='module')
+def svc_client_with_templates(svc_client, mock_redis):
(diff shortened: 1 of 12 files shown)


lorenzo-cavazzi:
superseded by #1254

Issue #1158, opened 2020-04-02
Warn user when using CLI outside the root directory


mohammad-sdsc:
Some renku CLI commands won't work when run in a directory other that project's root. This might result in a bug report message or even worse, a silent metadata corruption. Until we fix https://github.com/SwissDataScienceCenter/renku-python/issues/1143, we should warn users when running commands outside project's root directory

Issue #1143, opened 2020-03-26
CLI commands fail when run from non-root directory


mohammad-sdsc:
**Describe the bug**
Commands might fail when run from anywhere other than project's root directory.

**To Reproduce**
```
$ renku init --template 'Basic Python Project'
$ cd test-project/notebooks
$ renku dataset add my-data -c ~/path/to/a/file
```

**Screenshots and/or execution output**
```
Ahhhhhhhh! You have found a bug. 🐞

1. Open an issue by typing "open";
2. Print human-readable information by typing "print";
3. See the full traceback without submitting details (default: "ignore").

Please select an action by typing its name (open, print, ignore) [ignore]: 
Traceback (most recent call last):
  File "/home/mohammad/.local/bin/renku", line 11, in <module>
    load_entry_point('renku', 'console_scripts', 'renku')()
  File "/home/mohammad/.local/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/mohammad/sdsc/code/renku-python/renku/cli/exception_handler.py", line 128, in main
    self._handle_github()
  File "/home/mohammad/sdsc/code/renku-python/renku/cli/exception_handler.py", line 175, in _handle_github
    getattr(self, '_process_' + value)()
  File "/home/mohammad/sdsc/code/renku-python/renku/cli/exception_handler.py", line 119, in main
    result = super().main(*args, **kwargs)
  File "/home/mohammad/sdsc/code/renku-python/renku/cli/exception_handler.py", line 90, in main
    return super().main(*args, **kwargs)
  File "/home/mohammad/.local/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/mohammad/.local/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mohammad/.local/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mohammad/.local/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/mohammad/.local/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
[...]

PR #1162, opened 2020-04-03
feat(core): CLI commands warning when in non-root directory

mohammad-sdsc:
# Description

Prints a warning message in stderr when running CLI commands from any directory other than the project's root. It also fixes this issue where ever possible (partially addressing https://github.com/SwissDataScienceCenter/renku-python/issues/1143).

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/1158


renku/cli/__init__.py
```diff
@@ -200,6 +202,16 @@ def cli(ctx, path, renku_home, use_external_storage):
         use_external_storage=use_external_storage,
     )
 
+    if (
+        path != os.getcwd() and
+        ctx.invoked_subcommand not in WARNING_UNPROTECTED_COMMANDS
+    ):
+        click.secho(
```
jsam, line 209: One small improvement would be to show this message after the command output and not before. For example I have a repo with ~30 datasets and when I do `renku datasets` it's quite easy to miss this warning message.

mohammad-sdsc:
> Look's good code wise. I'm just wondering if we should also shadow the `--path` flag which allows execution in parent directories.

I agree. We should also remove this `--datadir` flag but they belong to a separate issue/PR I believe.

renku/cli/__init__.py
mohammad-sdsc: TBH, I don't know a clean way of showing the message at the end using Click (considering pagers as well). This is the case when there is an exception and the bug report prompt is show (where this warning is probably needed most).

Issue #1042, opened 2020-02-21
Prevent adding dataset folder to same dataset


Panaetius:
I had downloaded data to `data/sbb-raw-data` and created a  renku dataset called `raw-sbb-data` (which in turn creates the folder `data/raw-sbb-data`).

I then wanted to add the former folder to the dataset, but mistyped/picked the wrong folder, doing `rrenku dataset add raw-sbb-data data/raw-sbb-data/` which leads to renku adding a folder `data/raw-sbb-data/raw-sbb-data` which gets picked up and added as a subfolder, which gets pick up, etc.

```$ renku dataset add raw-sbb-data data/raw-sbb-data/

Ahhhhhhhh! You have found a bug. 🐞

Recorded in Sentry with ID: 3bee230ff4c74a1097abb5be65c7eb68

Traceback (most recent call last):
  File "/home/jovyan/.local/bin/renku", line 8, in <module>
    sys.exit(cli())
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/renku/cli/exception_handler.py", line 123, in main
    self._handle_sentry()
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/renku/cli/exception_handler.py", line 119, in main
    result = super().main(*args, **kwargs)
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/renku/cli/exception_handler.py", line 90, in main
    return super().main(*args, **kwargs)
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jovyan/.local/pipx/venvs/renku/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

PR #1163, opened 2020-04-03
fix(dataset): avoid recursive addition of data directory

mohammad-sdsc:
Avoids recursive addition of a dataset's data directory to it.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/1042


mohammad-sdsc:
@rokroskar Can you please re-approve? I've resolved a conflict and it removed your approval.

PR #1171, opened 2020-04-08
chore(core): set filelock logging to error

jsam:


rokroskar:
The code looks fine, but the title of the PR says "warning" while the logging is set to "error" - which one is intended?

jsam:
Sorry, its a typo - just realized it as well. :see_no_evil:

PR #1175, opened 2020-04-08
refactor(service): project clone

jsam:
This PR refactor project clone behavior into private function so it could be re-used in `templates.read_manifest`.


renku/service/views/cache.py
```diff
@@ -163,33 +163,15 @@ def upload_file_view(user, cache):
...
 @requires_cache
-@requires_identity
-@accepts_json
-def project_clone(user, cache):
-    """Clone a remote repository."""
+def _project_clone(cache, user_data, data=None):
+    """Clones the project for a given user."""
     ctx = ProjectCloneContext().load(
-        (lambda a, b: a.update(b) or a)(request.json, user),
+        (lambda a, b: a.update(b) or a)(data or request.json, user_data),
```
rokroskar, line 170: I'm a bit confused by what this code is doing. `a.update(b)` is always `None` so the lambda always returns `a`. But `a.update(b)` executes, which updates `a`, or in this case `data` or `request.json` with the contents of `user_data`. In this case, the `lambda` has a side-effect outside of its scope which is strange, but I also don't get the point of this code. If the intent is just to merge `data` (or `request.json`) with `user_data` why not just do it outside of the function call? I'm probably just not understanding the syntax or something here...

jsam: This is quite common one liner to update and return the dict so you understanding is absolutely correct. `data` was called `context` in the first draft which might be less confusing. In many patterns (e.g. circuit breaker, etc) its quite normal that function mutates call context (especially if the context might be large and you don't want to make a copy). In this particular case it does not matter at all - weird or not.

But I do agree that function signature might be slightly better without using call context and list the all parameters (but that's quite subjective in this case since the parameter number is non-trivial).  What we could do though is split the function to do less and have more API function which achieve the same goal.

I could spent more time on modeling this non-public function to make it nicer, but since it's private and view only function I did what's the quickest!

Issue #737, opened 2019-10-04
Decouple communication between presentation layer (CLI/Service) and renku core


Panaetius:
In #702 dataset import/export click dependencies were removed from `renku.core` code, by using callbacks and otherwise injecting the relevant code into the business layer.

This works, but can get a bit unwieldy, especially since many of the callbacks are still somewhat `click`-specific and the interface might not be applicable for other potential consumers of `core`. Additionally, each method has it's own callbacks that sometimes differ from other functions' callbacks, making them still conceptually tightly coupled.

Basically, the refactoring we did with the `cli`, `core` namespaces means our application now has a rather traditional `presentation`, `logic` and potentially `DAL` layered architecture, so we should follow patterns that make sense to decouple `presentation` from `logic`.

Something like the mediator pattern or event based communication could work well.

As this should be unified for all functions (not each having its own slightly different implementation), it needs some architecture work and is not done in #702

Panaetius:
This has been implemented

PR #1199, opened 2020-04-17
feat(cli): Adds warning messages for LFS, fix output redirection

### LINKED ISSUE #1099: improve user feedback with external storage (LFS)
Users should be notified that the data they are importing or producing is going into external storage. At the same time, we should not overwhelm the user with so many messages that it's annoying and they stop paying attention.

Panaetius:
This adds warning messages when files get added to LFS like:
```
Warning: Adding files to Git LFS:
        file.txt
To disable this warning in the future, run:
        renku config show_lfs_warnings False
```

This PR also fixes output redirection in `renku run`. Before, when doing something like `renku run command >out.txt 2>err.txt`, not just the output of command would get passed into the files, but also any renku specific messages (like errors, warnings etc.). With this PR, the output redirection is only applied to the subprocess call, and renku itself can log to the terminal normally. Note: this doesn't work on windows, where the old behaviour is still in place.

This PR also fixes a bug where dataset files would always get added to LFS, even if the user didn't have/want LFS.

Closes #1099

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -201,6 +201,28 @@ def client(project):
     """Return a Renku repository."""
     from renku.core.management import LocalClient
 
+    original_get_value = LocalClient.get_value
+
+    def mocked_get_value(
+        self, section, key, local_only=False, global_only=False
+    ):
+        """We don't want lfs warnings in tests."""
+        if key == 'show_lfs_warnings':
+            return 'False'
+        return original_get_value(self, section, key, local_only, global_only)
+
+    LocalClient.get_value = mocked_get_value
+
+    yield LocalClient(path=project)
+
+    LocalClient.get_value = original_get_value
+
+
+@pytest.fixture
+def client_with_lfs_warning(project):
+    """Return a Renku repository with lfs warnings active."""
+    from renku.core.management import LocalClient
+
     yield LocalClient(path=project)
 
 
--- a/conftest.py
+++ b/conftest.py
@@ -230,12 +230,16 @@ def no_lfs_warning(client):
     yield client
 
 
-@pytest.fixture
+@pytest.fixture(scope='function')
 def client_with_lfs_warning(project):
     """Return a Renku repository with lfs warnings active."""
     from renku.core.management import LocalClient
 
-    yield LocalClient(path=project)
+    client = LocalClient(path=project)
+    client.set_value('renku', 'lfs_threshold', '0b')
+    client.repo.git.add('.renku/renku.ini')
+    client.repo.index.commit('update renku.ini')
+    yield client
 
 
 @pytest.fixture
--- a/conftest.py
+++ b/conftest.py
@@ -207,7 +207,7 @@ def client(project):
         self, section, key, local_only=False, global_only=False
     ):
         """We don't want lfs warnings in tests."""
-        if key == 'show_lfs_warnings':
+        if key == 'show_lfs_message':
             return 'False'
         return original_get_value(self, section, key, local_only, global_only)
 
@@ -216,6 +216,18 @@ def client(project):
     yield LocalClient(path=project)
 
     LocalClient.get_value = original_get_value
+
+
+@pytest.fixture
+def no_lfs_warning(client):
+    """Sets show_lfs_message to False.
+
+    For those times in life when mocking just isn't enough.
+    """
+    with client.commit():
+        client.set_value('renku', 'show_lfs_message', 'False')
+
+    yield client
 
 
 @pytest.fixture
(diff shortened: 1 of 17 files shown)


emmjab:
> This PR also fixes a bug where dataset files would always get added to LFS, even if the user didn't have/want LFS.

I also tested this by not installing LFS into a local copy of a repository and trying to add data; message shows up that i can use `-S`, when I use `-S`, the data is added to dataset but not to LFS ✔️

Panaetius:
Good point on changing the message to `INFO`!

Also currently the tests hang indefinitely, I think because Github Actions has no `/dev/tty`

rokroskar:
I'm confused by this last point:

> This PR also fixes a bug where dataset files would always get added to LFS, even if the user didn't have/want LFS.

This seems to already be fixed in `v0.10.2`:

```bash
 renku init test
cd test
echo woop > woop
git add woop
git commit -am 'added woop'
renku dataset create test-data
renku --no-external-storage dataset add test-data woop
git show
```

Panaetius:
> I'm confused by this last point:
> 
> > This PR also fixes a bug where dataset files would always get added to LFS, even if the user didn't have/want LFS.
> 
> This seems to already be fixed in `v0.10.2` (in #1130):
> 
> ```shell
> renku init test
> cd test
> echo woop > woop
> git add woop
> git commit -am 'added woop'
> renku dataset create test-data
> renku --no-external-storage dataset add test-data woop
> git show
> ```

Hmm, I see. That PR solves it, though the code should have been cleaned up a bit I think.

All places tracking files in external storage were doing it like this:
https://github.com/SwissDataScienceCenter/renku-python/blob/6a17512e416737419b370c730172117b338dd528/renku/core/models/cwl/command_line_tool.py#L358
i.e. wrapping the `track_paths_in_storage` call with the if. Except for https://github.com/SwissDataScienceCenter/renku-python/blob/115e46227decd7a4597a0918f3bae279fc73e366/renku/core/management/datasets.py#L364 where this check was missing (This is changed in this PR).

We can do it by checking inside `track_paths_in_storage` as in the PR you linked (which is cleaner anyway), but then we should remove all those `if client.has_external_storage:` statements I think. 

Also is there a difference between `client.has_external_storage` and `use_external_storage`, are both needed? I guess one is for when local storage is not configured and the other is for when a user explicitly wants to exclude files from external storage even though it's configured.

Maybe the checks can be unified as a single property, or `client.has_external_storage` can be moved to inside `track_paths_in_storage` as well.

I think if you `dataset add` inside a repo that has no lfs configured, you would get an exception right now about having to set up external storage vs. doing a `renku run` which would no complain about lfs not being set up (and just not add the files to LFS)

conftest.py
```diff
@@ -218,6 +218,18 @@ def mocked_get_value(
     LocalClient.get_value = original_get_value
 
 
+@pytest.fixture
+def no_lfs_warning(client):
+    """Sets show_lfs_message to False.
+
+    For those times in life when mocking just isn't enough.
```
rokroskar, line 225: 😂

renku/cli/config.py
```diff
@@ -62,6 +62,29 @@
...
++------------------------+-------------------------------------+-----------+
+| dataverse.access_token | Access token for Dataverse API      | ``None``  |
++------------------------+-------------------------------------+-----------+
+| dataverse.server_url   | URL for the Dataverse API server    | ``None``  |
+|                        | to use                              |           |
++------------------------+-------------------------------------+-----------+
+| show_lfs_message       | Whether to show messages about      | ``True``  |
+|                        | files being added to git LFS        |           |
+|                        | or not                              |           |
++------------------------+-------------------------------------+-----------+
```
jsam, line 87: Great addition. :guitar:  Thanks! :rocket:

renku/core/commands/dataset.py
```diff
@@ -276,6 +276,10 @@ def add_to_dataset(
                     progress=progress,
                 )
 
+            if messages:
+                for msg in messages:
+                    click.echo(INFO + msg)
```
jsam, line 281: I know we had previous similar implementation and this is just addition on top of it - with that said I think we should not do `click.echo` inside core commands since it could lead to leaking of messages to logs when used in service (or any other SDK context). I would propose we simply return those messages and then do click output inside of the CLI (since core module should not strictly depend on click - if we want to use it as SDK).

Panaetius: Click was mostly removed in dataset.py after https://github.com/SwissDataScienceCenter/renku-python/pull/738 but has been slowly creeping back in. There's also https://github.com/SwissDataScienceCenter/renku-python/issues/737
So not sure if this PR is the right place to address this.

Panaetius: https://github.com/SwissDataScienceCenter/renku-python/pull/1225 would this do?

PR #1186, opened 2020-04-14
chore(service): endpoint for editing dataset metadata

jsam:
closes: #1184

Issue #1148, opened 2020-03-27
fix(service): creating a dataset with long name results in InvalidParameter


jsam:
When we try to create a dataset with a long name through service, core is raising exception InvalidParameter for a dataset name. This is most likely due to API change on the core and would need adjusting.

jsam:
Related: #1075

Issue #1075, opened 2020-03-02
feat: allow the user to specify metadata on dataset creation


rokroskar:
Renku users should be able to specify required and optional metadata at the moment of dataset creation. This is needed for the CLI as well as the UI.

mohammad-sdsc:
Related: https://github.com/SwissDataScienceCenter/renku-python/issues/515

rokroskar:
Ah good point - so is this even needed still?

mohammad-sdsc:
Depends if we need more metadata or not. ATM, it accepts description, title, and list of creators (name and email).

rokroskar:
Does the backend support the same? cc @jsam

jsam:
Service does support on dataset creation setting this metadata (name, creators, description). For any additional fields we should build support in the core first.

rokroskar:
@vfried do you require other metadata at the moment?

vfried:
I need to check, at the moment we are only sending names without spaces and it would be good to be able to send names that have spaces. I will check if this was fixed or if it's still causing problems.

rokroskar:
@vfried @mohammad-sdsc @jsam is there additional metadata that needs to be added on creation? 

Right now we have

```
renku dataset create --help
Usage: renku dataset create [OPTIONS] SHORT_NAME

  Create an empty dataset in the current repo.

Options:
  -t, --title TEXT        Title of the dataset.
  -d, --description TEXT  Dataset's description.
  -c, --creator TEXT      Creator's name, email, and affiliation. Accepted
                          format is 'Forename Surname <email> [affiliation]'.

  -k, --keyword TEXT      List of keywords or tags.
  -h, --help              Show this message and exit.
```

vfried:
I still didn't started working on implementing this changes inside the UI, if what's available in the UI is the same as for the CLI (what's shown in the comment above) i would say that it's good for now.

jsam:
Another followup, #1236.

rokroskar:
Can this be closed @jsam @vfried @mohammad-sdsc ?

mohammad-sdsc:
> Can this be closed @jsam @vfried @mohammad-sdsc ?

On CLI side nothing more is needed. As for the UI, I'd prefer to close this and create an issue there if it's hasn't done yet.

rokroskar:
agreed @mohammad-sdsc! closing for now - @vfried if something else is needed from the backend, please reopen this or  create a new issue. Thanks!

PR #1200, opened 2020-04-17
refactor: replaces dataset_name with short_name & title support

jsam:
This PR adds consistency between service and the core in regards with `dataset_name` => `short_name` and adds support for adding title on dataset creation. 

addresses: #1075 
addresses: #1148

Rebase for #1186 is needed after merging this PR.

### DIFF
--- a/renku/service/serializers/datasets.py
+++ b/renku/service/serializers/datasets.py
@@ -138,28 +138,6 @@ class DatasetListRequest(Schema):
     project_id = fields.String(required=True)
 
 
-class CreatorDetails(Schema):
-    """Serialize a creator of a dataset to a response object."""
-
-    name = fields.String()
-    email = fields.String()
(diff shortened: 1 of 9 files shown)


renku/service/serializers/datasets.py
```diff
@@ -25,26 +25,37 @@
...
 
-class DatasetCreateRequest(Schema):
-    """Request schema for a dataset create view."""
+class DatasetDetails(Schema):
+    """Serialize a dataset to a response object."""
 
-    creators = fields.List(fields.Nested(DatasetCreators))
-    commit_message = fields.String()
-    dataset_name = fields.String(required=True)
+    short_name = fields.String(required=True)
```
mohammad-sdsc, line 36: `short_name` has the attribute `schema:alternateName`.

Issue #1176, opened 2020-04-08
chore(core): add ability on clone to checkout specific ref


jsam:
Currently, there is a need on `project_clone` and `read_manifest` to checkout a specific ref after cloning. This functionality should probably be added to the core module to keep flexibility for adding such flag on the CLI down the road. 

Expected behaviour of `renku_clone` API is that it accepts `ref=<>` keyword argument. If ref is specified it will checkout after clone given ref, and if it's `None` it will checkout default branch.

PR #1204, opened 2020-04-21
feat(datasets): specify title on dataset creation

jsam:
closes: #1148

### DIFF
--- a/renku/service/serializers/datasets.py
+++ b/renku/service/serializers/datasets.py
@@ -38,7 +38,7 @@ class DatasetDetails(Schema):
     created_at = fields.String(allow_none=True, attribute='created')
 
     title = fields.String(attribute='name')
-    creator = fields.List(fields.Nested(DatasetCreators))
+    creators = fields.List(fields.Nested(DatasetCreators), attribute='creator')
     description = fields.String()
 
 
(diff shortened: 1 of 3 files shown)


rokroskar:
How is this addressing #1176?

jsam:
Ah, sorry wrong number. Updated.

renku/service/serializers/datasets.py
```diff
@@ -38,7 +38,7 @@ class DatasetDetails(Schema):
     created_at = fields.String(allow_none=True, attribute='created')
 
     title = fields.String(attribute='name')
-    creator = fields.List(fields.Nested(DatasetCreators))
+    creators = fields.List(fields.Nested(DatasetCreators), attribute='creator')
```
Panaetius, line 41: Given that `CreatorMixin` has a list of creators, naming it `creators` instead of `creator` makes perfect sense. Could we change it on `renku.core.models.datasets.Dataset` to make things consistent=

rokroskar: it was called `creator` on purpose because that's the schema.org field (in the `Dataset` model). From an RDF pov there are no creators, just multiple values for `creator` I suppose.

jsam: Yes, I was thinking the same. Since this is a core change and I would like to avoid change core api with service PR's, I will make a separate issue for that. 👍

Panaetius: Separate issue sounds good.

PR #1208, opened 2020-04-22
feat: project clone API support for revision checkout

jsam:
closes: #1176

### DIFF
--- a/renku/core/management/clone.py
+++ b/renku/core/management/clone.py
@@ -38,6 +38,7 @@ def clone(
     progress=None,
     config=None,
     raise_git_except=False,
+    checkout_rev=None,
 ):
     """Clone Renku project repo, install Git hooks and LFS."""
     from renku.core.management.client import LocalClient
@@ -62,6 +63,13 @@ def clone(
             ) from e
 
         raise e
+
+    branches = [Path(ref.abspath).name for ref in repo.references]
+
+    if checkout_rev in branches:
+        repo.git.checkout(checkout_rev)
+    elif checkout_rev:
+        repo.git.checkout(checkout_rev, b=checkout_rev)
 
     if config:
         config_writer = repo.config_writer()
--- a/renku/core/management/clone.py
+++ b/renku/core/management/clone.py
@@ -64,9 +64,9 @@ def clone(
 
         raise e
 
-    branches = [Path(ref.abspath).name for ref in repo.references]
+    remote_refs = [Path(ref.abspath).name for ref in repo.remote().refs]
 
-    if checkout_rev in branches:
+    if checkout_rev in remote_refs:
         repo.git.checkout(checkout_rev)
     elif checkout_rev:
         repo.git.checkout(checkout_rev, b=checkout_rev)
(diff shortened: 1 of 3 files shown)


renku/core/management/clone.py
```diff
@@ -63,6 +64,13 @@ def clone(
 
         raise e
 
+    branches = [Path(ref.abspath).name for ref in repo.references]
```
mohammad-sdsc, line 67: `references` include tags as well; you should rewrite this statement as `branches = [ref.name for ref in repo.branches]` to return only branches.

tests/cli/test_integration_datasets.py
```diff
@@ -1293,14 +1293,56 @@ def test_renku_clone_with_config(tmpdir):
...
+        )
+
+        assert '3d387e64ea25079df8dd43b8875058cf9f4b0315' == str(
+            repo.active_branch
+        )
+        reader = repo.config_reader()
+        reader.values()
+
+        lfs_config = dict(reader.items('filter.lfs'))
+        assert '0' == lfs_config.get('custom')
```
mohammad-sdsc, line 1328: Are these last 4 statements required for this test?

tests/cli/test_integration_datasets.py
```diff
@@ -1293,14 +1293,56 @@ def test_renku_clone_with_config(tmpdir):
...
+    """Test cloning of a Renku repo checking out a branch."""
+    remote = 'https://dev.renku.ch/gitlab/contact/no-renku.git'
+
+    with chdir(str(tmpdir)):
+        repo = project_clone(
+            remote,
+            checkout_rev='test-branch',
+        )
+
+        assert 'test-branch' == str(repo.active_branch)
```
mohammad-sdsc, line 1343: This test can be merged with previous one by using parameters. Also, consider to add a test for a tag reference.

tests/cli/test_integration_datasets.py
```diff
@@ -1293,14 +1293,56 @@ def test_renku_clone_with_config(tmpdir):
             }
         )
 
-        repo = git.Repo('datasets-test')
+        assert 'master' == str(repo.active_branch)
```
mohammad-sdsc, line 1296: I'd rather use `repo.active_branch.name` instead of relying on `str()` function.

renku/core/management/clone.py
jsam: `repo.branches` doesn't list the remote branches as well. I've updated this line and added a test with tags so this case should be covered as well now. Thanks!

tests/cli/test_integration_datasets.py
jsam: I would like to keep it and reason being is because this is exact case how API is being used from the service perspective so having an additional check for specific ref with custom git config gives us a nice safety net for that case.

tests/cli/test_integration_datasets.py
jsam: I kept the previous test, but I've converted this test to parametric as you suggested and added a test for the tag case which has been missing. Thanks!

tests/cli/test_integration_datasets.py
jsam: Done.

Issue #1150, opened 2020-03-27
Add keywords to metadata when creating/editing datasets


mohammad-sdsc:
Datasets should have keywords which can be set when creating them and must be editable.

PR #1209, opened 2020-04-22
feat(dataset): keywords in metadata

mohammad-sdsc:
# Description

Keywords for datasets when creating/editing.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/1150

### DIFF
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -31,7 +31,8 @@ You can provide a title and a description for the dataset by using ``--title``
 and ``--description`` options when creating a dataset. You can also set
 creator of a dataset by passing a ``--creator`` option along with a string
 that defines creator's name, email, and an optional affiliation. Pass multiple
-``--creator`` flags to add a list of creators.
+``--creator`` flags to add a list of creators. To pass a list of keywords for
+a dataset use ``--keyword`` flag.
 
 Listing all datasets:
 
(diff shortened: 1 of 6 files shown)


renku/cli/dataset.py
```diff
@@ -31,7 +31,8 @@
 and ``--description`` options when creating a dataset. You can also set
 creator of a dataset by passing a ``--creator`` option along with a string
 that defines creator's name, email, and an optional affiliation. Pass multiple
-``--creator`` flags to add a list of creators.
+``--creator`` flags to add a list of creators. To pass a list of keywords for
+a dataset use ``--keyword`` flag.
```
Panaetius, line 35: This is getting a bit crowded, maybe we could use a table instead? https://www.tablesgenerator.com/text_tables to save time :)

renku/cli/dataset.py
```diff
@@ -59,8 +60,8 @@
 Editing a dataset's metadata
 
 Use ``edit`` subcommand to change metadata of a dataset. You can edit title,
-description, and list of creators of a dataset by using ``--title``,
-``--description``, and ``--creator`` flags.
+description, keywords, and list of creators of a dataset by using ``--title``,
+``--description``, ``--keyword``, and ``--creator`` flags.
```
Panaetius, line 64: Also a table?

renku/cli/dataset.py
mohammad-sdsc: Good idea!

PR #1210, opened 2020-04-23
feat(core): Adds lfs file size limit and lfs ignore file

### LINKED ISSUE #919: cli: renku run lfs policy should be configurable

`renku run` puts everything except `.ipynb` files into lfs. This makes the files cumbersome to work with and cumbersome to display in the web UI.

Improve default behavior and make use of lfs configurable. My proposal for the default behavior is:

* Files < 100kB do not go into git-lfs

I think this will greatly improve the renku experience. This behavior should be configurable. I think it will be enough to just make the cutoff file size configurable, but a further enhancement would be to allow the file type to play a role as well.

Manually remove files from LFS and rewrite history.

**Additional context**

### LINKED ISSUE #1201: renku -S docs section
At the risk of accidentally providing users new opportunities to commit really large files into their repos, it might be nice to sometimes permit, for people who know what they're doing, maybe, the option to add files to a dataset without storing them in LFS.

What problem are we trying to solve?

*Use case 1: viewing files that are small enough to view*
I want to be able to view my dataset files in the file browser, so I & others can take a look without having to start an environment, with the LFS button checked, which might add in tons of extra data & take forever to start. If these files are super large, I know I won't be able to see them, and it would be ridiculous to try to render them (and equally ridiculous to commit them to the git repo not in LFS). So, I only expect to be able to view my files that are part of a dataset in the file browser when they're smaller than some reasonable threshold.

*Use case 2: unavoidable/desirable code files in datasets*
I want to import a dataset from a 3rd party platform, and that dataset contains code. Code files are kind of supposed to live in the git tree. I want to be able to share/view the contents of the file from the renkulab UI, and also I want to be able to run the code from that dataset on other data in other datasets, and i don't want to have to do anything special to `renku storage pull dataset/code.py` every time.

*Use case 3: ability to git diff changes to dataset files*
I want to view changes to my dataset files over time.

**why? why would we do this??**
Iterating towards using LFS as its meant to be used without confounding users more than we already are.

OPTION 1:
`renku dataset add <my dataset> <my file> --no-lfs`
`renku dataset import <blah blah> --no-lfs`

OPTION 2:
`renku dataset add <my dataset> <my file> --try-no-lfs`
`renku dataset import <blah blah> --try-no-lfs`

The second option might be nicer -- if can be implemented to check the size of the total amount of files, and each file about to be added: 
* If above some threshold, print `above LFS threshhold of 100MB, adding to LFS`
* If below the threshold, print  `under LFS threshhold of 100MB, not adding to LFS`
* If the total amount of data in the import exceeds some other threshhold, print `total import size of XMB exceeds threshhold of 500MB, adding to LFS. To later exclude files from LFS, <follow these docs>`

This behavior could be default behavior, but i think we should test using it as an option first to see if it solves the problem we want to solve.

**Additional context**
Part of the fix LFS epic.

Panaetius:
Adds a configurable file size limit, below which files are not tracked in git LFS.

Adds a `.renkulfsignore` file with the same syntax as `.gitignore` that keeps matching files from being tracked in LFS

We need to add default `.renkulfsignore` files to our templates (With just `*.ipynb` for starters?)

Closes #1201 (I think?)
Closes #919


Panaetius:
I'm not really sure if/where to best document these features @emmjab

emmjab:
Great! Yeah @Panaetius it should close #1201 when there are docs!-- 

> Adds a configurable file size limit, below which files are not tracked in git LFS.

What's the syntax for this -- since I think this is a command, there should be autodoc generated.

> Adds a .renkulfsignore file with the same syntax as .gitignore that keeps matching files from being tracked in LFS

I think the renku-python docs have something about managing LFS somewhere

We also need to add this to `renku-project-template` & document this in the templates, the renku docs that talk about the templates, the renku docs that talk about LFS & in the readme of the templates for starters.

renku/core/utils/file_size.py
```diff
@@ -0,0 +1,58 @@
...
+    'pi': 1024**5,
+    'ei': 1024**6,
+    'zi': 1024**7,
+    'yi': 1024**8,
+}
+
+
+def parse_file_size(size_str):
+    """Parse a human readable filesize to bytes."""
+    res = re.search(r'([0-9.]+)([a-zA-Z]){1,2}', size_str)
```
mohammad-sdsc, line 48: `([a-zA-Z]){1,2}` should be converted to `([a-zA-Z]{1,2})` or otherwise it matches only one character.

Panaetius: Oh you're right, thank you!

Issue #1190, opened 2020-04-15
test(service): add IT for protected branch sync


jsam:
We should craft an IT where we work with remote branch which is a protected branch. In this case service will fail and we could think about some simple coping mechanism for it:

* fork the change in a branch
* push the branch to remote 
* notify user that the changes were syncing through the branch (and displayed the generated branch name)

Issue #1189, opened 2020-04-15
test(service): add IT for changing master


jsam:
If project master changes on the remote, sync with remote action will fail. We should add a test for this case and think about solutions for this problems. Some of the possible things we could do is:
* pull & retry push
* reclone (depth=1) & retry push

Panaetius:
There is a test for this now (using local remotes)

PR #1222, opened 2020-04-27
feat: protected branches support

jsam:
closes: #1190 

This PR adds a feature for handling failures to sync with remote repositories. In case of such failures, it will create a new branch (named by generating `uuid4()` ) and push it to a remote. Because of this behaviour, we add to responses of `datasets.add` and `datasets.create`. In cases where we sync with remote within jobs, this `remote_branch` key will be written inside of job details which can be accessed via `/jobs`.

Additional key in HTTP responses can be found in:

```
{
  ...
  "remote_branch": "d059c132512846e88aa642e88a1bf6e2"
  ...
}
```

or in case of a job details:

```
{
  ...
  "extras": {
      "remote_branch": "d059c132512846e88aa642e88a1bf6e2"
  }
  ...
}
```

### DIFF
--- a/renku/service/cache/models/job.py
+++ b/renku/service/cache/models/job.py
@@ -62,3 +62,10 @@ class Job(Model):
 
         self.extras['error'] = error
         self.save()
+
+    def update_extras(self, key, value):
+        """Update extras field."""
+        if not self.extras:
+            self.extras = {key: value}
+        else:
+            self.extras[key] = value
--- a/renku/service/jobs/datasets.py
+++ b/renku/service/jobs/datasets.py
@@ -89,7 +89,7 @@ def dataset_import(
             )
 
             remote_branch = repo_sync(project.abs_path)
-            user_job.update_extras = remote_branch
+            user_job.update_extras('remote_branch', remote_branch.name)
 
             user_job.complete()
         except (
@@ -124,7 +124,7 @@ def dataset_add_remote_file(
             )
 
             remote_branch = repo_sync(project.abs_path)
-            user_job.update_extras('remote_branch', remote_branch)
+            user_job.update_extras('remote_branch', remote_branch.name)
 
             user_job.complete()
     except (HTTPError, BaseException, GitCommandError) as e:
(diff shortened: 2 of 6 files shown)


rokroskar:
@vfried I guess there will need to be some new type of interaction built into the "add dataset" flow where a merge request is automatically created from the master branch and the user is presented with the option of merging it?

vfried:
Just to see if i understand correctly... in cases where the dataset.add and dataset.create fail because the branch is protected what would happen is:

Backend:
- a new branch is created 
- we receive the branch id inside /jobs
- the user should merge this branch into master 

Inside the UI:
- after the user puts "create" if we get inside /jobs "remote_branch" we should inform the user that a branch was created and that they have to merge it to see the datasets in the protected branch.

If this is correct, i have only one question... in case several actions are taken (create and then add) does one branch get created or two? I think it would be important that the create and add are seen as only one operation in case of dataset create, and in the future in case we edit the dataset metadata and on top of that files get added.

jsam:
> If this is correct, i have only one question... in case several actions are taken (create and then add) does one branch get created or two? I think it would be important that the create and add are seen as only one operation in case of dataset create, and in the future in case we edit the dataset metadata and on top of that files get added.

Two branches will be created due to the fact that the remote upstream can change (the followup ticket to this one  #1189 . If the failure occurs, user needs to resolve it and merge the resolution manually through a pull request. 

There is a followup to improve this even further - with #1260 which I have partially implemented already we should offer a user a feature to work on a specific branch if he wants to (so I assume the UI should support this functionality as well).

Also one more thing to point out, the newly created branch is not only returned in `/jobs` endpoint, but it is also returned for sync operations in the response of it. In the `/jobs` endpoint, you can see that name for async operations.

vfried:
I think for now two branches are ok, I guess if we can allow the user to choose to work in a single branch we could also send inside the "add" files the branch that we receive from the "create"operation so we could allow the use of one branch even if the user doesn't select a branch to work with.

PR #1225, opened 2020-04-28
feat(core): Adds communication util to decouple click from core

### LINKED ISSUE #737: Decouple communication between presentation layer (CLI/Service) and renku core

Panaetius:
adds `renku.core.utils.communication` namespace that can be used to return messages (e.g. `click` ones for the CLI) without having to pass things around everywhere and without any coupling/dependencies.

Closes #737 

WIP: Needs to be added everywhere, Service needs its own implementation

### DIFF
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -368,6 +368,7 @@ import click
 import requests
 from tqdm import tqdm
 
+import renku.core.utils.communication as communication
 from renku.core.commands.dataset import add_file, create_dataset, \
     dataset_remove, edit_dataset, export_dataset, file_unlink, \
     import_dataset, list_datasets, list_files, list_tags, \
@@ -602,6 +603,7 @@ def add(
 ):
     """Add data to a dataset."""
     progress = partial(progressbar, label='Adding data to dataset')
+    communication.subscribe(communication.ClickCallback())
     add_file(
         urls=urls,
         short_name=short_name,
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -367,7 +367,6 @@ from functools import partial
 import click
 import requests
 
-from renku.cli.utils import click_callback_communication
 from renku.core.commands.dataset import add_file, create_dataset, \
     dataset_remove, edit_dataset, export_dataset, file_unlink, \
     import_dataset, list_datasets, list_files, list_tags, \
@@ -595,7 +594,6 @@ def edit(short_name, title, description, creator, keyword):
 @click.option(
     '--ref', default=None, help='Add files from a specific commit/tag/branch.'
 )
-@click_callback_communication
 def add(
     short_name, urls, external, force, overwrite, create, sources, destination,
     ref
@@ -680,7 +678,6 @@ def ls_files(short_names, creators, include, exclude, format, columns):
 @click.option(
     '-y', '--yes', is_flag=True, help='Confirm unlinking of all files.'
 )
-@click_callback_communication
 def unlink(short_name, include, exclude, yes):
     """Remove matching files from a dataset."""
     file_unlink(
@@ -805,7 +802,6 @@ def export_(
 @click.option(
     '-y', '--yes', is_flag=True, help='Bypass download confirmation.'
 )
-@click_callback_communication
 def import_(uri, short_name, extract, yes):
     """Import data from a 3rd party provider or another renku project.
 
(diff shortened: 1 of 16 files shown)


jsam:
Ok, I thought about it over coffee and I think what we could do is the following. We keep this proposal under the hood for easier management and just update the API part.

So instead of the:
```
with chdir(project.abs_path):
  with service_callback_communication(user_job):
      import_dataset(...)
```

we could have something like:

```
```
with project(path) as my_project:    # this should under the hood chdir to the cwd and create project context manager
  my_project.attach_output(stdout)   # logs to the stdout as they happend (mainly for CLI usage) and it could be ommited completely
  
  result = import_dataset(...)
  if my_project.warnings:
    pass # handle warnings
   
 ```
```

Panaetius:
I think that would be a workable and clean solution :+1: Also curious about what @mohammad-sdsc and @rokroskar have to say.

rokroskar:
I like the suggestion from @jsam - the only thing we need to make sure is that the user using renku-python as an SDK shouldn't need to do much to get a sensible default (i.e. messages and errors going to stdout/stderr). I think calling `attach_output` is already too much :)

Panaetius:
We can make stdout/stderr the default if no other listener is set

jsam:
I'm not really convinced that stdout should be default since 2 out of 3 use cases would need to have it disabled from stdout.

`my_project.attach_output(stdout)` is an optional configurable parameter for the cases where continuous output makes sense (as it is in CLI), in the SDK usage and the service we don't want this since it would just pollute the standard output so having control over this in those cases makes a lot of sense to me.

Internal communications should be disabled as default, and if the user wants to peek into it it should opt-in for such behaviour.

Panaetius:
> I'm not really convinced that stdout should be default since 2 out of 3 use cases would need to have it disabled from stdout.
> 
> `my_project.attach_output(stdout)` is an optional configurable parameter for the cases where continuous output makes sense (as it is in CLI), in the SDK usage and the service we don't want this since it would just pollute the standard output so having control over this in those cases makes a lot of sense to me.
> 
> IMHO, internal communications (output to stdout) should be disabled as default, and if the user wants to peek into it (via stdout) it should opt-in for such behaviour.
> 
> The way I see it that makes sense to me is, core module should always propagate exceptions to the top-level API and not just log them internally to stderr so the user of the core module can handle them correctly and the way he wants. Now if we talk about the CLI, on the other hand, I do agree that sensible defaults of stdout and stderr should be there since that is the best practice to do so.

Right now communication is disabled by default.

I'll still add an stdout/stderr implementation so someone using it as an SDK could easily enable the output without click stuff.

This is mainly about warnings/info and I don't intend to get rid of exceptions, they should still be raised in core, as you said.

Panaetius:
I changed it a bit. The way it works right now:
- Exceptions work as before and are untouched by this, so nothing changes execution flow in that area
- If you just use `renku.core` by itself, no warnings/infos get logged anywhere. You can do `communication.subscribe(StandardOutput())` somewhere to enable logging to stdout.
- If you use the CLI, you don't have to do anything. Click communication strategy is available/injected automatically in all CLI commands.
- In the service things are a bit more difficult since we need the user_job. But as seen here https://github.com/SwissDataScienceCenter/renku-python/blob/25ae5406f7227d1198b52ca4c06f83f1399c7599/renku/service/jobs/datasets.py#L42 there's now just a `ProjectContext` that does some stuff, and if you want to do something with logged messages, you have them available in `ctx.callback.infos`, `ctx.callback.warnings` and you can do whatever you want with those.

Panaetius:
Closing in favor of https://github.com/SwissDataScienceCenter/renku-python/pull/1744

Issue #1144, opened 2020-03-26
Check LFS availability as early as possible


mohammad-sdsc:
It happened that a user added a large file to a project (which involved a large download/calculation) and at the very end received an error that LFS is not installed/enabled.

We should do LFS checks before any downloads and warn users about its unavailability as early as possible.

PR #1239, opened 2020-04-30
feat(dataset): fail early when external storage not installed

mohammad-sdsc:
# Description

Check LFS availability ASAP.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/1144


renku/core/management/datasets.py
```diff
@@ -261,6 +261,8 @@ def add_data_to_dataset(
                 f'Destination is not a directory: "{destination}"'
             )
 
+        self.check_external_storage()
```
Panaetius, line 264: Since this is part of client, we could decorate this method with `@ensure_external_storage` since that is what is used for this purpose in client already.

renku/core/models/cwl/command_line_tool.py
```diff
@@ -271,6 +271,8 @@ def generate_tool(self):
     @contextmanager
     def watch(self, client, no_output=False):
         """Watch a Renku repository for changes to detect outputs."""
+        client.check_external_storage()
```
Panaetius, line 274: in a similar vein to the one mentioned above, we could decorate `with_workflow_storage` with `@ensure_workflow_storage` to achieve the same result, then `check_external_storage` is not needed

renku/core/management/datasets.py
mohammad-sdsc: I am not sure if I understand your comment; do you mean to decorate `add_data_to_dataset`:
```
@ensure_external_storage
def add_data_to_dataset(...)
```
If so, this won't work because `ensure_external_storage` does nothing (i.e. doesn't call the decorated method) if LFS is not enabled, but we still want to be able to add data in that case.

renku/core/models/cwl/command_line_tool.py
mohammad-sdsc: Same as above.

renku/core/management/datasets.py
Panaetius: You're right, I didn't think this through 😖 

My main motivation was cutting down on things related to external storage, it's getting a bit crowded with `has_external_storage`, `ensure_external_storage`, `check_external_storage`, `use_external_storage` .

At least I know I've already almost introduced a bug into master by mixing up has_external_storage and use_external_storage in the past, so it was already a bit confusing.

Maybe we could rework it to have a single decorator with parameters that handles all the cases? Like `@check_external_storage(required=True, raise=False)` or something along those lines?

I'm also fine with merging this as is and creating a followup issue.

mohammad-sdsc: Actually, I couldn't do much here other than using less-confusing names: `use_external_storage` is a command line option and `ensure_external_storage` is a wrapper and neither of them can be merged with the rest.

I've merged `has_external_storage` and `check_external_storage` into one and renamed `use_external_storage` and `ensure_external_storage` to `external_storage_requested` and `check_external_storage_wrapper` respectively.

Issue #983, opened 2020-02-04
feat(deps): factor out dependency lists by component


jsam:
Service dependency list is getting bigger and with it the need to decouple `renku[cli]` from `renku[service]`. 

We should cleanup our dependency lists and separate dependencies by components: `core`, `cli`, `svc`. Each of the components should have it's own dependency list and way to install (per component). For example, the following should be possible:

```
pip install renku[cli]
pip install renku[svc]
```

PR #1241, opened 2020-05-04
Login to renkulab through CLI

ableuler:
Revival of #193, see the original PR for details about the current status.


jsam:
I wonder if we should tackle #983 before merging this one. The way I see it this would add additional vertical to it, which would be:

```
pip install renku[lab]
```

ableuler:
> I wonder if we should tackle #983 before merging this one

Sounds reasonable to me. This issue isn't terribly urgent.

rokroskar:
It seems like this might have to be a pretty manual merge... the underlying code is completely changed afaict.

jsam:
I think we should close this PR. Any objections against it?

rokroskar:
I think the general idea is right, but the code is so different now that a rebase would be pretty difficult.

ableuler:
I just wanted to make sure we still have @ebiiii's code around once we finally tackle this issue. I think many of the problems he has solved back then are independent of the rest of the renku-python codebase (ie the whole interaction with the browser). Closing the PR but keeping the branch around and mentioning it in #1242 is fine for me.

PR #1243, opened 2020-05-04
docs: update readme instructions for installing in editable mode

lorenzo-cavazzi:
# Description

Update the instructions included in the `README` file for installing `renku` locally in editable mode with `pipx`.

Not sure if `python setup.py bdist_wheel` is mandatory but I couldn't install without manually building first.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -173,13 +173,20 @@ Developing Renku
 ================
 
 For development it's convenient to install ``renku`` in editable mode. This is
-still possible with ``pipx``. First clone the repository and then do:
+still possible with ``pipx``. First clone the repository and run the setup:
+
+::
+
+    $ python setup.py bdist_wheel
+
+Then install with ``pipx`` in editable mode:
 
 ::
 
     $ pipx install \
         --editable \
-        <path-to-renku-python>[all] \
+        --force \
+        --spec <path-to-renku-python> \
         renku
 
 This will install all the extras for testing and debugging.


lorenzo-cavazzi:
I find it convenient using `pipx` while debugging thanks to the possibility to easily inject `ptvsd` and later attaching the debugger from tools like VScode. Not that I am a great supporter of `pipx`, I just updated the current instructions for development.

If you think we should suggest a different strategy for developing, feel free to overhaul the content of that section :slightly_smiling_face:

lorenzo-cavazzi:
btw, how it comes that the tests fail with the error `lists of files in version control and sdist do not match` when I just modified the readme? :thinking:

jsam:
> btw, how it comes that the tests fail with the error `lists of files in version control and sdist do not match` when I just modified the readme?

No idea. I checked last night a bit what's going on, but was unsuccessful. We will probably fix it today.

jsam:
Ok, rebase on #1247 should make it work again. Look's like some libraries got updated and we required a bit of adjustment.

README.rst
```diff
@@ -173,13 +173,20 @@ Developing Renku
...
+
+Then install with ``pipx`` in editable mode:
 
 ::
 
     $ pipx install \
         --editable \
-        <path-to-renku-python>[all] \
+        --force \
+        --spec <path-to-renku-python> \
```
rokroskar, line 189: This is no longer the current `pipx` syntax - if you update your pipx the original instructions will work.

README.rst
```diff
@@ -173,13 +173,20 @@ Developing Renku
 ================
 
 For development it's convenient to install ``renku`` in editable mode. This is
-still possible with ``pipx``. First clone the repository and then do:
+still possible with ``pipx``. First clone the repository and run the setup:
+
+::
+
+    $ python setup.py bdist_wheel
```
rokroskar, line 180: I'm not sure this is needed? if you install in editable mode the wheel is ignored anyway

jsam:
Sorry, I just realized that my edit on the comment was not saved. This was the edit:

> I'm missing a point, why would somebody pipx install in editable mode? Is `bdist` installation really needed in this case? Could you add a bit of context to this? I think there is a ticket for restructuring readme file completely so maybe we should sit down and brainstorm a bit on the ways things are being recommended to the users. Readme is probably first touchpoint for many so it's important that things are clear in there.

README.rst
lorenzo-cavazzi: Oh my bad, I tought I installed the latest version but I was getting it from an outdated source.

README.rst
lorenzo-cavazzi: Yeah I wasn't sure about this. I kept getting an error until I manually run the setup but it looks strange we have to do it manually :thinking:

lorenzo-cavazzi:
> I think there is a ticket for restructuring readme file completely [...]

Oh nice! Considering that @rokroskar pointed out I was updating the command to support an outdated version of `pipx` ( :facepalm: ), I think we can close this PR.

Issue #1159, opened 2020-04-02
Moving files between datasets


ciyer:
**Is your feature request related to a problem? Please describe.**
I sometimes get the structure of my datasets wrong, and I want to be able to move files to a different dataset.

**Describe the solution you'd like**
A command `renku dataset mv`

**Describe alternatives you've considered**
Moving the files in git and deleting them from the old dataset.

**Additional context**

Issue #1069, opened 2020-02-28
[CLI man] Improve messages


pameladelgado:
**Is your improvement request related to a use case or a component design? Please describe.**
The use case for the proposed improvements is based on following the first steps tutorial and playing around with Renku CLI.

**Motivation**
For users that do not read the docs and want to enthusiastically just get started, their primary source of information is the CLI man documentation, as with any command. 
With this in mind, I think we should focus in maintaining a good CLI documentation at least as much as we do for the online documentation.

**Description**
The following is a set of examples of where, by changing the CLI man message, the UX could be improved. 

* When copying the DOI link in the tutorial example the https part is included, this is not a valid syntax for renku CLI but the provider exists. The error message could indicate the syntax instead of saying that the provider is not found.
```
$ renku dataset import https://www.doi.org/10.7910/DVN/WTZS4K
Error: Invalid parameter value: Could not process https://www.doi.org/10.7910/DVN/WTZS4K.
Provider not found for https://www.doi.org/10.7910/DVN/WTZS4K. Currently supporting following providers: dataverse, zenodo
```
* In the new release there's a distinction between the Title and the Name of a dataset. In my opinion, it would be much clearer to the user to just have NAME instead of SHORT_NAME displayed all thorough the renku dataset commands.

```
$ renku dataset tag mytag -h
Usage: renku dataset tag [OPTIONS] SHORT_NAME TAG
```
what does short name means? what would be the long name? (I was confused at the beginning then I understood that the title might be a long sentence)

* While adding a large dataset on another tab and trying to add a small one:
```
$renku dataset add notebooks/filtered-flights.csv 
Ahhhhhhhh! You have found a bug.
filelock.Timeout: The file lock '/work/qa-flights-tutorial/.renku.lock' could not be acquired.
```

* What does TEXT mean here? is a revision the same as commit sha? (a hint I got after checking the documentation)
```
Options:
  --revision TEXT
```

* Would it make sense to add examples in some -the most important/common- renku command cases?

* When importing a big dataset the user does not get any feedback, the cursor is there stuck. It would be good to know that its importing or see a progress bar if pertinent.

![waiting-import](https://user-images.githubusercontent.com/43743260/75553431-aae4a100-5a38-11ea-85fc-d8c9ac4bc78b.png)

* After importing a file the progress bar disappears, this was unexpected 

* why is one tag created by renku import? called 1? does it indicate a version? why do I have a different commit sha when I add a tag?

```
[...]

Issue #787, opened 2019-10-31
enable "renku run --input <dataset-name>"


rokroskar:
Renku users would like to use the `Dataset` abstraction in other renku commands. The most obvious example is to provide the ability for something like:

```bash
renku run --input-dataset <dataset> <command>
```
In this case, renku should make the contents of `<dataset>` available to `<command>`. 

Similarly, it would be desirable to enable this behavior:

```bash
renku run --output-dataset <dataset> <command>
```
Here, outputs would automatically be added to `<dataset>`.

Panaetius:
Maybe we should do a workshop since this is probably larger than the design meeting format

Panaetius:
It makes sense to think about this after we had a chance to discuss graph building changes in https://github.com/SwissDataScienceCenter/renku-python/issues/1312

ciyer:
This looks like a duplicate of #706. Those two should be consolidated.

Panaetius:
closing this in favor of https://github.com/SwissDataScienceCenter/renku-python/issues/706

Issue #445, opened 2019-03-07
update workflow Run template steps when a file is moved


rokroskar:
**Is your feature request related to a problem? Please describe.**
`renku mv` should update the relevant workflow steps to use the new file location

**Describe alternatives you've considered**
cursing every time `renku update` fails after I move a file

rokroskar:
With #1545 this does indeed become optional.

Panaetius:
@mohammad-sdsc What's your take on this?

mohammad-sdsc:
Updating workflows is done in https://github.com/SwissDataScienceCenter/renku-python/pull/1249 for files that belong to a dataset.
I agree that we don't need to do this once #1312 is done because users can modify workflow template (or delete and create a new one). 

Actually, we don't need `renku mv` anymore once #1249 is merged.

mohammad-sdsc:
This is implemented in https://github.com/SwissDataScienceCenter/renku-python/pull/1249 but requires rebasing and incorporating new changes to workflows.

Panaetius:
Closed in favor of #711

PR #1249, opened 2020-05-05
feat(dataset): move/rename files between datasets

mohammad-sdsc:
# Description

`renku dataset mv` command to move files between datasets or rename a single file. To choose which files to move we can use `-I` or `-X` to define a pattern similar to what other `renku dataset` commands accept.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/1159

### DIFF
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -190,6 +190,31 @@ symbolic link in renku commands as a normal file. To add an external file pass
 
     $ renku dataset add my-dataset -e /path/to/external/file
 
+Moving/renaming files
+
+It is possible to move files between datasets or rename a single file.
+``renku dataset mv`` command accepts a source and a destination dataset
+and moves all files that match a specific pattern from source to destination:
+
+.. code-block:: console
+
+    $ renku dataset mv src dst -I '*csv' -X '2020*'
+
+This command moves all CSV files that don't start with 2020 from ``src``
+dataset to ``dst``. If no pattern is given then all files are moved. Note that
+move command works only with files and it moves all files to the root of the
+destination's data directory. You can pass a ``--destination`` or ``-d`` flag
+to specify a different directory as the destination.
+
+If only one file is moved and a destination is passed, then the file will be
+renamed. The following command renames ``data.csv`` to ``old-data.csv`` in the
+same dataset. Note that the destination dataset can be omitted when
+moving/renaming whithin the same dataset:
+
+.. code-block:: console
+
+    $ renku dataset mv my-dataset -I data.csv -d old-data.csv
+
 Updating a dataset:
 
 After adding files from a remote Git repository, you can check for updates in
@@ -370,7 +395,7 @@ from tqdm import tqdm
 
 from renku.core.commands.dataset import add_file, create_dataset, \
     dataset_remove, edit_dataset, export_dataset, file_unlink, \
-    import_dataset, list_datasets, list_files, list_tags, \
+    import_dataset, list_datasets, list_files, list_tags, move_files, \
     remove_dataset_tags, tag_dataset_with_client, update_datasets
 from renku.core.commands.echo import WARNING, progressbar
 from renku.core.commands.format.dataset_files import DATASET_FILES_COLUMNS, \
@@ -591,7 +616,7 @@ def edit(short_name, title, description, creator, keyword):
     '--destination',
     'destination',
     default='',
-    help='Destination file or directory within the dataset path'
+    help='Destination directory within the dataset path'
 )
 @click.option(
     '--ref', default=None, help='Add files from a specific commit/tag/branch.'
@@ -889,3 +914,44 @@ def update(short_names, creators, include, exclude, ref, delete, external):
         progress_context=progress_context
     )
     click.secho('OK', fg='green')
+
+
+@dataset.command('mv')
+@click.argument('source-dataset')
+@click.argument('target-dataset', required=False)
+@click.option(
+    '-I',
+    '--include',
+    multiple=True,
+    help='Include files matching given pattern.'
+)
+@click.option(
+    '-X',
+    '--exclude',
+    multiple=True,
+    help='Exclude files matching given pattern.'
+)
+@click.option(
+    '-d',
+    '--dst',
+    '--destination',
+    'destination',
+    default='',
+    help='Destination file or directory within the target dataset'
+)
+@click.option(
+    '-y', '--yes', is_flag=True, help='Confirm overwriting of existing files.'
+)
+def move(source_dataset, target_dataset, include, exclude, destination, yes):
+    """Move matching files between datasets or rename a single file."""
+    target_dataset = target_dataset or source_dataset
+
+    move_files(
+        source_dataset=source_dataset,
+        target_dataset=target_dataset,
+        include=include,
+        exclude=exclude,
+        destination=destination,
+        yes=yes,
+        interactive=True
+    )
--- a/renku/core/management/datasets.py
+++ b/renku/core/management/datasets.py
@@ -1165,6 +1165,9 @@ class DatasetsApiMixin(object):
             used_files = dataset.find_files(sources)
             files_to_keep.update(used_files)
 
+        if self.storage_installed_locally:
+            self.pull_paths_from_storage(*[str(s) for s in sources])
+
         for src, dst in files:
             dst.parent.mkdir(parents=True, exist_ok=True)
 
@@ -1216,7 +1219,7 @@ class DatasetsApiMixin(object):
             self._replace_paths_in_workflows(files, commit=commit)
 
     def _replace_paths_in_workflows(self, files, commit):
-        """Replace file paths in dataflows after moving files."""
+        """Replace file paths in workflows after moving files."""
 
         def replace_entity_path(entity):
             replaced = False
@@ -1258,8 +1261,10 @@ class DatasetsApiMixin(object):
                 replaced |= replace_entity_path(entity)
 
             for generation in process.generated or []:
+                self.remove_from_activity_index(process, persist=False)
                 entity = generation.entity
                 replaced |= replace_entity_path(entity)
+                self.add_to_activity_index(process, persist=False)
 
             subprocesses = getattr(process, '_processes', [])
             for p in subprocesses:
@@ -1270,13 +1275,7 @@ class DatasetsApiMixin(object):
 
             plan = process.association.plan
 
-            for output in plan.outputs or []:
-                entity = output.produces
-                replaced |= replace_entity_path(entity)
-
-            for input_ in plan.inputs or []:
-                entity = input_.consumes
-                replaced |= replace_entity_path(entity)
+            replaced |= replace_paths_in_run(plan)
 
             subprocesses = getattr(plan, 'subprocesses', [])
             for p in subprocesses:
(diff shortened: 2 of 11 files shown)


emmjab:
Thanks @mohammad-sdsc ! It works as I would expect (e.g. similar to the familiar `mv` command, except dataset must exist first, and great error message when it doesn't already)


help text & commit messages look good: 

```
$ ▶ renku dataset mv --help
Usage: renku dataset mv [OPTIONS] SOURCE_DATASET TARGET_DATASET

  Move matching files from a dataset to another.

Options:
  -I, --include TEXT             Include files matching given pattern.
  -X, --exclude TEXT             Exclude files matching given pattern.
  -d, --dst, --destination TEXT  Destination file or directory within the
                                 target dataset

  -y, --yes                      Confirm overwriting of existing files.
  -h, --help                     Show this message and exit.
```
```
commit 09d8f3ff594c131ffefcd9f2a2f1a5cd6b1dd56f
Author: Emma Jablonski <e.jablonski@gmail.com>
Date:   Tue May 5 17:37:45 2020 +0000

    renku dataset: committing 1 moved files

commit 0130723636a94de2102574e8c277aea097a981b2
Author: Emma Jablonski <e.jablonski@gmail.com>
Date:   Tue May 5 17:37:07 2020 +0000

    renku dataset mv penguins new_penguins

commit 259a5e72bce1383f5dab6a859cc8ac1bfdedcbb0
Author: Emma Jablonski <e.jablonski@gmail.com>
Date:   Tue May 5 17:37:07 2020 +0000

    renku dataset: committing 1 moved files

commit d6cd6e053538b6a9df058bd1368bec3e29c69418
Author: Emma Jablonski <e.jablonski@gmail.com>
Date:   Tue May 5 17:36:41 2020 +0000

    renku dataset create new_penguins
```

```
$ ▶ renku dataset mv penguins doesnt_exist_yet
Error: Dataset "doesnt_exist_yet" is not found.
$ ▶ renku dataset mv doesnt_exist_yet penguins
Error: Dataset "doesnt_exist_yet" is not found.
```

And when no files are in the dataset, can't move them, helpful error message: 

```
$ ▶ renku dataset mv new_penguins penguins
Error: Invalid parameter value - No files found.
```

And also the files themselves get juggled between the `data` subdirs -- didn't actually expect this to happen but 👍 

```
$ ▶ ls data/penguins/
emperor.jpg
$ ▶ renku dataset mv penguins new_penguins
$ ▶ ls data/new_penguins/
emperor.jpg
```

If I had added an image to a dataset originally and that file was not in a subdirectory of `data`, if i move it to a new dataset, it gets put into the `data` subdirectory:

```
$ ▶ renku dataset mv penguins -I penguin2.jpg new_penguins
$ ▶ ls data/new_penguins/
emperor.jpg  penguin2.jpg
```

So if i want to put a file that's hanging around somewhere into a `data` subdirectory, can i call mv? Answer:yes.

```
$ ▶ renku dataset mv penguins -I chinstrap.jpg penguins
$ ▶ ls data/penguins/
chinstrap.jpg
```

I think this is all reasonable, but it's a little confusing that `renku dataset add` doesn't put the file into the `data` directory if it exists already in the repo, where `renku dataset mv` does. I think in general the behavior of `renku dataset mv`, moving actual files along with the metadata is more aligned with what new users expect.

mohammad-sdsc:
Thanks @emmjab and @Panaetius for the review!

We've decided to copy all data files to a dataset's data directory when adding files. I've created an issue for this: https://github.com/SwissDataScienceCenter/renku-python/issues/1259. The behavior will be the same for `add` and `mv` after this issue is done.

> Maybe since you can also rename files with this command, change to:

I updated the wording for this function and added the behavior when there is no argument (i.e. moving all) to the dataset documentation.

renku/cli/dataset.py
```diff
@@ -883,3 +907,42 @@ def update(short_names, creators, include, exclude, ref, delete, external):
         progress_context=progress_context
     )
     click.secho('OK', fg='green')
+
+
+@dataset.command('mv')
```
ciyer, line 912: I wonder if it isn't better to name the command `mv-files`? This would be more consistent with the commands like `rm`, which operates on datasets (but not files).

I tried to rename a dataset by `mv`, but this didn't work. If the command were `mv-files`, it would be clear that I couldn't rename the dataset using mv.

renku/cli/dataset.py
```diff
@@ -179,6 +179,30 @@
...
+directory. You can pass a ``--destination`` or ``-d`` flag to specify a
+different directory as the destination.
+
+If only one file is moved and a destination is passed, then the file will be
+renamed. The following command renames ``data.csv`` to ``old-data.csv`` in the
+same dataset:
+
+.. code-block:: console
+
+    $ renku dataset mv my-dataset my-dataset -I data.csv -d old-data.csv
```
ciyer, line 204: This is a helpful example because it's a little counterintuitive that to rename a file you have to pass in the same dataset as src and dst. Could you add this to the text that is shown with `renku dataset mv --help`?

renku/cli/dataset.py
```diff
@@ -883,3 +907,42 @@ def update(short_names, creators, include, exclude, ref, delete, external):
...
+
+
+@dataset.command('mv')
+@click.argument('source-dataset')
+@click.argument('target-dataset')
+@click.option(
+    '-I',
+    '--include',
+    multiple=True,
+    help='Include files matching given pattern.'
```
ciyer, line 919: An example in the help text would be nice. You could show that the pattern needs to be  quoted to prevent shell expansion.

renku/cli/dataset.py
mohammad-sdsc: We are trying (mostly) to follow git commands names; for example we also have `renku dataset add/unlink` that also work on files. I am not sure if `mv-files` helps here because we normally don't move datasets and can only rename them (renaming is not supported yet BTW).
The main point here is that we use the same `dataset` verb to act on files and datasets. Maybe we should resume the discussion regarding this to separate them (i.e. `renku file` vs `renku dataset`).

renku/cli/dataset.py
mohammad-sdsc: I've fixed this; now, if one doesn't specify a destination dataset then the source dataset is used as destination.

renku/cli/dataset.py
mohammad-sdsc: So far the convention was to keep these help messages short. An example is already in the dataset documentation. We also have an issue to improve UX regarding help in commands: https://github.com/SwissDataScienceCenter/renku-python/issues/1069.

mohammad-sdsc:
Update: The last commit allows omitting destination dataset where in that case source is used as destination. This prevents specifying the same dataset name twice when renaming. Also, if source and destination files are the same an error is raised.

mohammad-sdsc:
I've changed the wording according to your suggestion.

> Also occurred to me that I didn't test what happens when you try to renku update after moving a dataset that was a participant in a renku run. 

The update command won't work if a file is moved because it uses file's path to find a file. There won't be a trivial fix for that and we should come up with a solution which does not use file paths or one that uses datasets (e.g. #787).

rokroskar:
> The update command won't work if a file is moved because it uses file's path to find a file. There won't be a trivial fix for that and we should come up with a solution which does not use file paths or one that uses datasets (e.g. #787).

but then this effectively breaks the project, no? I think we need to find some sort of solution.

mohammad-sdsc:
> this effectively breaks the project, no? I think we need to find some sort of solution.

It does if the files are used on some workflows similar to when a file or dataset is removed. To me, a proper solution is to decouple workflows from file paths by assigning fixed IDs to files and refer to them with those IDs independent from their location/name/dataset/etc. #787 would be a good start in this direction.
We can also go through all CWL files and replace old path with new ones (and I guess commit them one-by-one) for the sake of this story. Honestly, I believe this is a patchy solution and is not worth the effort; we can just make it clear in the docs that if users move a file some workflows might break.
Let me know what do you think and/or if you have other solutions in mind.

ciyer:
> It does if the files are used on some workflows similar to when a file or dataset is removed. To me, a proper solution is to decouple workflows from file paths by assigning fixed IDs to files and refer to them with those IDs independent from their location/name/dataset/etc. #787 would be a good start in this direction.

I think using git content hashes to identify files is a good idea in the metadata stores, because the metadata apply to the content, the path is not so important, but I think paths make sense in workflows. For example, if you run a script as a workflow and then fix a bug in the script (changing its content hash), you would still want the workflow to run.

> We can also go through all CWL files and replace old path with new ones (and I guess commit them one-by-one) for the sake of this story. Honestly, I believe this is a patchy solution and is not worth the effort; we can just make it clear in the docs that if users move a file some workflows might break.

I agree that doing this automatically for the user is a bad idea. Is this feasible: If a user moves a file, we check if the file is used in any workflows, if so, we ask the user if they want to update the workflows as well?

renku/core/management/datasets.py
```diff
@@ -1209,6 +1211,99 @@ def move_files(self, source_dataset, destination_dataset, files):
...
 
+        if staged_files:
+            files = {
+                str(s.relative_to(self.path)): str(d.relative_to(self.path))
+                for s, d in files
+            }
+            self._replace_paths_in_workflows(files, commit=commit)
+
+    def _replace_paths_in_workflows(self, files, commit):
+        """Replace file paths in dataflows after moving files."""
```
Panaetius, line 1222: `s/dataflows/workflows/`?

renku/core/management/datasets.py
```diff
@@ -1209,6 +1211,99 @@ def move_files(self, source_dataset, destination_dataset, files):
...
+            subprocesses = getattr(process, '_processes', [])
+            for p in subprocesses:
+                replaced |= replace_paths_in_process(p)
+
+            if not process.association or not process.association.plan:
+                return replaced
+
+            plan = process.association.plan
+
+            for output in plan.outputs or []:
```
Panaetius, line 1276: Couldn't you use `replace_paths_in_run` here?

renku/core/management/datasets.py
```diff
@@ -1209,6 +1211,99 @@ def move_files(self, source_dataset, destination_dataset, files):
...
+
+            return replaced
+
+        for path in self.workflow_path.glob('*.yaml'):
+            workflow = WorkflowRun.from_yaml(path, client=self)
+
+            replaced = replace_paths_in_process(workflow)
+
+            if replaced:
+                workflow.to_yaml(path=path)
```
Panaetius, line 1296: you need to add the new paths to the activity index https://github.com/SwissDataScienceCenter/renku-python/blob/master/renku/core/management/repository.py#L486 which is used to find workflows that were involved in a path, and probably remove the old path entries from there as well.

renku/core/models/jsonld.py
```diff
@@ -632,10 +632,11 @@ def asjsonld(self):
         source.update(asjsonld(self))
         return source
 
-    def to_yaml(self):
+    def to_yaml(self, path=None):
         """Store an instance to the referenced YAML file."""
         jsonld_ = self.asjsonld()
-        write_yaml(path=self.__reference__, data=jsonld_)
+        path = path or self.__reference__
```
Panaetius, line 638: So far the intended use of this was to wrapp calls to `to_yaml` into `with with_reference():`. I'm not sure if that has any benefit over supplying the path directly (It does some check to make sure the reference wasn't changed). I think we should update to this new version everywhere that is currently wrapped in `with_reference` to make the code more readable.

mohammad-sdsc: I've created an issue for this: https://github.com/SwissDataScienceCenter/renku-python/issues/1361.

mohammad-sdsc:
A similar functionality with easier UX is implemented in https://github.com/SwissDataScienceCenter/renku-python/pull/1993. The other PR does not update workflow files; followup story: https://github.com/SwissDataScienceCenter/renku-python/issues/445.

PR #1254, opened 2020-05-06
feat(service): read template manifest

lorenzo-cavazzi:
# Description

Adds to the core service the endpoint `/renku/templates.read_manifest` to read the manifest file from a template repository (the equivalent of `renku init -l`).
It requires the query parameter `url` with the template repository url (`renku init -s`) and it's possible to pass `ref` for the target reference (`renku init -r`)

It's the first step in addressing #862, specifically it implements the `projects.template.options` endpoint suggested in the initial post (or the combination of `template.read_variables` and `template.read_manifest` from the second post).
I prefer an early feedback before implementing the `projects.create` endpoint.

**How to test it**
To manually test it, the reference docker image is `lorenzocavazzitech/renku-core:0.10.4-d93cda2`  currently deployed in https://lorenzo.dev.renku.ch/
You can query it with Postman at https://lorenzo.dev.renku.ch/api/renku/templates.read_manifest
It requires the following headers
```
x-requested-with: XMLHttpRequest
Content-Type: application/json
Cookie: session=<your-valid-session-cookie-here>
```
and at least the `url` query parameter (`ref` is the other optional)
```
url: https://github.com/SwissDataScienceCenter/renku-project-template
```


renku/service/views/cache.py
```diff
@@ -180,12 +180,13 @@ def _project_clone(cache, user_data, project_data):
     project_clone(
         project_data['url_with_auth'],
         local_path,
-        depth=project_data['depth'],
+        depth=project_data['depth'] if project_data['depth'] != 0 else None,
```
jsam, line 183:
```suggestion
        depth=project_data['depth'] if project_data.get('depth') != 0 else None,
```

PR #1268, opened 2020-05-12
[WIP]Make test fail on purpose

Panaetius:
Makes test fail on purpose to test flaky rerun issues


Panaetius:
Close as it's not needed anymore/the issue could be reproduced

PR #1272, opened 2020-05-13
feat: prompt for missing variables on renku init

### LINKED ISSUE #1126: Read variables list when using renku init

lorenzo-cavazzi:
# Description
The parameters required by the template selected during `renku init` were defaulted to `""` if not provided with the command through the `--parameter` options. This led to a bad UX, especially when the user selected a template after reading the manifest.

This PR implements a comparison between the provided parameters and the required ones. If they don't match:
* Missing parameters will be prompted.
* Useless parameters will be removed and a notification will be printed.

Fixes #1126


renku/cli/init.py
```diff
@@ -388,6 +389,30 @@ def init(
...
+                f'({template_variables[key]})'
+            ),
+            default='',
+            show_default=False
+        )
+        parameter[key] = value
+    useless_variables = input_parameters_keys - template_variables_keys
+    if (len(useless_variables) > 0):
+        click.echo(
+            'These parameters are not required by the template: {}'.format(
```
Panaetius, line 409:
```suggestion
            INFO + 'These parameters are not used by the template and were ignored:\n\t {}'.format(
```

together with `from renku.core.commands.echo import INFO`

renku/cli/init.py
```diff
@@ -388,6 +389,30 @@ def init(
...
+                ', '.join(useless_variables)
```
Panaetius, line 410:
```suggestion
                '\n\t'.join(useless_variables)
```

renku/cli/init.py
lorenzo-cavazzi: oh nice! I didn't know about `INFO`. Thanks!

renku/cli/init.py
lorenzo-cavazzi: done

PR #1273, opened 2020-05-14
feat(cli): Adds renku save command

### LINKED ISSUE #1227: renku save

opening on behalf of @cchoirat !

Users have trouble with the add-commit-push loop (but especially the push part) when working in interactive environments to save their work:
* because they're new to git
* because working with git locally just requires committing to "save", and pushing is something you can do as an afterthought when you want to share your stuff later
* because it's not clear how the interactive environments and gitlab and renkulab are connected (which parts are magic and which parts have to be user-managed)
* because so many programs nowadays just "autosave" (.. including renku :P)
* because renku commands add & commit without pushing, but you get no feedback that that's what happened and that there's more you need to do to not lose your work.

It's important to make it easy for users to save their work back to renkulab from the interactive environment because if the environment is killed they lose everything.

`renku save`, which (verbosely) adds, commits, and pushes.
`renku save [-m opt. commit msg, default timestamp] [-d destination, default renkulab]`

@rokroskar pointed out that `save` should make sure to add files above the `renku config` threshold size to `.gitattributes` so that they get stored in gitLFS (unless ignored by `.renkulfsignore`)

We should also make sure something reasonable happens in an anonymous interactive environment, e.g. print the steps: to make an account & configure ssh & add the remote & push (?) & also a link to docs that say the same thing.

I think someone just opened another related issue that I'll find and link here, for printing a reminder message to push after commits have been made.
It was this: https://github.com/SwissDataScienceCenter/renkulab-docker/issues/102

**Additional context**
@cchoirat made similar functionality for R and suggested this `renku save` as a feature for the CLI too.
One specific use case is for teaching, where git is in the way of the lesson.

Panaetius:
Adds a renku save command of the form `renku save [-m <message>] [-d remote] [<paths*>]`

Also sets the git `credential.helper` to `cache` so password the user types are stored in memory for a while (15 minutes by default, but we can change it to whatever we like), if no `credential.helper` was defined before.

Closes #1227


emmjab:
Thanks @Panaetius !

I tried it out and mostly think it's great! I have some comments about the default behavior.

First, the helptext:

```
$ ▶ renku save --help
Usage: renku save [OPTIONS] [PATHS]...

  Save and push local changes.

Options:
  -m, --message TEXT      The commit message to use
  -d, --destination TEXT  The git remote to push to
  -h, --help              Show this message and exit.
```
(0) I think this is pretty clear, but it would be nice to know that there is/what is the default destination, though maybe it's hard to concisely say this is already configured for interactive environments.

I added a new file, then called `renku save`.

```
$ ▶ renku save
Successfully saved: 
        notebooks/test_random.ipynb
OK
```

(1) I like the messaging! I know which file was added. I also checked that it was pushed 👍 

However, checking the commit messages after a few of these:

```
commit f734f61b1bae18f146dafde663840fd77fb97ded (HEAD -> master, origin/master, origin/HEAD)
Author: Emma Jablonski <e.jablonski@gmail.com>
Date:   Mon May 18 10:07:21 2020 +0000

    2020.05.18 10:07:21

commit a54a9b5dd2cb34d82772a18fe2eb39fe328f4def
Author: Emma Jablonski <e.jablonski@gmail.com>
Date:   Mon May 18 10:06:10 2020 +0000

    2020.05.18 10:06:10

commit 650151074aff1779b9eb05263a7339fd35a9faa4
Author: Emma Jablonski <e.jablonski@gmail.com>
Date:   Mon May 18 10:03:50 2020 +0000

    2020.05.18 10:03:49

commit 22211308f10c910900bf2907bb4209430f2fb986
Author: Emma Jablonski <e.jablonski@gmail.com>
Date:   Mon May 18 09:53:57 2020 +0000

    2020.05.18 09:53:57
```

(2) Maybe the default could also list the files that were "saved", like what the output to the `renku save` command shows? (I know user can specify a message, but for the default, the datetime is already recorded in the commit)

Panaetius:
Description of `-d` is a bit more verbose now.
Commit message is now `Saved changes to: file1 file2 ...`, limited to 100 characters on the first line, with the rest of files in the following lines as a list, like 
```
Saved changes to: file1 file2 file3
    file4
    file5
    file6
```
Also improved the test to check that commits were made.

renku/core/commands/save.py
```diff
@@ -0,0 +1,95 @@
...
+
+import git
+
+from renku.core import errors
+
+from .client import pass_local_client
+
+
+@pass_local_client
+def save_and_push(client, message=None, remote=None, paths=None):
```
jsam, line 31: This function has several blocks, could we add a comment for each block with short description. I think it would make things a bit easier to read.

Panaetius:
I unified it with the service `repo_sync`, let me know what you think.

PR #1280, opened 2020-05-18
chore: prep release v0.10.4

jsam:
Preparation for release v0.10.4.


rokroskar:
I think this needs to be updated for #1254 ?

Issue #1252, opened 2020-05-06
Use the calamus library for Dataset entities


rokroskar:
We now have a JSON-LD library [calamus](https://github.com/SwissDataScienceCenter/calamus) - we should use it for the (de)serialization (from)to JSON-LD and a good place to start is the Datasets. 

This should be done after https://github.com/SwissDataScienceCenter/renku-python/pull/1142 is completed.

PR #1281, opened 2020-05-19
feat(core): Calamus integration

mohammad-sdsc:
# Description

This is an initial integration of Calamus with Renku. It does not remove any existing `jsonld` constructs to achieve minimum change in the code. 

Since some changes on Calamus' side are required, it's dependency is on a remote branch (and not a PyPI released version).

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/1252


renku/core/models/datasets.py
```diff
@@ -615,6 +617,130 @@ def __attrs_post_init__(self):
                 self.name, self.version
             )
 
+    @classmethod
+    def from_yaml(cls, path, client=None, commit=None):
```
Panaetius, line 621: General question, do we still want to use `.yaml` or `.json`? There shouldn't be any performance difference, `.json` is probably better supported with external tools (e.g. you can just copy-paste the content to jsonld playground), if we ever switch to a DB approach (i.e. a key-value store) it'd be json anyways.

renku/core/models/jsonld.py
```diff
@@ -553,6 +553,8 @@ def from_jsonld(
     @classmethod
     def from_yaml(cls, path, client=None, commit=None):
         """Return an instance from a YAML file."""
+        assert False, 'this should not be called.'
```
rokroskar, line 556: isn't it still needed for the classes that have not been migrated to calamus?

renku/core/models/projects.py
```diff
@@ -69,6 +71,8 @@ class Project(object):
 
     client = attr.ib(default=None, kw_only=True)
 
+    commit = attr.ib(default=None, kw_only=True)
```
rokroskar, line 74: why is this needed?

renku/core/models/projects.py
```diff
@@ -140,6 +144,25 @@ def project_id(self):
...
+
+        extra = dict(client=client, commit=commit)
+        data.update(extra)
+        self = ProjectSchema().load(data, unknown=INCLUDE)
+        self.__reference__ = path
+
+        return self
+
+    def to_yaml(self):
+        """Store an instance to the referenced YAML file."""
```
rokroskar, line 162:
```suggestion
        """Write an instance to the referenced YAML file."""
```

renku/core/models/projects.py
```diff
@@ -183,3 +206,28 @@ def __iter__(self):
...
+
+class ProjectSchema(JsonLDSchema):
+    """Project Schema."""
+
+    class Meta:
+        """Meta class."""
+
+        rdf_type = [schema.Project, prov.Location]
+        model = Project
+        translate = {
```
rokroskar, line 223: does calamus actually use the `translate` fields?

renku/core/models/datasets.py
mohammad-sdsc: Will be addressed in https://github.com/SwissDataScienceCenter/renku-python/issues/1299.

renku/core/models/jsonld.py
mohammad-sdsc: No. Dataset and Project are the only classes that use this method for serialisation.

renku/core/models/projects.py
mohammad-sdsc: It was in the signature of `from_yaml(cls, path, client=None, commit=None)` and I've added it here to avoid a validation error. Since `Project` has its own `from_yaml` now I'll remove this from the signature and here.

renku/core/models/projects.py
mohammad-sdsc: It's not really needed and will be removed.

setup.py
```diff
@@ -143,10 +143,15 @@ def run(self):
...
 
+dependency_links = [
+    'git+https://github.com/SwissDataScienceCenter/calamus.git@000-renku-integration#egg=calamus',
+]
+
 install_requires = [
     'appdirs>=1.4.3',
     'apispec==3.3.1',
     'attrs>=19.3.0',
+    'calamus @ git+https://github.com/SwissDataScienceCenter/calamus.git@000-renku-integration#egg=calamus',
```
rokroskar, line 154: I guess the other PR should be merged before this one

renku/data/shacl_shape.json
```diff
@@ -529,6 +529,13 @@
                "datatype": {
                   "@id": "xsd:string"
                }
+            },
+            {
+               "nodeKind": "sh:Literal",
+               "path": "sdsc:external",
```
Panaetius: We should really add this to the `renku:` ontology in https://github.com/SwissDataScienceCenter/renku-ontology/tree/gh-pages

renku/core/models/datasets.py
```diff
@@ -615,6 +617,130 @@ def __attrs_post_init__(self):
...
+        """Return an instance from a YAML file."""
+        from marshmallow import INCLUDE
+
+        data = jsonld.read_yaml(path)
+
+        extra = dict(client=client, commit=commit)
+        data.update(extra)
+        self = DatasetSchema().load(data, unknown=INCLUDE)
```
Panaetius: For the future, it'd be possible to subclass the calamus JsonldSchema so it supports `client` and `commit` as parameters and passes them on to children if they are set (saving us quite a bit of headache). What do you think? We can do a followup issue.

renku/core/models/datasets.py
```diff
@@ -744,6 +727,39 @@ class Meta:
     same_as = fields.Nested(schema.sameAs, UrlSchema, missing=None)
     short_name = fields.String(schema.alternateName)
 
+    @pre_load
+    def fix_files_context(self, data, **kwargs):
```
Panaetius, line 731: Is this needed all the time or just initially for a migration? If the latter, could we add the `pre_load` code inside a relevant migration (copy/subclass the schema there) instead of here?

renku/core/models/projects.py
```diff
@@ -220,14 +203,32 @@ class Meta:
...
+
+    name = fields.String(schema.name, missing=None)
+    created = fields.DateTime(schema.dateCreated, missing=None)
+    updated = fields.DateTime(schema.dateUpdated, missing=None)
+    version = fields.String(schema.schemaVersion, missing=1)
+    creator = fields.Nested(schema.creator, PersonSchema, missing=None)
+    _id = fields.Id(init_name='id', missing=None)
+
+    @pre_load
+    def translate_old_project(self, data, **kwargs):
```
Panaetius, line 216: Same as above, can we use a schema inside a migration to do this or would this break things?

renku/core/models/datasets.py
mohammad-sdsc: That's a good idea! I've done it.

renku/data/shacl_shape.json
mohammad-sdsc: I've already fixed this, I guess you looked at an older commit.

renku/core/models/datasets.py
mohammad-sdsc: Actually, this one is not easy to get rid of; it is required for all project versions (2 ATM). Ideally we should refactor migrations to be independent from the core models so that these won't be required. I've created a followup story: https://github.com/SwissDataScienceCenter/renku-python/issues/1340.

renku/core/models/projects.py
mohammad-sdsc: This is not required by the migrations themselves but by the code that check the project version (which infinitely tries to create project causing an SO). I've changed that code and removed this from here.

PR #1282, opened 2020-05-19
test: fix add existing files

jsam:


tests/cli/test_datasets.py
```diff
@@ -1651,7 +1651,9 @@ def test_add_existing_files(
     assert 0 == result.exit_code
     assert 'These existing files were not overwritten' in result.output
     assert str(path) in result.output
-    assert 'Warning: No file was added to project' not in result.output
+    
+    if not external:
+        assert 'Warning: No file was added to project' not in result.output
```
mohammad-sdsc, line 1656: This should not fix the issue which seems to be related to LFS. It's better to remove this line altogether.

jsam: Yea, I'm just testing it and trying to narrow the issue down.

PR #1287, opened 2020-05-25
feat(service): create new projects from templates

### LINKED ISSUE #862: feat: service endpoint for project creation

lorenzo-cavazzi:
Adds the full `renku init` functionality to the service. It always requires a remote repository
containing valid template since there is no locally cached version. Fix #862

Please note that the user provided in the fixture `authentication_headers` must be able to create projects for the tests to succeed.

**How to test**
A temporary preview is deployed here: https://lorenzo.dev.renku.ch (service API reachable at `/api/renku/`)

You can use Postman to invoke the APIs. Always include the following headers
```
x-requested-with: XMLHttpRequest
Content-Type: application/json
Cookie: session=<your-valid-session-cookie-here>
```

READ THE MANIFEST
You can check the available templates.
```
GET https://lorenzo.dev.renku.ch/api/renku/templates.read_manifest?url=https://github.com/SwissDataScienceCenter/renku-project-template
```

CREATE PROJECT
Create a project by providing the proper data in the body.
```
POST https://lorenzo.dev.renku.ch/api/renku/templates.create_project
```

Assume you want to create a new project from the python template, here is a stub of the body (adjust `namespace` and `name`)
```
{
	"url": "https://github.com/SwissDataScienceCenter/renku-project-template",
	"ref": "0.1.10",
	"identifier": "python-minimal",
	"parameters": [{
		"key": "description",
		"value": "put your description here"
	}],
	"project_name": "test from postman",
	"project_namespace": "your namespace here",
	"project_repository": "https://dev.renku.ch/gitlab"
}
```


lorenzo-cavazzi:
question: should I update any service-specific documentation anywhere?

jsam:
There are diagrams for every endpoint in renku repository. You can open a ticket for updating them there or feel free to update them. All we have to do here is make sure that we register the endpoint into the docs generator (but I see you already did that so we are good). :taco:

lorenzo-cavazzi:
diagrams created in SwissDataScienceCenter/renku#1121

renku/service/utils/__init__.py
```diff
@@ -74,3 +84,13 @@ def repo_sync(repo_path, remote_names=('origin', )):
                 repo.git.push(remote.name, repo.active_branch)
 
     return pushed_branch
+
+
+def new_repo_push(
+    repo_path, source_url, source_name='origin', source_branch='master'
```
Panaetius, line 90: If https://github.com/SwissDataScienceCenter/renku-python/pull/1273 gets merged first, this should be updated to also use the `renku.core.commands.save.repo_sync` introduced there

renku/service/views/templates.py
```diff
@@ -65,3 +75,84 @@ def read_manifest_from_template(user, cache):
...
+        for template in templates if template['folder'] == ctx['identifier']
+    ), None)
+    if template is None:
+        return error_response(
+            INVALID_PARAMS_ERROR_CODE,
+            'invalid identifier for target repository'
+        )
+
+    # Verify missing parameters
+    template_parameters = template.get('variables', {})
```
Panaetius, line 122: in https://github.com/SwissDataScienceCenter/renku-python/pull/1272 this is cone in `renku/cli/init.py`, it'd be nice to have a single method for missing/ignored parameter check that can be used both in the service and the cli.

renku/service/views/templates.py
```diff
@@ -65,3 +75,84 @@ def read_manifest_from_template(user, cache):
...
+    new_project_path = make_new_project_path(user, ctx)
+    if new_project_path.exists():
+        shutil.rmtree(str(new_project_path))
+    new_project_path.mkdir(parents=True, exist_ok=True)
+
+    # prepare data and init new project
+    source_path = template_project.abs_path / ctx['identifier']
+    parameters_dict = {}
+    for parameter in ctx['parameters']:
+        parameters_dict[parameter['key']] = parameter['value']
```
Panaetius, line 144:
```suggestion
    parameters_dict = {p['key']: p['value'] for p in ctx['parameters']}
```

renku/service/utils/__init__.py
lorenzo-cavazzi: On nice. In that case, I should still set up a remote for a new project, right? Maybe that step could still go in this function where I invoke `repo_sync`

renku/service/views/templates.py
lorenzo-cavazzi: good point! Now that #1272 is merged, I have no excuses for not doing this :slightly_smiling_face:

lorenzo-cavazzi: I double-checked the code, and I think it makes sense to keep them separate (I slightly updated it as you suggested below). The logic is similar, but the input parameters are formatted in a different way (`{ key: <keyname>, value: <value> }` for API vs `{ <keyname>: <value> }` for cli) and missing one triggers an error in the API but does something different in the cli, so the common code is very limited.

I can still create a single method if you think that helps in reading the code better.

renku/service/utils/__init__.py
Panaetius: Yes that makes sense 👍  (Sorry for the late reply...)

lorenzo-cavazzi: oh I realize only now that I haven't addressed this :facepalm: 
I'll give it a look right now

renku/service/views/templates.py
lorenzo-cavazzi: Ok I tried to use the `repo_sync` function but I had to change a detail since it's not possible to fetch from a non-existing repo before pushing. I hope that change makes sense.

Issue #910, opened 2020-01-16
svc: enable chunked uploads for large files


jsam:
To make service uploads more robust (not allocating huge chunks of memory) and faster (by using more web workers for uploading chunks) for really big files, we should consider implementing chunked uploads. 

For inspiration on how to achieve this, we could look into existing chunking strategy:
  * [AWS S3 API](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingRESTAPImpUpload.html) 
  * [dropzone.js](https://codecalamity.com/uploading-large-files-by-chunking-featuring-python-flask-and-dropzone-js/)

PR #1288, opened 2020-05-25
feat(service): chunked upload API support

jsam:
closes: #910 

This PR adds new endpoint `/cache.chunked_upload` for file uploads which is much lower level than existing file upload and gives much more control to the client. It implements the API which is fully compatible with dropzone.js library and could be used with it out of the box.

[WIP]:

* uploading chunks in parallel
* finish testing with dropzone.js
* streamline compatibility with existing `/cache.files_upload`
* tests for parallel chunk upload


Panaetius:
Pretty out of date, we'll look at it in a new PR when the time comes

Issue #1219, opened 2020-04-27
chore: improve UX on project initialization


jsam:
Currently, I find `renku init` user experience quite clunky and very hard to read. It would be great if we could improve that UX so it would be a bit more readable.

Here is how it looks at the moment:
![image](https://user-images.githubusercontent.com/1253355/80374753-76099480-8897-11ea-8091-57a2cf7582ef.png)

I would expect less text in the choice - if the text is really needed for some, we could create docs on the web explaining each template or use external `README.md` for such purpose.

rokroskar:
I agree, this is definitely too verbose atm - I would probably only want to see "Basic Python Project" etc. and provide an option to give details if necessary.

PR #1294, opened 2020-05-27
chore(core): small UX cleanup on init of a project

jsam:
closes: #1219


tests/cli/test_init.py
```diff
@@ -81,14 +81,12 @@ def test_template_selection_helpers():
...
-    assert (
-        '1 folder_python Template Python: Description Python'
-    ) in stripped_sentence
-    assert '2 folder_R Template R: Description R' in stripped_sentence
+
+    assert '1 folder_python' in stripped_sentence
+    assert '2 folder_R' in stripped_sentence
     assert instructions not in stripped_sentence
-    full_sentence = create_template_sentence(templates, True)
+    full_sentence = create_template_sentence(templates, instructions=True)
```
Panaetius, line 88: Could use a test/assert with `describe=True` as well

jsam: Done.

PR #1300, opened 2020-05-28
fix: disable version check in githook calls

### LINKED ISSUE #1257: Renku hook and version check don't work together
We have a git hook that uses `renku show outputs` to check that we don't manually change a generated file.
If there is a enw version of renku available, the version check will output a corresponding message, leading the hook check to believe that there was a modified file (when there wasn't). 
Since we check the version once a day, this will happen every day and then suddenly work if rerun after, so we should fix this to not confuse our users.

```
$ git commit -m "..."
You are trying to update generated files.

Modified Files:

You are using renku version 0.10.3, however version 0.10.4 is available.
You should consider upgrading...

To commit anyways, use "git commit --no-verify".

$ git commit -m "..."
[master abcdefg] ...
1 file changed, 1 insertion (+)
$ 
```

Panaetius:
disables version check on renku commands used in our git hooks.

Closes #1257


renku/data/pre-commit.sh
```diff
@@ -38,7 +38,7 @@ if [ ${#MODIFIED_FILES[@]} -ne 0 ] || [ ${#ADDED_FILES[@]} -ne 0 ]; then
 fi
 
 if [ ${#MODIFIED_FILES[@]} -ne 0 ] ; then
-  MODIFIED_OUTPUTS=$(renku show outputs "${MODIFIED_FILES[@]}")
+  MODIFIED_OUTPUTS=$(renku --disable-version-check show outputs "${MODIFIED_FILES[@]}")
```
rokroskar, line 41: you could set the `RENKU_DISABLE_VERSION_CHECK` environment variable to avoid a regression in case more commands are added here and we forget about the flag?

Panaetius: I didn't know that env var exists, thanks for pointing it out. Makes more sense

renku/data/pre-commit.sh
```diff
@@ -22,6 +22,7 @@
 ######################################
 
 # Find all modified or added files, and do nothing if there aren't any.
+export RENKU_DISABLE_VERSION_CHECK=false
```
rokroskar, line 25: I think you want it set to 1

Issue #776, opened 2019-10-24
feat(svc): Enable unlinking a file from a dataset


jsam:
Add a feature to enable unlinking a file from a dataset.

PR #1314, opened 2020-06-03
feat(service): adds endpoints for unlinking files from a dataset

jsam:
closes: #776

### DIFF
--- a/renku/service/views/datasets.py
+++ b/renku/service/views/datasets.py
@@ -25,7 +25,7 @@ from flask_apispec import marshal_with, use_kwargs
 from git import GitCommandError, Repo
 
 from renku.core.commands.dataset import add_file, create_dataset, \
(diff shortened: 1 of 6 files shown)


renku/service/views/datasets.py
```diff
@@ -343,3 +347,56 @@ def edit_dataset_view(user_data, cache):
...
+    with chdir(project.abs_path):
+        records = file_unlink(
+            short_name=ctx['short_name'],
+            include=ctx.get('include_filters'),
+            exclude=ctx.get('exclude_filters'),
+            yes=True,
+            interactive=False,
+            commit_message=ctx['commit_message']
+        )
+        unlinked = [record.name for record in records]
```
mohammad-sdsc, line 400: Should we return `path` here instead of name to make the results uniquely identifiable?

PR #1323, opened 2020-06-09
fix(core): Fixes SoftwareAgent person context

Panaetius:
Adds type to SoftwareAgent `wasStartedBy` so it gets serialized with a correct context


jachro:
The graph team tested that locally and all works perfect!
Thanks for super fast fix!

Issue #1284, opened 2020-05-22
Allow to set dataset directory with renku config


Panaetius:
In #1197 we investigate the `--datadir` flag to commands. Currently this flag does not do anything except when running `renku dataset` without a subdirectory.

It doesn't make sense to have to specify a dataset dir on every renku command, if a user wants to use a directory other than `/data/`, it should be configured at the project level and used by all commands.
So it'd make sense to have it as a `renku config` setting that writes it to `renku.ini` and use it consistently in all commands.

The `--datadir` flag will be removed in #1197

Issue #570, opened 2019-06-28
Persist datadir information in project config


jsam:
Currently, we have no way of telling datadir for a dataset before instantiation of click context. This is needed to make all datasets commands able to operate on a dirty repository.

rokroskar:
should this be part of `.renku/metadata.yml` or a project-local config file? I'm not sure, for example, we need/want this information in the KG (and I don't really know what we'd call it).

Panaetius:
Related to https://github.com/SwissDataScienceCenter/renku-python/issues/1284

PR #1347, opened 2020-06-25
feat(core): configurable data dir

mohammad-sdsc:
# Description

Allows specifying a data directory at the project creation time using `--data-dir` flag. This is persisted in the project's config file (`.renku/renku.ini`) and cannot be modified afterwards.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/570
Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/1284


renku/cli/init.py
```diff
@@ -246,6 +246,21 @@ def is_path_empty(path):
...
 
+def resolve_data_directory(data_dir, path):
+    """Check data directory is within the project path."""
+    if not data_dir:
+        return
+
+    absolute_data_dir = Path(path) / data_dir
+
+    try:
+        return absolute_data_dir.relative_to(path)
```
Panaetius, line 257: Missing a `.resolve()`

Panaetius: We should also check that datadir is not `.`, `.renku` or `.git`,

PR #1366, opened 2020-07-06
feat(core): move templates to renku-python

### LINKED ISSUE #1122: Update cached templates
**Problem**
Currently, templates are downloaded and installed locally with the `renku` command line, but they are never updated.

**Proposal** (updated)
* Do *not* embed the templates in the `renku` package, but instead download them just after installing or whenever using `renku init` the first time
* When using `renku init`, check if there is an update in the remote repository and suggest users to download the latest template version. Do this only when `--template-source` is not specified and possibly with a reasonable cooldown (no need to check it frequently if multiple projects are initiated one after the other).
* We need a `minimum renku version` variable in the templates manifest file to avoid updating to an incompatible template version.
* [verify] we should probably move the templates cache folder to another location, like `~/.renku/templates`, to avoid any permission related problem when updating. Any obvious drawback with this approach?

### LINKED ISSUE #1306: move template responsibility to renku-python
# Proposal: renku-python should be the source of base templates

## Background

We have had tons of issues with the renku project templates stored in an external repository. There are circular dependencies that manifest at release time and the templates can't be properly tested from the codebase. 

## Proposal

We should make this repository the "source of truth" for the base templates and the templates repository should be read-only. The CI in this repository should push a new version of the templates since those can serve as a starting point for others who may want to create their own. 

## Compatibility

Nothing should change for the consumers of the template repo, so this change should not cause any incompatibilities with other components. 

## Implementation

1. create a template directory as a resource in the current renku-python codebase
2. make tests that check whether e.g. a new project using this template passes `renku doctor` and `renku migrate` checks. 
3. make a CI job that pushes the contents of this template directory to https://github.com/SwissDataScienceCenter/renku-project-template keeping the template in-sync with the version of renku-python

mohammad-sdsc:
# Description

Moves renku project templates to `renku-python` and creates a PR to updates `renku-project-template` repo whenever templates are updated.

Fixes #1306 
Fixes #1122


rokroskar:
I think there is no need to move templates to renku-python completely, but the implementation of the solution to https://github.com/SwissDataScienceCenter/renku-python/issues/1122 should probably still be kept?

mohammad-sdsc:
> I think there is no need to move templates to renku-python completely, but the implementation of the solution to #1122 should probably still be kept?

Solution to #1122 was to keep the templates in renku-python and update them together. I believe #1122 is partly addressed in https://github.com/SwissDataScienceCenter/renku-python/pull/1509. I'll close this PR.

Issue #1339, opened 2020-06-22
chore(service): add a version to migrations response


jsam:
Currently, the service does not give information about the source and destination versions of migrations. This information would be useful for UI.

PR #1372, opened 2020-07-07
feat(service): adds source and destination versions to migrations check

jsam:
closes: #1339 

* adds `agent_version` field to project metadata `.renku/metadata.yml`
* updates project metadata after the execution of the commands
* adds fields to migrations_check on the service


renku/core/management/repository.py
```diff
@@ -148,6 +149,11 @@ def __attrs_post_init__(self):
             except subprocess.CalledProcessError:
                 pass
 
+    @property
+    def latest_version(self):
+        """Returns latest `SoftwareAgent` used in the repository."""
+        return SoftwareAgent.from_commit(self.repo.head.commit)
```
Panaetius, line 155: This doesn't work if the newest commit was a commit done by the user (not renku), e.g.

```
>>> from renku.core.commands.migrate import migrations_check, migrations_versions
>>> migrations_versions()
('0.10.5.dev52', 'Panaetius')
```

jsam: Yea, this PR was supposed to be draft. There is also one more issue present on the service when the source and destination are the same. I'll address this asap.

renku/core/models/projects.py
```diff
@@ -163,6 +167,10 @@ def from_jsonld(cls, data, client=None):
 
     def to_yaml(self):
         """Write an instance to the referenced YAML file."""
+        from renku import __version__
+
+        self.agent_version = __version__
```
Panaetius, line 172: Should we do this in a @pre_dump marshmallow hook on `ProjectSchema` ? I'm not entirely sure if that's a good idea, on the one hand it'd always set it even if project is nested somewhere, but there might be some circumstances where we don't want that. But I an't think of any off the top of my head.

jsam: I personally do like fat-serializers & thinner models, however, if you look at our models currently and the current state we have thin-serializers and very fat models were we do all (r/w) model mutations on the model itself. If we want to change that, then we should create a refactoring epic and go for the switch. Not sure if mixing them is the best thing here, it just leads to confusion IMHO (having read mutation on models and then write on serializers). 

I would leave it like this for now and open a ticket to make serializers as fat as possible.

PR #1382, opened 2020-07-13
fix(core): Allow more than one activity to invalidate a file in SHACL

Panaetius:
More than one activity can take part in an invalidation (namely parent and a child workflow), this fixes the SHACL to reflect that.

Also changes calamus datetime serialisation to be always timezone aware


Panaetius:
This is now part of https://github.com/SwissDataScienceCenter/renku-python/pull/1386 closing this one

PR #1386, opened 2020-07-15
feat(core): Move workflow serialisation over to calamus

### LINKED ISSUE #1298: Use the calamus library for Workflow and provenance entities
We should use [Calamus](https://github.com/SwissDataScienceCenter/calamus) or the (de)serialization (from)to JSON-LD.

This should be done after #1252 and #1037.

### LINKED ISSUE #1418: fix(core): project metadata updated/schema:dateUpdated field not getting set on update
I've tried to address this problem in #1372 but due to quite challenging command lifecycle we have, this is turning out to be a bigger challenge than I expected. This ticket is about updating that field in project metadata every time renku project is meaningfully changed by renku commands.

Panaetius:
Also stores workflows as flattened jsonld, reducing the size of the files by 50-70%. And speeding up renku log by a factor of 4 (tested on the covid project with current master).

Needs https://github.com/SwissDataScienceCenter/calamus/pull/31 to work

Closes #1418 
Closes #1298


renku/core/models/workflow/parameters.py
```diff
@@ -17,106 +17,101 @@
...
-        # right now it's limited to a-zA-Z0-9 (-_ will work once it's fixed)
-        return "_:MappedIOStream-{}".format(str(uuid.uuid4())).replace("-", "")
+        """Generate an id for a mapped stream."""
+        host = "localhost"
+        if self.client:
+            host = self.client.remote.get("host") or host
+        host = os.environ.get("RENKU_DOMAIN") or host
+
+        return urllib.parse.urljoin(
+            "https://{host}".format(host=host), pathlib.posixpath.join("/iostreams", self.stream_type)
```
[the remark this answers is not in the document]
Panaetius: that's on purpose, all streams have the same id, since there's no difference between them (`stdout` on my machine is semantically the same as `stdout` on your machine. The mapped path is what differs but that is tracked on the CommandInput/Output, not on the stream). I discussed it with Kuba and we didn't see a reason why we should duplicate stream objects all over the place with different ids but the same semantics.

renku/core/models/datasets.py
```diff
@@ -122,7 +133,7 @@ def _convert_creators(value):
 class CreatorMixin:
     """Mixin for handling creators container."""
 
-    creators = jsonld.container.list(Person, kw_only=True, context="schema:creator", converter=_convert_creators)
+    creators = attr.ib(kw_only=True, converter=_convert_creators)
```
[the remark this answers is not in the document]
Panaetius: `factory=list` does not denote this as a list, instead it is almost the same as `default=[]`, but the latter can have issues that the `[]` is the same instance, so if instantiated twice and you add an entry to the first instance, it'll show up on the second instance as well (Which is why we should never use `default=[]` or `default={}` and always use factory instead).

So `factory=list` should only be added if there's not a default like `default=None` already and if the property is not a required property. In this case, I think creators is required, so we don't want a default here. This prevents `CreatorMixin` from being instantiated without passing `creators` to the constructor.

This might be true for the other cases you mention as well, but maybe we should double check.

renku/core/management/migrations/m_0005__2_cwl.py
```diff
@@ -227,6 +280,11 @@ def _migrate_single_step(client, cmd_line_tool, path, commit=None, persist=False
         run.path = path
         process_run = ProcessRun.from_run(run, client, path, commit=commit)
         process_run.invalidated = _invalidations_from_commit(client, commit)
+        if isinstance(process_run.association.agent, Person) or not process_run.association.agent.label.startswith(
+            "renku "
+        ):
+            # fix broken SoftwareAgent due to rebases
```
jsam, line 286: Is this a `NOTE:` or `TODO:`? Maybe add a mark here?

renku/core/management/migrations/m_0005__2_cwl.py
```diff
@@ -250,11 +309,17 @@ def _migrate_composite_step(client, workflow, path, commit=None):
             subrun = parse_cwl_cached(str(path))
 
         subprocess, _ = _migrate_single_step(client, subrun, path, commit=commit)
-        subprocess.path = run.path
         run.add_subprocess(subprocess)
 
     with with_reference(run.path):
         wf = WorkflowRun.from_run(run, client, run.path, commit=commit)
+
+        # fix broken SoftwareAgent due to rebases
```
jsam, line 317: Same as above.

renku/core/management/migrations/models/v3.py
```diff
@@ -41,20 +41,24 @@ def __init__(self, **kwargs):
...
+
     @classmethod
-    def from_git(cls, git):
+    def from_git(cls, git, client=None):
         """Create an instance from a Git repo."""
         name, email = get_user_info(git)
-        return cls(name=name, email=email)
+        instance = cls(name=name, email=email)
+
+        if not instance._id or "mailto:None" in instance._id:
```
jsam, line 52: We are doing this check multiple times. Maybe extracting it into a function makes more sense?

renku/core/models/calamus.py
```diff
@@ -17,31 +17,59 @@
...
+                and data["_label"]
+                and "@UNCOMMITTED" not in data["_label"]
+                and "@" in data["_label"]
+            ):
+                try:
+                    self._add_field_to_data(data, "commit", self._client.repo.commit(data["_label"].rsplit("@")[1]))
+                except ValueError as e:
+                    if "could not be resolved, git returned" not in str(e):
+                        raise
+                    # commit does not exist in local repository.
```
jsam, line 71:
```suggestion
                    # NOTE: Commit does not exist in a local repository.
```

renku/core/models/calamus.py
```diff
@@ -82,3 +118,86 @@ def _deserialize(self, value, attr, data, **kwargs):
...
+
+    def __init__(self, *args, propagate_client=True, **kwargs):
+        """Init method."""
+        super().__init__(*args, **kwargs)
+
+        self.propagate_client = propagate_client
+
+    @property
+    def schema(self):
+        """The nested Schema object.
```
jsam, line 134: When you generate docs, we could perserve links for code browsing by:
```suggestion
        """The nested ``marshmallow.Schema`` object.
```

renku/core/models/workflow/parameters.py
```diff
@@ -17,106 +17,101 @@
...
     def default_id(self):
-        """Set default id."""
-        # TODO: make bnode ids nicer once this issue is in a release:
-        # https://github.com/RDFLib/rdflib/issues/888
-        # right now it's limited to a-zA-Z0-9 (-_ will work once it's fixed)
-        return "_:MappedIOStream-{}".format(str(uuid.uuid4())).replace("-", "")
+        """Generate an id for a mapped stream."""
+        host = "localhost"
+        if self.client:
+            host = self.client.remote.get("host") or host
```
jsam, line 49: No need to branch, `get` internally supports default values.
```suggestion
            host = self.client.remote.get("host", host)
```

renku/core/models/workflow/parameters.py
```diff
@@ -17,106 +17,101 @@
...
+        host = os.environ.get("RENKU_DOMAIN") or host
```
jsam, line 50: Same as above.

renku/core/models/workflow/parameters.py
Panaetius: Actually this doesn't work. In some cases, remote is 
```
> client.remote
{'host': None, 'name': None, 'owner': None}
```
so get returns `None` instead of the default value

jsam: That smells like a serialization bug then. We should fix it in remote probably.

renku/core/models/workflow/parameters.py
jsam: This is pulling from the environment so you can utilize `get`.

renku/core/models/workflow/parameters.py
jsam: Also, I believe that the property `remote` could be cached.

Panaetius: it's what you get when you just renku init a repository locally with no remote. We could return `{}` instead but that's not any more or less correct than the current version. Ideally we'd return `None` if there is no remote, but then we'd have to guard that everywhere in code where it's used, so a separate ticket makes sense. --> https://github.com/SwissDataScienceCenter/renku-python/issues/1457

PR #1391, opened 2020-07-16
chore: Pins all versions in setup.py 

### LINKED ISSUE #785: Pin python dependencies and automatically update them
Since currently we use the most recent dependencies in our setup.py, but this can lead to things unexpectedly breaking due to downstream changes, we should pin our requirements to fixed versions.

To still keep them up-to-date, we should use a tool like pyup or dependabot to automatically create PRs for new versions and only merge the PRs if all the tests pass. That way our builds don't suddenly break without any change in code.

Panaetius:
so downstream dependency updates don't break things

Closes #785


jsam:
Thanks! :rocket:

PR #1397, opened 2020-07-22
test: core IT migrated & performance improvements

jsam:
This PR makes a few improvements on our testing setup:

* import datasets tests use 60x times smaller datasets for importing
* `integration` marked tests now run in parallel using `pytest-xdist` ( normal unit tests are still running sequentially due to flakiness, but by component)
* local execution of `pytest -m integration -n auto` leads to the execution time of <5 minutes (depends on the CPU/Threads count)
* speeds up CI execution (4 workers only, expandable with our infrastructure if we want faster):
    * `test-macos` & `test-linux`: ~20min (was ~40min)
    * `test-linux-integration`: ~15min (was 45min)
    * `test-macos-integration`: ~20min (was ~55min+) 
    * overall: ~25min (was ~55+min with collect & provisioning)
* unit tests are decoupled on the CI level so testing happens for each module individually (tried to used xdist on them but `cli` tests are really flaky, so such parallelization does not benefit us much)
  * `test-linux` are now run in parallel for each component: `test-linux-cli`, `test-linux-core`, `test-linux-service`
  * `test-macos` are now run in parallel for each component: `test-macos-cli`, `test-macos-core`, `test-macos-service`

Due to the big wait times during work hours for macos machines, I've disabled the `test-macos-integration`  to run only on post-merge. I've tried to run a test suite on my threadripper machine with 48 threads and the entire suite ran blazingly fast due to the parallelization introduced in this PR (to avoid `cli` flakiness locally I used `pytest-rerunfailures`). 

We could speed up this run even further by provisioning Github runners on our infrastructure to expand the computing power.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -110,6 +110,7 @@ tests_require = [
     'pytest-timeout>=1.3.4',
     'pytest-pep8>=1.0.6',
     'pytest-yapf>=0.1.1',
+    'pytest-rerunfailures>=9.0',
     'pytest>=4.0.0',
     'responses>=0.7.0',
     'unify>=0.4',
--- a/setup.py
+++ b/setup.py
@@ -110,7 +110,6 @@ tests_require = [
     'pytest-timeout>=1.3.4',
     'pytest-pep8>=1.0.6',
     'pytest-yapf>=0.1.1',
-    'pytest-rerunfailures>=9.0',
     'pytest>=4.0.0',
     'responses>=0.7.0',
     'unify>=0.4',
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -103,6 +103,7 @@ def run(self):
     'isort==5.1.4',
     'six>=1.13.0',
     'pydocstyle>=3.0.0',
+    'pytest-xdist>=1.33.0',
```
Panaetius, line 106: We shouldn't add versions that are open to new versions. dependabot takes care of that and it prevents tests suddenly failing in the future without a change on our side.

PR #1407, opened 2020-07-24
feat: decode base64 headers

jsam:
If renku fullname or email header is passed as base64, we will try to decode them if not we just use the raw value.


jsam:
@rokroskar Thanks! :+1:

Issue #723, opened 2019-10-01
Yapf to Black


jsam:
To streamline our code formatting with other repositories, we should migrate to Black.

Depends on: #722

PR #1408, opened 2020-07-24
chore: replace yapf formatter with black

mohammad-sdsc:
# Description

Creates necessary base for formatting with `black` and removes `yapf`. A github action is added to apply black format to open PRs; this action can be removed when the transition is done.

You just need to review first and third commits: https://github.com/SwissDataScienceCenter/renku-python/pull/1408/commits/f528cb70080be2f977448f04bb249f1b356a135c and https://github.com/SwissDataScienceCenter/renku-python/pull/1408/commits/2e0413dc664026c1b02929ff00c7fbd59ec9fe96

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/723

## NOTE

If you use `python3 -m pip install --user --editable .[all]` to install renku in editable mode, you need to include `--no-use-pep517` due to an incompatibility with `pyproject.toml` file. See https://discuss.python.org/t/pip-19-1-and-installing-in-editable-mode-with-pyproject-toml/1553/21.


CONTRIBUTING.rst
```diff
@@ -105,20 +105,18 @@ encourage usage of `pipenv` and `pyenv`:
...
+      $ black .
 
-   You may want to set up yapf_ styling as a pre-commit hook to do this
-   automatically:
-
-   .. code-block:: console
-
-      $ curl https://raw.githubusercontent.com/google/yapf/master/plugins/pre-commit.sh -o .git/hooks/pre-commit
-      $ chmod u+x .git/hooks/pre-commit
+   You may want to set up black_ styling as a pre-commit_ hook to do this
```
Panaetius, line 114: We should add the `.pre-commit-config.yaml` mentioned in https://github.com/psf/black#version-control-integration to our repo to make this easier

rokroskar: I agree - we already have it set up like this in the notebooks and gateway repos, would be nice to have consistency

Issue #1340, opened 2020-06-22
Refactor migrations to be independent of core models


mohammad-sdsc:
ATM, some migrations rely on the core models in statements like `for dataset in client.datasets.values()`. This tight coupling between the two requires extra migration steps when changing core models. Sometimes these steps must be part of the model itself (e.g. `pre_load` hooks in Calamus-based `Dataset` model).
We should refactor the migrations to be independent of the core models either by working directly with yaml/json-ld data or by having dedicated migration models.

PR #1431, opened 2020-07-31
feat(core): separate models for migrations

mohammad-sdsc:
# Description

Create new migration models for `Dataset`, `DatasetFile`, and `Project`.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/1340


renku/core/models/migration.py
```diff
@@ -0,0 +1,218 @@
...
+    from renku.core.models.datasets import DatasetFile
+
+    if isinstance(value, DatasetFile):
+        return value.as_jsonld()
+    if isinstance(value, dict):
+        return value
+    raise ValueError
+
+
+class Base:
```
Panaetius, line 46: We still have the issue here that we use one schema (the ones in this file) for all migrations, but schemas might change potentially leading to issues further down the line. I think it would be safest to have the schemas inside the individual migrations, even if it is code duplication. We could add the Schema to the first migration that uses it, import if from there in successive migrations if it didn't change, but add a modified copy to a migration if the schema changed.

And we could use calamus for this if we added a flag to calamus to not create an instance but instead just output a dict (like base marshmallow does). Maybe an option like `create_instance=False` passed to the schema?

And for instance workflows are represented as a flattened list instead of a nested jsonld structure, so this code would not be able to handle that, but calamus can.

mohammad-sdsc: I've ended up adding a generic base class that gets its attributes from the schema that is used for loading.

PR #1436, opened 2020-07-31
chore(service): add commit message to migration

jsam:


jsam:
I think that's fine. This PR fixes the initial commit in the chain so it does looks like this.

```
gunicorn server.py --w 4
renku migrate: committing migrated workflow
renku migrate: committing migrated workflow
renku migrate: committing migrated workflow
```

If there will be a need from the service side to affect other commit messages then we would have to adjust that in the core, for now it's limitation.

PR #1446, opened 2020-08-05
build: Changes linux tests to use self-hosted runners

### LINKED ISSUE #1417: tests: improvements on our tests infrastructure
This sprint we did some nice improvements to our tests infrastructure and by analyzing some aspects of it, I've come to conclusion that our testing suite is substantially large and if we want to speed things even further and make them more usable and developer-friendly, we should add custom GH actions runner to our infrastructure.

With custom runners, provisioning step should be much faster and by having a machine with more than 4 cores we now have the ability to utilize them to speed up tests.

This ticket is about exploring such a scenario in the Linux environment (potential similar improvements could be made with macos in the future).

Panaetius:
Changes linux tests and linux IT to use self hosted runners

Closes #1417

### DIFF
--- a/.github/workflows/test_deploy.yml
+++ b/.github/workflows/test_deploy.yml
@@ -61,7 +61,7 @@ jobs:
       run: twine check dist/*
 
   test-linux-cli:
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, linux]
     strategy:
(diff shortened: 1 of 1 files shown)


Panaetius:
Closing as I don't think it's relevant anymore

PR #1447, opened 2020-08-05
Service dependency info

travissarles:
Does the CLI call the gateway as well as GitLab directly?


rokroskar:
@travissarles the architecture and sequence diagrams for the service are here: https://renku.readthedocs.io/en/latest/developer/services/core_service.html

The CLI doesn't interact with any services - the user has to push/pull from the git server with regular git commands.

jsam:
@travissarles welcome onboard! 

For any questions related to the renku core or assistance feel free to talk to us on slack (might be faster). I propose we close this PR, and for any onboarding questions you might have, we can talk directly.

jsam:
I'm closing this issue as a part of a cleanup of PR's. Since service does not have any hard dependencies on any other service at this point, I think adding this doc comment would be slightly miss leading. Thanks for the PR though, if you have any other suggestion on how we can improve our docs we would be more than happy to implement them.

PR #1448, opened 2020-08-05
chore(service): add sentry

### LINKED ISSUE #1415: feat(service): add sentry support
Currently, service does not have many runtime troubleshooting ways which make debugging a bit difficult. To help with this pain we should add sentry support, so we have some vision into the exceptions which can occur during runtime.

rokroskar:
Adds sentry for flask, rq, and redis. 

Currently running on rok.dev.renku.ch. To test, use `kubectl` to log in to the `renku-core` pod (`core` container) and `curl localhost:8080/debug-sentry` from the command line). Then go to https://sentry.dev.renku.ch and check the `renku-python` project. 

Tested the flask and rq integrations (we don't have any custom code for redis, not even sure it's necessary to have it here). You can see the two events [here](https://sentry.dev.renku.ch/share/issue/794361b301114bb58e2f89511e46346e/) and [here](https://sentry.dev.renku.ch/share/issue/450faa2d24db44a8bc6ccae3dd7abae3/).

Fixes #1415


helm-chart/renku-core/values.yaml
```diff
@@ -16,8 +16,8 @@ cleanupProjectsTTL: 1800
 logLevel: INFO
 
 image:
-  repository: renku/renku-core
-  tag: 'latest'
+  repository: rrrrrok/renku-core
```
ableuler, line 19: These two lines should not be committed.

rokroskar: thanks - I should actually mark the PR as draft since I haven't had a chance to test the RQ piece yet

helm-chart/renku-core/templates/deployment.yaml
```diff
@@ -83,6 +87,10 @@ spec:
               value: {{ .Values.cleanupProjectsTTL | quote }}
             - name: RQ_WORKER_LOG_LEVEL
               value: {{ .Values.logLevel }}
+            - name: SENTRY_DSN
+              value: {{ .Values.sentry.dsn }}
+            - name: SENTRY_ENV
```
jsam, line 92: I think we should think about moving the names of the environment variables to the values file (since they could be considered values as well). Then here we could use `range` to iterate through all `env` entries. 

Something like this:

```
          env:          
            {{- range .Values.env }}
            - name: {{ .name }}
              value: "{{ .value }}"
            {{- end }}
```

and then in values.yaml we could have:

```
env:          
 - name: "SENTRY_DSN"
   value: ""
...
```

renku/service/entrypoint.py
```diff
@@ -62,6 +66,13 @@
 
 logging.basicConfig(level=os.getenv("SERVICE_LOG_LEVEL", "WARNING"))
 
+if os.getenv("SENTRY_DSN"):
```
jsam, line 69: Since both of them are used, should we check for both of them here?

```suggestion
if os.getenv("SENTRY_DSN") and os.getenv("SENTRY_ENV"):
```

rokroskar: not sure - the environment doesn't need to be set for sentry to work, it depends a bit on how sentry is used. The DSN otoh is required to activate sentry.

helm-chart/renku-core/templates/deployment.yaml
rokroskar: I could see this being useful for making certain things more flexible, but in general isn't it better/cleaner/less error prone to be more explicit about what is going into the environment? Because in the application the environment variable names are fixed.

jsam: IMHO, I think it's easier to reason about the helm package when all application-specific defining elements are in 1 place. That way, templates are templates and values contain all application-specific stuff. But maybe that's just me. This is off the scope for this PR anyway, I'm just saying that I think it might be nicer to have all off it in 1 place.

rokroskar: I'd say we could discuss this in the context of an overall chart refactoring - would be nice to stay consistent across all the charts we develop. We can open a general issue/epic for this.

PR #1450, opened 2020-08-07
build: Restrict build matrix on PRs

### LINKED ISSUE #1441: Reduce number of executed actions for pull requests
When creating/modifying a PR, we run all tests in renku-python for three versions of python: 3.6, 3.7, and 3.8. This requires a large amount of resources and so far we did not have any incompatibility issues. A proposal is to test with python 3.6 on PRs to speed up the development and check against all three versions only on merges to `master`.

Panaetius:
Only runs PR tests on linux 3.7 but runs tests on 3.6, 3.7 and 3.8 on master and tags.

Closes #1441


conftest.py
```diff
@@ -68,25 +88,25 @@ def instance_path(renku_path, monkeypatch):
         yield renku_path
 
 
-@pytest.fixture()
+@pytest.fixture(scope="function")
```
mohammad-sdsc, line 91: I believe default scope is `function` and we don't need to specify it here and other places in this file.

mohammad-sdsc (commented):
Thanks! This looks really good. Just a single comment.

conftest.py
Panaetius: I added them everywhere because I originally didn't trust it and thought it might inherit from other fixtures, but didn't remove it after that turned out to not be the case.

mohammad-sdsc (approved):
Great! Thanks!

Issue #1103, opened 2020-03-12
fix: wrong URL when adding to datasets via service


jsam:
```
"file://../../../../uploads/15eb65fd-3914-496a-861a-d0b8c2264cd8/101992.gif.pdf" 
```

Should be:
```
"file://101992.gif.pdf" 
```

rokroskar:
I think the purpose of the `url` field is to be able to identify where it came from - since there is no way of establishing the source automatically, should we even be setting it?

jsam:
I have 1 question about this URL topic. I agree completely that one of the potential cases is that we use that field to tell where the file came from but I think that is a miss use from my perspective (that's why I said that we should introduce a `source` property for that or newer version of it 'sameAs'). URL property as it is defined in schema stands for currently accessible URL of an item. So it's much more general in that sense. 

So instead of removing it, we should set it correctly. So for example in the case of file upload via service, that field could be something like: 
```
url: https://dev.renku.ch/projects/contact/integration-test/blob/9aaf8d0df1c0dec1b387250a399279e5d87ed7fa/data/ad35952ea8234123a7db829fc3834186/datafile1.txt
```

In case we opt for deletion this could be a property we build on the fly via `{project._id/file._id}`, but this will lead to potential problems down the road. Therefore, I'm not 100% sure that we should remove it.

rokroskar:
Yeah that's a good point! I agree it's a misuse of the term. I think we used it because there is no obvious `source` term - but maybe there is now? Would be worth looking into, thanks for pointing it out!

mohammad-sdsc:
After discussing with @Panaetius we decided that:
1. Introduce `renku:source` attribute to track source of the files added to a dataset. There is no other term defined in `schema.org` that carries the same meaning. We already use `schema:isBasedOn` to store file sources added from Git repos but it is meant to be used for derivative work and not tracking sources.
2. Value of `source` attribute for local and external files will be set to their path relative to the project's root directory.
3. `schema:url` attribute for `DatasetFile` will be set to `https://<host>/projects/<user>/<project>/files/blob/<path/to/file>`

PR #1451, opened 2020-08-07
feat(dataset): source and url for DatasetFile

mohammad-sdsc:
# Description

Fixes url for `DatasetFile` and sets its source. Just to be clear, `source` attribute for uploaded service files will be something like `"../../../../uploads/15eb65fd-3914-496a-861a-d0b8c2264cd8/101992.gif.pdf"` since uploaded files are added from a local dir in the server.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/1103


conftest.py
```diff
@@ -336,27 +337,53 @@ def data_repository(directory_tree):
...
+
+    bare_base_path = working_dir / "bare"
+    with tarfile.open(compressed_repo_path, "r") as fixture:
+        fixture.extractall(str(bare_base_path))
+
+    bare_path = bare_base_path / name
+    repository_path = working_dir / "repository"
+    repository = Repo(bare_path, search_parent_directories=True).clone(repository_path)
+
+    return repository
```
jsam, line 354: Should we yield here, and cleanup explicitly after execution?

conftest.py
```diff
@@ -336,27 +337,53 @@ def data_repository(directory_tree):
...
+    repository_path = old_repository.working_dir
 
-    shutil.rmtree(working_dir_path.strpath)
+    os.chdir(repository_path)
+
+    yield old_repository
+
+    os.chdir(repository_path)
+    old_repository.head.reset(commit, index=True, working_tree=True)
+    # remove any extra non-tracked files (.pyc, etc)
```
jsam, line 385: You could add an `INFO:` mark here.

conftest.py
```diff
@@ -388,74 +415,35 @@ def old_bare_repository(request, tmpdir_factory):
...
 
 
 @pytest.fixture
-def old_project(old_repository):
-    """Create a test project."""
-    repo = old_repository["repo"]
-    repository_path = repo.working_dir
+def old_dataset_project(tmp_path_factory):
+    """Prepares a testing repo created by old version of renku."""
+    name = "old-datasets-v0.9.1.git"
```
jsam, line 438: I was thinking about this before. Should we make this a parameter?

conftest.py
mohammad-sdsc: This is used in both session-/function-scoped fixtures so we cannot make it a function-scoped fixture. If we make it session-scoped then we need an extra fixture for the function-scoped fixtures to reset the state after each test. I believe it is cleaner/less code to leave it as a function.

conftest.py
mohammad-sdsc: This one is a bit different than the others because we do more checks on this repository after migration. This fixture is function-scoped (because it is used in only one test) while the others are session-scoped. I am not sure if we can parameterize this along with the others in a clean way. Do you have some solutions in mind?

jsam: I see. I guess it's fine then. I believe this might get refactored in the near future though since we might add more old projects created with different versions. The way I see it that we might refactor it as a parametrized test is that alongside testing fixture we also provide expected snapshots of metadata for that fixture. 

Then we can write a test which would:
1) ensure that fixture metadata and provided snapshot are actually different
2) run migration
3) ensure that fixture metadata and provided snapshot are the same

Usually, this kind of snapshot testing is done in the UI, but I think if it works fine for HTML it will work fine for our YAML files as well. But anyway, this is way out of scope here, so we can create a ticket for this if you agree.

conftest.py
jsam: Ah, I assumed it might not get cleaned up, but it does - all good! :+1: :see_no_evil:

PR #1468, opened 2020-08-14
chore(service): report checked exceptions to sentry

jsam:
Reports handled exceptions from the service endpoints to sentry. This change will report all exceptions which occur, please note that this might lead to possible spam of sentry. If that happens, we should disable the ones which we find irrelevant.


Panaetius (approved):
🚀

PR #1469, opened 2020-08-14
feat(service): Add additional template parameters

### LINKED ISSUE #1384: Send all the relevant variables to the templating function
As a follow up of #1287, it would be nice to provide to the templating function used by `renku init` all the available variables. Currently, only the variables provided as parameters are sent to the templating function, plus the project name.

The new variables could be:
* template url
* template reference
* project target id
* [namespace] -- when invoked from the service
* [repository] -- when invoked from the service
* [project_slug] -- when invoked from the service --> requires a little re-factory, maybe it's not needed since it's easy to derive from namespace + name

It would also be nice to use something different than `name` for the project name variable, so that it won't be a reserved word not usable as a parameter. Maybe adding `_` before the automatically provided variables?

Panaetius:
Adds new parameters (mainly for the service side):
- `__template_source__`
- `__template_ref__`
- `__template_id__`
- `__namespace__`
- `__repository__`
- `__project_slug__`

Values that are not available (when calling in the cli) are left empty.

Closes #1384


renku/cli/init.py
```diff
@@ -443,6 +446,14 @@ def init(
         except GitCommandError as e:
             click.UsageError(e)
 
+    # supply additional parameters
+    parameter["__template_source__"] = template_source
```
rokroskar, line 450: should this be renamed to `metadata` for consistency? Just a suggestion, feel free to ignore.

renku/cli/init.py
```diff
@@ -443,6 +446,14 @@ def init(
...
+    parameter["__template_ref__"] = template_ref
+    parameter["__template_id__"] = template_id
+    parameter["__namespace__"] = ""
+    parameter["__repository__"] = ""
+    parameter["__project_slug__"] = ""
+
```
rokroskar, line 456: `project_slug` includes the namespace - there are places where we want to use the escaped and stripped project name for naming certain files. Could we add `project_name_stripped` (from [here](https://github.com/SwissDataScienceCenter/renku-python/pull/1469/files#diff-35c86f980dbff80762ed1851377d73e2R71) but maybe give it a nicer name?) to the parameters that are passed to the template?

rokroskar (changes requested):

rokroskar (dismissed):

rokroskar (approved):

PR #1476, opened 2020-08-18
feat(workflow): avoid unnecessary parent runs

### LINKED ISSUE #1399: Don't create parent Runs if it is not necessary

### LINKED ISSUE #1400: Reuse parent Runs if possible

mohammad-sdsc:
# Description

Reuse parent runs in WorkflowRun when possible. Do not create parent runs when there is only a single step. Do not serialize WorkflowRun.

Fixes #1399 
Fixes #1400


Panaetius:
I tested this a bit and it looks very nice. I did find an issue/inconsistency in how rerun/update works, but I'm not sure if this is related to this PR or if it's how things always worked.

Let's say we have the following situation:
```
input1 ---dc4c--> intermediate1

**Comment by mohammad-sdsc** [comment]
I've tested your scenario on master and it has the same issue with `rerun` with and without `--from` option; it runs the last step or all steps in turns. I believe this is an issue with graph generation. I don't think it's in the scope of this PR to fix that. I'll create an issue for that and we will decide either to fix it or to wait for the redesigned graph generation.
I've fixed the issue with not reusing a parent run by checking all WorkflowRuns that generate an output if we cannot find one in the graph.
Regarding the `rerun` behavior, I also think it should run only the last step that generated a file when running `renku rerun file`. Using `--from` option users can specify a different path to execute. Rerun should not include updated inputs automatically because that what `update` is for. We should discuss these cases in more details.

File: renku/core/commands/graph.py [hunk 1]
```diff
@@ -484,13 +481,32 @@ def as_workflow(
...
 
-        return self._find_identical_parent_run(parent_process)
+        return self._find_identical_parent_run(run=parent_process, outputs=outputs)
+
+    def _find_identical_parent_run(self, run, outputs):
+        from marshmallow.exceptions import ValidationError
+
+        def workflow_has_identical_subprocesses(workflow_, subprocesses_ids_):
+            wf_subprocesses_ids = [step.process._id for step in workflow_.association.plan.subprocesses]
+            return wf_subprocesses_ids == subprocesses_ids_
```
Panaetius, line 491: Change to a `set()` instead of a list for comparison or sort the list by index of the `OrderedSubprocess` first. Unfortunately, we can't guarantee the order when deserializing from json-ld (Hence the `OrderedSubprocess` class).

Panaetius (changes requested):

renku/core/commands/graph.py
mohammad-sdsc: Good catch!

Panaetius (dismissed):

Issue #1419, opened 2020-07-28
chore(service): finish metrics tracking and displaying


jsam:
With #986, we implemented metrics which can be tracked using Prometheus and visualized using Grafana. There are a few things which we should implement to conclude the topic, those are:

1. Publish rqmetrics to pypi and dockerhub
2. Add sidecar to core service deployment
3. Finish drawing dashboard and persisting them

PR #1480, opened 2020-08-19
chore(service): add rqmetrics to deployment

jsam:
closes: #1419

TODO:
- merge on https://github.com/SwissDataScienceCenter/rqmetrics/pull/4


helm-chart/renku-core/templates/deployment.yaml
```diff
@@ -27,6 +27,13 @@ spec:
         - name: shared-volume
           emptyDir: {}
       containers:
+        - name: {{ .Chart.Name}}-rqmetrics
+          image: "justsam/rqmetrics:latest"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          env:
+            - name: RQ_REDIS_HOST
+              value: {{ include "call-nested" (list . "redis" "redis.fullname") }}-master
+
```
rokroskar, line 36: does this need annotations to be picked up by prometheus?

rokroskar: actually I guess they're needed in the service: https://github.com/SwissDataScienceCenter/renku-graph/blob/master/helm-chart/renku-graph/templates/event-log-service.yaml#L10-L13

helm-chart/renku-core/templates/deployment.yaml
```diff
@@ -27,6 +31,13 @@ spec:
         - name: shared-volume
           emptyDir: {}
       containers:
+        - name: {{ .Chart.Name}}-rqmetrics
+          image: "justsam/rqmetrics:latest"
```
rokroskar, line 35: This needs to be templated, e.g.

```
image: {{ .Values.metrics.image }}
```

helm-chart/renku-core/templates/deployment.yaml
```diff
@@ -27,6 +31,13 @@ spec:
...
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
```
rokroskar, line 36:
```suggestion
          imagePullPolicy: {{ .Values.metrics.image.pullPolicy }}
```

helm-chart/renku-core/templates/deployment.yaml
```diff
@@ -27,6 +31,13 @@ spec:
         - name: shared-volume
           emptyDir: {}
       containers:
+        - name: {{ .Chart.Name}}-rqmetrics
```
rokroskar, line 34: This should be in an `if` block, e.g.

```
{{ if .Values.metrics.enabled }}
```

rokroskar (changes requested):

helm-chart/renku-core/templates/deployment.yaml
```diff
@@ -14,6 +14,10 @@ spec:
       app.kubernetes.io/instance: {{ .Release.Name }}
   template:
     metadata:
+      annotations:
+        prometheus.io/scrape: 'true'
+        prometheus.io/path: '/metrics'
+        prometheus.io/port: '9005'
```
rokroskar, line 20:
```suggestion
        prometheus.io/port: '8765'
```

rokroskar (changes requested):

Panaetius (approved):
🚀

helm-chart/renku-core/values.yaml
```diff
@@ -31,6 +31,13 @@ service:
 ingress:
   enabled: false
 
+metrics:
+  enabled: true
```
rokroskar, line 35: usually the convention is to have metrics exporting default to false - can we do the same here?

rokroskar (changes requested):

rokroskar (approved):
Thanks @jsam, nothing quite as satisfying as watching a dashboard do its thing! :)

PR #1509, opened 2020-09-07
feat(service): adds template and dockerfile migration to migration endpoint

### LINKED ISSUE #1370: implement a command to update the version of renku-python in interactive environments
After https://github.com/SwissDataScienceCenter/renku-project-template/pull/86 is merged, it should be straightforward to update the renku-python version installed in users' docker images. We should have a command that checks if the current version needs to be updated and update it if that is the case. In the case of released versions this is trivial, it's not so obvious what to do when `renku migrate` is called from a development version, since the corresponding development release may not exist on PyPI.

Panaetius:
Closes #1370 

Requires/includes https://github.com/SwissDataScienceCenter/renku-python/pull/1469


rokroskar:
The PR description references itself - update?

Panaetius:
> The PR description references itself - update?

Ah, meant to link the other template parameter PR, fixed now.

vfried (approved):
All works well!

renku/core/management/repository.py
```diff
@@ -540,7 +559,36 @@ def import_from_template(self, template_path, metadata, force=False):
...
+
+        immutable_paths = []
+
+        for path in paths:
+            relative_path = Path(path)
+
+            if str(relative_path) in self.project.immutable_template_files:
+                immutable_paths.append(path)
+
+        return immutable_paths
```
mohammad-sdsc, line 586: This would be a shorter version of the same function:
```
immutable_template_files = self.project.immutable_template_files or []
return [p for p in paths if p in immutable_template_files]
```

tests/core/commands/test_init.py
```diff
@@ -193,6 +198,252 @@ def test_template_filename(local_client):
...
+        ]
+        template_files = []
+        for project_file in project_files:
+            expected_file = template_path / project_file.relative_to(local_client.path)
+            template_files.append(expected_file)
+            assert expected_file.exists()
+
+        # NOTE: Update from template
+        fetch_template = mocker.patch("renku.core.commands.init.fetch_template")
+        fetch_template.return_value = (manifest, temppath, "renku", "0.0.2")
```
mohammad-sdsc, line 236: It's better to factor the common setup code for this test and the following in a separate function to make the tests shorter and more readable.

mohammad-sdsc (changes requested):
Looks pretty good to me! Just made a few comments!

renku/core/management/migrate.py
```diff
@@ -82,7 +137,153 @@ def migrate(client, progress_callback=None):
...
+    return n_migrations_executed != 0, template_updated, docker_updated
+
+
+def _update_template(client, check_only=False, progress_callback=None):
+    """Update local files from the remote template."""
+    from renku.core.commands.init import fetch_template
+
+    project = client.project
+
+    if not project.template_version:
```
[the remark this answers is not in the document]
Panaetius: the latter. a project created from a template before this change won't have a version set, this way we know that we can't update it.

tests/core/commands/test_init.py
Panaetius: Well, each setup code is different, but I tried to hide it away in a function as a fixture.

renku/core/management/migrate.py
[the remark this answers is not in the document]
Panaetius: And we couldn't fetch templates in this case as the old version of init doesn't store what template was used/where the template was from.

mohammad-sdsc (approved):
Thanks for looking into this 👍!

PR #1511, opened 2020-09-08
chore: add style checks back to GH actions

### LINKED ISSUE #1458: pytest addons don't run for python files
We used to use `run-tests.sh` in `.github/workflows/test_deploy.yml` to run all tests but we refactored this workflow to select tests based on their tag. As a result, pytest pluging (e.g. flake8, black) and isort checks are not executed on the python files in `renku` directory.
We need to add two jobs to the workflows to run pytest and isort on that directory:
```
pytest -v -o testpaths="renku conftest.py" --ignore=renku/version.py
isort -c -df .
```

Panaetius:
closes #1458


.github/workflows/test_deploy.yml
```diff
@@ -32,6 +32,25 @@ jobs:
         GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
     if: "!startsWith(github.ref, 'refs/tags/') && github.ref != 'refs/heads/master'"
 
+  style-check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+    - name: Set up Python ${{ matrix.python-version }}
```
mohammad-sdsc, line 40: `${{ matrix.python-version }}` -> `3.8`

test.sh
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
```
mohammad-sdsc, line 1: What is the purpose of this file? It seems that we can delete it since it's not used anywhere.

mohammad-sdsc (commented):

test.sh
Panaetius: I think i needed it to test restricting the build matrix but ended up going with a different solution. I'll remove it

mohammad-sdsc (approved):
👍

PR #1516, opened 2020-09-10
feat(core): follow URL redirections for dataset files

Panaetius:
Follows URL redirections for dataset import and dataset add

Closes #1416

PR #1517, opened 2020-09-10
chore: add Flake8 test

Panaetius:
Adds flake8 tests since Black doesn't cover everything.

### DIFF
--- a/renku/service/views/datasets.py
+++ b/renku/service/views/datasets.py
@@ -23,15 +23,7 @@ from flask import Blueprint, request
 from flask_apispec import marshal_with, use_kwargs
 from git import GitCommandError, Repo
 
-from renku.core.commands.dataset import (
-    add_file,
-    create_dataset,
-    dataset_remove,
-    edit_dataset,
-    file_unlink,
-    list_datasets,
-    list_files,
-)
+from renku.core.commands.dataset import add_file, create_dataset, dataset_remove, edit_dataset, file_unlink
 from renku.core.commands.save import repo_sync
 from renku.core.models import json
 from renku.core.utils.contexts import chdir
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -39,6 +39,7 @@ check_styles(){
     pydocstyle renku tests conftest.py docs
     black --check --diff renku tests conftest.py
     isort -c --df .
+    flake8 renku tests conftest.py setup.py
     check-manifest --ignore ".travis-*,renku/version.py,renku/templates,renku/templates/**"
     find . -path ./.eggs -prune -o -iname \*.sh -print0 | xargs -0 shellcheck
 }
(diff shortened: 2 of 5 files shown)


conftest.py
```diff
@@ -324,6 +324,9 @@ def request_callback(request):
         rsps.add_callback(responses.GET, "http://example.com/file", callback=request_callback)
         rsps.add_callback(responses.GET, "https://example.com/file", callback=request_callback)
         rsps.add_callback(responses.GET, "http://example.com/file.ext?foo=bar", callback=request_callback)
+        rsps.add_callback(responses.HEAD, "http://example.com/file", callback=request_callback)
+        rsps.add_callback(responses.HEAD, "https://example.com/file", callback=request_callback)
+        rsps.add_callback(responses.HEAD, "http://example.com/file.ext?foo=bar", callback=request_callback)
```
mohammad-sdsc, line 329: Is this required for this PR or is left from somewhere else?

renku/core/commands/dataset.py
```diff
@@ -482,6 +482,11 @@ def import_dataset(
     client, uri, name="", extract=False, with_prompt=False, yes=False, commit_message=None, progress=None,
 ):
     """Import data from a 3rd party provider or another renku project."""
+    u = urllib.parse.urlparse(uri)
+    if u.scheme not in ("", "file", "git+https", "git+ssh"):
+        # NOTE: Check if the url is a redirect.
+        uri = requests.head(uri, allow_redirects=True).url
+
```
mohammad-sdsc, line 489: Same as above.

renku/core/management/datasets.py
```diff
@@ -251,7 +251,7 @@ def add_data_to_dataset(
             )
         else:
             for url in urls:
-                is_remote, is_git = _check_url(url)
+                is_remote, is_git, url = _check_url(url)
```
mohammad-sdsc, line 254: Same as above for the whole file.

renku/service/views/datasets.py
```diff
@@ -29,8 +29,6 @@
     dataset_remove,
     edit_dataset,
     file_unlink,
-    list_datasets,
-    list_files,
```
mohammad-sdsc, line 33: This requires a re-formatting.

mohammad-sdsc (commented):
I guess this include codes from another PR. Looks good other than that!

run-tests.sh
```diff
@@ -39,6 +39,7 @@ check_styles(){
     pydocstyle renku tests conftest.py docs
     black --check --diff renku tests conftest.py
     isort -c --df .
+    flake8 renku tests
```
mohammad-sdsc, line 42: We should include `conftest.py` and `setup.py` too.

conftest.py
Panaetius: Weird that that ended up here. It's part of https://github.com/SwissDataScienceCenter/renku-python/pull/1516 I'll remove it here

mohammad-sdsc (approved):
Thanks! Looks good!

PR #1518, opened 2020-09-10
feat(dataset): flattened JSON-LD metadata

### LINKED ISSUE #1253: use a flat JSON-LD file for Dataset metadata
After #1252 it should be beneficial to save the dataset metadata as a flat json-ld file. This will also enable additional features like easy invalidation (see #709)

mohammad-sdsc:
# Description

Flattened JSON-LD for dataset's metadata.

An issue was Person with the same email but different names/affiliation that would create a list for those properties in the flattened JSON-LD. In such cases, this PR only loads the element in the list because I did not want to change the @id field to be something other than email.

Fixes #1253


Panaetius (approved):
:rocket:

PR #1520, opened 2020-09-11
feat(service): move user identification to jwt

### LINKED ISSUE #1414: feat(service): move user identification completely to JWT
Service currently uses several headers to determine users identity, and all headers are proxied from the gateway which leads to some unexpected encoding issues. This issue has been temporarily fixed, so the headers are now encoded into base64 and then decoded (if sent to the service as such) on the service.

To come up with more stable solution, we should move all those identity headers into a JWT and extract them from there. This issue should be synced with the gateway maintainers, since there are some open questions which we should flush out, like:

1) Can we trust the gateway for JWT verification?
2) Should JWT be used as Authorization header or do we want to move it into other possible locations like: `query`, `param`, custom `header`?
3) Should Gitlab token still be outside of the token and used as `Authorization`?

This ticket needs further design.

jsam:
#closes #1414 

This PR addresses some additional issues and support:
  * Support for phased rollout (old header support should still work)
  * Adds optional identity on list dataset and list dataset files for public repository
  * Fixes correct messaging if the client does not provide `project_id` or `git_url` (or if provided `git_url` is not public)

Testing Deployment with both services: https://sam.dev.renku.ch/


renku/service/serializers/headers.py
```diff
@@ -32,42 +36,108 @@ def decode_b64(value):
...
         if not rfc_compliant:
             raise ValidationError("authorization value contains invalid value")
 
         return components[-1]
 
-    @pre_load()
+    @staticmethod
+    def decode_user(data):
+        """Extract renku user from a JWT."""
+        decoded = jwt.decode(data, JWT_SECRET_TOKEN, algorithms=["HS256"], audience="renku",)
```
Panaetius, line 101: This could raise `ExpiredSignatureError`, `ImmatureSignatureError`, `InvalidIssuedAtError` , should we handle that?

Panaetius (changes requested):
This is great! Thank you. I added some remarks regarding security and usage in k8s.

Panaetius (approved):
Thank you, great work!

PR #1523, opened 2020-09-14
feat(cli): adds consistent behaviour for cli commands

### LINKED ISSUE #691: renku dataset ls
The current command for showing the list of datasets that have been created is `renku dataset`, but I never remember that (especially because the helptext for `renku dataset` is "handle datasets").

Since the command for listing dataset files is `ls-files` it would be more analogous to have the command `renku dataset ls` or `renku dataset list`, which was the command originally asked for (https://github.com/SwissDataScienceCenter/renku-python/issues/417).

### LINKED ISSUE #1263: print a warning when working on non-existing dataset

Currently, trying to list files for a non-existing dataset, prints an empty table. This may be confusing since misspelling the dataset name is not uncommon and the result is the same also for an empty (but existing) dataset. I would suggest printing a `dataset <wrong-name> doesn't exist` warning to make it more obvious.

Example of current behavior
```
▶ renku dataset
ID                                    SHORT_NAME    TITLE        VERSION
------------------------------------  ------------  -----------  ---------
f5a3a233-7d7e-4b79-98d8-a1a0ca597905  air-quality   Air quality

▶ renku dataset ls-files air-quality
DATASET SHORT_NAME    ADDED                  SIZE  PATH
--------------------  -------------------  ------  ----------------------------------------------------------
air-quality           2020-05-07 13:59:37  501 KB  data/air-quality/ugz_luftqualitaetsmessung_seit-2012.csv
air-quality           2020-05-11 12:15:12  1.8 MB  data/air-quality/ugz_luftqualitaetsmessungen_1983-2011.csv

▶ renku dataset ls-files I-DO-NOT-EXIST
DATASET SHORT_NAME    ADDED    SIZE    PATH
--------------------  -------  ------  ------

▶ renku dataset create empty
Use the name "empty" to refer to this dataset.
OK

▶ renku dataset ls-files empty
DATASET SHORT_NAME    ADDED    SIZE    PATH
--------------------  -------  ------  ------
```

Also, the command `renku dataset` shows an empty table when no datasets are available. We may want to change that as well, maybe by suggesting an action, something like `No datasets available. You can create one using renku dataset create`

Panaetius:
Cleans up commands so they present a consistent user interface.

`renku config` now shows help when called without a subcommand. Functionality is split into new `config show`, `config set` and `config remove` subcommands.
`renku dataset` now shows help when called without subcommand. new `ls` subcommand added for old functionality. `renku dataset ls-files` will now show an error if it's called on a non-existing dataset.
`renku rm` now shows help when called without a path.
`renku run` will now show an error when called without a command.
`renku update` will now show help when called without arguments/options. a new `--all`/`-a` flag was added to support the old behaviour of updating everything.
`renku workflow` now shows help when called without a subcommand, added a `ls` subcommand to list workflows. Also updated docs to remove documentation for the `set-name` subcommand, as that isn't really useful from what I saw and the docs were outdated. Instead added docs for the `ls` and `create` subcommands. (Maybe we should remove the `set-name` subcommand? I'm not sure if `LinkReference`s are still useful.)

Closes #691 
Closes #1263


renku/cli/dataset.py
```diff
@@ -420,7 +420,13 @@ def prompt_tag_selection(tags):
...
 
 
-@click.group(invoke_without_command=True)
+@click.group()
+def dataset():
+    """Dataset commands."""
+    pass
+
+
+@dataset.command("list")
```
mohammad-sdsc, line 429: Should we use `ls` instead of `list`? It is shorter and is in par with `renku dataset ls-files`.

mohammad-sdsc: It would be good if we could have command aliases out of the box but it has some hassles and I'm not sure if it worths the effort: https://stackoverflow.com/questions/46641928/python-click-multiple-command-names

renku/cli/workflow.py
```diff
@@ -53,24 +104,28 @@ def _deref(ref):
...
 
 
-@click.group(invoke_without_command=True)
+@click.group()
+def workflow():
+    """Workflow commands."""
+    pass
+
+
+@workflow.command("list")
```
mohammad-sdsc, line 113: Same as above: `ls` might be a better alternative.

renku/core/commands/dataset.py
```diff
@@ -659,8 +659,11 @@ def _filter(client, names=None, creators=None, include=None, exclude=None):
         creators = set(creators)
 
     records = []
+    unused_names = list(names)
```
mohammad-sdsc, line 662: Maybe a `set` is better here in case user provides the same name twice.

renku/service/serializers/jobs.py
```diff
@@ -16,7 +16,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Renku service cache serializers for jobs."""
-from marshmallow import Schema, fields, post_dump, pre_dump
```
mohammad-sdsc, line 19: I believe this is from the other PR which is already merged 😄.

mohammad-sdsc (commented):
Looks great, thanks! Just a comment about `list` or `ls`.

mohammad-sdsc:
Please also update the prompt for LFS: `renku config show_lfs_message False`

renku/cli/dataset.py
Panaetius: https://github.com/SwissDataScienceCenter/renku-python/issues/1530

renku/cli/dataset.py
```diff
@@ -61,7 +61,7 @@
 
 .. code-block:: console
 
-    $ renku dataset
+    $ renku dataset list
```
mohammad-sdsc, line 64: `list` -> `ls`

renku/cli/dataset.py
```diff
@@ -72,7 +72,7 @@
 
 .. code-block:: console
 
-    $ renku dataset --columns id,name,date_created,creators
+    $ renku dataset list --columns id,name,date_created,creators
```
mohammad-sdsc, line 75: `list` -> `ls`

renku/cli/dataset.py
```diff
@@ -85,7 +85,7 @@
 
 .. code-block:: console
 
-    $ renku dataset --revision=1103a42bd3006c94efcaf5d6a5e03a335f071215
+    $ renku dataset list --revision=1103a42bd3006c94efcaf5d6a5e03a335f071215
```
mohammad-sdsc, line 88: `list` -> `ls`

renku/cli/workflow.py
```diff
@@ -17,17 +17,68 @@
...
-``renku workflow set-name <name>``. The name can be added to an arbitrary
-file in ``.renku/workflow/*.cwl`` anytime later.
+Manipulating workflows
+~~~~~~~~~~~~~~~~~~~~~~
+
+Listing workflows:
+
+.. code-block:: console
+
+    $ renku workflow list
```
mohammad-sdsc, line 27: `list` -> `ls`

mohammad-sdsc (commented):
Looks good! Some comments need an update. Also you should fix `test_dataset_import_real_doi`:
```
result = runner.invoke(cli, ["dataset", "-c", "name,creators"]) -> result = runner.invoke(cli, ["dataset", "ls", "-c", "name,creators"])
```

mohammad-sdsc (dismissed):
Looks great! Thanks!

mohammad-sdsc (approved):
Looks good, thanks!

Issue #428, opened 2019-02-25
cli: show (inputs|outputs) with more info


jirikuncar:
**Is your feature request related to a problem? Please describe.**

Display workflow, creation/modification date for inputs and outputs (`-v|--verbose` or `--format=table`).

**Describe the solution you'd like**
Use `tabular` output.

**Additional context**
Make sure the `pre-commit` hooks are working.

PR #1524, opened 2020-09-15
feat(cli): verbose output for renku show

mohammad-sdsc:
# Description

Detailed output for `renku show inputs/outputs`

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/428

### DIFF
--- a/renku/cli/show.py
+++ b/renku/cli/show.py
@@ -86,7 +86,19 @@ respectively.
    $ echo $?  # last command finished with an error code
    1
 
+You can use the ``-v`` or ``--verbose`` flag to print detailed information
+in a tabular format.
+
+.. code-block:: console
+
+   $ renku show inputs -v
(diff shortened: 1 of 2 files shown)


renku/cli/show.py
```diff
@@ -151,29 +168,28 @@ def siblings(client, revision, flat, verbose, paths):
 
 @show.command()
 @click.option("--revision", default="HEAD")
-@click.argument(
-    "paths", type=click.Path(exists=True, dir_okay=False), nargs=-1,
-)
+@click.option("-v", "--verbose", is_flag=True)
```
Panaetius, line 171: should we add a `-c`/`--columns` flag similar to other commands?

mohammad-sdsc: I don't think it's required here because the information that we show are limited where all of them fit in the screen.

Panaetius (approved):
Thank you!

PR #1525, opened 2020-09-15
chore(core): remove dependency on gitdb/gitdb2

### LINKED ISSUE #1502: Collision in requirements
The the requirements specified by the wheel distribution of renku 0.11.1 result in a python environment with colliding packages `gitdb` and `gitdb2`. See here: https://github.com/DavHau/mach-nix/issues/101

The python package manager of your choice might ignore this collision, but that doesn't solve the issue itself
There are two different libraries with the same import path.
Therefore one doesn't really know which package is being used during runtime.

The problematic part of the dependency tree is:
```
renku - 0.11.1 - wheel
│
├── gitdb2 - 3.0.1 - wheel
├── gitpython - 3.1.0 - wheel
│   └── gitdb - 4.0.5 - wheel
│       └── smmap - 3.0.4 - wheel
```

The problem can either be fixed by removing the requirement `gitdb2` or my patching `gitpython` upstream to also require `gitdb2` instead of `gitdb`

mohammad-sdsc:
# Description

Removes dependency on `gitdb` from `renku-python`.

Fixes #1502

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -91,36 +91,44 @@
         },
         "cffi": {
             "hashes": [
-                "sha256:0da50dcbccd7cb7e6c741ab7912b2eff48e85af217d72b57f80ebc616257125e",
-                "sha256:12a453e03124069b6896107ee133ae3ab04c624bb10683e1ed1c1663df17c13c",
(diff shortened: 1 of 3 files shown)


Panaetius (approved):
Glad it was actually this easy 😄

Issue #1416, opened 2020-07-28
feat(cli): add support to follow link on dataset add and import


jsam:
When adding or importing data behind shorthand URL versions, renku does not follow the link to the data but reports that no data is found. 

To reproduce the issue:

1. Create a link to a data using any custom URL shortener service (I hear https://bitly.com/ is lately popular) 
2. Try to import your data using the link

Proposed solution:
* Renku should try to follow redirects in such cases

PR #1528, opened 2020-09-17
chore: follow redirects

jsam:
closes: #1416


Panaetius:
Uhm, this is a duplicate of https://github.com/SwissDataScienceCenter/renku-python/pull/1516 no?

jsam:
I think it doesn't duplicate since you have that part for HEAD, and I have for GET. I think we can merge both as is and close the ticket. :)

Panaetius (approved):
Thanks :)

PR #1533, opened 2020-09-22
fix(core): fix workflow migration triples

### LINKED ISSUE #1532: Fix duplicate renku:Run triples being generated due to non-unique id's
Currently migrations create new id's for `renku:Run` each time they are run, causing duplication of triples in the triples-generator. `renku:Run` id's should be unique to prevent this duplication.

Panaetius:
Adds deterministic id's for migrate `renku:Run`s. The id's are now based on the hash of the original `*.cwl` path and its commit.

Closes #1532


mohammad-sdsc (approved):
Looks good! Thanks!

PR #1538, opened 2020-09-23
fix(core): fixes git encoding of paths with unicode characters

### LINKED ISSUE #1474: Space in filenames cause invalid id's
we use the filename in the default ids of entities, but if the ids contain spaces, we get invalid rdf ids due to the spaces not being escaped properly.

The addition of `quote()` in https://github.com/SwissDataScienceCenter/renku-python/blob/master/renku/core/models/entities.py#L268 was intended to fix that, but we still get errors like https://sentry.dev.renku.ch/organizations/swiss-datascience-center/issues/69/?project=3

This might be due to missing a migration for this or some edge case.

Panaetius:
Git encodes path with unicode characters, like `filé.txt` as `"fil\u00e9.txt"`. This undoes that encoding when loading file paths from gitpython.

Closes #1474

### DIFF
--- a/renku/core/commands/graph.py
+++ b/renku/core/commands/graph.py
@@ -31,6 +31,7 @@ from renku.core.models.git import Range
 from renku.core.models.provenance.activities import Activity, ProcessRun, Usage, WorkflowRun
 from renku.core.models.provenance.qualified import Generation
 from renku.core.models.workflow.run import Run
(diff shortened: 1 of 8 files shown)


mohammad-sdsc (dismissed):
Looks great! Thanks!

mohammad-sdsc (approved):
🚀

PR #1542, opened 2020-09-24
fix: template test cleanup

### LINKED ISSUE #1290: template tests cleanup
After #1272 and #1287 will be merged, it makes sense to move all the template constants used for test (`TEMPLATE_ID`, `METADATA`, ...) from `tests/core/commands/test_init.py` to a fixture in `conftest.py` since they are imported in multiple test files.

lorenzo-cavazzi:
# Description

Template and init tests cleanup.
* move template and init related costants and commands to fixtures
* updated and cleanup test affected by the aforementioned changes

fix #1290

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -191,6 +191,51 @@ def repository(tmpdir):
 
 
 @pytest.fixture()
+def template():
+    """Yield template data."""
+    template = {
+        "url":
+        "https://github.com/SwissDataScienceCenter/renku-project-template",
+        "id": "python-minimal",
+        "index": 1,
+        "ref": "0.1.11",
+        "metadata": {
+            "description": "nodesc"
+        },
+    }
+
+    yield template
+
+
+@pytest.fixture()
+def project_init(template):
+    """Yield template data."""
+    data = {
+        "test_project": "test-new-project",
+        "test_project_alt": "test-new-project-2",
+    }
+
+    commands = {
+        "init": ["init", "."],
+        "init_test": ["init", data["test_project"]],
+        "init_alt": ["init", data["test_project_alt"]],
+        "remote": [
+            "--template-source", template["url"], "--template-ref", template["ref"]
+        ],
+        "id": ["--template-id", template["id"]],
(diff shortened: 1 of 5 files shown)


conftest.py
```diff
@@ -190,6 +190,46 @@ def repository(tmpdir):
...
+        "index": ["--template-index", template["index"]],
+        "force": ["--force"],
+        "list": ["--list-templates"],
+        "parameters": ["--parameter", "p1=v1", "--parameter", "p2=v2"],
+        "parameters_equal_missing": ["--parameter", "p3:v3"],
+        "parameters_equal_early": ["--parameter", "=p4v3"],
+        "confirm": len(set(template["metadata"].keys())) * "\n",
+    }
+
+    yield template, data, commands
```
Panaetius, line 230: You don't need to yield `template`, you can just add it to the arguments of the tests and it will reuse the same fixture.

tests/cli/test_init.py
```diff
@@ -195,47 +209,66 @@ def test_init_on_cloned_repo(isolated_runner, data_repository):
...
+        f'"{ commands["parameters_equal_missing"][1]}"' in result.output
+    )
+
+    result = isolated_runner.invoke(
+        cli, commands["init_test"] + commands["id"] + commands["parameters"] + commands["parameters_equal_early"]
+    )
     assert 0 != result.exit_code
-    assert "Error: Invalid parameter value for --parameter " f'"{PARAMETER_EARLY_EQUAL[1]}"' in result.output
+    assert (
+        "Error: Invalid parameter value for --parameter " f'"{commands["parameters_equal_early"][1]}"' in result.output
```
Panaetius, line 250: Join the two strings to be a single f-string

Panaetius (changes requested):
Looks very nice. Just two minor comments

tests/cli/test_init.py
lorenzo-cavazzi: done, also in another occurence a couple of lines above

Panaetius (approved):

PR #1547, opened 2020-09-28
chore: fix repeat pipfile lock updates

Panaetius:
fixes repeat pipfile lock updates due to hashes being non-deterministic


rokroskar (approved):
I guess this _should_ work!

PR #1549, opened 2020-09-29
feat: add click plugin support

vigsterkr:
As discussed this adds the option to add renku cli subcommands via click plugins.
required for new renku-mls features.

updating calamus version in order to support blank node id generation

### DIFF
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -19,9 +19,12 @@
 Plugin Support
 ==============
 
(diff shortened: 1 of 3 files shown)


vigsterkr:
in case you want to have the calamus dependency change in a separate pr lemme know

Panaetius:
now in https://github.com/SwissDataScienceCenter/renku-python/pull/1604

Issue #1442, opened 2020-08-04
implement a `log` service endpoint 


rokroskar:
# Proposal 

The renku core service should expose a `/log` endpoint that would take a project id and commit range and optionally the format. It should return the graph in the specified format. 

## Rationale

At the moment, the graph service triples-generator installs a version of renku-python and runs `renku log` for each new commit of projects it's monitoring. This is leading to dependency incompatibility issues, complicates deployments and requires the triples-generator service to implement a lot of ad-hoc logic to deal with some peculiarities of the KG building process. Instead, this should be done by the core service, since it's already using this code for other tasks. 

With the current changes in renku-python requiring migration of workflows etc. the generation of triples in the graph service is becoming incredibly slow. We can optimize this quite a bit if we do it "natively" in the python-based service instead. 

## Compatibility

Not much should change for the graph services - the TG should continue to receive events from GitLab and process them, but instead of shelling out to run `renku log` it will send the same information (project, commit range) to the service endpoint. (cc @SwissDataScienceCenter/renku-graph-maintainers )

It's likely that we will need a new/different/adapted middleware in the gateway for this endpoint. Since the request will no longer be coming from the UI there won't be a cookie - but the graph services already have knowledge of some user credentials so it should be possible to use those in this case. 

## Open issues (if applicable)

The addition of this endpoint will greatly increase the load on the core service. We will probably want to make horizontally-(auto)scalable pods of workers for this queue/endpoint in order to respond to load spikes. Since they will only be reading the repository and not performing any writes, it would be possible to avoid using the cache to make scaling easier (we cannot have multi-attach volumes with shared caches at the moment).

jachro:
I quite like the idea as I see things becoming simpler and better structured in terms of responsibility. However, there are few things we’d need to consider/change before we do the switch:
1. Compatibility between renku python on the Core and TG - as of now, TG is quite tightly coupled to a certain version of renku-python as TG does not simply take the output of renku log and sticks in the Triples Store but curates it (e.g. adds info about project hierarchy, decouples Person object, adds additional data for dataset queries). For these TG has to know exactly what version of CLI is used as it has to know how to work and modify the data; Maybe assuring the versions compatibility on the deployment process is good enough?
[...]

PR #1571, opened 2020-09-30
feat(service): added build graph endpoint

jsam:
closes: #1442 

/deploy


renku/service/controllers/graph_build.py
```diff
@@ -0,0 +1,72 @@
...
+        """Renku operation for the controller."""
+        result = migrations_check().build().execute().output
+
+        if not result[1]:
+            raise RenkuException("project not supported")
+
+        with enqueue_retry(GRAPH_JOB_QUEUE) as queue:
+            queue.enqueue(
+                graph_build_job,
+                self.ctx.get("revision"),
```
Panaetius, line 59: `--revision` doesn't actually do what it's name implies and I consider is useless at best.

Firstly, it takes a single revision or a range `a..b`, but in the case of a single revision `rev` if interprets it as `rev..HEAD`, which isn't what we want in 99% of the cases.

Then, it doesn't actually check out that revision and generate triples for it, instead if looks at the paths that were changed in the range mentioned above, and then amounts to calling `renku log` with those paths supplied for the `paths` argument, BUT at the HEAD commit. So if the commit edited e.g. `.renku/datasets/uuid/metadata.yaml`, but the dataset was changed after, it will generate triples for that dataset, but with the state it is currently in.

That's why currently the KG checks out each commit and calls `renku log` instead of just checking out the HEAD and calling `renku log --revision`. If we want to support that usecase, we need to use a different approach here.

Also I think we should just get rid of `--revision` altogether, it causes way more problems than it's worth.

renku/service/jobs/graph.py
```diff
@@ -0,0 +1,127 @@
...
+
+        with chdir(tmpdir):
+            try:
+                command = migrate_project().with_commit().build()
+
+                result = command.execute(skip_template_update=True, skip_docker_update=True)
+                result, _, _ = result.output
+
+                if result:
+                    graph = build_graph(revision=callback_payload["commit_id"])
```
Panaetius, line 96: Same as above about revision. We'd need to checkout `revision` instead, run migrate on top of that, then call `build_graph` without revision.

Panaetius (changes requested):
Looks really good. There's just the unfortunate thing about `--revision` being broken (I hate that flag...)

renku/core/commands/format/graph.py
```diff
@@ -333,16 +333,21 @@ def makefile(graph, strict=False):
...
 
     if strict:
         r, _, t = validate_graph(ld, format="json-ld")
 
         if not r:
             raise SHACLValidationError("{}\nCouldn't get log: Invalid Knowledge Graph data".format(t))
-    click.echo(ld)
+
+    if to_stdout:
+        click.echo(ld)
```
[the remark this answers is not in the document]
jsam: I agree 100% we should clean this up. Only issue is that this is not the only occurrence of it in format module and cleaning entire format module in this PR might be slightly out of scope. I propose we open a ticket for this and then cleanup all of the core click dependency leftovers.

renku/service/controllers/graph_build.py
jsam: Thanks for the extended explanation, I agree. But this specific line is just a request parameter with which client specifies which branch/commit should be checkout and graph build of. See comment bellow.

renku/service/jobs/graph.py
jsam: Done. :rocket:

Panaetius (dismissed):
Thank you! :tada:

renku/core/commands/format/graph.py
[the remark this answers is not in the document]
Panaetius: https://github.com/SwissDataScienceCenter/renku-python/issues/1865

Panaetius (approved):

PR #1573, opened 2020-10-01
fix(core): make Run migration ids unique by relative path instead of absolute

Panaetius:
makes path for id calculation relative instead of absolute.


mohammad-sdsc (dismissed):
LGTM! Nice catch! 🐞

mohammad-sdsc (approved):
👍

PR #1574, opened 2020-10-02
fix(core): fix importing a dataset referenced from non-existent projects

### LINKED ISSUE #1569: Fix dataset import where project doesn't exist anymore
If a dataset is in multiple projects, but one of those projects gets deleted, dataset import functionality will fail with a 404 error because the dataset is still listed as being part of that project.

We should handle this case gracefully and download from a different project if one is available.

For instance https://dev.renku.ch/datasets/428c3626-1c56-463d-8753-336470cc6917/ was part of the repo https://dev.renku.ch/knowledge-graph/projects/lorenzo.cavazzi.tech/Project-11 but that repo got deleted, and now the dataset can't be imported anymore.

Panaetius:
Don't raise an exception if a dataset is referenced from a project in the KG that does not exist on gitlab anymore.

closes #1569

### DIFF
--- a/renku/core/commands/providers/renku.py
+++ b/renku/core/commands/providers/renku.py
@@ -55,15 +55,17 @@ class RenkuProvider(ProviderApi):
         same_as, kg_urls = self._get_dataset_info(uri)
         project_url = None
         failed_urls = []
+        non_existing_projects = []
 
         for kg_url in kg_urls:
             try:
                 kg_datasets_url, ssh_url, https_url = self._get_project_urls(kg_url)
             except errors.OperationError as e:
                 # NOTE: Project was likely deleted, but still referenced in the KG
-                if "project not found" not in e.message:
+                if "project not found" not in str(e):
                     raise
-                failed_urls.append(kg_url)
+                non_existing_projects.append(kg_url)
+                continue
 
             # Check if the project contains the dataset
             if same_as is None:  # Dataset is in the project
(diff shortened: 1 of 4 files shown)


renku/core/commands/providers/renku.py
```diff
@@ -57,7 +57,13 @@ def find_record(self, uri, client=None):
         failed_urls = []
 
         for kg_url in kg_urls:
-            kg_datasets_url, ssh_url, https_url = self._get_project_urls(kg_url)
+            try:
+                kg_datasets_url, ssh_url, https_url = self._get_project_urls(kg_url)
+            except errors.OperationError as e:
+                # NOTE: Project was likely deleted, but still referenced in the KG
+                if "project not found" not in e.message:
+                    raise
```
mohammad-sdsc, line 65: I'd suggest we raise a specific exception (e.g. `ProjectNotFound`) and catch it here instead of relying on error messages (specially because we might not have a test for this case). Using an existing exception (e.g. `DatasetNotFound`) should be ok too since raise and catch points are close.

renku/core/commands/providers/renku.py
```diff
@@ -57,7 +57,13 @@ def find_record(self, uri, client=None):
...
+                failed_urls.append(kg_url)
```
mohammad-sdsc, line 66: It's better to have a separate error list (e.g. `non_existing_projects`) and then have a separate error message for them.

mohammad-sdsc: In this case, it should be possible to have a failing test that tries to import a non-existing project and then check for the proper error message.

mohammad-sdsc (commented):
Some minor comments. It would be good if we can have a failing test for this case.

renku/core/commands/providers/renku.py
mohammad-sdsc: We should also ` continue` at this point to avoid running the rest of the code outside the try/except block.

mohammad-sdsc (approved):
LGTM, thanks!

PR #1575, opened 2020-10-02
feat(cli): show lfs status of dataset files

### LINKED ISSUE #1283: ls-files should show which files are in LFS
ATM, `renku dataset ls-files` shows list of all files and their sizes. It would be nice to also distinguish files that are tracked by LFS (by using a `*` for example).

Panaetius:
Adds an `LFS` column to `dataset ls-files` showing if a file is tracked in LFS or not

Closes #1283


tests/cli/test_datasets.py
```diff
@@ -602,6 +602,42 @@ def test_datasets_ls_files_check_exit_code(output_format, runner, project):
...
+    # NOTE: create some data
+    paths = []
+
+    new_file = tmpdir.join("file_1")
+    new_file.write(str(1))
+    paths.append(str(new_file))
+
+    new_file = tmpdir.join("file_2")
+    new_file.write(str(2) * 200000)
+    paths.append(str(new_file))
```
mohammad-sdsc, line 621: There is also a `large_file` fixture where you can use for this purpose.

mohammad-sdsc (approved):
Loog really good! Just made a comment for tests!

tests/cli/test_datasets.py
Panaetius: Thanks, I didn't know about that one!

mohammad-sdsc (approved):
🚀

jsam (approved):

PR #1576, opened 2020-10-02
fix(dataset): broken directory hierarchy for renku imports

mohammad-sdsc:
# Description

Preserves original dataset's directory hierarchy when importing a renku dataset. If dataset includes files that were not in its data directory, those files will be copied to the root of its new data directory.

Fixes https://github.com/SwissDataScienceCenter/renku-python/issues/1550


Panaetius (approved):
Thank you 🚀

PR #1581, opened 2020-10-05
fix(core): fix project migration getting overwritten with old metadata 

### LINKED ISSUE #1580: Migration overwrites migrated project metadata with old metadata
When migrating a project, the migrations correctly update project metadata in `.renku/metadata.yml`, but the last step of the migrations overwrite this with the incomplete, original metadata from before the migrations ran, in https://github.com/SwissDataScienceCenter/renku-python/blob/master/renku/core/management/migrate.py#L79

This causes `name` to be missing on older projects.

We should reload the project metadata before updating the project version.

Panaetius:
Closes #1580

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,25 @@
 Changes
 =======
 
+`0.11.4 <https://github.com/SwissDataScienceCenter/renku-python/compare/v0.11.3...v0.11.4>`__ (2020-10-05)
+----------------------------------------------------------------------------------------------------------
+
+Bug Fixes
+~~~~~~~~~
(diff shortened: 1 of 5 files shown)


mohammad-sdsc (approved):
Thanks for looking into this! LGTM. It includes some changes from `0.11.3` but that should be ok.

Panaetius:
> Thanks for looking into this! LGTM. It includes some changes from `0.11.3` but that should be ok.

Yeah, for some reason the Changelog was not included in the 0.11.3 hotfix, so I added it here. better late than never 🤷

PR #1582, opened 2020-10-05
fix(dataset): update local files metadata

### LINKED ISSUE #1531: Fix updating dataset with local files
We currently seemingly offer the functionality to add a file to a dataset and then later change the file and update(re-add) it to the dataset by using `renku dataset add --overwrite`.

This does not work in many ways.

**Filesize is not stored in the dataset**

```
$ renku dataset add testset testfile
$ renku dataset ls-files testset
DATASET NAME    ADDED                  SIZE  PATH
--------------  -------------------  ------  --------
testset         2020-09-21 11:59:05    3  B  testfile
[...modify testfile without commiting or running a renku command...]
$ renku dataset ls-files testset
DATASET NAME    ADDED                  SIZE  PATH
--------------  -------------------  ------  --------
testset         2020-09-21 11:59:05    20  B  testfile
```
Here, the file size in the output changes, even though the file was committed or added to the dataset in any way.

**Overwrite-adding the file only changes the label**

```
$ renku dataset add testset testfile
$ cat .renku/datasets/924ff13a-9ed8-4797-9cb3-e66a7d87477f/metadata.yml 
[...]
- '@id': https://localhost/blob/86ffe02d0092e37b2924135de5944d4bf15975d8/testfile
  [...]
  http://schema.org/dateCreated:
  - '@value': '2020-09-21T09:59:05.153679+00:00'
  http://schema.org/name:
  - '@value': testfile
  http://www.w3.org/2000/01/rdf-schema#label:
  - '@value': testfile@86ffe02d0092e37b2924135de5944d4bf15975d8
[...]

[...modify testfile...]
$ renku dataset add --overwrite testset testfile

$ git show ac25c026eb25bf42ff3f7857e470
[...]
--- a/.renku/datasets/924ff13a-9ed8-4797-9cb3-e66a7d87477f/metadata.yml
+++ b/.renku/datasets/924ff13a-9ed8-4797-9cb3-e66a7d87477f/metadata.yml
@@ -12,7 +12,7 @@
   http://schema.org/url:
   - '@value': https://localhost/projects/user/test_datasetaddforce/files/blob/testfile
   http://www.w3.org/2000/01/rdf-schema#label:
-  - '@value': testfile@86ffe02d0092e37b2924135de5944d4bf15975d8
+  - '@value': testfile@64b7eec3edb2d230bafc28cb58551060b6248a9c
   http://www.w3.org/ns/prov#atLocation:
   - '@value': testfile
   https://swissdatasciencecenter.github.io/renku-ontology#external:
```

only the label was changed with the new commit information. The id still has the old commit as part of its path, the dateCreated still has the old creation date.

Looking at the code, this would also not update `source` or `external` on the DatasetFile (`url` also, though I'm not sure if that would change).

Other than the changed file size (which is wrong due to the previously mentioned bug), `renku dataset ls-files` doesn't show any changed data, making it impossible for a user to tell that the dataset was changed.

This whole usecase might need https://github.com/SwissDataScienceCenter/renku/issues/696 to be properly supported, but I think we can do better than we do now already, without implementing that first.

- We can return filesize based on the committed object, not the file in the working directory.
- we can make `add --overwrite` work like `renku dataset unlink && renku dataset add` would work.
- We could also add `commit` to the `--columns` of `renku dataset ls-files`.
- We should investigate that `renku dataset export` actually exports the correct file.

mohammad-sdsc:
# Description

Update metadata after adding and overwriting a local file to a dataset. Adds a `commit` column to show file's commit in `renku dataset ls-files`.

Fixes #1531


Panaetius (dismissed):
This is great, thank you! Also for all the cleanup in the tests!

Panaetius (approved):

PR #1588, opened 2020-10-06
feat(dataset): list dataset description

### LINKED ISSUE #1494: Add "description" to DATASETS_COLUMNS so it can be display
`renku dataset <name> -c <columns>` currently does not support `description` as a column. We should add `description` to DATASETS_COLUMNS so it can be displayed.

mohammad-sdsc:
# Description

Adds an option to list dataset description in CLI. It truncates the description to a maximum of 64*5 characters.

Fixes #1494


renku/core/commands/format/datasets.py
```diff
@@ -27,9 +28,17 @@ def tabular(client, datasets, *, columns=None):
...
         columns = "id,date_created,name,creators,tags,version"
 
+    _create_dataset_short_description(datasets)
+
     return tabulate(collection=datasets, columns=columns, columns_mapping=DATASETS_COLUMNS)
 
 
+def _create_dataset_short_description(datasets):
+    for dataset in datasets:
+        lines = textwrap.wrap(dataset.description, width=64, max_lines=5) if dataset.description else []
```
Panaetius, line 38: Could we calculate the width by using the `COLUMNS` env variable (and falling back to 64 if it's not set)? like `width = $COLUMNS - <width of other columns>`?

mohammad-sdsc: The problem here is that we don't have the width of other columns and `tabulate` cannot set column width as well.

Panaetius (approved):

Issue #1566, opened 2020-09-30
refactor(service): move dataset endpoints to controllers


jsam:
With the new controller abstraction recently introduced for the needs to differentiate remote vs local execution, we need to refactor the service a bit to stay consistent.

This ticket is about moving cache endpoint to the new structure. The following endpoints should be refactored with this ticket:
  * list datasets view
  * list dataset files view
  * add a file to dataset view
  * create dataset view
  * remove dataset view
  * import dataset view
  * edit dataset view
  * unlink file view

PR #1592, opened 2020-10-09
refactor(service): moved edit and unlink endpoints to controllers

jsam:
This PR fixes syncing problems in edit and unlink endpoints and also moves them to the new structure using controllers. In addition, it applies small fixes to activity schema and cwl migrations.

addresses: #1566


tests/service/views/test_dataset_views.py
```diff
@@ -902,7 +902,7 @@ def test_add_remote_and_local_file(svc_client_with_repo):
 
 @pytest.mark.service
 @pytest.mark.integration
-@flaky(max_runs=10, min_passes=1)
+@flaky(max_runs=1, min_passes=1)
```
Panaetius, line 905: left-over from debugging?

Panaetius (changes requested):
Looks great, just one minor comment

tests/service/views/test_dataset_views.py
jsam: Thanks. Fixed.

Panaetius (approved):
🎉

PR #1606, opened 2020-10-12
build: optional homebrew update and limit macos threads

Panaetius:
ignores `brew update` failures
limits macos threads to 2 when running tests


mohammad-sdsc (approved):
LGTM, thanks!

PR #1607, opened 2020-10-13
chore: add missing sentry env

### LINKED ISSUE #1605: chore: make sure sentry init has environment specified
The sentry initialization does not always make use of the environment - it should be set everywhere. It's passed as an environment variable.

rokroskar:
Closes #1605


Panaetius (approved):
:tada:

jsam (changes requested):
Could we also add those variables to `.env-example`?

rokroskar:
@jsam not sure - what should they be set to? I don't want to have the default point to our sentry, we could get some random events from someone else developing the service. What should the environment be in this case?

jsam:
Just leave them blank

```
SENTRY_DSN=
SENTRY_ENV=
```

The point of that file is that we have a complete list of all environment variables used by the service which we can copy and use in docker compose stack or just use as a reference list.

jsam (approved):
tyvm :rocket:

rokroskar:
@jsam done!

PR #1609, opened 2020-10-13
docs: combine docs of renku and renku-python

### LINKED ISSUE #1552: Integrate renku-python docs with main renku docs
It is currently confusing for users to have two readthedocs repos with documentation, the `renku` and `renku-python` ones, with some CLI functionality explained in the main `renku` one and other documentation here. This makes it difficult to know where to look for things.

We should keep the two docs repos separate, but generate the sidebar on the left the same in both projects, so users can seamlessly navigate between the two projects (without even noticing it).

An example of how this can be done can be found here: https://mlbench.readthedocs.io/en/latest/ clicking on the `Core API` menu entry actually links to a separate RTD repository, but since the styling and menu are the same in both repos, a user does not visually notice this. And it allows to easily navigate back and forth between the projects.

These projects use a python file that creates the sidebar and is shared between all projects: https://github.com/mlbench/mlbench-docs/blob/develop/gensidebar.py

Panaetius:
In conjunction with https://github.com/SwissDataScienceCenter/renku/pull/1587

Closes #1552


docs/gettingstarted.rst
```diff
@@ -0,0 +1,55 @@
...
+    limitations under the License.
+
+Getting Started
+===============
+.. _gettingstarted-reference:
+
+Interaction with the platform can take place via the command-line
+interface (CLI).
+
+Start by creating for folder where you want to keep your Renku project:
```
rokroskar, line 25: This can be simplified with

```
renku init my-renku-project
cd my-renku-project
```

I think it was written before `renku init` could take a directory as an argument...

rokroskar (approved):
minor change could be done to simplify the getting started example a little... otherwise 👍

rokroskar (approved):

rokroskar (approved):

Issue #1608, opened 2020-10-13
chore(service): disable RQ from leaking tokens


jsam:
Currently, RQ is leaking tokens to logs. There is an option on RQ builder with which we can disable this again.

PR #1610, opened 2020-10-13
chore(service): disable argument logging on workers

jsam:
closes: #1608


Panaetius (approved):

rokroskar:
Thanks @jsam - I imagine that having this enabled would help debugging in some cases so would it make sense to make it contingent on the desired log level?

jsam:
@rokroskar I think all needed logging will still be there if we activate DEBUG level and if there will be some additional needs we can always look for those arguments in the job metadata, since that metadata will be cached until expiry which is quite some time.

PR #1614, opened 2020-10-15
fix(service): push to protected branches

jsam:


renku/core/commands/save.py
```diff
@@ -97,18 +97,19 @@ def repo_sync(repo, message=None, remote=None, paths=None):
...
     try:
-        # push local changes to remote branch
+        # NOTE: Push local changes to remote branch.
         if origin.refs and repo.active_branch in origin.refs:
             origin.fetch()
             origin.pull(repo.active_branch)
 
-        origin.push(repo.active_branch)
+        result = origin.push(repo.active_branch)
+        if result and "[remote rejected] (pre-receive hook declined)" in result[0].summary:
```
Panaetius, line 106: It's a bit weird that git itself returns something like `! [remote rejected]   master -> master (protected branch hook declined)` but GitPython returns `[remote rejected] (pre-receive hook declined)` without the branch->branch part.

I wonder if there's a case where GitPython might also return the longer message?

jsam: I think GitPython only captures the last part of the message. It's also very strange that they don't raise exception on git error anymore but return the InfoResult object. :shrug:

Panaetius (approved):

PR #1617, opened 2020-10-16
fix(dataset): set isBasedOn for renku datasets

### LINKED ISSUE #1615: Dataset imported from renku are not updated
**Describe the bug**
When a dataset is imported from another renku project and original dataset files are modified, `renku dataset update` does not bring over new content.

**Expected behavior**
Existing files should be updated.

**Additional context**
This happens because `schema:isBasedOn` is missing from imported dataset's files.

mohammad-sdsc:
# Description
Fix updating of renku datasets.

Fixes #1615

### DIFF
--- a/renku/core/commands/dataset.py
+++ b/renku/core/commands/dataset.py
@@ -150,7 +150,6 @@ def add_file(
     sources=(),
     destination="",
     ref=None,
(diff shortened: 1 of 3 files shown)


Panaetius (approved):
:rocket:

PR #1618, opened 2020-10-16
fix(core): fix bug where remote_cache caused project ids to leak

Panaetius:
the `remote_cache` class attribute was getting shared between `LocalClient` instantiations in the service, causing duplicate project `@id`s


jsam (approved):
LGTM. Thanks 🚀

PR #1623, opened 2020-10-20
fix(core): re-raise renku handled exception on network failure

jsam:
Tested with:
  - `doi:10.7910/DVN/5YQDU9`
  - `https://doi.org/10.7910/DVN/WTZS4K`


renku/core/utils/requests.py
```diff
@@ -33,4 +36,7 @@ def retry(total_requests=10, backoff_factor=1, statuses=(500, 502, 503, 504, 429
     _session.mount("http://", HTTPAdapter(max_retries=retries))
     _session.mount("https://", HTTPAdapter(max_retries=retries))
 
-    yield _session
+    try:
+        yield _session
+    except (ConnectionError, RequestException) as e:
```
mohammad-sdsc, line 41: `RequestException` is the base exception class and includes exceptions which are related to connection failure (e.g. `HTTPError`). Maybe here we should catch `ConnectionError` and `Timeout`: https://requests.readthedocs.io/en/master/_modules/requests/exceptions/

jsam: Yea, this evolved without too many reviews on my side. I think this should capture only `RequestException` actually because the problem is that we are not re-raising network issues as renku exceptions. It is much simpler to deal with the exception on the client-side if the client knows that any exception which will bubble up from the renku core module will be of type `RenkuException`, hence the reraise here fixes the problem (I believe).

mohammad-sdsc: > I think this should capture only `RequestException` actually

mohammad-sdsc (approved):
Looks good! Thanks!

PR #1624, opened 2020-10-20
fix(service): raise exception on uninitialized projects

### LINKED ISSUE #1594: renku-core should initialize projects properly when cloning them
If we create a plain project on gitlab, without any renku metadata or anything like that and then activate the KG and add a dataset through the UI, the dataset gets added but project metadata and other renku specific things don't get set up properly, leading to datasets without a project.

Panaetius:
raises an exception with `project_initialization_required` set to true on operations on uninitialized projects.

returns `initialized=False` on `cache.project_clone` endpoint

Closes #1594


jsam (approved):
Look's great! Thank you 🚀

PR #1625, opened 2020-10-21
fix(core): Fix graph building for nodes with same subpath

Panaetius:
fixes a bug where `rerun` would not rerun all steps if one of the output paths of a step was a subpath of an output of a previous step.
E.g. step1 -> `hello` and step2 -> `hello.wc`


mohammad-sdsc (approved):
LGTM, thanks!

PR #1627, opened 2020-10-21
fix(core): fix update creating a commit when showing help

Panaetius:
Fixes a commit being created when `renku update` is called with no arguments.


tests/cli/test_update.py
```diff
@@ -321,3 +321,38 @@ def test_relative_path_for_directory_input(client, run, renku_cli):
     assert 1 == len(cwl.inputs)
     assert isinstance(cwl.inputs[0].consumes, Collection)
     assert "data" == cwl.inputs[0].consumes.path
+
+
+def test_update_no_args(runner, project, renku_cli, no_lfs_warning):
+    """Test automatic file update."""
```
mohammad-sdsc, line 327: I'd suggest: `Test calling update with no args does nothing`.

tests/cli/test_update.py
```diff
@@ -321,3 +321,38 @@ def test_relative_path_for_directory_input(client, run, renku_cli):
...
+    repo = git.Repo(project)
+
+    update_and_commit("1", source, repo)
+
+    exit_code, run = renku_cli("run", "wc", "-c", stdin=source, stdout=output)
+    assert 0 == exit_code
+    assert 0 == len(run.subprocesses)
+
+    with output.open("r") as f:
+        assert f.read().strip() == "1"
```
mohammad-sdsc, line 343: I would remove such intermediate asserts because these are all setup code and we only care about the last assert: `assert before_commit == repo.head.commit`. It's enough to just test the exist code (e.g. `assert 0 == runner.invoke(cli, ["status"]).exist_code`) to avoid cluttering the test.

mohammad-sdsc (commented):
Thanks! Just some comments about (shortening) the test.

mohammad-sdsc (approved):
Thanks Ralf!

PR #1636, opened 2020-10-23
feat(dataset): use Git object hash in ID

### LINKED ISSUE #1568: Commit-independent identifiers for metadata entities
We should use file content hashes (e.g. a git object hash) to identify different versions of entities instead of relying on git commits.

**Note**
* We can store commit information in the metadata or we can retrieve and cache it when needed.
* For entities that do not have a corresponding git object we can use uuid

mohammad-sdsc:
# Description

PoC to use Git object hash instead of commit sha in DatasetFile identifiers.

Fixes #1568

### DIFF
--- a/renku/core/management/datasets.py
+++ b/renku/core/management/datasets.py
@@ -108,12 +108,12 @@ class DatasetsApiMixin(object):
             result[path] = self.load_dataset_from_path(path)
         return result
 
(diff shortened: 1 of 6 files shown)


mohammad-sdsc:
This is already implemented in https://github.com/SwissDataScienceCenter/renku-python/pull/1799 and not needed anymore.

PR #1639, opened 2020-10-27
fix(dataset): error when adding same file multiple times

### LINKED ISSUE #1577: Cannot load a dataset when multiple copies of the same file added from git
When adding the same file from the same revision in a git repo twice, dataset deserialization fails.

**To Reproduce**
1. `renku dataset create remote`
2. `renku dataset add remote --ref 0.3.0 -s Makefile https://github.com/SwissDataScienceCenter/renku-jupyter.git`
3. `renku dataset add remote --ref 0.3.0 -s Makefile -d new-Makefile https://github.com/SwissDataScienceCenter/renku-jupyter.git`
4. `renku dataset (ls)`

**Note**
This happens because two copies of the same `isBasedOn` entity get flattened while they have different `dateCreated` values. Therefore, the value will be an array and cannot be deserialized as DateTime.
A proper solution would be to create a separate class for the `isBasedOn` field (currently it uses `DatasetFile`) and remove meaningless fields for it.
A possible workaround would be to store only one value for `dateCreated` if it is an array.
We should not store `dateCreated` as `None` because SHACL validation will fail.

mohammad-sdsc:
# Description

Fixes a bug where dataset deserialization fails after adding multiple copies of the same file from a git repo.

Fixes #1577

### DIFF
--- a/renku/core/models/calamus.py
+++ b/renku/core/models/calamus.py
@@ -157,6 +157,7 @@ class DateTimeList(fields.DateTime):
         value = normalize_value(value)
 
         if isinstance(value, (list, tuple, set)):
+            value = sorted(value)
             value = value[0] if len(value) > 0 else None
 
         return super()._deserialize(value, attr, data, **kwargs)
--- a/renku/core/models/calamus.py
+++ b/renku/core/models/calamus.py
@@ -146,6 +146,22 @@ class StringList(fields._JsonLDField, marshmallow.fields.String, marshmallow.fie
             raise ValueError("Invalid type for field {}: {}".format(self.name, type(value)))
 
 
+class DateTimeList(fields.DateTime):
+    """A DateTime field that might be a list when deserializing."""
+
+    def __init__(self, *args, **kwargs):
+        """Create an instance."""
+        super().__init__(*args, **kwargs)
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        value = normalize_value(value)
+
+        if isinstance(value, (list, tuple, set)):
+            value = value[0] if len(value) > 0 else None
+
+        return super()._deserialize(value, attr, data, **kwargs)
+
+
 class Nested(fields.Nested):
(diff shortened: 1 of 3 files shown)


renku/core/models/calamus.py
```diff
@@ -146,6 +146,22 @@ def _deserialize(self, value, attr, data, **kwargs):
...
+
+    def __init__(self, *args, **kwargs):
+        """Create an instance."""
+        super().__init__(*args, **kwargs)
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        value = normalize_value(value)
+
+        if isinstance(value, (list, tuple, set)):
+            value = value[0] if len(value) > 0 else None
```
Panaetius, line 160: this would now pick a somewhat random date (Since the list isn't sorted). We short sort the list and have an explicit flag that says if we pick the oldest or latest date. For instance, for a creation date, the oldest makes sense, but for a modification date, the latest makes sense.

tests/cli/test_integration_datasets.py
```diff
@@ -788,6 +788,23 @@ def test_add_data_from_git_with_wildcards(runner, client, params, files):
     assert files == set(os.listdir("data/remote/new"))
 
 
+@pytest.mark.integration
+@flaky(max_runs=1, min_passes=1)
```
Panaetius, line 792: Is flaky needed here? If so, max_runs should be adjusted to a sensible number.

Panaetius (changes requested):
This is the second or third time we ran into a bug related to this and flattening. Should we spend some time investigating if there's other places in the calamus Schemes where this might become an issue?

renku/core/models/calamus.py
mohammad-sdsc: This is used for creation date atm and I am not sure if we want to use it for modification date ever. I sort the values to get the first one but leave the flag implementation when the need comes up.

Panaetius (approved):

PR #1641, opened 2020-10-28
refactor: remove reference context manager

### LINKED ISSUE #1361: Remove ReferenceMixin
Remove `ReferenceMixin` and pass file paths explicitly instead of relying on implicit `__reference__` variable.

mohammad-sdsc:
# Description

Remove implicit context manager for references and pass metadata file paths explicitly.

Fixes #1361


Panaetius (approved):
Thank you! Looks like it wasn't that difficult after all :smile:

mohammad-sdsc:
This whole reference management mechanism was something completely different. There is this `with_reference ` context manage that pushes a path in a thread-safe stack; each of `ReferenceMixin` sub-classes that gets instantiated within the context uses this path as its `__reference__`. I believe it was doing more stuff before we switch to using calamus but atm that's basically it.
The `refs` directory has nothing to do with `ReferenceMixin`. We still need it because it links dataset names to their UUID directories (using symlinks) which is used to lookup datasets by name in LocalClient.

jsam:
Ah, sorry I thought we are getting rid of the whole thing (hence the bigger estimate from my side). Thanks for clarification!

PR #1643, opened 2020-10-28
test(service): added structure for controller unit testing

jsam:
Additionally, adds project create controller unit test.


Panaetius (dismissed):
Thank you very much, this looks great and I think similar tests will be quite an improvement over the current e2e ones :+1:

Panaetius (approved):

PR #1672, opened 2020-11-02
chore: security update for cryptography

jsam:
Updates libraries.


setup.py
```diff
@@ -144,15 +144,15 @@ def run(self):
     "click-completion>=0.5.0,<=0.5.3",
     "click>=7.0,<=7.1.2",
     "click-plugins==1.1.1",
-    "cryptography>=2.7,<=2.9.2",
+    "cryptography>=3.2",
```
Panaetius, line 147: we should always limit the upper bound of version to prevent upstream libraries from adding breaking changes in the future, even if it means doing something like `>=3.2,<=3.2` (dependabot will make that make sense once there's a new version)

Panaetius (changes requested):

Panaetius (approved):
Thank you!

Panaetius (approved):

PR #1675, opened 2020-11-02
fix(dataset): deserialization error

mohammad-sdsc:
# Description

Fixes an issue where multiple project instances in the same dataset metadata file causes a deserialization error. Also removes blank node ids for non-existing files.


Panaetius (approved):
Thank you, looks great!

PR #1685, opened 2020-11-03
refactor(core): change pass_local_clients internals to Command builder

### LINKED ISSUE #1648: Change pass_local_client to use new Command builder under the hood
`pass_local_client` should use command builder internally.

Panaetius:
Adds Command builder inside pass_local_client

Closes #1648


renku/core/commands/client.py
```diff
@@ -66,41 +59,30 @@ def pass_local_client(
...
-            client = ctx.ensure_object(LocalClient)
+        cmd = Command().command(method)
 
-        stack = contextlib.ExitStack()
+        if not ignore_std_streams:
+            cmd = cmd.track_std_streams()
 
-        # Handle --isolation option:
-        if get_git_isolation():
-            client = stack.enter_context(client.worktree())
```
mohammad-sdsc, line 80: What happens to isolation then? If we don't want to support, we should remove the command line option and update the docs.

mohammad-sdsc (commented):
LGTM. I've made a comment about isolation support.

renku/core/commands/client.py
Panaetius: It was in the `Command` class, but I move it around to make it more explicit.

Honestly, `--isolation` is one the features in renku that I don't fully understand (how it works/when it is useful). So I wouldn't feel comfortable making a decision to drop it or not. We could open an issue to investigate.

mohammad-sdsc (approved):
Thanks!

PR #1691, opened 2020-11-04
fix(service): correctly set project_name slug on project create

jsam:


renku/core/models/git.py
```diff
@@ -105,13 +105,16 @@ def __attrs_post_init__(self):
 
     @classmethod
     def parse(cls, href):
-        """Derive basic informations."""
+        """Derive URI components."""
+        if href.isascii() is False:
```
mohammad-sdsc, line 109: Is there is reason for not using `if not href.isascii()`?

renku/core/utils/scm.py
```diff
@@ -19,9 +19,19 @@
 import re
 
 
-def strip_and_lower(input):
+def normalize_to_ascii(input_string, sep="-"):
     """Adjust chars to make the input compatible as scm source."""
-    return re.sub(r"\s", r"-", input.strip()).lower()
+    return (
+        f"{sep}".join(
```
mohammad-sdsc, line 25: sep is a string so `f"{sep}"` can be replaced with `sep`.

mohammad-sdsc (approved):
LGTM, thanks! I've made two minor comments, feel free to ignore!

mohammad-sdsc (approved):
🚀

lorenzo-cavazzi:
This comes late but the logic for deriving slugs is still slightly different from the GitLab.
Specifically, multiple separators `-` should be collapsed.

You can verify this in the preview here: https://lorenzo.dev.renku.ch/projects/new
A title like `test --üäüaua` should create a slug like `test-aua` instead of `test----aua`

PR #1693, opened 2020-11-04
docs: pipx installation

tcchrist:
# Description

I think you should adjust the instruction about the installation of pipx:
The current instructions use the following command to ensure pipx is included in $PATH:

`pipx ensurepath`

I think the correct way to do this is the following:

`python3 -m pipx ensurepath`

Also, this is how the official pipx documentation suggests it.
(https://github.com/pipxproject/pipx#install-pipx)

## Type of change
docs

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -84,7 +84,7 @@ and make sure that the ``$PATH`` is correctly configured.
 ::
 
     $ python3 -m pip install --user pipx
-    $ pipx ensurepath
+    $ python3 -m pipx ensurepath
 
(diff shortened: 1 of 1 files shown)


Panaetius:
Thank you for bringing this to our attention!

Since our test suite doesn't run on PRs created from external repositories, I'll move this PR over so we can merge it.

Panaetius:
Closing in favor of https://github.com/SwissDataScienceCenter/renku-python/pull/1693

PR #1694, opened 2020-11-05
refactor(dataset): use command builder

### LINKED ISSUE #1650: Switch dataset commands to use Command builder
Dataset commands should use the Command builder directly instead of being decorated with `pass_local_client`.

mohammad-sdsc:
# Description

Refactor dataset commands to use command builder pattern

Fixes #1650


Panaetius:
My original thought was that we expose the command builder in `renku.core` and then build&execute the command in `renku.cli` or `renku.service` as necessary.

```
# in core:
def _add_files_to_dataset(...):
    [...]

add_files_to_dataset = Command().command(_add_files_to_dataset).with_commit(...).require_migration()


#################################################
# in cli:
from renku.core.datasets import add_files_to_dataset

@dataset.command()
@click.argument("name")
[...]
def add(name, urls, external, force, overwrite, create, sources, destination, ref):
    """Add data to a dataset."""
    command = (add_files_to_dataset
        .with_click_communication() #echos output/warnings to stdout/err
        .lock_project().build())
    result = command.execute(name, [...])
    if result.status = CommandResult.FAILURE:
        raise result.error
    click.secho("OK", fg="green")


########################################################
# in service
from renku.core.datasets import add_files_to_dataset

class DatasetsAddFileCtrl(ServiceCtrl, ReadWithSyncOperation):
    """Controller for datasets add endpoint."""

    def __init__(self, cache, user_data, request_data):
        [...]

    def renku_op(self):
        """Renku operation for the controller."""
        command = (add_files_to_dataset
            .with_service_communication(RESPONSE_SERIALIZER)
            .lock_dataset([...])
            .build()
       result = command.execute(self.name, [...])
       return result
```

renku/core/incubation/command.py
```diff
@@ -123,20 +123,23 @@ def execute(self, *args, **kwargs):
 
         output = None
         error = None
+
         try:
             with context["stack"]:
                 output = context["click_context"].invoke(self._operation, context["client"], *args, **kwargs)
         except errors.RenkuException as e:
             error = e
+            if raise_exceptions:
```
Panaetius, line 132: The original idea was that the post hooks can handle exceptions if they want (and even remove them from the result if they don't matter). This is important for e.g. being able to reset the working directory to the previous state on error to not leave it dirty and things like that.

But I see the need to raise exceptions in general, especially unhandled ones. It probably makes sense to have a post-hook that raises any unhandled exception and runs as the last hook of the whole chain. 

What do you think?

renku/core/management/datasets.py
```diff
@@ -460,18 +450,24 @@ def _add_from_local(self, dataset, path, external, destination):
             }
         ]
 
-    def _add_from_urls(self, urls, destination, destination_names, extract, progress):
+    def _add_from_urls(self, urls, destination, destination_names, extract):
+        listeners = communication.get_listeners()
+
+        def add_from_url(function, **kwargs):
```
Panaetius, line 456: A name like `subscribe_communication_listeners` would be more clear.

Panaetius (changes requested):
Thank you very much, this is a big step for being more flexible and moving away from the monolithic client.

I added some mostly minor change requests.

The DatasetLock is intended to be a dummy to be implemented properly later on?

renku/core/commands/dataset.py
```diff
@@ -414,69 +369,34 @@ def export_dataset(
...
-        yes=yes,
-        commit_message=commit_message,
-        progress=progress,
-        previous_dataset=previous_dataset,
-        delete=delete,
-    )
+def export_dataset():
+    """Command for exporting a dataset to 3rd party provider."""
+    command = Command().command(_export_dataset).lock_dataset()
+    return command.require_migration().require_clean().with_commit(commit_only=DATASET_METADATA_PATHS)
```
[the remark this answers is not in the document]
mohammad-sdsc: 🤦

Panaetius (approved):
Thank you!

renku/core/incubation/command.py
mohammad-sdsc: That's a great idea! I added a _post_hook to do this. However, I've changed the order that post-hooks execute to be descending, so when command's pre-hook runs first it's post-hook runs last. Let me know if you think it should not be done in this way.

renku/core/commands/dataset.py
```diff
@@ -204,7 +164,7 @@ def _add_to_dataset(
     if sources and len(urls) > 1:
         raise UsageError('Cannot use "--source" with multiple URLs.')
 
-    if interactive:
+    if communication.has_prompt():
```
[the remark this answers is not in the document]
mohammad-sdsc: TBH, I did not want to break service unexpectedly but what you say make sense to me. I've enabled this check for service as well.

mohammad-sdsc:
> The DatasetLock is intended to be a dummy to be implemented properly later on?

Right, implementing it does not belong to this story. But it makes sense to use the right lock type in the dataset so that when we implement it, we won't change dataset commands.

Panaetius (dismissed):

Panaetius (approved):

PR #1695, opened 2020-11-05
fix(service): correctly determine resource age

jsam:


renku/service/cache/models/file.py
```diff
@@ -52,6 +51,15 @@ def abs_path(self):
         """Full path of cached file."""
         return CACHE_UPLOADS_PATH / self.user_id / self.relative_path
 
+    @property
+    def age(self):
+        """Returns file's age in seconds."""
+        # NOTE: `created_at` field is aligned to UTC timezone.
+        created_at = (self.created_at - datetime.utcfromtimestamp(0)).total_seconds()
+        utc_since = (datetime.utcnow() - datetime.utcfromtimestamp(0)).total_seconds()
+        age = int(utc_since - created_at)
```
mohammad-sdsc, line 60: Aren't the three above lines the same as `int((datetime.utcnow() - self.created_at).total_seconds())`?

renku/service/cache/models/project.py
```diff
@@ -52,6 +51,15 @@ def abs_path(self):
         """Full path of cached project."""
         return CACHE_PROJECTS_PATH / self.user_id / self.owner / self.name
 
+    @property
+    def age(self):
+        """Returns project's age in seconds."""
+        # NOTE: `created_at` field is aligned to UTC timezone.
+        created_at = (self.created_at - datetime.utcfromtimestamp(0)).total_seconds()
+        utc_since = (datetime.utcnow() - datetime.utcfromtimestamp(0)).total_seconds()
+        age = int(utc_since - created_at)
```
mohammad-sdsc, line 60: Same as above.

mohammad-sdsc (commented):
LGTM, just a single comment.

renku/service/cache/models/file.py
jsam: You are absolutely right. Don't know what I was thinking. :eyes:

mohammad-sdsc (approved):
LGTM, thanks!

PR #1696, opened 2020-11-05
test(service): support more cases for project_name

jsam:


lorenzo-cavazzi (approved):
Works well now! :rocket: 

If anyone wants to try a preview: https://lorenzotest.dev.renku.ch/projects/new

rokroskar:
So we're just removing non-ascii characters and replacing them with `-`? Not doing character translation?

jsam:
@rokroskar Yes. We are doing exactly the same as GitLab does when you create a new project.

lorenzo-cavazzi (approved):
I tested this a bit while developing the UI side SwissDataScienceCenter/renku-ui#1115 and it works well :+1: 

It's currently used in this UI preview: https://lorenzotest.dev.renku.ch/

mohammad-sdsc (approved):
Thanks 🚀

PR #1701, opened 2020-11-09
refactor(cli): use command builder

### LINKED ISSUE #703: Decouple workflow commands
With #676 first step was done to decouple implementation from click command but workflow commands have not been fully refactored/decoupled. To enable usage of core renku commands in service, those commands need to be decoupled.

### LINKED ISSUE #1651: Switch run/rerun/update/show/status/log commands to use Command builder
Workflow commands (run/rerun/update/show/status/log) should use the Command builder directly instead of being decorated with `pass_local_client`.

mohammad-sdsc:
# Description

Refactor commands to use Command builder pattern.

Fixes #703 
Fixes #1651


renku/core/commands/run.py
```diff
@@ -0,0 +1,155 @@
...
+        if factory.warnings:
+            communication.echo(factory.warnings)
+
+    finally:
+        if system_stdout:
+            sys.stdout = old_stdout
+            system_stdout.close()
+        if system_stderr:
+            sys.stderr = old_stderr
+            system_stderr.close()
```
Panaetius, line 155:
```suggestion
        if system_stdout:
            system_stdout.flush()
            sys.stdout = old_stdout
            system_stdout.close()
        if system_stderr:
            system_stderr.flush()
            sys.stderr = old_stderr
            system_stderr.close()
```

Not really related to this PR, but I noticed sometimes renku errors will still get written to file instead of printed (in `renku run echo 'test' > test.txt` you could end up with a renku error in test.txt instead of printed in the terminal).

I think this could be solved by flushing `system_stdout` and `system_stderr` here, so would be nice if we could add that.

Panaetius (dismissed):
Thank you! I have a comment about an improvement but we can also do that in a separate issue.

Panaetius (approved):

PR #1708, opened 2020-11-10
fix(service): set template version and metadata correctly

Panaetius:
Sets template version, immutable_files and automated_update correctly when creating a project through the service.

Fixes bug where `progress_callback` was instead interpreted as `check_only` on template/docker_update methods.

Adds support for template update handling template files deleted by the user.

Fixes `renku check-immutable-template-files` command.

Adds additional tests


jsam (approved):
LGTM. Thank you :rocket:

renku/cli/__init__.py
```diff
@@ -196,6 +196,7 @@ def help(ctx):
 cli.add_command(init_command)
 cli.add_command(log)
 cli.add_command(migrate)
+cli.add_command(check_immutable_template_files)
```
mohammad-sdsc, line 199: We should add `hidden=True` in `migrate.py` for `check_immutable_template_files` to hide it from normal help output.

mohammad-sdsc (commented):
LGTM, thanks! Just a minor comment.

renku/cli/__init__.py
Panaetius: 👍 
Done

mohammad-sdsc (approved):
Thanks!

PR #1714, opened 2020-11-11
chore: ascii check compatible with py36

jsam:


Panaetius (commented):
Thank you! All tests pass on 3.6 now

Panaetius (approved):

rokroskar:
3.6 is unsupported in 12 months: https://www.python.org/dev/peps/pep-0494/#id28 - I thought we support >= 3.7 anyway?

jsam:
@rokroskar Technically, we still support it cause some distros of ubuntu which launched with 3.6 are still on the LTS. We can reconsider this and drop the support, I'm all for it.

Issue #1043, opened 2020-02-21
Graph paradigm shift - store it, don't build it


rokroskar:
Our current implementation rebuilds the graph on demand for every renku command that requires it. Because of the nature of the graph this requires expensive git tree traversals and a lot of file i/o. 

The proposal is to store the graph on disk instead - this way, all the renku commands would query the generated graph instead of re-creating it first. The graph service could also consume this data directly, greatly speeding up the graph building.

jsam:
This ticket is same as  #864, what we are missing is not the general idea but more concrete implementation implications for this. I propose we use one of those 2 tickets as a more general breakdown which is needed to achieve this.

Panaetius:
I think most of the pieces are in place to really start working on the new graph building. We can do this while still keeping `renku log` the same.

The steps as I see them are:
- Store metadata in a single file
- Use that single file for our commands
- Store metadata from all commits in the single file
- Add a command similar to log that queries this file instead of walking commits

we can do this sequence independently for workflows and datasets (not sure if it's needed for Project metadata? might need to create issues for that as well). Then for the last step we can join/concat the implementations to get a new `renku log` (`renku graph dump` ?) that works solely on the 2-3 single files (one for datasets, 1-2 for provenance/workflows).
Once that works well, we can retire `renku log` and remove old metadata files in a further migration.

Panaetius:
Done :tada:

PR #1723, opened 2020-11-13
feat(api): parameters

### LINKED ISSUE #1560: Specify explicit parameters from inside python code
Users should be able to declare parameters inside a script as such, without passing them on the command line. This would later allow us to show them on the KG and open other use-cases like showing changes to parameters between runs.

This might take the form of 
```
lr = renku.parameter('learning_rate', 0.9)
```

mohammad-sdsc:
# Description

Documentation for API.
Allows defining run parameters for workflows.

Fixes #1560

### DIFF
--- a/renku/core/models/workflow/converters/cwl.py
+++ b/renku/core/models/workflow/converters/cwl.py
@@ -279,6 +279,34 @@ class CWLConverter(object):
         for argument in step.arguments:
             tool_object.inputs.append(CWLConverter._convert_argument(argument))
 
+        workdir_req.listing.append(
+            cwlgen.InitialWorkDirRequirement.Dirent(
+                entry="$(inputs.input_renku_metadata)", entryname=".renku", writable=False
+            )
+        )
+        tool_object.inputs.append(
+            cwlgen.CommandInputParameter(
+                "input_renku_metadata",
+                param_type="Directory",
+                input_binding=None,
+                default={"path": os.path.abspath(os.path.join(basedir, ".renku")), "class": "Directory"},
+            )
+        )
+
+        workdir_req.listing.append(
+            cwlgen.InitialWorkDirRequirement.Dirent(
+                entry="$(inputs.input_git_directory)", entryname=".git", writable=False
+            )
+        )
+        tool_object.inputs.append(
+            cwlgen.CommandInputParameter(
+                "input_git_directory",
+                param_type="Directory",
+                input_binding=None,
+                default={"path": os.path.abspath(os.path.join(basedir, ".git")), "class": "Directory"},
+            )
+        )
+
         if workdir_req.listing:
             tool_object.requirements.append(workdir_req)
         if jsrequirement:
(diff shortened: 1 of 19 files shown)


renku/core/models/workflow/converters/cwl.py
```diff
@@ -279,6 +279,34 @@ def _convert_step(step, tmpdir, basedir, filename=None):
...
+                "input_renku_metadata",
+                param_type="Directory",
+                input_binding=None,
+                default={"path": os.path.abspath(os.path.join(basedir, ".renku")), "class": "Directory"},
+            )
+        )
+
+        workdir_req.listing.append(
+            cwlgen.InitialWorkDirRequirement.Dirent(
+                entry="$(inputs.input_git_directory)", entryname=".git", writable=False
```
Panaetius, line 298: Did you check what the performance implication is of copying over the `.git` directory for each workflow rerun? Especially on large projects.

renku/core/models/workflow/parameters.py
```diff
@@ -265,6 +265,53 @@ def as_jsonld(self):
...
+
+    value = attr.ib(default=None, type=str, kw_only=True)
+
+    type = attr.ib(default=None, type=str, kw_only=True)
+
+    @staticmethod
+    def generate_id(run_id, name):
+        """Generate an id."""
+        name = urllib.parse.quote(name, safe="")
+        return "{}/arguments/{}".format(run_id, name)
```
Panaetius, line 286: should we change the id to `[...]/parameter/[...]` ?

Panaetius (changes requested):
This looks really nice, thank you! I have a minor change request, otherwise just some questions.

Right now a user could not change the value of a Parameter on `renku update`/`rerun`, correct? Only by modifying the script, but not e.g. passing it on the cmdline `renku update --param param-2 41`? Should we add something like this? It might be error prone if two scripts define `param-2` but then again, maybe a user wants to set the same value in multiple places.

renku/core/models/workflow/converters/cwl.py
mohammad-sdsc: It creates only a symlink and does not do an actual copy. This is not needed once we can load metadata without traversing git history. I'll add a TODO to remove when https://github.com/SwissDataScienceCenter/renku-python/issues/1043 is done.

renku/core/models/workflow/parameters.py
mohammad-sdsc: Was waiting for a second thought on that 😄

mohammad-sdsc:
> Right now a user could not change the value of a Parameter on `renku update`/`rerun`, correct? Only by modifying the script, but not e.g. passing it on the cmdline `renku update --param param-2 41`? Should we add something like this? It might be error prone if two scripts define `param-2` but then again, maybe a user wants to set the same value in multiple places.

I believe this requires more thought and I prefer to keep the use-case limited to API for this story. I'll create a follow up story to enable users to pass parameters via command line.

Panaetius (approved):
Thank you!

PR #1730, opened 2020-11-17
fix(core): pre-commit hook message for unsupported project

### LINKED ISSUE #1697: Disable pre-hook check for unsupported project version
**Describe the bug**
When trying to commit in a project that is modified by a more recent version of renku than what is currently installed, we get a meaningless error message:
```
You are trying to update generated files.

Modified files:
Error: Project is not supported by this version of Renku.
Upgrade to the latest version of Renku.

To commit anyway, use "git commit --no-verify".
```

**To Reproduce**
1. Go to a renku project
2. Edit `.renku/metadata.yaml` and increase `schemaVersion`
3. `git commit -am 'updated project version'`

**Additional context**
This happens because a git pre-hook executes renku which fails due to newer project version. We need to skip this pre-hook check if project is unsupported by the installed renku version.

mohammad-sdsc:
# Description

Print proper message in git pre-commit hooks when project version is unsupported or requires a migration.

Fixes #1697

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -23,6 +23,7 @@ import pathlib
 import re
 import secrets
 import shutil
+import subprocess
 import tarfile
 import tempfile
 import time
@@ -33,14 +34,13 @@ from copy import deepcopy
 from pathlib import Path
 
 import fakeredis
-import git
 import pytest
 import requests
 import responses
 import yaml
 from _pytest.monkeypatch import MonkeyPatch
 from click.testing import CliRunner
-from git import Repo
+from git import GitCommandError, Repo
 from walrus import Database
 
 from tests.utils import make_dataset_add_payload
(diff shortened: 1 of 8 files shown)


renku/data/pre-commit.sh
```diff
@@ -40,6 +40,12 @@ fi
```
Panaetius, line 40: It'd make sense to add explicit commands that check if a project is newer or outdated and check for it here, after the check if renku is installed. Something like `renku project-status` or `renku migrate --check` that returns a message regarding the project state and an exit code that goes with it (like 2= migration required, 3= project too new).
Then we can check for the exit code and write a message "Project metadata is outdated, migrate it with `renku migrate`" or "This project was created with a newer version of renku, update renku to continue working on it" or something to that effect.

Just to make it more explicit to the user what is happening and what steps they need to take to solve the problem.

Panaetius (changes requested):

renku/data/pre-commit.sh
mohammad-sdsc: I've separated outdated and unsupported cases and also added `renku migrate --check`. However, I did not use it in the pre-hook because running another renku command there would slow-down commits.

conftest.py
```diff
@@ -604,40 +586,53 @@ def old_workflow_project(request, tmp_path_factory):
...
 
     repo_path = working_dir / name
     repo = Repo(repo_path)
 
-    os.chdir(repo_path.strpath)
+    os.chdir(repo_path)
     yield repo
 
-    shutil.rmtree(repo_path.strpath)
+    shutil.rmtree(repo_path)
```
Panaetius, line 619: I think we should chdir back to the original directory to make sure we end up in a good state.

mohammad-sdsc: I've changed them all to use the `chdir` context manager. However, I believe working directory gets reset at the beginning of the each test and this should not matter much for function-scoped fixtures.

Panaetius (approved):

PR #1738, opened 2020-11-17
feat(cli): better error message for renku clone

### LINKED ISSUE #1715: support renku clone over SSH
**feature (problem related)**
If you join renku using a auth provider such as SWITCH edu-ID or GitHub, you most probably won't have set a password in your renku gitlab account. If you then try to clone a renku project using `renku clone` you'll receive a generic exception with a long stack trace.

I think the failed `renku clone` command should provide a clean error message, similar to the one you get by using "git glone".
Irregardless of this, I think the default way to access git repositories should be SSH not HTTPS. Unfortunately the `renku clone` command seems not to support git clone via SSH. 

The current workaround is to not use renku cli and just use git clone, followed by the lfs setup / checkout.
However, this somehow vitiate the purpose of the renku cli to simply these steps for untrained users.

mohammad-sdsc:
# Description

Print clean git error message when `renku clone` fails.

Fixes #1715

### DIFF
--- a/renku/core/management/clone.py
+++ b/renku/core/management/clone.py
@@ -56,7 +56,10 @@ def clone(
         repo = Repo.clone_from(url, path, recursive=recursive, depth=depth, progress=progress)
     except GitCommandError as e:
         if not raise_git_except:
-            raise errors.GitError("Cannot clone remote Renku project: {}".format(url)) from e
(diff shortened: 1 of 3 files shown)


Panaetius (approved):
Thank you, looks great!

PR #1744, opened 2020-11-19
feat(core): add communication to command builder

### LINKED ISSUE #1658: Add logging and callback functionaliy to Command builder
With the new command builder, we can potentially improve how we handle messaging and callback quite a bit.

We could have a builder for click logging (Info/Error/Warning) that can be set when building a Command for the cli and gets injected where needed (ideally without having to be passed around manually). And then in the service there can be a different builder that does the same, but in a service compatible way. 

User prompts could potentially be handled this way as well.

We might use this to log to sentry as well.

mohammad-sdsc:
# Description

Includes a common interface for logging and interacting with users.

Fixes #1658 
Partially includes https://github.com/SwissDataScienceCenter/renku-python/pull/1225


Panaetius (approved):
Thank you, looks great! Glad we can finally solve this in a nice way.

And really happy the ThreadLocal solves the issue with threading so neatly!

Panaetius (approved):

PR #1746, opened 2020-11-20
fix(core): correctly generate project id for gitlab (sub)groups

### LINKED ISSUE #1640: Projects in gitlab subgroups do not have correct URL
**Describe the bug**
When a project is created in a Gitlab subgroup, project's url is not correct.

**To Reproduce**
1. Create a group and subgroup in gitlab
2. Create a project under this subgroup on dev
3. Look at project ID in `.renku/metadata`

**Expected behavior**
A project created under `group/sub-group` should have a URL like `https://dev.renku.ch/projects/group/sub-group/project` but the URL is `https://dev.renku.ch/projects/sub-group/project`.

**Note**
We use project's ID as project's URL so for fixing this issue one option is to fix the ID.

Panaetius:
Closes #1640


mohammad-sdsc (approved):
LGTM, thanks!

PR #1747, opened 2020-11-20
chore: install container w/o dirty repo; run as non-root

### LINKED ISSUE #1743: chore: fix the renku-python version used in the core service
At the moment, because the core service docker container runs a command that modifies the state of the repo, the version of the service is `0.x.y-dirty` - this looks particularly bad now with migrations implemented in the UI - see, for example:

image

There is no need to pollute the repo in the container - we can either not use the requirements builder or we can export the requirements file to a temporary path outside the repo before installing.

rokroskar:
* fixes #1743 by saving the requirements file in a temporary directory instead of in the repository directory.

* refactors the service requirements in setup.py so they are not installed in the CLI

* side benefit - runs the app as a non-privileged user.

### DIFF
--- a/Dockerfile.svc
+++ b/Dockerfile.svc
@@ -1,4 +1,4 @@
-FROM python:3.7-alpine
+FROM python:3.7-alpine as base
 
 RUN apk add --update --no-cache alpine-sdk g++ gcc linux-headers libxslt-dev python3-dev build-base openssl-dev libffi-dev git bash && \
     pip install --no-cache --upgrade pip setuptools pipenv requirements-builder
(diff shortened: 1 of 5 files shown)


Dockerfile.svc
```diff
@@ -8,15 +8,22 @@ RUN apk add --no-cache --allow-untrusted \
     --repository http://dl-cdn.alpinelinux.org/alpine/latest-stable/main \
     --repository http://nl.alpinelinux.org/alpine/edge/community \
     git-lfs && \
-    git lfs install
+    git lfs install && \
+    addgroup -g 1000 shuhitsu && \
+    adduser -S -u 1000 -G shuhitsu shuhitsu && \
```
Panaetius, line 13: 🤣 @ shuhitsu

Dockerfile.svc
```diff
@@ -1,4 +1,4 @@
-FROM python:3.7-alpine
+FROM python:3.7-alpine as base
```
Panaetius, line 1: You don't use `base` anywhere, is this needed?

Dockerfile.svc
rokroskar: our notebook service is kyaku ;)

Dockerfile.svc
rokroskar: no, I was trying a few different things and this was left over, thanks!

Panaetius (dismissed):
Thank you 🚀

Panaetius (dismissed):

Panaetius (approved):

PR #1749, opened 2020-11-23
feat(core): shorten commit messages to 100 characters for readability

### LINKED ISSUE #1633: Overwrite default commit when adding files in bunch to renku
`renku dataset add` on a folder automatically commits the updated dataset, with a commit message that list all the files in the folder. We don't have the option to edit the commit.

If the folder contains many files, the commit message can be very long and break the image build with an `argument line too long` error.

I also suspect that it slows down the jupyterlab interface, in particular its git plugin.

I'd like a `-m` option that allows me to overwrite the default commit message.

If the option exists, it should be documented in the `renku dataset add --help`.

* Renku should by default complain and ask me if I want to edit the commit if it exceeds a reasonable size unless I specify a _batch_ mode option.

* If we really want to keep the long commit message, we could append the longer message to our own log file and cross-reference it with the git logs (e.g. using commit IDs). At least that would isolate those long messages from the git log, so that they don't break the build or cause other issues.

**Additional context**

At the moment I must either add the file one at the time, which is very slow, or tune the command in order to add the files in larger batch while respecting a reasonable commit size, or rebase/amend and reword the commit messages.

Panaetius:
closes #1633


rokroskar (commented):
This is a great improvement, but aren't most recommendations (from e.g. github) to keep the first line to 50 characters?

Panaetius:
hmm I don't know about that. I used 100 chars in `renku save` and wanted to keep thibgs consistent. And something like `renku dataset add mydataset` is already 28 characters, leaving not that much room for arguments.

rokroskar:
I guess some places use 72 characters. Is it really more useful to have 100 characters vs 50 when it's truncated anyway? You don't see the full list either way and have to use some other git command to get the complete file listing.

Panaetius:
Well the more characters there are the more hints you have to remind you what you did. Like `renku dataset import mydataset https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/PQX0XG` with 50chars is `renku dataset import mydataset https://dataverse.`, with 100 `renku dataset import mydataset https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910` but only the full message would be useful in knowing which dataset was imported.

I don't have a strong opinion either way, I can change it to 50 if you want.

rokroskar:
Well, I always thought we could do a short and sweet first line and then add additional info to the rest of the commit message. In the case of files being added, I would only include the first couple of files e.g. something like

```
renku dataset add <datasetname> file1, file2...

Files added:

file1
file2
file3
file4
file5
...
(1005 additional files, use "git show --name-only <commit>" for the full list)
```

rokroskar:
I realize the current solution is already a nice improvement and my suggestion could be implemented as a follow-up. I think the length should be whatever renders nicely in gitlab, i.e. to make sure that it doesn't get clipped here:

image

rokroskar (approved):

PR #1753, opened 2020-11-24
chore: fix dirty version on chartpress build

rokroskar:
This is hopefully the last in the series of "quick fix to dirty version" PRs. 

The solution was a bit more involved than previously thought - because we use `chartpress` to publish our images and because `chartpress` modifies some files in the repo before building the image, it will necessarily result in a "dirty" version of the installed package. I circumvent this here by introducing a docker build argument which can be set to in order to `git reset --hard` the repo before the installation in the `Dockerfile`. The default behavior is still to build with the local (uncommitted) changes in order to not surprise any unsuspecting developers who might be otherwise left wondering what happened to their code in the docker image. 

To test:

```
$ chartpress
$ docker run --rm -ti --entrypoint bash <image> renku --version
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,8 +6,6 @@ __pycache__
 
 build
 dist
-docs
-tests
 
 *.mo
 *.pyc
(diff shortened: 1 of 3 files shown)


Panaetius (approved):
🎉  thanks!

PR #1754, opened 2020-11-24
fix: use project_id as part of project filesystem path

jsam:
This PR changes old caching path of 

```
CACHE_PROJECTS_PATH / self.user_id / self.owner / self.name
```

to 

```
CACHE_PROJECTS_PATH / self.user_id / self.project_id / self.owner / self.name
```

This improves couple of things:

- Prevents potential concurrent problems (in case of errors) which could arise due to the nature of cloning the same project by the same user multiple times in parallel. 
- Easier debugging time. Since project_id is the identifier for the Project model instance and contains project metadata stored in the cache, having this reference also on the filesystem pointer makes debugging concurrent problems easier.
- If the project got deleted and re-cloned it will have new generated `project_id` and this property will be visible immediately by inspecting the filesystem.

### DIFF
--- a/renku/service/controllers/cache_list_projects.py
+++ b/renku/service/controllers/cache_list_projects.py
@@ -40,7 +40,11 @@ class ListProjectsCtrl(ServiceCtrl, ReadOperationMixin):
     def list_projects(self):
         """List locally cache projects."""
         projects = [project for project in self.cache.get_projects(self.user) if project.abs_path.exists()]
-        return {"projects": projects}
+        projects.sort(key=lambda p: p.created_at)
+
+        latest = {p.git_url: p for p in projects}
+
+        return {"projects": list(latest.values())}
 
     def renku_op(self):
         """Renku operation for the controller."""
--- a/renku/service/controllers/cache_list_projects.py
+++ b/renku/service/controllers/cache_list_projects.py
@@ -16,6 +16,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Renku service cache list cached projects controller."""
+import itertools
+
 from renku.service.controllers.api.abstract import ServiceCtrl
 from renku.service.controllers.api.mixins import ReadOperationMixin
 from renku.service.serializers.cache import ProjectListResponseRPC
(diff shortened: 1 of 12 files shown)


Panaetius (approved):
Thank you 🚀

renku/service/controllers/cache_list_projects.py
```diff
@@ -40,7 +40,11 @@ def context(self):
     def list_projects(self):
         """List locally cache projects."""
         projects = [project for project in self.cache.get_projects(self.user) if project.abs_path.exists()]
-        return {"projects": projects}
+        projects.sort(key=lambda p: p.created_at)
+
+        latest = {p.git_url: p for p in projects}
+
+        return {"projects": list(latest.values())}
```
Panaetius, line 47:
```suggestion

        return {
            "projects": [
                 max(g, key=lambda p: p.created_at) 
                     for _, g in itertools.groupby(projects, lambda p: p.git_url)
             ]
         }
```
Is maybe a bit more readable and doesn't rely on a side-effect of how dictionaries work in python.

renku/service/controllers/utils/project_clone.py
```diff
@@ -16,46 +16,32 @@
...
         raise_git_except=True,
         config={"user.name": project_data["fullname"], "user.email": project_data["email"],},
         checkout_rev=project_data["ref"],
     )
-    project_data["initialized"] = initialized
+    project.save()
 
     service_log.debug(f"project successfully cloned: {repo}")
-    service_log.debug(f"project folder exists: {local_path.exists()}")
+    service_log.debug(f"project folder exists: {project.exists()}")
```
Panaetius, line 45: Shouldn't this be `project.abs_path.exists()` ?

Panaetius (changes requested):

renku/service/controllers/utils/project_clone.py
jsam: Project is a model instance with a method `.exists()` which ensures the correct filesystem path assembly and check based on the data found in the model instance.

Panaetius (approved):

PR #1764, opened 2020-11-25
docs: update renku config documentation

lorenzo-cavazzi:
# Description

Update the documentation around `renku config` to include the full list of supported options.
It was pointed out that there isn't a docs page for that in renku, and this seems to be the best place where to add this information now that both `renku` and `renku-python` docs are unified.

Any reference to the `registry` config has been removed since it's not used anywhere and it was used in the examples.
If you prefer to keep it there for any reason, feel free to push changes directly to this PR or to request the change.

re SwissDataScienceCenter/renku#1647


rokroskar (approved):
Thanks!!

PR #1766, opened 2020-11-26
fix(service): fix old header compatibility

Panaetius:


rokroskar (approved):
Tested this on rok.dev.renku.ch and seems to fix the problem 👍

PR #1767, opened 2020-11-26
fix(service): remove chdir

Panaetius:
Removes chdir need by explicitely setting working directory.


renku/service/controllers/api/mixins.py
```diff
@@ -91,7 +92,12 @@ def local(self):
 
         self.project_path = project.abs_path
 
-        with chdir(self.project_path):
+        with click.Context(
+            click.Command("create_from_template"),
+            obj=LocalClient(
+                path=default_path(self.project_path), renku_home=RENKU_HOME, external_storage_requested=True,
+            ),
+        ).scope():
```
mohammad-sdsc, line 100: I'd suggest that we put this code as a function in `core.utils.context` and use it instead of repeating this implementation:
```
@contextlib.contextmanager
def click_context(path, command):
    """Provide a click context."""
    import click

    from renku.core.management import LocalClient
    from renku.core.management.config import RENKU_HOME
    from renku.core.management.repository import default_path

    with click.Context(
        click.Command(command),
        obj=LocalClient(path=default_path(path), renku_home=RENKU_HOME, external_storage_requested=True),
    ).scope():
        yield
```

Then here it would be:
```
with click_context(path=project.abs_path, command="create_from_template"):
    ...
```

mohammad-sdsc (commented):
LGTM, just a refactoring comment.

mohammad-sdsc (approved):
Thanks!

PR #1769, opened 2020-11-26
chore: do not pull LFS in service

rokroskar:
This PR adds the `GIT_LFS_SKIP_SMUDGE=1` environment variable to the service containers to prevent pulling LFS data when not needed. The operations that require LFS objects do so explicitly anyway (e.g. dataset import from renku projects). This significantly improves the performance of common operations like `dataset.list`.


Panaetius (commented):
Should we make this configurable in the values.yaml but with a default of 1? Just as a backup if we do notice a case where this breaks things, then we can easily change it back.

rokroskar:
good point - I can add a value for it

Panaetius (approved):
Thank you!

PR #1772, opened 2020-11-27
chore(service): specify gunicorn config via values

rokroskar:
this PR allows the renkulab admin to set the gunicorn worker config via helm values. 

~~edit: I can't get the values to work if I pass them via `--set core.nWorkers=4` with `helm`, for example... but they work if I set them in the values file~~


Panaetius (approved):
thanks! :smile:

PR #1775, opened 2020-12-01
feat(cli): automatically track files in git-lfs if necessary

### LINKED ISSUE #1725: automatically add large files to LFS in pre-commit hook
Pre-commit hooks can alter commits, as long as they `git add` the modified/added files.

We should streamline our git lfs hook to automatically track large files in LFS on commit using this, instead of failing and printing a warning.

We should decide if the user should confirm this or if it should just be done automatically.

Panaetius:
Closes #1725

### DIFF
--- a/renku/data/pre-commit.sh
+++ b/renku/data/pre-commit.sh
@@ -84,17 +84,35 @@ fi
 if [ ${#ADDED_FILES[@]} -ne 0 ]; then
   UNTRACKED_PATHS=$(renku storage check-lfs-hook "${ADDED_FILES[@]}")
   if [ "$UNTRACKED_PATHS" ]; then
+
     echo 'You are trying to commit large files to Git instead of Git-LFS.'
     echo
-    echo 'Large files:'
-    for file in "${UNTRACKED_PATHS[@]}"; do
-      echo "$file"
-    done
-    echo
-    echo 'To track these files in Git LFS use "git lfs track <FILENAMES>".'
-    echo
-    echo 'To commit anyway, use "git commit --no-verify".'
-    exit 1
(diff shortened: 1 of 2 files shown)


rokroskar:
Hmm I wonder if this is going to make some people unhappy as well - maybe it needs to be configurable via e.g. an environment variable? Then in places like the autosave we can use this and not ask questions but otherwise the user still has some control. What do you think?

Panaetius:
> Hmm I wonder if this is going to make some people unhappy as well - maybe it needs to be configurable via e.g. an environment variable? Then in places like the autosave we can use this and not ask questions but otherwise the user still has some control. What do you think?

Yes that's a good point, I'll add it.

Panaetius:
It now needs to be enabled by doing `renku config set autocommit_lfs true`

renku/data/pre-commit.sh
```diff
@@ -84,17 +84,35 @@ fi
...
-    echo 'Large files:'
-    for file in "${UNTRACKED_PATHS[@]}"; do
-      echo "$file"
-    done
-    echo
-    echo 'To track these files in Git LFS use "git lfs track <FILENAMES>".'
-    echo
-    echo 'To commit anyway, use "git commit --no-verify".'
-    exit 1
+    AUTOCOMMIT_LFS=$(renku config show autocommit_lfs)
```
rokroskar, line 90: This should allow us to set it outside of the project configuration so we don't need to make changes to the user's project in the autosave hook. E.g.

```
AUTOCOMMIT_LFS=${AUTOCOMMIT_LFS:=$(renku config show autocommit_lfs)}
```

I _think_ that should work but I'm not sure...

rokroskar (approved):
🚀

PR #1777, opened 2020-12-02
chore: use sensible defaults for worker parallelism

rokroskar:
Sets the default to use threading so the service can respond to k8s probes while serving requests.


mohammad-sdsc (approved):
LGTM, thanks!

PR #1780, opened 2020-12-03
test: fix git commit hook tests on python3.6

Panaetius:
Fixes locale problem with pytest, gitpython, click and python3.6
Fixes GitHookException stdout in case where messages are too long.

### DIFF
--- a/renku/data/pre-commit.sh
+++ b/renku/data/pre-commit.sh
@@ -66,7 +66,7 @@ if [ ${#MODIFIED_FILES[@]} -ne 0 ] ; then
     echo 'To commit anyway, use "git commit --no-verify".'
     exit 1
   fi
-  IMMUTABLE_TEMPLATE_FILES=$(renku migrate check-immutable-template-files "${MODIFIED_FILES[@]}")
+  IMMUTABLE_TEMPLATE_FILES=$(renku check-immutable-template-files "${MODIFIED_FILES[@]}")
   if [ "$IMMUTABLE_TEMPLATE_FILES" ]; then
     echo 'You are trying to update files marked as immutable in your project template.'
     echo 'This would prevent the project from being updated with new versions of the template in the future.'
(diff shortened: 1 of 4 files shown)


renku/data/pre-commit.sh
```diff
@@ -27,7 +27,7 @@ IFS=$'\n' read -r -d '' -a MODIFIED_FILES \
   <<< "$(git diff --name-only --cached --diff-filter=M)"
 IFS=$'\n' read -r -d '' -a ADDED_FILES \
   <<< "$(git diff --name-only --cached --diff-filter=A)"
-
+echo "a"
```
mohammad-sdsc, line 30: Debugging left-over?

renku/data/pre-commit.sh
```diff
@@ -66,7 +66,7 @@ if [ ${#MODIFIED_FILES[@]} -ne 0 ] ; then
     echo 'To commit anyway, use "git commit --no-verify".'
     exit 1
   fi
-  IMMUTABLE_TEMPLATE_FILES=$(renku migrate check-immutable-template-files "${MODIFIED_FILES[@]}")
+  IMMUTABLE_TEMPLATE_FILES=$(renku check-immutable-template-files "${MODIFIED_FILES[@]}")
```
mohammad-sdsc, line 69: 😱

renku/data/pre-commit.sh
mohammad-sdsc: In some other places as well.

mohammad-sdsc (commented):
Made some comments! I was wondering if we should use Python 3.6 as default in PR CI instead of 3.7.

renku/data/pre-commit.sh
Panaetius: right... :fearful:

mohammad-sdsc (approved):
LGTM, thanks!

PR #1781, opened 2020-12-03
chore: add acceptance test action

rokroskar:
Adds an action to automatically deploy renku to a clean namespace and run acceptance tests. Only runs on PRs that affect the service. 

Points left to do:

1. determine when this should run
2. how to do the teardown

For 1. it was discussed that this should run on all PRs that push changes to `service` and `core` but to not be required for PRs to pass because of possible issues with dependencies with other components. It should also run on `master` and tags. 

For 2. I would leave it as it is done in the main renku repo atm, which is that the deployment gets destroyed after the PR is closed. But maybe in the near future this should change to doing a `helm uninstall` if the tests pass to release some of the resources followed by a namespace delete after the PR is merged. If we delete the namespace right away we run into issues with LetsEncrypt rate limiting if the tests have to run again. 

/deploy

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -58,7 +58,7 @@ jobs:
         RENKU_BOT_DEV_PASSWORD: ${{ secrets.RENKU_BOT_DEV_PASSWORD }}
         RENKU_ANONYMOUS_SESSIONS: false
         RENKU_TESTS_ENABLED: true
-        renku: "@development"
+        renku: "@000-deploy-action"
         renku_core: "@${{ github.head_ref }}"
   test-pr:
     runs-on: ubuntu-latest
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -22,7 +22,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - id: deploy-comment
-      uses: SwissDataScienceCenter/actions/check-pr-description@000-deploy-action
+      uses: SwissDataScienceCenter/renku/actions/check-pr-description@000-deploy-action
       with:
         string: /deploy
         pr_ref: ${{ github.event.number }}
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -19,6 +19,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - run: echo $REPO_NAME
+      env:
+        REPO_NAME: ${{ github.repository.name }}
     - name: renku build and deploy
       uses: SwissDataScienceCenter/renku/actions/deploy-renku@000-deploy-action
       env:
@@ -27,9 +30,9 @@ jobs:
         DOCKER_USERNAME: ${{ secrets.RENKU_DOCKER_USERNAME }}
         GITLAB_TOKEN: ${{ secrets.DEV_GITLAB_TOKEN }}
         KUBECONFIG: "${{ github.workspace }}/renkubot-kube.config"
-        RENKU_NAMESPACE: ci-renku-python-${{ github.event.number }}
-        RENKU_TMP_NAMESPACE: ci-renku-python-${{ github.event.number }}-tmp
-        RENKU_RELEASE: ci-renku-python-${{ github.event.number }}
+        RENKU_NAMESPACE: ci-rp-${{ github.event.number }}-renku
+        RENKU_TMP_NAMESPACE: ci-rp-${{ github.event.number }}-renku-tmp
+        RENKU_RELEASE: ci-rp-${{ github.event.number }}-renku
         RENKU_VALUES_FILE: "${{ github.workspace }}/values.yaml"
         RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES }}
         RENKUBOT_KUBECONFIG: ${{ secrets.RENKUBOT_DEV_KUBECONFIG }}
@@ -52,10 +55,10 @@ jobs:
     - name: Test the PR
       env:
         RENKU_BOT_DEV_PASSWORD: ${{ secrets.RENKU_BOT_DEV_PASSWORD }}
-        RENKU_RELEASE: "ci-renku-python-${{ github.event.number }}"
+        RENKU_RELEASE: "ci-rp-${{ github.event.number }}-renku"
       run: |
         RENKU_PYTHON_VERSION=$(git rev-parse HEAD)
-        echo "Passing renku-python version $RENKU_PYTHON_VERSION"
+        echo "Passing rp version $RENKU_PYTHON_VERSION"
 
         git clone https://github.com/SwissDataScienceCenter/renku.git /tmp/renku
         cd /tmp/renku/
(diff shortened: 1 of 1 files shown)


.github/workflows/acceptance-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+        RENKU_RELEASE: ci-renku-python-${{ github.event.number }}-renku
+        RENKU_VALUES_FILE: "${{ github.workspace }}/values.yaml"
+        RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES }}
+        RENKUBOT_KUBECONFIG: ${{ secrets.RENKUBOT_DEV_KUBECONFIG }}
+        RENKUBOT_RANCHER_BEARER_TOKEN: ${{ secrets.RENKUBOT_RANCHER_BEARER_TOKEN }}
+        renku: "@development"
+        renku_core: "@${{ github.head_ref }}"
+  test-pr:
+    runs-on: ubuntu-latest
+    if: github.event.action != 'closed'
```
Panaetius, line 31: I think we should instead only run on `opened,edited,synchronize,reopened`

.github/workflows/acceptance-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+        python-version: 3.7
+    - name: Install dependencies
+      run: |
+        sudo apt-get update -y && sudo apt-get install -y grep
+        pip install yq
+    - name: Test the PR
+      env:
+        RENKU_BOT_DEV_PASSWORD: ${{ secrets.RENKU_BOT_DEV_PASSWORD }}
+      run: |
+        export RENKU_RELEASE="ci-renku-python-${{ github.event.number }}-renku"
```
Panaetius, line 46: We could set this in the `env:` above instead.

.github/workflows/acceptance-tests.yml
```diff
@@ -0,0 +1,76 @@
+name: Acceptance tests
+
+on:
+  pull_request: []
+    # paths: ['renku/service']
```
Panaetius, line 5:
```suggestion
    paths: ['renku/service', 'renku/core','helm-chart']
```

Panaetius (changes requested):

.github/workflows/acceptance-tests.yml
rokroskar: Thanks, I haven't finalized this yet since I'm still testing so I'll keep this here as a reminder

Panaetius:
Regarding 2. (and more generally): it should teardown when the PR is closed and at the beginning of the workflow (i.e. if a new commit gets pushed, teardown the previous deployment before deploying again)

rokroskar:
Why teardown if a new commit is pushed? It just updates the deployment in that case.

Panaetius:
Ah I thought there were some issues in the renku repo with acceptance tests when rerun and wanted to prevent that happening here.

rokroskar:
no usually it works great. There are occasional glitches but usually when some component is swapped to something incompatible. Then a helm uninstall and a rerun of the workflow usually does the trick.

Panaetius (dismissed):

rokroskar:
@Panaetius I added the spec about paths for the acceptance tests - can you re-approve?

rokroskar:
Actually we should wait with merging this until https://github.com/SwissDataScienceCenter/renku/pull/1751 is merged.

.github/workflows/acceptance-tests.yml
```diff
@@ -0,0 +1,112 @@
...
+    if: github.event.action == 'closed'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: docker://alpine/k8s:1.16.8
+      name: k8s teardown
+      env:
+        KUBECONFIG: ${{ github.workspace }}/renkubot-kube.config
+        RENKUBOT_KUBECONFIG: ${{ secrets.RENKUBOT_KUBECONFIG }}
+        RENKU_RELEASE: "ci-${{ github.event.number }}-renku"
+        RENKU_NAMESPACE: "ci-${{ github.event.number }}-renku"
```
lorenzo-cavazzi, line 96: The values of `RENKU_RELEASE` and `RENKU_NAMESPACE` here appear to be different from the one used in the `deploy-pr` job `RENKU_RELEASE: ci-renku-python-${{ github.event.number }}-renku`

.github/workflows/acceptance-tests.yml
```diff
@@ -0,0 +1,112 @@
...
+  pull_request:
+    types:
+    - opened
+    - edited
+    - synchronize
+    - reopened
+    - closed
+    paths:
+    - renku/service
+    - renku/core
```
lorenzo-cavazzi, line 13: I believe this should be `renku/core/**`, otherwise it won't match on subfiles/folders. At least this was the case in the UI.
Reference: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

.github/workflows/acceptance-tests.yml
rokroskar: thanks @lorenzo-cavazzi for spotting that! btw, I forgot that I wrote a teardown action as well... I've changed this now to use the common action - maybe you want to do the same in https://github.com/SwissDataScienceCenter/renku-ui/pull/1166

.github/workflows/acceptance-tests.yml
```diff
@@ -0,0 +1,119 @@
...
+        path: test-artifacts
+  cleanup:
+    if: github.event.action == 'closed'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: SwissDataScienceCenter/renku/actions/teardown-renku@000-deploy-action
+      env:
+        GITLAB_TOKEN: ${{ secrets.GITLAB_DEV_TOKEN }}
+        KUBECONFIG: ${{ github.workspace }}/renkubot-kube.config
+        RENKUBOT_KUBECONFIG: ${{ secrets.RENKUBOT_KUBECONFIG }}
```
lorenzo-cavazzi, line 116:
```suggestion
        RENKUBOT_KUBECONFIG: ${{ secrets.RENKUBOT_DEV_KUBECONFIG }}
```

.github/workflows/acceptance-tests.yml
```diff
@@ -0,0 +1,121 @@
...
+        RENKU_TESTS_ENABLED: true
+        renku: "@000-deploy-action"
+        renku_core: "@${{ github.head_ref }}"
+    - name: Check existing renkubot comment
+      uses: peter-evans/find-comment@v1
+      id: findcomment
+      with:
+        issue-number: ${{ github.event.pull_request.number }}
+        comment-author: 'RenkuBot'
+    - name: Create comment pre deploy
```
lorenzo-cavazzi, line 54: Are you re-posting the link on purpose or did an `if` get lost?
```suggestion
    - name: Create comment pre deploy
      if: steps.findcomment.outputs.comment-id == 0
```

.github/workflows/acceptance-tests.yml
```diff
@@ -0,0 +1,121 @@
...
+        RENKU_ANONYMOUS_SESSIONS: false
+        RENKU_TESTS_ENABLED: true
+        renku: "@000-deploy-action"
+        renku_core: "@${{ github.head_ref }}"
+    - name: Check existing renkubot comment
+      uses: peter-evans/find-comment@v1
+      id: findcomment
+      with:
+        issue-number: ${{ github.event.pull_request.number }}
+        comment-author: 'RenkuBot'
```
lorenzo-cavazzi, line 53: This could be further filtered, to be sure an unrelated message from RenkuBot doesn't steal the scene

```suggestion
        comment-author: 'RenkuBot'
        body-includes: "You can access the deployment of this PR at"
```

Panaetius (approved):
Thank you! :tada: :rocket: :clinking_glasses:

Panaetius (approved):
Looks like any problem is solved by having enough `renku`

Panaetius (approved):

PR #1784, opened 2020-12-04
feat(service): gzip decompression support

### LINKED ISSUE #1776: Support for gzip decompression in service
As a service user, I want service to support `application/x-gzip` MIME type so that it can decompress gzip compressed files.

mohammad-sdsc:
# Description

Support for decompressing gzip files after upload.

Fixes #1776


Panaetius (approved):
Thank you!

PR #1792, opened 2020-12-08
chore: Set RENKU_DOMAIN env var in service

### LINKED ISSUE #1768: service: set the `RENKU_DOMAIN` env var for proper host in ids
The projects created in the renku core service are initialized with `localhost` as the hostname for all of the entity ids. We should set the `RENKU_DOMAIN` environment variable to correspond to the actual hostname so it's used in the ids.

Panaetius:
closes #1768


helm-chart/renku-core/templates/deployment.yaml
```diff
@@ -92,6 +92,8 @@ spec:
             {{ end }}
             - name: GIT_LFS_SKIP_SMUDGE
               value: {{ .Values.gitLFSSkipSmudge | quote }}
+            - name: RENKU_DOMAIN
+              value: {{ .Values.global.renku.domain }}
```
jsam, line 96: Can we write down the default values for this somewhere? I don't see them in renku repository.

Also, let's please add this to `.env-example` so when we use the service with docker compose it gets picked up.

Panaetius: It doesn't really have a default value, as it depends on the deployment. But it is here https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/renku/values.yaml#L54

It's not really needed outside of a kubernetes deployment (it just defaults to `localhost` in code), so it doesn't need to be set when running with e.g. docker compose. But I can add it to the .env-example.

jsam: Ah, okay I'm just blind then. Thanks!

Yes, please let's add it to `.env-example` and :rocket:

jsam (approved):

PR #1793, opened 2020-12-08
feat(service): use jwt sub claim as user-id

### LINKED ISSUE #1770: service: the JWT sub claim should be used to identify users
The cache is currently using a base64-encoded email as the user identifier in the cache (and elsewhere?). Since we now have the possibility to handle JWTs, we should use the `sub` claim from the JWT token to identify users instead. This is the only stable, immutable user identifier across the platform.

Panaetius:
closes #1770


jsam (approved):
I'm not really sure that having a fully immutable identifier for this is really needed here, cause we have cleanups in place and clients should work in a way to safely handle this case. What problems did it cause? The reason why it was an email was to make troubleshooting easier.

Would be nice to know what problems have you find. 

Other than that :rocket:

PR #1799, opened 2020-12-09
feat(dataset): dataset provenance in a single file

### LINKED ISSUE #1710: Store all datasets in single file
Similar to what was done in https://github.com/SwissDataScienceCenter/renku-python/pull/1545 we should add dataset metadata for the current commit in a single file. Dataset commands should write to this new file as well as to the old location. Commands still use the old format.

We keep the old files for compatibility with `renku log` but exclude the new file from it.

mohammad-sdsc:
# Description

Parses git history and puts all datasets' metadata in a single file: `.renku/dataset.json`. To generate this file use: `renku graph generate-dataset`. To export add `-d` flag to the export command: `renku graph export -d --format jsonld`.

Fixes #1710

### DIFF
--- a/renku/core/incubation/graph.py
+++ b/renku/core/incubation/graph.py
@@ -17,6 +17,7 @@
 # limitations under the License.
 """Dependency and Provenance graph building."""
 
+import shutil
 from collections import defaultdict
 from pathlib import Path
 from typing import Dict
@@ -27,15 +28,18 @@ from renku.cli.update import execute_workflow
 from renku.core import errors
 from renku.core.incubation.command import Command
 from renku.core.management.config import RENKU_HOME
+from renku.core.management.migrate import migrate
 from renku.core.management.repository import RepositoryApiMixin
 from renku.core.models.entities import Entity
 from renku.core.models.provenance.activities import Activity as ActivityRun
 from renku.core.models.provenance.activity import ActivityCollection
+from renku.core.models.provenance.datasets import DatasetProvenance
 from renku.core.models.provenance.provenance_graph import ProvenanceGraph
 from renku.core.models.workflow.dependency_graph import DependencyGraph
 from renku.core.models.workflow.run import Run
 from renku.core.utils import communication
 from renku.core.utils.contexts import measure
+from renku.core.utils.migrate import read_project_version_from_yaml
 from renku.core.utils.scm import git_unicode_unescape
 
 GRAPH_METADATA_PATHS = [
@@ -68,7 +72,7 @@ def _generate_graph(client, force):
             pass
     else:
         if client.has_graph_files():
-            raise errors.OperationError("Graph files exist. Use --force to regenerate the graph.")
+            raise errors.OperationError("Graph files exist. Use ``--force`` to regenerate the graph.")
 
     create_empty_graph_files()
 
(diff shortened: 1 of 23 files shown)


renku/core/models/provenance/datasets.py
```diff
@@ -0,0 +1,515 @@
...
+
+        rdf_type = schema.Dataset
+        model = Dataset
+        unknown = EXCLUDE
+
+    creators = Nested(schema.creator, PersonSchema, many=True)
+    date_created = fields.DateTime(
+        schema.dateCreated, missing=None, allow_none=True, format="iso", extra_formats=("%Y-%m-%d",),
+    )
+    date_deleted = fields.DateTime(prov.invalidatedAtTime, missing=None, allow_none=True, format="iso")
```
Panaetius, line 481: If we want to be consistent with PROV-O, then to use `invalidatedAtTime`, dataset also needs to be a `prov.entity`, as that's the domain of that property.

renku/core/models/provenance/datasets.py
```diff
@@ -0,0 +1,515 @@
...
+            u = urlparse(derived_from_url)
+            derived_from_url = u._replace(netloc=host).geturl()
+            self.derived_from = Url(id=derived_from_id, url_id=derived_from_url)
+
+        for file_ in self.files:
+            file_.client = client
+
+    def update_from(self, dataset, client, revision, date):
+        """Update metadata from a new version of the dataset."""
+        self._set_identifier(dataset.identifier)
```
Panaetius, line 274: Does this go against dataset immutability? I.e. using this you can change the properties of a dataset without generating a new ID?

renku/core/incubation/graph.py
```diff
@@ -217,3 +230,120 @@ def _get_modified_paths(client, plans_usages):
...
+    for n, commit in enumerate(commits, start=1):
+        communication.echo(f"\rProcessing commits {n}/{n_commits}\r")
+
+        files_diff = list(commit.diff(commit.parents or NULL_TREE, paths=".renku/datasets/*"))
+        paths = [git_unicode_unescape(f.a_path) for f in files_diff]
+        deleted_paths = [git_unicode_unescape(f.a_path) for f in files_diff if f.change_type == "A"]
+
+        datasets, deleted_datasets = _fetch_datasets(client, commit.hexsha, paths=paths, deleted_paths=deleted_paths)
+
+        revision = commit.hexsha
```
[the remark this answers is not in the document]
mohammad-sdsc: We use git object hashes and not git commit hashes. Revision is passed only to enable calculating the correct content hash in git history at a specific revision. For example, when a file was added to a dataset at revision xyz we want to calculate file's content hash at that revision (and not its current content).

renku/core/models/provenance/datasets.py
mohammad-sdsc: Right. Actually, when dataset has a non-existing identifier we insert it as a new dataset and this function won't get called. This is called only when identifier is the same as the current identifier, so, we don't need to set it again.

renku/core/incubation/graph.py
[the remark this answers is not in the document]
Panaetius: Ah I see :+1:

Panaetius (dismissed):
Thank you, looks great!

Panaetius (approved):

PR #1801, opened 2020-12-11
chore: add debug mode to core service

### LINKED ISSUE #1758: allow core service to be run in debug mode
We should add an env variable to start the core service in debug mode, allowing hot-reloading of changed scripts in a pod, like `app.run(debug=os.environ["RENKU_SVC_DEBUG"])`.

This should help with debugging problems more easily.

Dev/CI deployments should set this to True by default, but we should set it to False in production deployments.

### LINKED ISSUE #1759: support kubernetes debugging of core service
We should set up the core service to allow debugging using VSCode, as mentioned in https://github.com/Azure/vscode-kubernetes-tools/blob/master/debug-on-kubernetes.md#5-python-debug-prerequisites on dev/CI deployments (And prod?)

This allows faster response to difficult to reproduce issues on the service.

Panaetius:
Adds debug mode that allows editing&reloading files in a running pod and attaching to a running pod with VSCode.

closes #1758 
closes #1759 

/deploy


rokroskar (dismissed):
works great, thanks!

rokroskar (approved):

PR #1806, opened 2020-12-15
feat(cli): add path to graph export

### LINKED ISSUE #1732: Add --revision to new workflow log
change `renku graph log` to `renku graph workflow dump` and add a `--revision` flag and `<paths>` argument to output triples for a certain revision and/or paths.

Only support `--revision` for commits done after new graph building, with a message to users for earlier commits to checkout the commit themselves, run migrations, `renku graph generate` and then generate the log.

mohammad-sdsc:
# Description

Allows `renku graph export` to accept `paths` which would limit resulting graph to those that generate these paths.

Fixes #1732 (partially)


renku/core/models/provenance/provenance_graph.py
```diff
@@ -149,6 +169,39 @@ def query(self, query):
...
+                activities.append(parent_activity)
+                todo.append(parent_activity)
+
+        return self._create_graph(activities=activities)
+
+    def _get_latest_activity(self, path=None, entity_id=None):
+        """Return the latest activity that generated a path or entity."""
+        for activity in reversed(self.activities):
+            for generation in activity.generated:
+                if generation.entity.path == path or generation.entity._id == entity_id:
```
Panaetius, line 202: This would fail if I specify a path of `dir/file1` but the activity had a directory output and generated just `dir/`, no? Since the generation would be a Collection, but with the file in one of its `.entities`.

Even if it works, we should add a test for this case, as things often break with directory outputs.

Panaetius: also the opposite case if a user specifies a folder path but the latest activity changed a file in the folder.

vigsterkr:
@mohammad-sdsc great PR! is there any way i can help this moving along?

mohammad-sdsc:
> @mohammad-sdsc great PR! is there any way i can help this moving along?

Thanks Viktor! This PR is almost ready. We need to have a meeting to discuss about the implementation details of the new graph and make a decision (see https://github.com/SwissDataScienceCenter/renku-python/issues/1732#issuecomment-789329510). After that, we either merge this PR or discard it (i.e. implementing revision export using diff between commits).

PR #1809, opened 2021-01-04
chore: fix exclusion of pipfile lock action from dependabot PRs

Panaetius:
Apparently the branches-ignore keyword does not apply to the originating branch of a PR but to the target branch of a PR


mohammad-sdsc (approved):
Thank!

PR #1823, opened 2021-01-05
fix(core): fix gitlab url parsing when GITLAB_BASE_URL is set without port

Panaetius:
Fixes issue with RE string concatenation when one part is `None`


mohammad-sdsc (approved):
LGTM, thanks!

PR #1828, opened 2021-01-07
fix(cli): improve display of dataset show command and minor fixes

Panaetius:
Reorders fields and fixes issue if some values are missing


mohammad-sdsc (commented):
Looks quite good! Please update the docs at the top of this file.

mohammad-sdsc (approved):
Thanks!

PR #1830, opened 2021-01-08
feat(dataset): dataset create and add using new metadata

### LINKED ISSUE #1735: Add dataset commands for new graph metadata structure
Add all dataset commands under `renku graph dataset` (e.g. `renku graph dataset add`), but using the new metadata structure, keeping the old behavior.

mohammad-sdsc:
# Description

Use new metadata in `graph dataset create` and `graph dataset add` commands.

Fixes #1735

### DIFF
--- a/renku/cli/exception_handler.py
+++ b/renku/cli/exception_handler.py
@@ -86,7 +86,7 @@ class RenkuExceptionsHandler(click.Group):
         try:
             return super().main(*args, **kwargs)
         except RenkuException as e:
-            click.echo("Error: {}".format(e))
+            click.echo("Error: {}".format(e), err=True)
             if e.__cause__ is not None:
                 click.echo("\n{}".format(traceback.format_exc()))
             exit_code = 1
@@ -147,7 +147,7 @@ class IssueFromTraceback(RenkuExceptionsHandler):
             with capture_internal_exceptions():
                 from git import Repo
 
-                from renku.core.commands import get_git_home
+                from renku.core.commands.git import get_git_home
                 from renku.core.models.provenance.agents import Person
 
                 repo = Repo(get_git_home())
--- a/renku/cli/exception_handler.py
+++ b/renku/cli/exception_handler.py
@@ -86,9 +86,9 @@ class RenkuExceptionsHandler(click.Group):
         try:
             return super().main(*args, **kwargs)
         except RenkuException as e:
-            click.echo("Error: {}".format(e), err=True)
+            click.echo(f"Error: {e}", err=True)
             if e.__cause__ is not None:
-                click.echo("\n{}".format(traceback.format_exc()))
+                click.echo(f"\n{traceback.format_exc()}")
             exit_code = 1
             if isinstance(e, (ParameterError, UsageError)):
                 exit_code = 2
--- a/renku/core/management/datasets.py
+++ b/renku/core/management/datasets.py
@@ -123,7 +123,7 @@ class DatasetsApiMixin(object):
     @property
     def datasets_provenance(self):
         """Return dataset provenance if available."""
-        if not self.has_datasets_provenance_file():
+        if not self.has_datasets_provenance():
             return
         if not self._datasets_provenance:
             self._datasets_provenance = DatasetProvenance.from_json(self.datasets_provenance_path)
@@ -132,7 +132,7 @@ class DatasetsApiMixin(object):
 
     def update_datasets_provenance(self, dataset, remove=False):
         """Update datasets provenance for a dataset."""
-        if not self.has_datasets_provenance_file():
+        if not self.has_datasets_provenance():
             return
 
         if remove:
@@ -142,7 +142,7 @@ class DatasetsApiMixin(object):
 
         self.datasets_provenance.to_json()
 
-    def has_datasets_provenance_file(self):
+    def has_datasets_provenance(self):
         """Return true if dataset provenance exists."""
         return self.datasets_provenance_path.exists()
 
@@ -223,6 +223,8 @@ class DatasetsApiMixin(object):
         """Yield an editable metadata object for a dataset."""
         dataset = self.load_dataset(name=name)
         clean_up_required = False
+        dataset_ref = None
+        path = None
 
         if dataset is None:
             if not create:
@@ -232,6 +234,7 @@ class DatasetsApiMixin(object):
             dataset, path, dataset_ref = self.create_dataset(name=name)
         elif create:
             raise errors.DatasetExistsError('Dataset exists: "{}".'.format(name))
+
         dataset_path = self.path / self.data_dir / dataset.name
         dataset_path.mkdir(parents=True, exist_ok=True)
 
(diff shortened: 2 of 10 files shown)


mohammad-sdsc:
Loading all dataset provenance is not needed frequently when working with datasets. In common case, we only need to load the latest version of metadata for a dataset and make some modifications in it. Therefore, loading the whole history is not needed and only adds to data loading time.

renku/core/models/provenance/datasets.py
```diff
@@ -226,6 +269,20 @@ def from_dataset(cls, dataset, client, revision):
...
         )
 
+    @staticmethod
+    def _convert_from_dataset_files(files, client, revision):
+        dataset_files = []
+        files = {f.path: f for f in files}  # NOTE: To make sure there are no duplicate paths
+        for path in files:
+            file = files[path]
+            dataset_file = DatasetFile.from_dataset_file(file, client=client, revision=revision)
+            if not revision:
```
Panaetius, line 279:
```suggestion
            if not dataset_file:
```

renku/core/models/provenance/datasets.py
```diff
@@ -377,13 +443,30 @@ def add(self, dataset):
...
+            if dataset.derived_from:
+                datasets_ids.remove(dataset.derived_from.url_id)
+
+        assert len(datasets_ids) <= 1, f"There are more than one latest versions with name `{name}`"
+
+        if not datasets_ids:
+            return None
+
+        url = datasets_ids.pop()
+        return next(d for d in self.get_by_name(name) if d.id == url)
```
Panaetius, line 469:
```suggestion
        datasets = {d.id: d for d in self.get_by_name(name)}

        for dataset in datasets.values():
            if dataset.derived_from:
                datasets.pop(dataset.derived_from.url_id, None)

        assert len(datasets) <= 1, f"There are more than one latest versions with name `{name}`"

        if not datasets:
            return None

        _, dataset = datasets.popitem()
        return dataset
```

would reduce the number of times we need to iterate over this.

tests/cli/test_datasets.py
```diff
@@ -521,7 +521,8 @@ def test_repository_file_to_dataset(runner, project, client, subdirectory):
     client.repo.git.commit(message="Added file a", no_verify=True)
 
     # add data
-    result = runner.invoke(cli, ["dataset", "add", "dataset", str(a_path)], catch_exceptions=False,)
+    command = ["graph"] if use_graph else []
```
Panaetius, line 524: Cool idea!

Panaetius (changes requested):

renku/core/models/provenance/datasets.py
mohammad-sdsc: We cannot iterate over a dict and remove items from it; I've just changed `for dataset in datasets.values()` to `for dataset in self.get_by_name(name)`

Panaetius: right, `for dataset in list(datasets.values())` would work though. I'm just thinking in a project with a large provenance, we want to loop over the provenance as little as possible.

renku/cli/exception_handler.py
```diff
@@ -86,7 +86,7 @@ def main(self, *args, **kwargs):
         try:
             return super().main(*args, **kwargs)
         except RenkuException as e:
-            click.echo("Error: {}".format(e))
+            click.echo("Error: {}".format(e), err=True)
```
jsam, line 89: Should we switch this to f-string?

renku/core/management/datasets.py
```diff
@@ -247,6 +250,51 @@ def with_dataset(self, name=None, create=False):
...
+            raise errors.DatasetExistsError('Dataset exists: "{}".'.format(name))
+        else:
+            dataset = dataset.to_dataset(self)
+
+        dataset_path = self.path / self.data_dir / dataset.name
+        dataset_path.mkdir(parents=True, exist_ok=True)
+
+        try:
+            yield dataset
+        except Exception:
```
jsam, line 277: Should we use `RenkuException` here?

renku/core/models/provenance/datasets.py
```diff
@@ -128,6 +152,25 @@ def _extract_identifier(self):
...
+                revision=revision,
+                added=self.date_added,
+                based_on=self.based_on,
+                external=self.is_external,
+                id=None,
+                path=self.entity.path,
+                source=self.source,
+                url=None,
+            )
+        except KeyError:  # path does not exists at revision
```
jsam, line 171: Could we add a tag `NOTE:` and also which line from `.from_revision` is expected to raise this?

renku/core/models/provenance/datasets.py
```diff
@@ -226,6 +269,20 @@ def from_dataset(cls, dataset, client, revision):
             version=dataset.version,
         )
 
+    @staticmethod
+    def _convert_from_dataset_files(files, client, revision):
```
jsam, line 273: Should we add a doc string here?

renku/core/models/provenance/datasets.py
```diff
@@ -316,37 +373,46 @@ def _update_files(self, dataset, client, revision, date):
...
+            license=self.license,
+            same_as=self.same_as,
+            tags=self.tags,
+            title=self.title,
+            url=self.url,
+            version=self.version,
         )
 
-        dataset_files.append(dataset_file)
+    def _convert_to_dataset_files(self, client):
```
jsam, line 406: Should we add a doc string here?

renku/core/models/provenance/datasets.py
```diff
@@ -377,13 +443,30 @@ def add(self, dataset):
 
     def get(self, identifier):
         """Return a dataset by its original identifier."""
-        datasets = [d for d in self._datasets if d.identifier == identifier]
-        assert len(datasets) <= 1, f"Found more than one with identifier `{identifier}`."
-        return datasets[0] if datasets else None
+        datasets = (d for d in self._datasets if d.identifier == identifier)
+        dataset = next(datasets, None)
+        assert next(datasets, None) is None, f"Found more than one dataset with identifier `{identifier}`."
```
jsam, line 448: Should replace this assert with explicit check and raise?

jsam (commented):
Thank you, it look's great! 

I just added a few comments/questions, otherwise we can :rocket:

renku/core/management/datasets.py
mohammad-sdsc: No, this should catch non-Renku exceptions to do a cleanup.

renku/core/models/provenance/datasets.py
mohammad-sdsc: We've added these checks as safety-nets here and in other places to make sure that they never happen. I'll keep the format as it is to be in par with other places. End users should not see this (or the repo is really broken).

renku/core/models/provenance/datasets.py
mohammad-sdsc: I've changed the wording to make it clear which part of that function raises this error.

Panaetius (approved):

PR #1833, opened 2021-01-12
docs(cli): renku clone

### LINKED ISSUE #1673: Improve `renku clone` documentation

Currently, there is no documentation for the command `clone` on the docs page https://renku-python.readthedocs.io/en/latest/commands.html
It would be nice to
* Add the command to the list, so we can use that as a reference for renku users asking us how to properly clone projects (maybe also adding a link in the UI where we show the `renku clone` command for a project)
* Add a few lines to describe a couple of common cases. Examples are 1) making a complete clone locally (it requires `--pull-data`, right?) and 2) clone a project locally and push it to another renkulab deployment, lfs objects included (I got asked a few times how to do that)

mohammad-sdsc:
# Description

Docs for `renku clone`. Changes an option to pull LFS data by default.

Fixes #1673

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -27,6 +27,13 @@ Renku Command Line
 
 .. automodule:: renku.cli.init
 
+.. _cli-clone:
+
+``renku clone``
+---------------
+
+.. automodule:: renku.cli.clone
+
 .. _cli-config:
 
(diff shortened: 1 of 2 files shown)


Panaetius (approved):
Great work, thank you :)

PR #1834, opened 2021-01-13
feat(svc): add endpoints for renku config

### LINKED ISSUE #1783: renku config set committing files other than renku.ini
```
$ touch file
$ renku config set something
```
commits `.renku/renku.ini` AND `file`. It should only commit the renku.ini and leave other dirty files unchanged.

### LINKED ISSUE #1819: Expose renku config command through core service
The UI now uses configurations stored in the `.renku/renku.ini` file. Currently, we access them in read-only mode by invoking a GitLab API to read the raw file.
This solution is acceptable, but now we want to provide the users with an interface to modify and remove those values, and we prefer to keep it in line with the renku-python `renku config` command.

A simple API structure that exposes the `renku config` commands with very similar syntax would be enough.
Here is a proposal [updated]:
* 0.1.0: https://app.swaggerhub.com/apis-docs/lorenzo-cavazzi/renku-core/0.1.0
* 0.1.1: https://app.swaggerhub.com/apis-docs/lorenzo-cavazzi/renku-core/0.1.1 --> set/remove more than 1 key at the same time
* 0.1.3: https://app.swaggerhub.com/apis-docs/lorenzo-cavazzi/renku-core/0.1.3 --> single action to set or remove, remove `project_id` and use instead `git_url` + `branch` , change properties naming

Feel free to point out improvements or changes.

Writing directly to the file through GitLab seems to be the only alternative. We prefer not to go down that way since the UI would have its own custom implementation of `renku config`.

**Additional context**
This is the UI reference issue where the discussion started swissdatasciencecenter/renku-ui/issues/1114
Feel free to chime in! :smiley:

Panaetius:
closes #1819 
closes #1783

/deploy


renku/cli/utils/click.py
```diff
@@ -31,3 +31,37 @@ def convert(self, value, param, ctx):
...
+        for mutex in mutually_exclusive:
+            if type(mutex) == tuple:
+                self.mutually_exclusive.add(mutex[0])
+                self.mutually_exclusive_names.append(mutex[1])
+            else:
+
+                self.mutually_exclusive.add(mutex)
+                self.mutually_exclusive_names.append(mutex)
+
+        help = kwargs.get("help", "")
```
jsam, line 53: Minor thing, but I believe we should stay away from shadowing built-in names. Maybe rename this to `_help`?

renku/cli/utils/click.py
```diff
@@ -31,3 +31,37 @@ def convert(self, value, param, ctx):
...
+        if self.mutually_exclusive:
+            ex_str = ", ".join(self.mutually_exclusive_names)
+            kwargs["help"] = help + (" NOTE: This argument is mutually exclusive with " " arguments: [" + ex_str + "].")
```
jsam, line 56: Another minor, I think everywhere we are using `.format` or f-strings?

renku/core/management/config.py
```diff
@@ -37,6 +39,15 @@ def _get_global_config_dir():
     return click.get_app_dir(APP_NAME, force_posix=True)
 
 
+class ConfigFilter(Enum):
```
jsam, line 42: I think we have a couple of places where we should roll out more enum based approaches. I really like this approach, only thing I'm wondering if we should do the same with them as we did with errors. Should we make a `enums.py` for the project and put all enums into one place? That way might be easier to reason about all possible options.

renku/core/management/config.py
```diff
@@ -69,29 +80,40 @@ def local_config_path(self):
...
+            config_files += [self.global_config_path, self.local_config_path]
 
-        if not local_only:
+        if config_filter != ConfigFilter.LOCAL_ONLY:
             with self.global_config_lock:
                 config.read(config_files)
         else:
             config.read(config_files)
+
+        # transform config section for backwards compatibility
```
jsam, line 109: Should we add a `NOTE: ` ?

renku/core/management/config.py
```diff
@@ -69,29 +80,40 @@ def local_config_path(self):
...
+        for s in config.sections():
+            if not s.startswith('renku "'):
```
jsam, line 111: Does black not complain here? :thinking:

renku/core/management/config.py
```diff
@@ -69,29 +80,40 @@ def local_config_path(self):
...
+                continue
+
+            config[s[7:-1]] = dict(config.items(s))
```
jsam, line 114: Should we add a small comment on what `s[7:-1]` is selecting?

renku/core/management/config.py
Panaetius: About the `s` ?

renku/service/controllers/config_set.py
```diff
@@ -0,0 +1,51 @@
...
+# limitations under the License.
+"""Renku service cache list cached projects controller."""
+from renku.core.commands.config import update_multiple_config
+from renku.service.controllers.api.abstract import ServiceCtrl
+from renku.service.controllers.api.mixins import ReadOperationMixin
+from renku.service.serializers.config import ConfigSetRequest, ConfigSetResponseRPC
+from renku.service.views import result_response
+
+
+class SetConfigCtrl(ServiceCtrl, ReadOperationMixin):
```
jsam, line 26: Do we want to sync this repository change back to the remote?

```suggestion
class SetConfigCtrl(ServiceCtrl, ReadWithSyncOperation):
```

renku/service/controllers/config_set.py
```diff
@@ -0,0 +1,51 @@
...
+    def renku_op(self):
+        """Renku operation for the controller."""
+        update_config_command = update_multiple_config().build()
+        update_config_command.execute(self.ctx["config"])
+
+        return {}
+
+    def to_response(self):
+        """Execute controller flow and serialize to service response."""
+        return result_response(SetConfigCtrl.RESPONSE_SERIALIZER, self.execute_op())
```
jsam, line 51: In case you want to sync this update to the remote repository.
```suggestion
        return result_response(SetConfigCtrl.RESPONSE_SERIALIZER, self.execute_and_sync())
```

renku/service/views/config.py
```diff
@@ -0,0 +1,69 @@
...
+
+@use_kwargs(ConfigShowRequest, location="query")
+@marshal_with(ConfigShowResponseRPC)
+@header_doc(description="Show renku config for a project.", tags=(CONFIG_BLUEPRINT_TAG,))
+@config_blueprint.route(
+    "/config.show", methods=["GET"], provide_automatic_options=False,
+)
+@handle_common_except
+@requires_cache
+@requires_identity
```
jsam, line 51: I'm wondering if there is any use case to add support for anonymous session on this endpoint. Maybe preview of configuration shown in the UI?

jsam (commented):
Look's amazing! Thank you very much.

I just posted a couple of questions before we merge. :+1:

renku/core/management/config.py
jsam: No I mean for the single quotes :D

Panaetius: Actually, black is the one that changes it from `"renku \""` to `'renku "'` :man_shrugging:

jsam:
Thanks, look's great! Let me know once you rebase it so we can merge it! :rocket:

jsam (approved):
Thank you. :rocket:

PR #1835, opened 2021-01-14
feat(svc): add helm 3 values schema to chart

### LINKED ISSUE #1755: Add JSON schema validation for helm chart
We should add a validation schema for our helm chart: https://helm.sh/docs/topics/charts/#schema-files

This allows us to enforce the structure of the values file needed, along with constraints that make sense (for instance min length of 32 for `jwtTokenSecret`)

Panaetius:
closes #1755


helm-chart/renku-core/values.schema.json
```diff
@@ -0,0 +1,209 @@
...
+          "description": "Kubenetes resource requests/limites (cpu/memory/etc.) to request",
+          "type": "object"
+      },
+      "nodeSelector": {
+          "description": "Node selector for deployment",
+          "type": "object"
+      },
+      "tolerations": {
+          "description": "Tolerations for deployment",
+          "type": "object"
```
mohammad-sdsc, line 162: Shouldn't this be an array? In the values files it is: `tolerations: []`

helm-chart/renku-core/values.schema.json
```diff
@@ -0,0 +1,209 @@
...
+      },
+      "affinity": {
+          "description": "Affinity for deployment",
+          "type": "object"
+      },
+      "redis": {
+          "description": "Settings for the redis subchart",
+          "type": "object"
```
mohammad-sdsc, line 170: redis properties are missing. Do we need them?

helm-chart/renku-core/values.yaml
```diff
@@ -7,7 +7,7 @@ replicaCount: 1
 cacheDirectory: /svc/cache
 cleanupInterval: 3600
 projectCloneDepth: 1
-templateCloneDepth: 0
+templateCloneDepth: 1
```
mohammad-sdsc, line 10: We should also update the docs in `helm-chart/README.rst`: 
```
`templateCloneDepth` git clone depth for template projects (default: `0`)
```

mohammad-sdsc (commented):
Looks good! Just some minor comments/questions.

helm-chart/renku-core/values.schema.json
Panaetius: They're for the redis subchart, so it doesn't make sense to validate values of another chart here.

mohammad-sdsc (approved):
Thank you!

jsam (approved):
:rocket:

PR #1836, opened 2021-01-14
fix(service): reset cache after failed push

### LINKED ISSUE #1794: Improve migration for projects with protected branches
When a project with a protected branch is migrated we should receive back the "remote_branch" that contains the changes so that we can to notify the user that the master was not really migrated. 

Also, when this happens if i do a migration_check after i should see again that the project hasn't been migrated (since what i should see is what is inside master).

### LINKED ISSUE #1795: git reset project in service cache in case of protected branches
If we do an action like migration on a protected branch, we will push the changes made to a new branch, but the state of the protected branch in the cache will contain the migrations, making it out of sync with the state of the protected branch in gitlab.

We should git-reset the project in the cache back to the previous state if we're dealing with a protected branch, to keep the states in the cache and gitlab in sync.

mohammad-sdsc:
# Description

Reset service cache state if pushing to remote branch fails. The push will be done to another remote branch which its name is returned as `remote_branch` in the JSON response.

Fixes #1795 
Fixes #1794


Panaetius (approved):
Thank you!

PR #1838, opened 2021-01-15
fix(core): output git lfs error messages when there is an error

### LINKED ISSUE #1814: Errors not reported with renku storage commands
**Describe the bug**
If there's an issue when executing commands like `git lfs pull` the corresponding `renku storage pull` command does not output the stack trace. The renku command exits (successfully) instead of reporting the error.

**Additional context**
This happened in a couple of deployments when testing LFS and it was not working properly.

Panaetius:
closes #1814

### DIFF
--- a/renku/core/management/storage.py
+++ b/renku/core/management/storage.py
@@ -24,7 +24,7 @@ import tempfile
 from collections import defaultdict
 from pathlib import Path
 from shutil import move, which
-from subprocess import PIPE, STDOUT, call, check_output, run
+from subprocess import PIPE, STDOUT, check_output, run
 
 import attr
 import pathspec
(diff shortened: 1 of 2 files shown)


renku/core/management/storage.py
```diff
@@ -182,23 +189,37 @@ def track_paths_in_storage(self, *paths):
...
+            result = run(
+                self._CMD_STORAGE_UNTRACK + list(paths),
+                stdout=PIPE,
+                stderr=STDOUT,
+                cwd=self.path,
+                universal_newlines=True,
             )
+
+            if result.returncode != 0:
+                raise errors.GitLFSError("fError executing 'git lfs untrack: \n {result.stdout}")
```
mohammad-sdsc, line 220: "fError ... -> f"Error ...

mohammad-sdsc (commented):
Look good! Just a minor typo.

renku/core/management/storage.py
Panaetius: Good catch! Thank you

mohammad-sdsc (approved):
Thank you!

PR #1839, opened 2021-01-15
refactor(core): workflow migration speedup and progress feedback

### LINKED ISSUE #1837: Migration progress feedback to users
As a renku user, I want to see activity progress on long migrations tasks (e.g. workflow migrations) so that I know things are working and have an estimate of when they are going to finish.

mohammad-sdsc:
# Description

Shows a progress report when migrating workflows and makes pre-processing CWL files faster by traversing git history once.

Fixes #1837

### DIFF
--- a/renku/cli/init.py
+++ b/renku/cli/init.py
@@ -349,7 +349,7 @@ def init(
         )
 
     template_manifest, template_folder, template_source, template_version = fetch_template(
-        template_source, template_ref, click.echo
+        template_source, template_ref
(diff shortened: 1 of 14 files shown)


Panaetius (changes requested):
Thank you, great improvement and good to have the command pattern in migration as well! Minor comment that might improve performance a bit more, at least if there's large commits.

Panaetius (approved):

PR #1849, opened 2021-01-21
fix(core): fix renku save with deleted files

Panaetius:
Fixes issue with getting staged paths and also dealing with deleted&staged files in `renku save`

### DIFF
--- a/renku/core/commands/save.py
+++ b/renku/core/commands/save.py
@@ -93,7 +93,9 @@ def repo_sync(repo, message=None, remote=None, paths=None):
     if paths:
         # commit uncommitted changes
         try:
-            staged_files = {d.a_path for d in repo.index.diff("HEAD")} if repo.head.is_valid() else set()
+            staged_files = (
+                {git_unicode_unescape(d.a_path) for d in repo.index.diff("HEAD")} if repo.head.is_valid() else set()
+            )
 
             path_to_save = set(paths) - staged_files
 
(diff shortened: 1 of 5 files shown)


renku/core/commands/save.py
```diff
@@ -93,7 +93,13 @@ def repo_sync(repo, message=None, remote=None, paths=None):
     if paths:
         # commit uncommitted changes
         try:
-            repo.git.add(*paths)
+            staged_files = {d.a_path for d in repo.index.diff("HEAD")} if repo.head.is_valid() else set()
```
mohammad-sdsc, line 96: Should we call `git_unicode_unescape` on `d.a_path`? If so, then we should also call it on `paths` as well.

renku/core/management/git.py
```diff
@@ -128,7 +128,7 @@ def modified_paths(self):
     def dirty_paths(self):
         """Get paths of dirty files in the repository."""
         repo_path = self.repo.working_dir
-        staged_files = self.repo.index.diff("HEAD") if self.repo.head.is_valid() else []
+        staged_files = [d.a_path for d in self.repo.index.diff("HEAD")] if self.repo.head.is_valid() else []
```
mohammad-sdsc, line 131: Same about calling `git_unicode_unescape`.

mohammad-sdsc (commented):
LGTM, just two comments about unicode file names.

renku/core/commands/save.py
Panaetius: We only need to call it on paths that come from git, so yes about adding it here, but `paths` is supplied by us/the user, so there it's not necessary.

mohammad-sdsc (approved):
Thanks!

PR #1850, opened 2021-01-21
feat(svc): add support for adding images to datasets

### LINKED ISSUE #1707: feat(core): allow users to associate an image with a dataset
The UI would like to be able to show an image in dataset listings/details, helping to make the dataset UI more visually pleasing and usable.

We can use https://schema.org/image to store this.

An open question is how to store this. Ideally we'd store it as an URL to the image in the repository, but that might need some path rewriting to display properly in the UI.

We should discuss which format makes the most sense for both teams, and how a user would set and/or change this, both through the cli and through the core service.

Panaetius:
closes #1707 

/deploy


renku/core/management/datasets.py
```diff
@@ -266,13 +278,82 @@ def create_dataset(self, name=None, title=None, description=None, creators=None,
...
+                img_path = image_folder / f"{position}{ext}"
+                shutil.copy(path, img_path)
+
+            dataset.images.append(
+                ImageObject(
+                    str(img_path.relative_to(self.path)), position, id=ImageObject.generate_id(dataset, position)
+                )
+            )
+            images_updated = True
+
```
mohammad-sdsc, line 354: Should we delete image files of images that have been deleted?

tests/service/views/test_dataset_views.py
```diff
@@ -1036,6 +1211,117 @@ def test_edit_datasets_view(svc_client_with_repo):
     assert {"title": "my new title", "keywords": ["keyword1"]} == response.json["result"]["edited"]
 
 
+@pytest.mark.service
+@pytest.mark.integration
+@flaky(max_runs=10, min_passes=1)
+def test_edit_dataset_with_images(svc_client_with_repo):
+    """Create a new dataset with metadata."""
```
mohammad-sdsc, line 1218: `Edit a dataset with images`

mohammad-sdsc (commented):
Thanks Ralf for looking into this! I've made a few comments.

renku/core/management/datasets.py
```diff
@@ -266,13 +278,82 @@ def create_dataset(self, name=None, title=None, description=None, creators=None,
...
+                continue
+
+            path = content_url
+            if not os.path.isabs(path):
+                path = os.path.normpath(os.path.join(self.path, path))
+
+            if not os.path.exists(path) or (
+                not os.path.commonprefix([path, self.path]) == str(self.path)
+                and not os.path.commonprefix([path, CACHE_UPLOADS_PATH]) == str(CACHE_UPLOADS_PATH)
+            ):
```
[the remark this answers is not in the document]
Panaetius: We don't want someone to upload ```
{
  ...
  "images":[{
    "content_url": "/etc/passwd",

mohammad-sdsc (dismissed):
Thank you!

mohammad-sdsc (approved):

PR #1851, opened 2021-01-21
refactor(core): unify new graph generation

mohammad-sdsc:
# Description

Use `renku graph generate` to generate new graph metadata for both workflows and datasets. Use `renku graph export` to generate triples for workflows and datasets.


Panaetius (approved):
Thank you, looks great!

PR #1861, opened 2021-02-04
chore: updates license headers for 2021

jsam:


tests/core/commands/providers/test_dataverse.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright 2020 Swiss Data Science Center (SDSC)
+# 2021Swiss Data Science Center (SDSC)
```
Panaetius, line 3: This one is borked

Panaetius (changes requested):

tests/core/commands/providers/test_dataverse.py
jsam: Oops. Thanks for catching it.

Panaetius (approved):
Thank you :pray:

PR #1862, opened 2021-02-04
chore: sort dependencies in setup.py

jsam:


setup.py
```diff
@@ -247,14 +247,14 @@ def _get_disribution_url():
     author_email="contact@datascience.ch",
     url="https://github.com/swissdatasciencecenter/renku-python",
     project_urls={
-        "Changelog": ("https://github.com/swissdatasciencecenter/renku-python" "/blob/master/CHANGES.rst"),
+        "Changelog": "https://github.com/swissdatasciencecenter/renku-python" "/blob/master/CHANGES.rst",
```
Panaetius, line 250:
```suggestion
        "Changelog": "https://github.com/swissdatasciencecenter/renku-python/blob/master/CHANGES.rst",
```

setup.py
```diff
@@ -150,13 +149,14 @@ def run(self):
     extras_require["nodocs"].extend(reqs)
 
 install_requires = [
-    "appdirs>=1.4.3,<=1.4.4 ",
+    "PyYAML>=3.12,<=5.3.1",
```
Panaetius, line 152: Did you use isort for this? I wonder if case-insensitive sort would be better. At least I could see myself searching for pyyaml and being confused when it's not with the other `py*` entries. But just a minor detail

Panaetius (changes requested):

setup.py
jsam: Fair point. Used case-insensitive sort.

Panaetius (approved):
:rocket:

Panaetius (approved):

PR #1863, opened 2021-02-04
chore: fixes usages of deprecated attrs parameter 'cmp='

jsam:
`attrs` has deprecated `cmp=` flag and it's forcing users to move to `eq`and `order`. This PR updates those usages to stay forward compatible and removes those deprecation warnings when we run tests locally.


Panaetius (approved):
Thanks :smile:

PR #1864, opened 2021-02-05
feat(core): renku login

### LINKED ISSUE #677: Log in to remote renku platform with the CLI
A user should be able to log in to the hosted renku platform with the command-line client. This 

`renku login <endpoint>`

optionally, the default endpoint could be set in the configuration, e.g. 

`renku config --set endpoint=renkulab.io`

The specific use-case is to implement a credentials helper to make pushing/pulling from git repositories and docker registries less painful. For example, one could do something like

```
renku login https://renkulab.io
> Username: abcd
> Password: 1234
```

This would store my JWT for renkulab.io and when I did
```
git push https://renkulab.io/...
```

it would swap that JWT with the gateway for a scoped oauth2 token that it could use for the git commands. A similar credentials helper could (presumably) be implemented for logging in to the docker registry, making it much easier to work locally with images produced by the CI pipelines. 

Writing really long documentation about ssh keys and private tokens. 

This has been partially (or almost fully?) implemented in #193 

This one is a good example of a good UX https://concourse-ci.org/fly.html#fly-login and more-or-less what was attempted in #193.

mohammad-sdsc:
# Description

Adds a `renku login` command that communicates with Renku's gateway and receives and stores an oauth access token. ATM, this token is used when sending request to access KG.

Fixes #677

### DIFF
--- a/renku/cli/login.py
+++ b/renku/cli/login.py
@@ -19,7 +19,7 @@
 
 You can use ``renku login`` command to authenticate with a remote Renku
 deployment. This command will bring up a browser window where you can log in
-using your credentials. Renku CLI receives and stores a secure token where will
+using your credentials. Renku CLI receives and stores a secure token that will
 be used for future authentications.
 
 .. code-block:: console
--- a/renku/cli/login.py
+++ b/renku/cli/login.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2018-2021 Swiss Data Science Center (SDSC)
+# A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+# Eidgenössische Technische Hochschule Zürich (ETHZ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Logging in to a Renku deployment.
+
+TODO: Write Documentation
+"""
+
+import click
+
+from renku.cli.utils.callback import ClickCallback
+from renku.core.commands.login import login_command, logout_command
+
+
+@click.command()
+@click.argument("endpoint", required=False, default=None)
+def login(endpoint):
+    """Log in to the platform."""
+    communicator = ClickCallback()
+    login_command().with_communicator(communicator).build().execute(endpoint=endpoint)
+
+    # TODO: Print a warning that token is stored in plain-text
+
+
+@click.command()
+def logout():
+    """Logout from the platform and delete credentials."""
+    communicator = ClickCallback()
+    logout_command().with_communicator(communicator).build().execute()
--- a/renku/cli/login.py
+++ b/renku/cli/login.py
@@ -45,7 +45,10 @@ Logging out from Renku removes the secure token from your system:
 
 .. code-block:: console
 
-    $ renku logout
+    $ renku logout <endpoint>
+
+If you don't specify an endpoint when logging out, credentials for all
+endpoints are removed.
 """
 
 import click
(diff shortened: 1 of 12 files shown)


rokroskar:
running `renku login` outside of a renku repo works, but then crashes at the end because there isn't a renku repo. I wonder if it _should_ work regardless of whether you are in a repo or not (I can see it being useful in some future cases) or do we want to fail gracefully and tell the user they need to be in a repo if they want to log in?

mohammad-sdsc:
> running `renku login` outside of a renku repo works, but then crashes at the end because there isn't a renku repo. I wonder if it _should_ work regardless of whether you are in a repo or not (I can see it being useful in some future cases) or do we want to fail gracefully and tell the user they need to be in a repo if they want to log in?

ATM, this works but it does not do much: We need to modify git config in order to allow access to gitlab; outside a git repo users will get a warning that git config is not modified. We only store the token to global renku config.

rokroskar:
That's fine for now I think - we can iterate on this if needed. i don't see why we wouldn't add it to the user git-config, but lets stick with this for now.

renku/cli/login.py
```diff
@@ -0,0 +1,75 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Logging in to a Renku deployment.
+
+You can use ``renku login`` command to authenticate with a remote Renku
+deployment. This command will bring up a browser window where you can log in
+using your credentials. Renku CLI receives and stores a secure token where will
```
Panaetius, line 22:
```suggestion
using your credentials. Renku CLI receives and stores a secure token that will
```

renku/core/commands/login.py
```diff
@@ -0,0 +1,116 @@
...
+
+    communication.echo(
+        f"Please log in at {parsed_endpoint.geturl()} on your browser.\n"
+        "Once completed, you may close the browser window."
+    )
+
+    login_url = _get_url(parsed_endpoint, "/api/auth/login", query)
+    webbrowser.open_new_tab(login_url)
+
+    info_url = _get_url(parsed_endpoint, "/api/auth/info", query)
```
Panaetius, line 56: The login URL should really not just show the regular login page, but something like "Application 'Renku Commandline' wants access to your renkulab.io account. Authorize? Yes | No". Otherwise this is a pretty big security risk.

I could just post a link to facebook or something with "Hey, checkout my cool renku project" and it links to the /api/auth/login URL with an UUID that I (the attacker) know, and then I can just monitor the endpoint to get a users token and wait for someone to click the link. Even worse if the user is already logged in, in which case clicking the link is already enough and they don't need to even enter their credentials. As soon as they click the link, they would have basically lost.

With the more traditional approach of having a local webserver as a callback, this is somewhat mitigated as you can limit the callback URL to being `localhost`.

But either way, we need a page that tells the user what is going to happen and where they have to explicitly confirm that this is what they want.

Panaetius (changes requested):

Panaetius (dismissed):
Thank you!

mohammad-sdsc:
> Really nicely done! I have just one tiny request: When I first typed `renku login` without providing a RenkuLab URL, I got the infamous `aaaaahhh, you've found a bug`. It would be nice to display the output of `renku login -h` as you do it for other commands.

Nice catch! There is a test for it but this corner case was not covered. Users should be able to run `renku login` with no argument which will read the endpoint from the config file.

renku/core/utils/urls.py
```diff
@@ -70,6 +70,9 @@ def parse_authentication_endpoint(client, endpoint):
     if not endpoint:
         endpoint = client.get_value(section="renku", key="endpoint")
         if not endpoint:
+            print("NO ENDPOINT", use_remote)
```
Panaetius, line 73: Debug statement accidentally left in?

tests/cli/test_login.py
```diff
@@ -0,0 +1,146 @@
...
+ENDPOINT = "renku.deployment.ch"
+USER_CODE = "valid_user_code"
+
+
+def test_login(runner, client, mock_login):
+    """Test login command."""
+    result = runner.invoke(cli, ["login", ENDPOINT], input=USER_CODE)
+
+    assert 0 == result.exit_code
+    assert "jwt-token" == read_renku_token(client, ENDPOINT)
```
Panaetius, line 32: I think `mock_login` should yield the token strings ('jwt-token', 'other-token') along with the mock, then we can compare to that here (not have the string defined in two places)

Panaetius (changes requested):

Panaetius (approved):
LGTM! :rocket:

Panaetius (approved):

ableuler (approved):

ableuler:
Nice work @mohammad-sdsc! 💪

PR #1871, opened 2021-02-10
feat: adds root redirect to swagger docs

jsam:


rokroskar (approved):
works as advertised - thanks!

PR #1872, opened 2021-02-10
fix: correctely address HTTP server errors

jsam:
This PR adds correct management of HTTP server errors which can bubble up from Flask/Werkzeug. Note, these exceptions are bubbling up from different component and are not Renku core related.

It aligns them to the JSON-RPC status codes in such way that HTTP `404` becomes JSON-RPC `-32404` and HTTP 50x becomes JSON-RPC `-3250x`.

Before, for non-existing routes and server related failures service would return "Internal server error". Note, different from `{"error": -32100, "reason": "internal error"}` which is related to wrong setup for renku operation command.

This PR can be checked with:
image


mohammad-sdsc (approved):
LGTM, thanks!

PR #1878, opened 2021-02-15
feat(svc): add support for storing remote dataset images in the repo

### LINKED ISSUE #1842: feat(svc): allows users to force copying dataset images from absolute urls
Following the changes in https://github.com/SwissDataScienceCenter/renku-python/issues/1707 a user should be able to tell the core service to download an image and copy it to the `.renku/datasets/<UUID>/images/` folder when adding an image to a dataset.

This follows the same semantics like copying the image when passing a relative path.

This way, users can ensure that remote urls are not suddenly dead and that we don't hammer third party hosts with requests.

The original request should be adjusted with an optional field `force_download` like so 
```
"images": [
    {
        "content_url": "...absolute or relative url...",
        "file_id": "...id of a cache.file_upload...",
        "position": integer,
        "force_download": bool
  }
]
```

Panaetius:
closes #1842


renku/core/management/datasets.py
```diff
@@ -331,12 +333,31 @@ def set_dataset_images(self, dataset, images, safe_image_paths=[]):
...
                 dataset.images.append(existing)
                 continue
-
+            image_type = None
             if urlparse(content_url).netloc:
                 # NOTE: absolute url
-                dataset.images.append(ImageObject(content_url, position, id=ImageObject.generate_id(dataset, position)))
-                images_updated = True
-                continue
+                if "download" not in img or not img["download"]:
```
mohammad-sdsc, line 339: I would use `if not img.get("download", False)`.

mohammad-sdsc (commented):
Looks pretty good!

renku/core/models/datasets.py
```diff
@@ -985,6 +985,7 @@ class ImageObjectRequestJson(marshmallow.Schema):
     file_id = marshmallow.fields.String()
     content_url = marshmallow.fields.String()
     position = marshmallow.fields.Integer()
+    download = marshmallow.fields.Bool(default=False)
```
jsam, line 988: Should we hint the type through better naming here? Maybe `is_downloaded` is better suited?

tests/service/views/test_dataset_views.py
```diff
@@ -222,6 +222,78 @@ def test_create_dataset_with_images(svc_client_with_repo):
...
+@pytest.mark.service
+@pytest.mark.integration
+@flaky(max_runs=10, min_passes=1)
+def test_create_dataset_with_image_download(svc_client_with_repo):
+    """Create a new dataset with metadata."""
+    svc_client, headers, project_id, _ = svc_client_with_repo
+
+    payload = {
+        "project_id": project_id,
+        "name": "{0}".format(uuid.uuid4().hex),
```
jsam, line 234:
```suggestion
        "name": f"{uuid.uuid4().hex}",
```

tests/service/views/test_dataset_views.py
```diff
@@ -222,6 +222,78 @@ def test_create_dataset_with_images(svc_client_with_repo):
...
+        "images": [{"content_url": "https://renkulab.io", "position": 1, "download": True},],
+    }
+
+    response = svc_client.post("/datasets.create", data=json.dumps(payload), headers=headers,)
+    assert response
+    assert {"error"} == response.json.keys()
+
+    payload = {
+        "project_id": project_id,
+        "name": "{0}".format(uuid.uuid4().hex),
```
jsam, line 247:
```suggestion
        "name": f"{uuid.uuid4().hex}",
```

tests/service/views/test_dataset_views.py
```diff
@@ -222,6 +222,78 @@ def test_create_dataset_with_images(svc_client_with_repo):
...
+        "images": [{"content_url": "https://renkulab.io/doesnt_exist.png", "position": 1, "download": True},],
+    }
+
+    response = svc_client.post("/datasets.create", data=json.dumps(payload), headers=headers,)
+    assert response
+    assert {"error"} == response.json.keys()
+
+    payload = {
+        "project_id": project_id,
+        "name": "{0}".format(uuid.uuid4().hex),
```
jsam, line 260:
```suggestion
        "name": f"{uuid.uuid4().hex}",
```

tests/service/views/test_dataset_views.py
```diff
@@ -222,6 +222,78 @@ def test_create_dataset_with_images(svc_client_with_repo):
...
+        "title": "my little dataset",
+        "creators": [{"name": "name123", "email": "name123@ethz.ch", "affiliation": "ethz"}],
+        "description": "my little description",
+        "images": [
+            {
+                "content_url": "https://raw.githubusercontent.com/SwissDataScienceCenter/calamus/master/docs/reed.png",
```
jsam, line 266: Can we extract this touch point as a parameter and add a version with a redirect (i.e. made with url shortnerer)?

jsam (commented):
Look's great!

Added few small suggestions/questions.

renku/core/models/datasets.py
Panaetius: `is_downloaded` would be a bit misleading, as it's more a `should_be_downloaded`, i.e. no download has happened at the time the request is sent. Maybe `do_download`, `force_download` or `should_be_downloaded` ?

tests/service/views/test_dataset_views.py
Panaetius: Actually, 
```
>>> a = uuid.uuid4().hex
>>> a == "{0}".format(a)
True
```

Why do we even have that in code

renku/core/models/datasets.py
Panaetius: I named it `mirror_locally`. Couldn't really think of a name that's not overly verbose, like `should_be_mirrored_to_repository` and I think this should be at least a bit more clear than just `download`.

renku/core/management/datasets.py
```diff
@@ -383,12 +385,29 @@ def set_dataset_images(self, dataset, images, safe_image_paths=[]):
...
+
+                # NOTE: mirror the image locally
+                try:
+                    path, _ = urlretrieve(content_url)
+                except error.URLError as e:
+                    raise errors.DatasetImageError(f"Dataset image with url {content_url} couldn't be mirrored") from e
+
+                image_type = imghdr.what(path)
+                if not image_type:
+                    raise errors.DatasetImageError(f"Dataset image with url {content_url} is not a valid image type")
```
mohammad-sdsc, line 406: Should we raise here? It seems that `imghdr` does not work all the time: https://stackoverflow.com/questions/36870661/imghdr-python-cant-detec-type-of-some-images-image-extension. This might cause unnecessary bug reports in future and I don't really see any value in doing this. UI can do the check as it needs to do it anyway for image URLs.
Maybe just use `imghdr` to detect the extension.

mohammad-sdsc (commented):

renku/core/management/datasets.py
```diff
@@ -383,12 +385,29 @@ def set_dataset_images(self, dataset, images, safe_image_paths=[]):
...
+                if not img.get("mirror_locally", False):
+                    dataset.images.append(
+                        ImageObject(content_url, position, id=ImageObject.generate_id(dataset, position))
+                    )
+                    images_updated = True
+                    continue
+
+                # NOTE: mirror the image locally
+                try:
+                    path, _ = urlretrieve(content_url)
```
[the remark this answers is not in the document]
Panaetius: Not really any reason, other than it being shorter/a bit more readable than:
```
r = requests.get(url, stream=True)
if r.status_code == 200:
    path = tmpdir.mkdir() / create_temp_filename_somehow()
    with open(path, 'wb') as f:
        r.raw.decode_content = True
        shutil.copyfileobj(r.raw, f)       
```
or
```
r = requests.get(url, stream=True)
if r.status_code == 200:
    path = tmpdir.mkdir() / create_temp_filename_somehow()
    with open(path, 'wb') as f:
        for chunk in r:
            f.write(chunk)
```

We don't have a generic download method that uses requests anywhere, to my knowledge. `urlretrieve` just deals with quite a bit of logic that we'd have to otherwise implement ourselves.

mohammad-sdsc (approved):
Thank you!

Issue #1877, opened 2021-02-12
test: improve testing fixtures


jsam:
To split up dependencies of our testing modules, we should cleanup and improve our testing fixtures. Currently, all of our testing fixtures are a part of `conftest.py`, and quite a few of them don't have local imports of dependencies which have been moved to extras. 

I propose that we keep global fixtures in `conftest.py` and move all the rest into the corresponding module it's testing. Global fixtures are the ones which are used in all 3 components, while the fixtures which are only related to `cli`, `core` or `service` we move into those testing modules through the use of `@pytest.mark.usefixtures`.

This should help with keeping things more clean and we should gain separation of concerns on the module level. Meaning that on the CI we don't have to install all dependencies anymore, but we can have `cli` tests install only dependencies which it needs and `service` tests install `.[service]` and on integration `.[all]`. 

edit: 
Once we do this cleanup, we should update our CI runners accordingly so we cover cases which might occur due to extras depending commands, ie. `service`.

rokroskar:
integration tests don't usually need the service, right? So they should still just install the base dependencies.

jsam:
Service also have integration tests. Those tests are double marked both with `integration` and `service`.

Panaetius:
Also check that the tests/ subfolders are properly split and there's no e.g. cli tests in `test/core`.

PR #1881, opened 2021-02-15
feat(dataset): oauth authentication for KG access

### LINKED ISSUE #1854: Allow passing a OAuth2/gitlab private token on dataset import from Renkulab
With the changes in the KG to secure projects ( https://github.com/SwissDataScienceCenter/renku/issues/1533 ) we will need to authenticate users somehow when importing datasets from renku, to have access to the relevant KG endpoints.

In a first iteration, we can do this by having the user supply an OAuth2 token or gitlab private key to the import command (similar to tokens used for export to zenodo and dataverse).

At some point in the future, we might implement a `renku login` command that simplifies this step, but for now, this should be enough to be compatible with the KG security changes.

mohammad-sdsc:
# Description

Reads and includes an OAuth token when querying KG for dataset imports.

Fixes #1854

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -493,6 +493,21 @@ def missing_kg_project_responses():
         yield rsps
 
 
+@pytest.fixture
+def mock_kg():
+    """Mock KG responses for dataset import."""
+    with responses.RequestsMock(assert_all_requests_are_fired=False) as requests_mock:
+
+        def callback(request):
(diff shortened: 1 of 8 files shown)


jsam (commented):
Look's pretty good. Let's rebase and merge it before #1877 gets in. :rocket: 

Let me know once you rebase.

jsam (approved):

PR #1884, opened 2021-02-16
fix: rdflib.ConjunctiveGraph.bind API of 4.2.2

### LINKED ISSUE #1882: Custom bindings for ProvenanceGraph are broken
By setting ProvenanceGraph.custom_bindings renku will throw an exception when creating the graph, as the `bind` method uses an argument which is only available in rdflib==5.0.0. Namely, `replace=True` kwarg.
The implementation here: https://github.com/SwissDataScienceCenter/renku-python/blob/master/renku/core/models/provenance/provenance_graph.py#L141

related to https://github.com/SwissDataScienceCenter/renku-python/issues/1859

vigsterkr:
# Description

If the ProvenanceGraph.custom_bindings property is being set, the `rdflib.ConjunctiveGraph.bind` function is called with `replace=True` function argument. This is only available in rdflib==5.0.0 or later.

Fixes #1882

### DIFF
--- a/renku/core/models/provenance/provenance_graph.py
+++ b/renku/core/models/provenance/provenance_graph.py
@@ -138,7 +138,7 @@ class ProvenanceGraph:
         self._graph.bind("wfprov", "http://purl.org/wf4ever/wfprov#")
 
         for prefix, namespace in self._custom_bindings.items():
-            self._graph.bind(prefix, namespace, replace=True)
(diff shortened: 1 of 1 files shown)


mohammad-sdsc (approved):
LGTM, thanks!

jsam (approved):

PR #1896, opened 2021-02-23
chore: upgrade rdflib to 5.0.0

### LINKED ISSUE #1859: update rdflib to >5.0.0
With https://github.com/common-workflow-language/schema_salad/pull/373 done, we should be able to update RDFLib to 5.0.0 as well (as cwltool was the only thing keeping us back). We need to make sure this doesn't break anything and see if we need to do any changes on our end.

mohammad-sdsc:
Fixes #1859


setup.py
```diff
@@ -154,13 +154,13 @@ def run(self):
...
     "attrs>=19.3.0,<=20.2.0",
-    "calamus>=0.3.6,<0.3.7",
+    "calamus>=0.3.7,<0.3.8",
     "click-completion>=0.5.0,<=0.5.3",
     "click-plugins==1.1.1",
     "click>=7.0,<=7.1.2",
     "cryptography>=3.4.1,<3.5",
     "cwlgen>=0.4.0,<=0.4.2",
-    "cwltool>=3.0.20200724003302,<=3.0.20200807132242",
+    "cwltool==3.0.20210124104916",
```
Panaetius, line 163: it makes sense to have requirements like `something>=3.0.1,<=3.0.1` or `something>=3.0.1,<3.1`, even if it's weird to pin a single version like this, but over time, dependabot will change those to dependency ranges that make sense, rather than just updating the number (if you have `something==3.0.1`, dependabot will just change it to e.g. `something==3.0.2`, whereas `something>=3.0.1,<=3.0.1` would get updated to `something>=3.0.1,<=3.0.2`). Same below with rdflib. 

I think in general, we always want to have an upper bound to prevent running into issue where a new version breaks things, but to only constrain it down to the minor version. e.g. `rdflib>=5.0.0,<5.1` for rdflib. That reasonably safe but still allows for patch releases.

mohammad-sdsc: Thanks for the explainataion!

Panaetius (approved):

PR #1898, opened 2021-02-24
feat(core): exclude renku metadata from lfs

### LINKED ISSUE #1866: prevent renku files from getting added to git lfs
We currently have the `lfs_threshold` config flag and the `.renkulfsignore` file to handle which files get added to git lfs. If a user sets the `lfs_threshold` to e.g. `0b`, files in `.renku/` could potentially be added to git lfs.

We should prevent this from happening in code (I think .renkulfsignore is the wrong place for this, as we can't make sure third party templates respect this).

Two solutions come to mind:
- Append a respective rule to `StorageApiMixin.renku_lfs_ignore` after the file is loaded, which should take care of all places this could happen
- Add manual checks for this everywhere where we could add files to lfs

The first of which is probably the better solution.

mohammad-sdsc:
# Description

Prevents `.renku` directory from being tracked in LFS.

Fixes #1866

### DIFF
--- a/renku/core/management/storage.py
+++ b/renku/core/management/storage.py
@@ -41,10 +41,6 @@ from renku.core.utils.git import add_to_git, run_command
 from .git import _expand_directories
 from .repository import RepositoryApiMixin
 
-# Batch size for when renku is expanding a large list
-# of files into an argument string.
-ARGUMENT_BATCH_SIZE = 100
-
 
 def check_external_storage_wrapper(fn):
     """Check availability of external storage on methods that need it.
--- a/renku/core/management/storage.py
+++ b/renku/core/management/storage.py
@@ -18,6 +18,7 @@
 """Client for handling a data storage."""
 import csv
 import functools
+import itertools
 import os
 import re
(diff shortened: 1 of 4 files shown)


renku/core/management/storage.py
```diff
@@ -124,10 +121,14 @@ def check_external_storage(self):
     def renku_lfs_ignore(self):
         """Gets pathspec for files to not add to LFS."""
         ignore_path = self.path / self.RENKU_LFS_IGNORE_PATH
+        renku_protected_paths = [".renku"]
```
Panaetius, line 124: This would also match e.g. `data/mydataset/.renku`, which might not be intended (`/.renku` would be to only match the .renku directory). But then again, I can't think of a case where we'd add a file or directory called `.renku` somewhere else and want it to end up in lfs, so it probably doesn't matter.

Panaetius (dismissed):
Thank you! I had two minor comments, but nothing of any importance.

Panaetius (approved):

PR #1899, opened 2021-02-24
feat: renku service up/down/ps/restart/logs commands

jsam:
This PR adds the following commands.

```
$ renku service up --help                              
Usage: renku service up [OPTIONS]

  Start ALL service components in daemon mode.

Options:
  -d, --daemon             Starts ALL processes in daemon mode.
  -rd, --runtime-dir TEXT  Directory for runtime metadata in daemon mode.
  -h, --help               Show this message and exit.

```

```
$ renku service ps
create_time       pid  cmdline                  status      mem_perct    cpu_perct    num_threads    num_fds
-------------  ------  -----------------------  --------  -----------  -----------  -------------  ---------
24.Feb 16:03   119592  renku service api        sleeping     0.547716            0              1          7
24.Feb 16:03   119710  renku service api        sleeping     0.495603            0              1         10
24.Feb 16:03   119593  renku service scheduler  sleeping     0.511912            0              1          4
24.Feb 16:03   119594  renku service worker     sleeping     0.509696            0              2          5
```

```
$ renku service logs --help
Usage: renku service logs [OPTIONS]

  Check logs of ALL running daemonized service components.

Options:
  -f, --follow      Follows logs of damonized service components.
  -a, --output-all  Outputs ALL logs of damonized service components.
  -e, --errors      Outputs ALL errors of damonized service components.
  -h, --help        Show this message and exit.

```

```
$ renku service down
Shutting down [119592] `renku service api`
Shutting down [119710] `renku service api`
Shutting down [119593] `renku service scheduler`
Shutting down [119507] `renku service up --runtime-dir /tmp/tmp76f6ohc_`
Shutting down [119594] `renku service worker`
OK
```


renku/cli/service.py
```diff
@@ -139,3 +207,145 @@ def scheduler_start():
...
+
+    click.echo(output)
+
+
+@service.command(name="up")
+@click.option("-d", "--daemon", is_flag=True, default=False, help="Starts ALL processes in daemon mode.")
+@click.option("-rd", "--runtime-dir", default=".", help="Directory for runtime metadata in daemon mode.")
+@click.pass_context
+def all_start(ctx, daemon, runtime_dir):
+    """Start ALL service components in daemon mode."""
```
rokroskar, line 233:
```suggestion
    """Start all service components."""
```

renku/cli/service.py
```diff
@@ -71,6 +85,93 @@ def run_worker(queues):
...
+def check_cmdline(cmdline, include=None):
+    """Check `cmdline` command of a process."""
+    include = include or []
+    service_components = include + SERVICE_COMPONENT_TAGS
+
+    for cmd in service_components:
+        if cmd in cmdline:
+            return True
+
+    return False
```
Panaetius, line 97: Maybe it would be safer/more robust if we set an env variable, like `IS_RENKU=1` on the running subprocesses, then we could check for that with `psutil.Process(...).environ()`?

tests/cli/test_service.py
```diff
@@ -0,0 +1,39 @@
...
+
+    assert "Using runtime directory" in result.output
+    assert 0 == result.exit_code
+
+    result = runner.invoke(cli, ["service", "down"], catch_exceptions=False)
+    assert 0 == result.exit_code
+
+    # NOTE: Instead of booting/waiting/check/terminating, we are bringing the processes immediately
+    # down and asserting the output. This would be the same as the (1), but much quicker due to lack of waiting.
+    assert "Shutting down" in result.output
```
Panaetius, line 39: We should add more tests. Do all_start, all_stop, ps, all_logs work. 

Also check that the processes are actually down after a `down` command.

Panaetius (changes requested):

renku/cli/service.py
jsam: I do see this as one way to achieve portability across different platforms, but I don't see why would that be more safer? 

To me it seems less safer cause environment variables of the process, could be, a mutable values, while the command is the signature of the process and as such is immutable. And even, if we do switch to this way we would still need to factor in for the service component tags and check the command line to determine which exact renku command we found, to exclude the self process. With that said, I believe current solution is fully portable and sufficiently safe (at least on the part of determining the process).

tests/cli/test_service.py
jsam: Seem's like while debugging macos runner, we lost few tests. I'm bringing them back! Thanks for the catch!

renku/cli/service.py
Panaetius: They're not mutable outside the process from what I understand i.e. psutil only gives you the env variables as they were set when the process was started.

My concern was mostly that the current implementation would also pick up `python myscript.py --dataset api` or something like that, and with the env variables we could cut down on something like that happening. 

Not that there is a high chance of this, but if we spawn processes anyways and we can set their initial env variables easily, that is an easy to implement additional sanity check.

Panaetius (approved):

PR #1900, opened 2021-02-25
feat(dataset): improve naming for imported datasets

### LINKED ISSUE #1698: Better naming for imported datasets
When importing a dataset, we create a dataset name (slug) using title of the dataset and its version which contains only alphanumeric characters. This is a bit confusing in the version part because we remove separator dots from version (e.g. `1.0` becomes `10`).

```
renku dataset import -y 10.7910/DVN/WTZS4K
ls data  # 201901_us_flights_10
```

We should keep the dots in version part; for example the above name will be `201901_us_flights_1.0`.

mohammad-sdsc:
# Description

Allow more characters in generated names for imported datasets.

Fixes #1698

### DIFF
--- a/renku/core/models/datasets.py
+++ b/renku/core/models/datasets.py
@@ -21,6 +21,7 @@ import datetime
 import os
 import pathlib
 import re
+import unicodedata
 import uuid
 from pathlib import Path
 from urllib.parse import ParseResult, quote, urljoin, urlparse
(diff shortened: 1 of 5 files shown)


renku/core/models/datasets.py
```diff
@@ -898,23 +899,26 @@ def generate_default_name(dataset_title, dataset_version=None):
...
+    slug = get_slug(dataset_title)
+    slug = slug.lower()[:24]
 
     if dataset_version:
-        version = to_unix(dataset_version)
-        return "{0}_{1}".format("_".join(name), version)
-
-    return "_".join(name)
+        version_slug = get_slug(dataset_version)
+        return f"{slug}_{version_slug}"
```
Panaetius, line 907: Should we change this to use something like `slug[:-len(version_slug)]` and limit version_slug length to something like 10 to ensure a consistent slug max length?

Panaetius (changes requested):
Minor comment, but I think it makes sense if we define a max length for slugs and never exceed that length (e.g. 25 or 30 characters).

Panaetius (approved):

PR #1901, opened 2021-02-26
chore: remove pass_local_client

### LINKED ISSUE #1652: Switch rest of commands to use Command builder
Switch non-workflow and non-dataset commands over to using the Command builder.

image

Panaetius:
Removes `pass_local_client` and changes everything over to command builder.

Closes #1652 

/deploy


mohammad-sdsc (approved):
Great work! I've tried so hard to find something but didn't go so far :D. Thanks!

One thing which can be a followup story is to also remove `@click.pass_context`. It seems that we don't need it anymore.

PR #1924, opened 2021-03-01
test: pytest fixtures breakdown to support better structure

jsam:
closes: #1877


tests/cli/test_isolation.py
```diff
@@ -96,7 +94,3 @@ def test_file_modification_during_run(
 
     with output.open("r") as fp:
         assert "test" == fp.read().strip()
-
-    diff = previous.diff(client.repo.head.commit)
-    modifications = [modification for modification in diff if modification.change_type == "M"]
-    assert 0 == len(modifications)
```
Panaetius, line 102: Why remove this? I think the whole point was that the change done in the line `fp.write('print("edited")')` does not get committed and this tests for that?

tests/cli/fixtures/cli_old_projects.py
```diff
@@ -0,0 +1,153 @@
...
+    repository = Repo(bare_path, search_parent_directories=True).clone(repository_path)
+
+    return repository
+
+
+@pytest.fixture(
+    params=["old-datasets-v0.3.0.git", "old-datasets-v0.5.0.git", "old-datasets-v0.5.1.git", "test-renku-v0.3.0.git"]
+)
+@pytest.mark.usefixtures("request", "tmp_path")
+def old_project(request, tmp_path):
```
Panaetius, line 48: As I understand this, `pytest.mark.usefixtures` only makes sense when a fixture needs to be run for a test to work, but the test itself doesn't access the fixture (going by https://docs.pytest.org/en/stable/fixture.html#use-fixtures-in-classes-and-modules-with-usefixtures ). E.g. if the fixture changes cwd to a temp directory, but the test doesn't care about what the path is.

So adding `usefixtures` and then also specifying the fixture as an argument doesn't do anything if I understand this correctly. Is there something else that needs this set?

tests/cli/test_isolation.py
jsam: Ah, good catch. Rebase went wild. I'll fix it asap.

tests/cli/fixtures/cli_old_projects.py
jsam: When it comes to tests, you're absolutely right (but here we have chained fixtures). This is more of a refactoring trick and could be removed if we don't want to refactor fixtures further. One thing which I see it could be useful to keep it in case if we want to move to multiple `conftest.py` with different configurations, otherwise I can remove it.

Reason why I added it is because during refactoring, pytest collector sometimes did not know about chained fixtures (cause I would move things around), and in those cases I've used this as a hinting mechanism to the collector that this is suppose to be a fixture, even though at the time it was not included and recognized in the runtime as one, this allowed me to easier move things around.

Panaetius (approved):

tests/cli/fixtures/cli_old_projects.py
jsam: Done.

PR #1951, opened 2021-03-08
feat(dataset): locking

### LINKED ISSUE #1653: Add dataset locking mechanism to Command builder

mohammad-sdsc:
# Description

After trying for a few days I could not make this story work in a reliable and clean way. The main problem is that concurrent renku operations can stage files concurrently and then one of them might commit files from the other process. I've tried the two following approaches to allow concurrent atomic commits:
1. Use a global git lock when committing to do add and commit at the same time: This approach breaks when requesting automatic commit at the end of a command because we don't know which metadata files we need to commit unless the command is executed. Therefore, we cannot commit files atomically (see https://github.com/SwissDataScienceCenter/renku-python/pull/1951/commits/3ede9c4d8fec23d4d01fa31e835193a23dfdb4f7).
2. Use git work trees to isolate dataset operations: This approach kinda works but with the following drawbacks:
    1. Untracked files cannot be added to datasets because they won't be copied to a worktree.
    2. Concurrent operations result in a 3-way merge. I've tried to use `rebase` to avoid this but I could not make it work in tests.
    3. Dataset provenance won't work because of conflicts.

Fixes #1653


mohammad-sdsc:
Cannot be done: See the description.

PR #1954, opened 2021-03-10
chore: mock user home and fix renku config concurrency issues

### LINKED ISSUE #1936: Global renku ini lock timeout
Most of our tests seem to be running pretty smoothly now, but there's still the occasional concurrency issue. One issue that I've been seeing more often is https://github.com/SwissDataScienceCenter/renku-python/pull/1925/checks?check_run_id=2020376036#step:7:126
```
____________ ERROR at setup of test_list_datasets_outside_a_context ____________
[gw1] linux -- Python 3.7.10 /opt/hostedtoolcache/Python/3.7.10/x64/bin/python

client = LocalClient(path=PosixPath('/tmp/pytest-of-runner/pytest-0/popen-gw1/test_list_datasets_outside_a_c0/964f2919bdddc5b9'... {'host': None, 'owner': None, 'name': None}}, _migration_type=<MigrationType.ALL: 7>, external_storage_requested=True)
directory_tree = PosixPath('/tmp/pytest-of-runner/pytest-0/popen-gw1/test_list_datasets_outside_a_c0/directory_tree')
/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/pep8.py:110: FutureWarning: Possible nested set at position 1
  EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')

    @pytest.fixture
    def client_with_datasets(client, directory_tree):
        """A client with datasets."""
        from renku.core.models.provenance.agents import Person
    
        person_1 = Person.from_string("P1 <p1@example.com> [IANA]")
        person_2 = Person.from_string("P2 <p2@example.com>")
    
        client.create_dataset(name="dataset-1", keywords=["dataset", "1"], creators=[person_1])
    
        with client.with_dataset("dataset-2", create=True) as dataset:
            dataset.keywords = ["dataset", "2"]
            dataset.creators = [person_1, person_2]
    
>           client.add_data_to_dataset(dataset=dataset, urls=[str(p) for p in directory_tree.glob("*")])

/home/runner/work/renku-python/renku-python/conftest.py:432: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/runner/work/renku-python/renku-python/renku/core/management/datasets.py:558: in add_data_to_dataset
    lfs_paths = self.track_paths_in_storage(*files_to_commit)
/home/runner/work/renku-python/renku-python/renku/core/management/storage.py:57: in wrapper
    return fn(self, *args, **kwargs)
/home/runner/work/renku-python/renku-python/renku/core/management/storage.py:198: in track_paths_in_storage
    if file_size >= self.minimum_lfs_file_size:
/home/runner/work/renku-python/renku-python/renku/core/management/storage.py:136: in minimum_lfs_file_size
    size = self.get_value("renku", "lfs_threshold")
/home/runner/work/renku-python/renku-python/conftest.py:370: in mocked_get_value
    return original_get_value(self, section, key, config_filter=config_filter)
/home/runner/work/renku-python/renku-python/renku/core/management/config.py:142: in get_value
    config = self.load_config(config_filter=config_filter)
/home/runner/work/renku-python/renku-python/renku/core/management/config.py:96: in load_config
    with self.global_config_lock:
/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/filelock.py:323: in __enter__
    self.acquire()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <filelock.UnixFileLock object at 0x7fdb7a00d310>, timeout = 0.0
poll_intervall = 0.05

    def acquire(self, timeout=None, poll_intervall=0.05):
        """
        Acquires the file lock or fails with a :exc:`Timeout` error.
    
        .. code-block:: python
    
            # You can use this method in the context manager (recommended)
            with lock.acquire():
                pass
    
            # Or use an equivalent try-finally construct:
            lock.acquire()
            try:
                pass
            finally:
                lock.release()
    
        :arg float timeout:
            The maximum time waited for the file lock.
            If ``timeout < 0``, there is no timeout and this method will
            block until the lock could be acquired.
            If ``timeout`` is None, the default :attr:`~timeout` is used.
    
        :arg float poll_intervall:
            We check once in *poll_intervall* seconds if we can acquire the
            file lock.
    
        :raises Timeout:
            if the lock could not be acquired in *timeout* seconds.
    
        .. versionchanged:: 2.0.0
    
            This method returns now a *proxy* object instead of *self*,
            so that it can be used in a with statement without side effects.
        """
        # Use the default timeout, if no timeout is provided.
        if timeout is None:
            timeout = self.timeout
    
        # Increment the number right at the beginning.
        # We can still undo it, if something fails.
        with self._thread_lock:
            self._lock_counter += 1
    
        lock_id = id(self)
        lock_filename = self._lock_file
        start_time = time.time()
        try:
            while True:
                with self._thread_lock:
                    if not self.is_locked:
                        logger().debug('Attempting to acquire lock %s on %s', lock_id, lock_filename)
                        self._acquire()
    
                if self.is_locked:
                    logger().info('Lock %s acquired on %s', lock_id, lock_filename)
                    break
                elif timeout >= 0 and time.time() - start_time > timeout:
                    logger().debug('Timeout on acquiring lock %s on %s', lock_id, lock_filename)
>                   raise Timeout(self._lock_file)
E                   filelock.Timeout: The file lock '/home/runner/.renku/renku.ini.lock' could not be acquired.
```

Which is us locking the global renku config (in `$HOME/.renku/renku.ini`). This could be solved by either mocking home directory to a temp directory for each test, which is probably safer for tests overall anyways, and/or by changing the way we lock things to have a read lock and a write lock, where the file being locked for read just prevents writes from happening, but not other reads, and being locked for write prevents reads from happening.

Panaetius:
Set `$HOME` to a temporary directory on the `repository` fixture. Adds exclusive write, shared read locking for the `~/.renku/renku.ini` file.

closes #1936

### DIFF
--- a/tests/cli/fixtures/cli_kg.py
+++ b/tests/cli/fixtures/cli_kg.py
@@ -54,6 +54,7 @@ def mock_kg():
             status_code = 404 if authorization == "Bearer renku-token" else 401
             return status_code, {"Content-Type": "application/json"}, ""
 
+        response.add_passthru("https://pypi.org/")
         response.add_callback(
             responses.GET, re.compile("http(s)*://renku.ch/knowledge-graph/datasets/.*"), callback=callback
         )
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -83,7 +83,7 @@
                 "sha256:5311c5c1769d16f46a4270e68a72147f6141fe8b3a1e10c718ac9559925b6624",
                 "sha256:981db6ecd8483c48725bf505c1164d5a531b698c78b14b9e9e8be693f6c9d63f"
             ],
-            "markers": "python_version < '4' and python_full_version >= '3.6.1'",
+            "markers": "python_version < '4.0' and python_full_version >= '3.6.1'",
             "version": "==0.3.7"
         },
         "certifi": {
@@ -215,7 +215,7 @@
                 "sha256:161b841809860cc5b9212c23a5f70db8cac932895208c4bcdf1e4c11be2517f1",
                 "sha256:db6ef5018c84ccff1856bca06b4a6065e02a50b4b21363527c26f9d584cd3889"
             ],
-            "markers": "python_version >= '3.6' and python_version < '4'",
+            "markers": "python_version >= '3.6' and python_version < '4.0'",
             "version": "==3.0.20210124104916"
         },
         "decorator": {
@@ -694,7 +694,7 @@
                 "sha256:13cdbb6b72798be9ef30caf460d3949bf057d9f073f73cc853908d50cc31229f",
                 "sha256:ca1008c18e91c2d0345764ddb871cc284feadc467241fbb468f14280b791388a"
             ],
-            "markers": "python_version >= '3.6' and python_version < '4'",
+            "markers": "python_version >= '3.6' and python_version < '4.0'",
             "version": "==9.3.0"
         },
         "ruamel.yaml": {
@@ -823,7 +823,7 @@
                 "sha256:8d7eaa5a82a1cac232164990f04874c594c9453ec55eef02eab885aa02fc17a2",
                 "sha256:f5321fbe4bf3fefa0efd0bfe7fb14e90909eb62a48ccda331726b4319897dd5e"
             ],
-            "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3, 3.4' and python_version < '4'",
+            "markers": "python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3, 3.4' and python_version < '4.0'",
             "version": "==1.25.11"
         },
         "wcmatch": {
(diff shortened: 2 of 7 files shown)


tests/cli/fixtures/cli_kg.py
```diff
@@ -54,6 +54,7 @@ def callback(request):
             status_code = 404 if authorization == "Bearer renku-token" else 401
             return status_code, {"Content-Type": "application/json"}, ""
 
+        response.add_passthru("https://pypi.org/")
```
mohammad-sdsc, line 57: Why is this needed?

mohammad-sdsc (approved):
Looks very good! Thanks! Just a minor comment, feel free to ignore.

tests/cli/fixtures/cli_kg.py
Panaetius: I got an error running tests, it hit the code that checks if an update is available for renku, but then responses complained that there was no hook defined for pypi.org url that gets used in the version check.

Issue #1870, opened 2021-02-10
service: write operations & porcelain


jsam:
So far all endpoints implemented on the service were designed to achieve most atomic renku operations. This approach is important to achieve some of the initial design goals, one of them being able to write clients in other languages which can be on par with the python client (e.g. R client) and offline mode. On the other hand core service also has stateless clients (i.e. UI), which currently struggle to maintain states and would like to push state management to the back end. With state management pushed to the back end, core service gains ability to execute multiple renku operation in sequence for those client (i.e batch execution). With the ability to batch multiple operations behind one interface, we could achieve simpler interactions with stateless clients. Hence the idea to add "porcelain" to the core service.

To understand better how service is designed, we should look at its renku operations through 3 main features - type of operation, possibility of remote execution, possibility of local execution. In short, types of operation are split into write or read operations (renku scoped operations). Local execution refers that we are executing renku operation on a project which is somewhere on the filesystem where the service is running. This could be in service defined cache or somewhere in user path. Remote execution means that user provides a remote address of a project and service needs to figure out the way to access that project, execute renku operations and sync it back to the remote.

Some of the work to support stateless clients has already been made, but mostly to support remote execution for renku **read** operations. The following table shows overview of current endpoints implementing renku operations:

Endpoint | Type of operations | Remote execution | Cache execution | Note
-- | -- | -- | -- | --
/datasets.create | Write | No | Yes |  
/datasets.add | Write | No | Yes |  
/datasets.remove | Write | No | Yes |  
/datasets.import | Write | No | Yes |  
/datasets.edit | Write | No | Yes |  
/datasets.unlink | Write | No | Yes |  
/cache.migrate | Write | No | Yes | Once remote execution is supported this endpoint should be renamed to /project.migrate
/datasets.files_list | Read | Yes | Yes |  
/datasets.list | Read | Yes | Yes |  
/cache.migrations_check | Read | Yes | Yes | Should probably be renamed to project.migrations_check
/graph.build | Read | Yes | Yes |  
/config.show | Read | Yes | Yes |
/config.set | Write | No | Yes |

As we can see, all **read** operations already fully support remote execution. Reason for this is because those read operations are relatively fast, cheap and require immediate response to the client. Write operations  on the other hand are more complex, slower and time of response doesn't matter, therefore should be implemented with asynchronous execution (life cycles for those executions is TBD). This gives us ability to implement them with batch execution in mind down the road.

[...]

PR #1957, opened 2021-03-10
feat: delayed write operations, ie. porcelain

jsam:
closes: #1870 

/deploy


renku/service/jobs/delayed_ctrl.py
```diff
@@ -0,0 +1,48 @@
...
+from renku.service.logger import worker_log
+from renku.service.views.decorators import requires_cache
+
+
+def to_ctrl(cache, context, user_data, renku_module, renku_ctrl):
+    """Based on renku operation return valid controller."""
+    ctrl_module = importlib.import_module(renku_module)
+    cls = getattr(ctrl_module, renku_ctrl)
+
+    _ = context.pop("is_delayed", None)  # NOTE: Prevent recursive invocation.
```
Panaetius, line 30:
```suggestion
    context.pop("is_delayed", None)  # NOTE: Prevent recursive invocation.
```

renku/service/controllers/api/mixins.py
```diff
@@ -128,7 +219,7 @@ def remote(self):
             return self.renku_op()
 
 
-class ReadWithSyncOperation(ReadOperationMixin, metaclass=ABCMeta):
+class RenkuOpSyncMixin(RenkuOperationMixin, metaclass=ABCMeta):
```
Panaetius, line 222: Would be nice if this had a more verbose description as well

Panaetius (approved):

rokroskar:
Since this is supposed to be a high-level API, is it documented somewhere that it's easy to find?

Panaetius:
> Since this is supposed to be a high-level API, is it documented somewhere that it's easy to find?

Only in the swagger docs. It's not really a different api, just an extension of existing endpoints to improve usability.

Panaetius:
I created https://github.com/SwissDataScienceCenter/renku-python/issues/2016 for docs

mohammad-sdsc (dismissed):
LGTM, thank you!

renku/service/controllers/api/mixins.py
```diff
@@ -203,27 +222,43 @@ def execute_op(self):
...
+        # NOTE: Only do a fetch every >30s to get eventual consistency but not slow things down too much
+        if project.fetch_age < PROJECT_FETCH_TIME:
+            return
+
+        lock = project.write_lock()
+
+        if self.skip_lock:
+            lock = contextlib.suppress()
+        try:
+            with lock:
```
mohammad-sdsc, line 235: Should we again check if `project.fetch_age < PROJECT_FETCH_TIME` again here and do nothing if it's the case? This makes sense if multiple operations are waiting for the lock to do the reset.

mohammad-sdsc (dismissed):
Thanks! This looks great! I've made a comment for a possible optimization; feel free to ignore it if it's not applicable.

mohammad-sdsc (approved):
🚀

PR #1958, opened 2021-03-10
Update docs on how to add local files to renkulab and create a datase…

olevski:
So this question came up a lot during my QA round. And even I did not know that you can use the `renku dataset` command in this way until someone told me. 

I hope the addition makes sense and it is done in the right file.

### DIFF
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -189,6 +189,24 @@ To add a specific version of files, use ``--ref`` option for selecting a
 branch, commit, or tag. The value passed to this option must be a valid
 reference in the remote Git repository.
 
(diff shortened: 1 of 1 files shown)


Panaetius:
Thank you for adding this! I am wondering if https://renku.readthedocs.io/en/0.7.4/user/cli-installation.html#using-the-cli-on-renkulab might be a better place for it, as the cli itself doesn't really know anything about jupyterlab?

olevski:
@Panaetius I dont mind moving this section elsewhere. I will do it - I agree that Jupyterlab shows up out of nowhere in the docs.

olevski:
Ok I see this should go to the renku repo, not renku-python.

PR #1965, opened 2021-03-16
chore: update scheduler intervals and logging redirect

jsam:
/deploy

### DIFF
--- a/tests/service/scheduler/test_scheduler.py
+++ b/tests/service/scheduler/test_scheduler.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2021 Swiss Data Science Center (SDSC)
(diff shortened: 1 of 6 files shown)


tests/service/scheduler/test_scheduler.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def test_enqueue_jobs(with_scheduler):
+    """Enqueue jobs at a interval."""
+    queues = Queue.all()
+    assert queues
+
+    assert 2 == len(queues)
+    for q in queues:
+        assert 1 == q.count
```
Panaetius, line 29: Could we add a project and file with expired ttl, and then check that they get cleaned up after this?

jsam: This is testing only the queuing up the jobs, not the job execution itself. The job execution itself is tested in `test_cache.py`

Panaetius (approved):

PR #1972, opened 2021-03-18
fix(service): sync cache with remote before operations to prevent cache getting out of sync

Panaetius:
Tries to do a `git fetch && git reset --hard origin` before executing an operation on the cache to ensure the cache is in sync.

Meant as a band-aid until porcelain commands are done and the UI switched over to them.

### DIFF
--- a/renku/service/controllers/api/mixins.py
+++ b/renku/service/controllers/api/mixins.py
@@ -20,7 +20,7 @@ from abc import ABCMeta, abstractmethod
 from functools import wraps
 from pathlib import Path
 
-import git
+from git import Repo
 
 from renku.core.errors import RenkuException, UninitializedProject
 from renku.core.management.config import RENKU_HOME
@@ -83,7 +83,7 @@ class ReadOperationMixin(metaclass=ABCMeta):
 
     def reset_local_repo(self, project):
         """Reset the local repo to be up to date with the remote."""
-        repo = git.Repo(self.project_path)
+        repo = Repo(self.project_path)
         origin = None
         if repo.active_branch.tracking_branch():
             origin = repo.remotes[repo.active_branch.tracking_branch().remote_name]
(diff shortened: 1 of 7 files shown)


renku/service/controllers/api/mixins.py
```diff
@@ -123,7 +138,7 @@ def sync(self, remote="origin"):
         if self.project_path is None:
             raise RenkuException("unable to sync with remote since no operation has been executed")
 
-        _, remote_branch = repo_sync(Repo(self.project_path), remote=remote)
+        _, remote_branch = repo_sync(git.Repo(self.project_path), remote=remote)
```
jsam, line 141: Could we keep the old notation? Why are we changing it here?

jsam (commented):
Look's good, thanks.

Maybe we should stash the local changes and push it to the remote branch (which could be then deleted by the user if necessary)?

tests/service/fixtures/service_integration.py
```diff
@@ -18,13 +18,35 @@
...
+import shutil
 import uuid
 from copy import deepcopy
 
 import pytest
 from git import GitCommandError
 
 
+@contextlib.contextmanager
+def _mock_cache_sync(repo):
```
jsam, line 31: Should we add a doc string?

renku/service/controllers/api/mixins.py
Panaetius: Oh good point. I think this was leftover code from debugging something

Panaetius:
I'm not sure I understand where this implementation would have an issue where it makes sense to stash changes before the git reset so nothing is lost

- If the remote has changes before a service call, we will always work on top of those changes, so everything works
- if the remote gets changed during a service call, repo_sync already takes care of pushing the changes to a temporary branch
- if the remote gets changed during a service call, and repo_sync fails for some reason (should only happen for permission or connection issues), you'd lose the changes made, but to me that's the correct behavior (along with raising an error)
- if a user tries to change the remote after a service call, they'll get an exception in their console on push, which I think is also correct behavior and they wouldn't lose any work


The only issue I could imagine is with the first point, if a user goes to e.g. the dataset edit page, the UI shows the fields, while they're filling out the form, the dataset gets changed via cli, and now when the user clicks "submit", the datasets.edit will be done on top of the newer version, not on top of what the user saw in the UI. And it might work, in which case nothing is lost (I can't think of a case with the current service where this causes a serious issue, everything is still in git). Or it might fail (like if the dataset was deleted while the user was filling out the form). But then the service will raise an error anyways and there's no new commit to keep

jsam (approved):
:rocket:

PR #1974, opened 2021-03-19
fix(core): fix template update if same filename was added locally and…

### LINKED ISSUE #1729: return template url in service on template migration
As requested by the UI team https://github.com/SwissDataScienceCenter/renku-ui/pull/1068#discussion_r524379534 we should return the template repository url, template reference and template id in the cache.migrations_check and /cache.migrate endpoints.

### LINKED ISSUE #1953: Update template reference even if there is no change
In some cases, a user might have changed a file locally and the template might have changed remotely with the same change (e.g. renku version in the Dockerfile).

In this case, the migration endpoint will say that the template can be updated (since there's is a different version on the remote than what was used locally), but running the migration won't update the project, as there is no file to change.

In these cases, we should still update the project metadata to the new template reference, so that the project gets shown as up-to-date.

Panaetius:
I couldn't reproduce this issue (even when there was no actual change in the template, the reference got updated and committed correctly). But I did find an issue if a file got added locally and also in the template, where template_update would fail as there was no checksum saved for the file.

I'm guessing the original issue was due to a stale cache?

closes #1953 (I guess)

[Edit]
After further consideration, the problem was likely ref checkouts not working in user_project_clone, which was fixed already. But there is also an issue with us not returning `automated_template_update`.

While we're at it, also closes #1729


jsam (approved):
:rocket:

jsam (approved):
:rocket:

PR #1975, opened 2021-03-19
chore: add better user messaging on service api start error

jsam:
Adds a ping check before executing service commands through the CLI. In case redis is not found running, we show nice message to the user instead of panicking.

### DIFF
--- a/renku/cli/service.py
+++ b/renku/cli/service.py
@@ -182,6 +182,10 @@ def service(ctx, env):
         import rq  # noqa: F401
         from dotenv import load_dotenv
 
+        from renku.service.cache.base import BaseCache
+
+        BaseCache.cache.ping()
+
         load_dotenv(dotenv_path=env)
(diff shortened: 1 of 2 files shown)


renku/cli/service.py
```diff
@@ -193,6 +197,13 @@ def service(ctx, env):
 
         ctx.exit(1)
 
+    except redis.exceptions.ConnectionError:
+        # NOTE: Cannot connect to the service dependencies, ie. Redis.
+
+        click.echo(ERROR + "Cannot connect to Redis")
```
Panaetius, line 203: What do the `ConnectionError`s look like? If they contain any useful information that could help a user to solve an issue, we might want to include the original error in this message.

jsam: It just says that redis could not be reached, nothing too useful.

Panaetius (approved):

PR #1979, opened 2021-03-22
fix(core): Add error handling if push of temporary branch fails

Panaetius:
Adds proper error handling if pushing the temporary branch after a failed push to a protected branch fails.

/deploy


jsam (approved):
:rocket:

PR #1992, opened 2021-03-25
fix(core): Fix annotations serialization in ProvenanceGraph

### LINKED ISSUE #1952: ProvenanceGraph update is not including Annotations
**Describe the bug**

When a project is using renku-mls (that adds Annotations to the run) and ProvenanceGraph (i.e. `renku graph generate` has been run within the project) the information is being lost on update of the graph, e.g. running `renku run`.

if one regenerates the graph (`renku graph generate --force`) the annotation information is correctly added.

**Expected behavior**
`renku mls leaderboard` returns all the ML models ran in the project. OR `.renku/provenance.json` contains `http://www.w3.org/ns/oa#Annotation` for the runs.

 - renku version: 0.14.0
 - Python version: 3.8.0

vigsterkr:
# Description

Fixes #1952

depends on https://github.com/SwissDataScienceCenter/calamus/pull/58

### DIFF
--- a/renku/core/models/provenance/activities.py
+++ b/renku/core/models/provenance/activities.py
@@ -551,7 +551,6 @@ class ProcessRun(Activity):
 
         process_run.generated = generated
 
-        process_run.plugin_annotations()
         return process_run
 
     @property
--- a/renku/core/models/cwl/annotation.py
+++ b/renku/core/models/cwl/annotation.py
@@ -59,5 +59,5 @@ class AnnotationSchema(JsonLDSchema):
         unknown = EXCLUDE
 
     _id = fields.Id(init_name="id")
-    body = fields.Raw(oa.hasBody)
+    body = fields.RawJsonLD(oa.hasBody)
     source = fields.Raw(dcterms.creator)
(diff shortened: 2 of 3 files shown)


vigsterkr:
unfortunately this is not yet fixing the whole problem. for some reason the `ProcessRun` does not contain the annotations from the hook implementation when running a `renku run ...` command. but using `renku graph generate --force` one can get the annotations in the `ProcessRun` and that way the annotations are going to be included in the graph properly

renku/core/models/provenance/activities.py
```diff
@@ -550,7 +550,6 @@ def from_run(cls, run, client, path, commit=None, subprocess_index=None, update_
 
         process_run.generated = generated
 
-        process_run.plugin_annotations()
```
vigsterkr, line 553: @Panaetius tagging you because based on git blame you've added this to the src. i think this is an unnecessary extra call of the plugin as it has been handled already in the post init function. or maybe i'm misinterpreted things here...?

mohammad-sdsc: This seems unnecessary to me as well since we don't store the results anywhere.

mohammad-sdsc (approved):
LGTM, thanks!

Issue #1737, opened 2020-11-17
Add a file resource that stores file metadata


Panaetius:
Once https://github.com/SwissDataScienceCenter/renku-python/issues/1711 and https://github.com/swissdatasciencecenter/renku-python/issues/1709 are done, we can add a dedicated file resource to track files in a project and reduce duplication between dataset and workflow metadata.

See https://github.com/SwissDataScienceCenter/renku/issues/696#issuecomment-586291175 for details.

Issue #671, opened 2019-09-11
Enable importing of files outside of datasets


rokroskar:
Currently we can only import files directly into datasets. This can be restrictive if a user is just starting a project and all they want to do is import some files. We should enable this with a command like

```
renku file add <URL>
```

The same semantics can be followed as for `renku dataset add` - one should be able to import anything from the local filesystem, a remote URL or a remote or local git repository. The same `--src` and `--dst` flags as proposed in https://github.com/SwissDataScienceCenter/renku-python/issues/670 should be used. 

One outstanding question is what to do about the metadata - we should store the information about the source similar to how it is stored for `DatasetFile`.

rokroskar:
That could potentially work - however the annoying thing I can foresee is that atm everything with `renku dataset add` gets added to the `data/<dataset>` directory and the `--destination` option allows you to specify the subpath relative to the dataset path, *not* relative to the project root. So if you wanted to add something to a custom path it wouldn't really be possible.

mohammad-sdsc:
What you've said @rokroskar is true; `--destination` is specified relative to the dataset directory. I also don't like this `renku file add` syntax/command but I believe asking users to use `renku dataset add` with no dataset name is a bit confusing and error-prone. Maybe we can use the same implementation behind the scene.

rokroskar:
It's not obvious anymore to me that this is needed - I will put this issue in the backlog and we can come back to it if someone requests this functionality. It seems that we can achieve most of what this would do with some mechanism that prevents users from committing large files to the git repo directly, e.g. via a git pre-commit hook.

rokroskar:
I will remove this from the datasets work/epic for now - I think it should be discussed in the context of how we want to handle and abstract away external storage choices. We should make a new epic for that asap in the next design meeting. cc @emmjab

schymans:
I was pointed to this issue from here: https://renku.discourse.group/t/add-code-from-different-project/89
So if it matters, I need this functionality, not for data files but for code files. One of the problems with `renku dataset add` is, as pointed out by @cramkri, that the file is put into git-lfs, which is not a good idea for code files. I want to be able to transparently import someone else's code file and then modify it for my needs.

Panaetius:
@schymans We now support an lfs file size limit as well as an lfs ignore file that can prevent this from happening, see https://renku.readthedocs.io/en/latest/user/lfs.html?highlight=renkulfsignore#renku-lfs-configuration for more details

[...]

PR #2001, opened 2021-04-06
chore: RFC-001 workflow ux

Panaetius:
RFC for new workflow commands to improve user experience.


rokroskar:
Is the plan to merge this and then carry on discussion [here](https://github.com/SwissDataScienceCenter/renku-python/discussions/categories/request-for-comment)?

Panaetius:
No the plan is to discuss it and to change it here, until we decide to accept the RFC, then merge it (otherwise we need to make a PR each time we want to edit it). We can discuss it there, though, while the PR is open

jachro:
After some checks on both sides it looks we the prov#hadRole we currently have on both Usage and Generation is not really used anymore. Let's drop it then :)

Panaetius:
> should Dataset Part's @id be extending Dataset's @id; currently, it doesn't extend;
>should PublicationEvent has @id like this: https://localhost/datasettags/1.0%40e732668c4eeec0ddd8871290f9484e68605c0048 or https://renku-host/datasettags/1.0%40e732668c4eeec0ddd8871290f9484e68605c0048

I think we could discuss this in a separate issue, as this RFC is on regards to workflows.

> should DatasetProvenance be a blank node? It is at the moment 
We won't use the single-file DatasetProvenance in the future as it didn't work out, and stick with the current format (or something very close to it).

> should Entity has the isPartOf property pointing to the project? It's a no-brainer but we're missing the property now
We're slowly moving away from having every node point to the project it is in, as this gets complicated, especially with flattened jsonld. E.g. if a projects `agent` changes, and you reuse an entity, now you have entities pointing to the same project node but with differing properties, which need to be then merged in some way. Ideally we'd only like to have top-level nodes point to the project (Dataset, Plan, Activity) to simplify things. I've updated the docs accordingly but it's certainly up for discussion.

jachro:
@Panaetius I updated my comment according to the latest discussion.

schymans:
I am looking at this file: https://github.com/SwissDataScienceCenter/renku-python/blob/rfc-001-workflow-ux/design/001-new-workflow-commands/001-new-workflow-commands.md

I see this for the first time, so sorry if my input below is useless due to lack of understanding. Just ignore it if this is the case.

The description starts with a specific workflow ID, but I wonder how to get the workflow IDs relevant for a specific file, e.g. where file `data.csv` was either created or used as input, and then how to extract the respective `renku run` command used to generate that workflow. As stated, the `renku run` commands can be quite complex, so it would be nice to copy a previous one and adapt for a new use case. 

Another question, how does `renku dataset add` integrate into the workflow ecosystem? One of the examples given above includes:
`$ renku run --name download-data --description "Downloads the daily data csv" --output datafile="data.{%Y-%m-%d}.csv" -- bash download.sh`
But I guess that the above would not add `data.{%Y-%m-%d}.csv` as a dataset unless `download.sh` has a command `renku dataset add` in it. Generally, I would like to preserve the link to the origin of the data, but I don't see how this is done in this example.

Panaetius:
@schymans In regards to how datasets figure into this, we decided to not design this in this RFC, but rather as a followup, as it's already complex enough. So we also don't know yet how datasets integrate into the workflow ecosystem (though we have some ideas). We discussed in the past that a run can automatically put files into a dataset or that a dataset can be specified as input and then path of the command are relative to the dataset, but there's many details and edge cases that we haven't settled on yet (Can a command produce more than 1 dataset, what happens if the command is used as part of a larger command, can a dataset be used that isn't in the current project, to just name a few open questions).

A python script could specify the URL as a parameter using https://renku.discourse.group/t/renku-python-api/340 so it's not hidden inside the script, so that would be a workaround in this case. Though it wouldn't track any information about whats in the URL like a checksum, just the URL itself. Same if you did `renku run -- curl http://...`.
Adding the file to a dataset by URL should preserve it's provenance with the current implementation. There has been discussion of adding a file to a repo by URL with tracking provenance but outside of a dataset, https://github.com/SwissDataScienceCenter/renku-python/issues/1737 and https://github.com/SwissDataScienceCenter/renku-python/issues/671 (I guess you know about this one :slightly_smiling_face: ), but this will have to wait until we've done the switch to new metadata (we need #1737 for that).

As for the other question, the intention is that `renku workflow history` allows showing executions of template by some filters. Not this could be a workflow ID or a file, or something else. But as I mentioned in the meeting this morning, that is the command that is least defined so far, just that it should behave similar to a `git log` in look and feel. The command itself is definitely useful, but not integral to how the new workflow system works, so not much time was spent discussing it and it's left to whoever implements it to come up with a good design. Here we'd be really grateful if the community could provide use case/examples of what they'd like it to do, so we can take those into account.

vigsterkr (approved):

schymans:
Thanks for the explanation, @Panaetius. Where would be the best place to continue this discussion? From a user's perspective, I would love to be able to click on an input file in the KG and see the command how it was imported into the repo, and if it was important from an external source, to follow the link and see if there are any updates on the external site that might make it worthwhile to update the version of the file I am using.

PR #2008, opened 2021-04-07
fix(service): fix project_clone with git ref specified

Panaetius:
Fixes project clone to correctly checkout the remote ref.

/deploy

### DIFF
--- a/renku/core/management/clone.py
+++ b/renku/core/management/clone.py
@@ -53,29 +53,51 @@ def clone(
         os.environ["GIT_LFS_SKIP_SMUDGE"] = "1"
 
     try:
-        repo = Repo.clone_from(url, path, recursive=recursive, depth=depth, progress=progress)
+        kwargs = {}
+
+        if checkout_rev:
+            kwargs["branch"] = checkout_rev
+
+        repo = Repo.clone_from(url, path, recursive=recursive, depth=depth, progress=progress, **kwargs)
     except GitCommandError as e:
-        if not raise_git_except:
-            lines = progress.other_lines + progress.error_lines if progress else []
-            error = "".join([f"\n\t{line}" for line in lines if line.strip()])
-            message = f"Cannot clone remote Renku project: Git exited with code {e.status} and error message:\n {error}"
-            raise errors.GitError(message)
+        # NOTE: clone without rev in case it was not a branch
+        if not checkout_rev:
+            if not raise_git_except:
+                lines = progress.other_lines + progress.error_lines if progress else []
+                error = "".join([f"\n\t{line}" for line in lines if line.strip()])
+                message = (
+                    f"Cannot clone remote Renku project: Git exited with code {e.status} and error message:\n {error}"
+                )
+                raise errors.GitError(message)
 
-        raise e
+            raise e
+        try:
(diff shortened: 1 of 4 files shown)


renku/core/management/clone.py
```diff
@@ -53,29 +53,51 @@ def clone(
...
-            raise errors.GitError(message)
+        kwargs = {}
+
+        if checkout_rev:
+            kwargs["branch"] = checkout_rev
 
-        raise e
+        repo = Repo.clone_from(url, path, recursive=recursive, depth=depth, progress=progress, **kwargs)
+    except GitCommandError as e:
+        # NOTE: clone without rev in case it was not a branch
```
rokroskar, line 63: this comment should be lower on L75

renku/core/management/clone.py
```diff
@@ -53,29 +53,51 @@ def clone(
...
+                    f"Cannot clone remote Renku project: Git exited with code {e.status} and error message:\n {error}"
+                )
+                raise errors.GitError(message)
 
-    remote_refs = [Path(ref.abspath).name for ref in repo.remote().refs]
+            raise e
 
-    try:
-        if checkout_rev in remote_refs:
+        try:
```
rokroskar, line 87: I find these repeated try/excepts a bit confusing - could you maybe add a few more inline comments to document the flow a bit better? Maybe something at the top of this whole block like:

1. try a clone assuming the `checkout_rev` is a branch
2. if this fails, try a clone without the branch
3. if this succeeds, try to checkout the `checkout_rev`

tests/cli/test_integration_cli.py
```diff
@@ -93,7 +93,7 @@ def test_renku_clone_checkout_rev(tmp_path, url):
             )
         ).output
 
-        assert "97f907e1a3f992d4acdc97a35df73b8affc917a6" == str(repo.active_branch)
+        assert "97f907e1a3f992d4acdc97a35df73b8affc917a6" == str(repo.head.commit)
```
rokroskar, line 96: 👍

rokroskar (changes requested):
This works as advertised - however, given how problematic this functionality has been it would be great to make the logic painfully obvious to anyone who might need to investigate this code in the future :)

renku/core/management/clone.py
```diff
@@ -53,29 +53,51 @@ def clone(
...
-            error = "".join([f"\n\t{line}" for line in lines if line.strip()])
-            message = f"Cannot clone remote Renku project: Git exited with code {e.status} and error message:\n {error}"
-            raise errors.GitError(message)
+        kwargs = {}
+
+        if checkout_rev:
+            kwargs["branch"] = checkout_rev
 
-        raise e
+        repo = Repo.clone_from(url, path, recursive=recursive, depth=depth, progress=progress, **kwargs)
```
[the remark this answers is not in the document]
Panaetius: Because this is one of those kwargs that's just passed along to the actual `git clone` command. I believed if it was `branch=None` this would result in `git clone --branch git@github.com/org/repo` (so no branch specified) which would error, in which case this is cleaner than an If statement with the command repeated with/without branch.

But actually I just tested that hypothesis and it doesn't work that way, `branch=None` seems to be fine. So I'll change it to your suggestion :smile:

rokroskar (approved):
Thanks @Panaetius !!

PR #2019, opened 2021-04-12
feat(cli): support template and docker migration

### LINKED ISSUE #1860: Enable Dockerfile and template update in the cli
Now that the service has support to update the renku version in the dockerfile and to update the template, we should add support for it in the CLI as well.

We should add flags for it to `renku migrate`, but the default should probably still be to only migrate metadata (Or we should prompt the user). Additionally, we should also add `check_migration` support on the CLI.

Panaetius:
closes #1860

### DIFF
--- a/renku/cli/migrate.py
+++ b/renku/cli/migrate.py
@@ -79,8 +79,8 @@ def migrate(check, no_commit, skip_template_update, skip_docker_update):
     if check:
         return
 
-    skip_docker_update = skip_docker_update or not template_update_possible
-    skip_template_update = skip_template_update or not docker_update_possible
+    skip_docker_update = skip_docker_update or not docker_update_possible
+    skip_template_update = skip_template_update or not template_update_possible
 
     communicator = ClickCallback()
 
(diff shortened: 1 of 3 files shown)


rokroskar (changes requested):
Thanks @Panaetius this is going to be a useful feature! I can't quite get it to work yet... so here is what I tried:

* in my local clone of the template repo, make a branch that is a few commits behind master
* initialize my project with `renku init -s file://<path-to-template-repo> -r test-branch test-template-update`
* move the tip of `test-branch` in the local template repo to `master` i.e. a few commits ahead - presumably `renku migrate` should now offer to update the template
* run `renku migrate -c` in the project - I get `Info: The Dockerfile refers to an older version of renku and can be updated using 'renku migrate'.` 
* running `renku migrate` doesn't seem to do anything

Not sure what to make of it?

rokroskar (approved):
this works now :) 

There is the wording suggestion still left, but if you disagree feel free to disregard.

Panaetius:
Also added some docs

rokroskar (approved):
🥇

PR #2020, opened 2021-04-13
feat(cli): detect filename from content-disposition header

### LINKED ISSUE #1724: Use the content-disposition header to set the filename on dataset add
File download links provided by some websites may include in the response a [`content-disposition` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) which instructs the client what to do with the content. In the case of file downloads it communicates e.g. the filename that the content should be stored under. 

We should use this header to automatically generate correct filenames with `renku dataset add`. Here is an example:

```
curl -I https://drive.switch.ch/index.php/s/euqXnfdvH3e2H30/download
HTTP/1.1 200 OK
Date: Thu, 12 Nov 2020 23:57:11 GMT
Server: Apache/2.4.18 (Ubuntu)
Set-Cookie: oc641cdd42e0=700fbdc845f551d9f696a9b3fa1a0962; path=/; secure; HttpOnly
Set-Cookie: oc_sessionPassphrase=NCcmH%2Fz6554LDYjZgzD22IwBJlbM8HKCrDvU7Mp%2BwKOCytZV9nJ3jSA2vSyaaszdmmkae%2BbvHU%2FdIfSXKNSQ%2FJ6UIz1H9AzKsuQTo9%2BwxHqXAqdaVIxr433wu4fQG04m; path=/; secure; HttpOnly; SameSite=strict
Content-Security-Policy: default-src 'self'; script-src 'self' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; frame-src *; img-src * data: blob:; font-src 'self' data:; media-src *; connect-src *
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-Robots-Tag: none
X-Download-Options: noopen
X-Permitted-Cross-Domain-Policies: none
Content-Disposition: attachment; filename*=UTF-8''SWITCHdrive%20Help.pdf; filename="SWITCHdrive%20Help.pdf"
Content-Transfer-Encoding: binary
Pragma: public
Expires: 0
Cache-Control: no-cache, no-store, must-revalidate, post-check=0, pre-check=0
Strict-Transport-Security: max-age=0; includeSubDomains; preload
Content-Length: 234464
Content-Type: application/pdf
```

Panaetius:
closes #1724


renku/core/management/datasets.py
```diff
@@ -1400,6 +1403,26 @@ def extract_dataset(filepath):
...
 
 
+def _filename_from_headers(request):
+    """Extract filename fromcontent-disposition headers if available."""
+    content_disposition = request.headers.get("content-disposition", None)
+
+    if not content_disposition:
+        return None
+    entries = content_disposition.split(";")
+    name_entries = [e.strip() for e in entries if e.strip().lower().startswith("filename*=")]
```
mohammad-sdsc, line 1413: `name_entries = next((e.strip() for e in entries if e.strip().lower().startswith("filename*=")), None)` is a bit cleaner.

renku/core/management/datasets.py
```diff
@@ -1400,6 +1403,26 @@ def extract_dataset(filepath):
...
+    if name_entries:
+        name = name_entries[0].split("=", 1)[1].strip()
+        encoding, _, name = name.split("'")
+        return unquote(name, encoding, errors="strict")
+
+    name_entries = [e.strip() for e in entries if e.strip().lower().startswith("filename=")]
```
mohammad-sdsc, line 1419: same as above: `name_entries = next((e.strip() for e in entries if e.strip().lower().startswith("filename=")), None)`

renku/core/management/datasets.py
```diff
@@ -1400,6 +1403,26 @@ def extract_dataset(filepath):
...
+    filename = name_entries[0].split("=", 1)[1].strip()
```
mohammad-sdsc, line 1420: It seemed to me that `filename=` is optional. Maybe check `name_entries` before using it.

mohammad-sdsc (commented):
Thanks! Looks pretty good. I've made some minor comments.

mohammad-sdsc (approved):
Thank you!

PR #2030, opened 2021-04-16
chore: release v0.14.2

Panaetius:


rokroskar (commented):
Maybe the template migration should be highlighted at the top as well?

rokroskar (approved):
Looks good, thanks! I think the highlights section makes sense!

PR #2033, opened 2021-04-19
feat(workflow): workflow naming metadata

### LINKED ISSUE #1986: add --name and --description to renku run
As a first step to improved workflow UX, we should add `--name` and `--description` parameter to the `renku run` command, that get set on the Run template. This is necessary for most of the new functionality to work and can be done without functional changes to the current metadata. We should use `schema:name` and `schema:description`, respectively. This probably also means that `renku:Run` nodes also need the `schema:Action` type for this to be valid.

I believe `--name` is already available on `renku graph run`, but we need to verify that this fulfills the criteria.

As always, check with the KG team before merging any schema changes.

mohammad-sdsc:
# Description
Allows adding name, description, and keywords to renku runs.

Fixes #1986


renku/cli/run.py
```diff
@@ -208,6 +213,8 @@
 
 @click.command(context_settings=dict(ignore_unknown_options=True,))
 @click.option("--name", help="A name for the workflow step.")
+@click.option("--description", help="Workflow step's description.")
+@click.option("--keyword", multiple=True, help="List of tags for the workflow.")
```
Panaetius, line 217: We should not say "tags" in the description, "keywords" and "tags" is already confusing for users on datasets, and the distinctions isn't always clear. so we be as consistent as possible on our end.

renku/core/models/workflow/plan.py
```diff
@@ -44,13 +44,26 @@
...
+    ):
         """Initialize."""
         self.arguments = arguments or []
         self.command = command
+        self.description = description
         self.id_ = id_
         self.inputs = inputs or []
-        self.name = name or "{}-{}".format(secure_filename(self.command), uuid.uuid4().hex)
+        self.keywords = keywords or []
+        self.name = name or f"{secure_filename(self.command)}-{uuid.uuid4().hex}"
```
Panaetius, line 66: This will result in many "python-abcdefg" runs. I wonder if we should use self.to_argv() and limit it to 15-20 characters, so that it's a bit more descriptive, at least potentially.

renku/core/utils/git.py
```diff
@@ -68,3 +70,16 @@ def split_paths(*paths):
 
     for index in range(batch_count):
         yield paths[index * ARGUMENT_BATCH_SIZE : (index + 1) * ARGUMENT_BATCH_SIZE]
+
+
+def get_slug(name):
```
Panaetius, line 75: `renku.core.utils.git` might be a better place for this, as a "slug" is an URL concept.

Panaetius (changes requested):
A few minor comments but overall it looks great, thank you!

jachro:
After some checks on both sides it looks we the `prov#hadRole` we currently have on both `Usage` and `Generation` is not really used anymore. Let's drop it then :)

renku/core/models/workflow/plan.py
mohammad-sdsc: This is a bit challenging to be done the right because we want the name to be unique as well while we don't load the whole DependencyGraph. For now, I've modified the `Plan.from_run` to pass a name to this constructor according to your suggestion. This constructor is not called directly.

Panaetius (approved):

PR #2042, opened 2021-04-24
chore: remove pipfile action and pipfile.lock

rokroskar:
We are not using `pipenv` anywhere in testing and all of the relevant dependencies are specified in `setup.py` so there is no reason to keep neither the action nor `Pipfile.lock` around... this should make it much less annoying to push to PRs :)


Panaetius (approved):
Great change!

PR #2044, opened 2021-04-27
chore: fix pypi release process and adjust versioning

### LINKED ISSUE #2043: Create a pypi pre-release on every commit
Currently we only create pre-releases for commits to master and tags, but we should do pre-releases on every commit do the `/deploy` accptance tests can successfully install renku-python in the Docker image.

We should also make the release not dependent on the tests passing, to ensure that there's always a matching release available.

We need to handle pre-release versioning in a way that we don't end up with duplicate versions

Panaetius:
Change version numbers to x.y.z.devN+sha256

Change pypi release to run on every pushed commit

closes #2043


mohammad-sdsc (approved):
LGTM. Thank you!

PR #2049, opened 2021-04-29
feat(dataset): update local files

### LINKED ISSUE #2039: Update local dataset files
As a renku user, I want `renku dataset update` command to update dataset's metadata if there is a change in local dataset files so that other projects that imported this dataset can see the update.

# Acceptance Criteria
- DatasetFile metadata is updated if there is a change in the corresponding Entity
- Dataset is mutated
- If a file is deleted, its metadata is deleted if `--delete` argument is passed

mohammad-sdsc:
# Description

`renku dataset update` also looks for changes in local files and update dataset metadata accordingly.

Fixes #2039


renku/core/commands/dataset.py
```diff
@@ -670,17 +673,26 @@ def _update_datasets(client, names, creators, include, exclude, ref, delete, ext
...
+    if local_files:
+        updated, deleted = client.update_dataset_local_files(records=local_files, delete=delete)
+        updated_files.extend(updated)
+        deleted_files.extend(deleted)
 
     if deleted_files and not delete:
         communication.echo(
-            "Some files are deleted from remote. To also delete them locally "
-            "run update command with '--delete' flag."
+            "Some files are deleted in the repo. To also delete them locally run update command with '--delete' flag."
```
Panaetius, line 689:
```suggestion
            "Some files are deleted in the repo. To also delete them from dataset metadata run update command with '--delete' flag."
```

Might be more clear. otherwise users might be confused if they work with a local dataset.

renku/cli/dataset.py
```diff
@@ -217,6 +217,9 @@
 with those datasets. Modifying those datasets locally will prevent them from
 being updated.
 
+The update command also checks for files changes in the project and updates
+dataset's metadata accordingly.
```
Panaetius, line 221:
```suggestion
The update command also checks for file changes in the project and updates
datasets' metadata accordingly.
```

I think it's either `datasets' metadata` or `each dataset's metadata` if I'm not mistaken about english posessives (which I might well be, it's complicated). At least I think if you have a singular form it has to be determined by something like "the" or "each" or "this". Or maybe just `dataset metadata`? Not that it really matters, just "updates
dataset's metadata" sounds weird somehow.

Panaetius (approved):
This looks great! @ksanao will be very happy :smile: 

Two spelling suggestions, but otherwise good to go.

Panaetius (approved):

PR #2050, opened 2021-04-29
fix(core): add url validator utility function

### LINKED ISSUE #1988: poor template URL handling
**Describe the bug**
If a template repo URL contains a trailing slash, the repo fetch fails - see [here](https://renku.discourse.group/t/community-contributed-project-templates/284/2)

**To Reproduce**

Go to https://renkulab.io/projects/new and put in `https://github.com/SwissDataScienceCenter/contributed-project-templates/` as the custom template URL. If you remove the trailing slash, all is well in the world. 

**Expected behavior**
The templates are fetched

**Screenshots and/or execution output**

image

**Additional context**

This seems to only be a problem with the service, i.e. 

```
renku init -s https://github.com/SwissDataScienceCenter/contributed-project-templates.git -r main test-template
```

seems to work fine.

vigsterkr:
# Description
This PR introduces a small utility function that validates a user provided URL and replace the use of GitURL by this function where it's unnecessary to actually parse the user provided URL itself more than actually just making sure that it's a valid URL.
Further more it includes a mini-fix for the GitURL regexps to allow trailing slashes.

Fixes #1988 #2051

/deploy


renku/core/management/datasets.py
```diff
@@ -1296,14 +1295,14 @@ def checkout(repo, ref):
 
         depth = 1 if not ref else None
         ref = ref or renku_branch
-        u = GitURL.parse(url)
-        path = u.pathname
-        if u.hostname == "localhost":
+        u = validate_url(url)
+        path = u.path
+        if "localhost" in u.netloc:
```
Panaetius, line 1300: Shouldn't it be `u.netloc = "localhost"` ? In case of myawesomelocalhost.com (Domain still available if someone wants it!)

renku/service/serializers/templates.py
```diff
@@ -69,12 +69,12 @@ def create_new_project_url(self, data, **kwargs):
         """Set owner and name fields."""
         try:
             project_name_stripped = normalize_to_ascii(data["project_name"])
+            if len(project_name_stripped) == 0:
+                raise ValidationError("Project name contontains only unsupported characters")
```
Panaetius, line 73:
```suggestion
                raise ValidationError("Project name contains only unsupported characters")
```

tests/service/controllers/test_templates_create_project.py
```diff
@@ -149,10 +149,10 @@ def test_except_project_name_handler(project_name, ctrl_init, svc_client_templat
...
     cache, user_data = ctrl_init
-    svc_client, headers, payload, rm_remote = svc_client_templates_creation
+    _svc_client, _headers, payload, _rm_remote = svc_client_templates_creation
     payload["project_name"] = project_name
 
     with pytest.raises(ValidationError) as exc_info:
         TemplatesCreateProjectCtrl(cache, user_data, payload)
 
-    assert "Invalid `git_url`" in str(exc_info.value)
+    assert "Project name contontains only unsupported characters" in str(exc_info.value)
```
Panaetius, line 158:
```suggestion
    assert "Project name contains only unsupported characters" in str(exc_info.value)
```

Panaetius (changes requested):
Looks good, only two minor comments.

renku/core/management/datasets.py
vigsterkr: yeah that was my first thought, but the idea behind this is that to support `localhost:<PORT>`. but indeed it fails big time, so i'll replace it with a regex. i guess `127.*` should be here as for check, or?

Panaetius: Honestly I'm not sure what case this part covers. it's not the `file://path/to/.git` case. But I don't see how for `git://localhost/myfolder` we can just assume that `str(Path(path).resolve())` actually resolves to anything on disk?

vigsterkr: true that!

renku/core/utils/urls.py
```diff
@@ -110,3 +111,10 @@ def get_slug(name):
     valid_end = re.sub(r"[._-]$", "", valid_start)
     no_dot_lock_at_end = re.sub(r"\.lock$", "_lock", valid_end)
     return no_dot_lock_at_end
+
+
+def validate_url(repo: str) -> ParseResult:
+    """Validates the supplied url and returns the parsed URL if valid."""
+    if re.search(_URL_VALIDATOR, repo, re.ASCII):
+        raise errors.ParameterError(f"Invalid url: `{repo}`")
+    return urllib.parse.urlparse(repo)
```
Panaetius, line 120: We use url.netloc in a couple locations, but for instance `"not-a-url"` would pass through this function and return a netloc of `""`. Maybe we should check that there is a netloc and if not raise a renku.ParameterError as well?

The exception would be with `file:///some/path/to/a/file` which also doesn't have a netloc but could potentially be valid. But current, we don't support these kinds of urls anyways.

renku/service/controllers/templates_read_manifest.py
```diff
@@ -44,8 +43,8 @@ def context(self):
 
     def template_manifest(self):
         """Reads template manifest."""
-        project = user_project_clone(self.user_data, self.ctx)
-        return read_template_manifest(project.abs_path)
+        template_manifest, _, _, _ = fetch_template(self.ctx["git_url"], self.ctx["ref"])
```
Panaetius, line 46: :+1:

tests/service/controllers/test_templates_read_manifest.py
```diff
@@ -17,16 +17,17 @@
...
+from renku.core.errors import GitError
+
 
 def test_template_read_manifest_ctrl(ctrl_init, svc_client_with_templates, mocker):
     """Test template read manifest controller."""
     from renku.service.controllers.templates_read_manifest import TemplatesReadManifestCtrl
 
     cache, user_data = ctrl_init
-    svc_client, headers, template_params = svc_client_with_templates
+    _svc_client, _headers, template_params = svc_client_with_templates
```
Panaetius, line 30:
```suggestion
    _, _, template_params = svc_client_with_templates
```
Since we don't use these values anyways

tests/service/controllers/test_templates_read_manifest.py
```diff
@@ -37,13 +38,13 @@ def test_template_read_manifest_ctrl(ctrl_init, svc_client_with_templates, mocke
...
 
-@pytest.mark.parametrize("git_url", ["https://github.com", "https://github.com/SwissDataScienceCenter",])
+@pytest.mark.parametrize("git_url", ["https://github.com/`test", "https://github.com/SwissDataScienceCenter/}",])
 def test_validation_exc_template_read_manifest_ctrl(git_url, ctrl_init, svc_client_with_templates, mocker):
     """Test validation exception on template read manifest controller."""
     from renku.service.controllers.templates_read_manifest import TemplatesReadManifestCtrl
 
     cache, user_data = ctrl_init
-    svc_client, headers, template_params = svc_client_with_templates
+    _svc_client, _headers, template_params = svc_client_with_templates
```
Panaetius, line 47: Same as above

tests/service/controllers/test_templates_read_manifest.py
```diff
@@ -58,10 +59,10 @@ def test_found_exc_template_read_manifest_ctrl(git_url, ctrl_init, svc_client_wi
...
```
Panaetius, line 62: Same as above

Panaetius (changes requested):
Some minor comments, but looks really great! I love the modified_environ utility function, that's a great addition!

renku/core/models/git.py
```diff
@@ -44,11 +44,11 @@
 
 _RE_PORT = r":(?P<port>\d+)"
 
-_RE_PATHNAME = r"(?P<pathname>(([\w\-\~\.]+)/)*?(((?P<owner>([\w\-\.]+/?)+)/)?(?P<name>[\w\-\.]+)(\.git)?)?)"
+_RE_PATHNAME = r"(?P<pathname>(([\w\-\~\.]+)/)*?(((?P<owner>([\w\-\.]+/?)+)/)?(?P<name>[\w\-\.]+)(\.git)?)?)/*"
```
rokroskar, line 47: I'm wondering why you need a match for anything after the `/`? I thought this is specifically a slash at the end of the URL, i.e. by definition nothing will come after it.

Panaetius: I think this is for multiple slashes at the end, which is a valid url (like http://github.com/project/repo/////)

rokroskar: 🤦 of course. I shouldn't review PRs on my phone...

Panaetius (approved):
Looks great, thank you :+1:

Panaetius (approved):

PR #2057, opened 2021-04-30
feat(core): add default value to all Run parameters

### LINKED ISSUE #1987: add default values to `renku:Run` template parameters
`renku:CommandParameter` should have a default value field (inherited in the CommandInput/Output/Argument and RunParameter classes) that is set to the values passed in the original `renku run` execution. We should be able to use https://schema.org/defaultValue for this, although that would mean that `CommandParameter` has to also be of type `schema:Thing` or one of its descendants.

This is needed to enable the new workflow UX changes.

Panaetius:
I added it to old and new metadata, but on the old metadata, I left the old way of doing things alongside the defaultValue to not reinvent the wheel. At some point we need to finalize this and add a migration that switches over to the new metadata, and gets rid of the old properties.

closes #1987


mohammad-sdsc (approved):
Thanks Ralf! This looks great! I've noticed some places where `default_value` should be used and fixed them in my PR. I've pushed a commit here as well; feel free to edit it if needed.

PR #2071, opened 2021-05-10
feat(workflow): naming metadata for command parameters

### LINKED ISSUE #2035: Add additional metadata to renku:CommandParameter
Add `schema:name`, `schema:description` to `renku:CommandParameter`, generate sensible default for name, leave description empty

mohammad-sdsc:
# Description

Adds `name` and `description` to `CommandParameter` and it's subclasses.

Fixes #2035

### DIFF
--- a/renku/core/models/workflow/plan.py
+++ b/renku/core/models/workflow/plan.py
@@ -120,7 +120,7 @@ class Plan:
         if not run:
             return uuid.uuid4().hex[:MAX_GENERATED_NAME_LENGTH]
 
-        name = "-".join(run.to_argv())
+        name = "-".join(str(a) for a in run.to_argv())
         name = secure_filename(name)
         rand_length = 5
         return f"{name[:MAX_GENERATED_NAME_LENGTH - rand_length -1]}-{uuid.uuid4().hex[:rand_length]}"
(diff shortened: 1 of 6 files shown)


renku/core/models/workflow/plan.py
```diff
@@ -192,11 +194,13 @@ def convert_output(output: CommandOutputTemplate) -> CommandOutput:
 
             return CommandOutput(
                 id=output._id.replace(self.id_, run_id),
-                produces=entity,
+                create_folder=output.create_folder,
```
Panaetius, line 197: Do we intend to merge my default_value PR before this or to merge this one instead of the default_value PR?

In the latter case, I think we should set default_value here instead of produces and remove produces.
What do you think?

Panaetius (commented):
This looks really great, can't wait to move forward with all these changes!

renku/core/models/workflow/plan.py
mohammad-sdsc: That PR is already merged. This is `CommandOutput` and I believe we did not remove `produces` field there since it belongs to the old metadata.

Panaetius: Right.. For some reason in my head that was already gone. Wishful thinking :sweat_smile:

Panaetius (approved):

PR #2090, opened 2021-05-18
fix(service): fix service project creation

Panaetius:
Fixed git user not being set in newly created project.

Also fixes `owner` being taken from the template repository (e.g. `SwissDataScienceCenter` instead of the actual project namespace)

Now returns `project_id` in the `templates.create_project` request. As the project is in cache anyways, I don't understand why this wasn't returned before.

/deploy

### DIFF
--- a/renku/service/controllers/templates_create_project.py
+++ b/renku/service/controllers/templates_create_project.py
@@ -87,7 +87,7 @@ class TemplatesCreateProjectCtrl(ServiceCtrl, RenkuOperationMixin):
             "name": self.ctx["project_name_stripped"],
             "fullname": self.ctx["fullname"],
             "email": self.ctx["email"],
-            "owner": self.ctx["owner"],
+            "owner": self.ctx["project_namespace"],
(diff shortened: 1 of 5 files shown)


renku/service/controllers/templates_create_project.py
```diff
@@ -87,7 +87,7 @@ def setup_new_project(self):
             "name": self.ctx["project_name_stripped"],
             "fullname": self.ctx["fullname"],
             "email": self.ctx["email"],
-            "owner": self.ctx["owner"],
+            "owner": self.ctx["project_namespace"],
```
mohammad-sdsc, line 90: Can subgroup cause a problem here?

mohammad-sdsc (approved):
Thanks for looking into this! Just a single question; feel free to ignore if not relevant.

renku/service/controllers/templates_create_project.py
Panaetius: I don't think so.

PR #2093, opened 2021-05-18
feat(core): ParameterValue in provenance

### LINKED ISSUE #2036: Add renku:ParameterValue on the provenance graph
Add `renku:ParameterValue`, `renku:PathParameterValue` and `renku:VariableParameterValue` nodes on the provenance graph when doing renku run, mapping to the `renku:CommandParameter` used on the dependency graph.

This information should be filled by `renku graph generate`

renku run/rerun/update should also set this in the new metadata.

mohammad-sdsc:
# Description

Adds `ParameterValue` and it's subclasses to `Activity` when a command is run. Also creates equivalent classes for the new graph metadata instead of using those from the existing metadata (`Association`, `Usage`, `Generation`, `CommandParameter` and its subclasses). Some refactoring and addition of type hints.

Fixes #2036

### DIFF
--- a/renku/core/commands/run.py
+++ b/renku/core/commands/run.py
@@ -147,9 +147,7 @@ def _run_command(
             if return_code not in (success_codes or {0}):
                 raise errors.InvalidSuccessCode(return_code, success_codes=success_codes)
 
-        client.process_and_store_run(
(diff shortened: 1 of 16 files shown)


Panaetius (approved):
This is great, thank you!

PR #2099, opened 2021-05-19
fix(core): remove locking from core read operations

### LINKED ISSUE #2098: service: locking timeout on concurrent API requests
Context: in #1870 a locking mechanism was introduced for the service, which handles read and write locks separately as well as retries. However, a separate locking mechanism is also present in the core methods that the service uses. This causes a locking timeout when multiple API calls are sent by a client concurrently:

image

Sentry Issue: [RENKU-PYTHON-EQ](https://sentry.dev.renku.ch/organizations/swiss-datascience-center/issues/5610/?referrer=github_integration)

```
Timeout: The file lock '/svc/cache/projects/am9obi5kb2VleGFtcGxlLmNvbQ==/e933bc7b74184afcbc38f973a4874f81/renku-qa/core-integration-test/.renku.lock' could not be acquired.
(14 additional frame(s) were not displayed)
...
  File "renku/core/incubation/command.py", line 125, in execute
    hook(self, context, *args, **kwargs)
  File "renku/core/incubation/command.py", line 406, in _pre_hook
    context["stack"].enter_context(context["client"].lock)
  File "contextlib.py", line 427, in enter_context
    result = _cm_type.__enter__(cm)
  File "filelock.py", line 323, in __enter__
    self.acquire()
  File "filelock.py", line 278, in acquire
    raise Timeout(self._lock_file)
```

Panaetius:
closes #2098

/deploy


mohammad-sdsc (approved):
LGTM, thank you!

PR #2117, opened 2021-05-28
Create user-story-workflow-execution-with-extensive-parameter-space.md

almutlue:
Use case for new design/features of the workflow commands in renku, specifically targeting the design of renku run --map

### DIFF
--- a/design/001-new-workflow-commands/user-story-workflow-execution-with-extensive-parameter-space.md
+++ b/design/001-new-workflow-commands/user-story-workflow-execution-with-extensive-parameter-space.md
@@ -1,0 +1,59 @@
+## RenkuLab use case for workflow execution while exploring parameter space: 
+Reference repo: [mnn_omni_batch](https://renkulab.io/projects/omnibenchmark/omni_batch/mnn-omni-batch)
+
+### Context
+
+Our framework consists of multiple repositories connected and coordinated using knowledge graph queries. 
+Each method workflow repository runs a method as part of a benchmark with a set of predefined parameter combinations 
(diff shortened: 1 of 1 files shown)


design/001-new-workflow-commands/user-story-workflow-execution-with-extensive-parameter-space.md
```diff
@@ -0,0 +1,59 @@
...
+compare them to all data file-parameter combinations and generate new workflows for new combinations by running `src/workflows/define_workflow.sh`, 
+while updating the existing ones using `renku update`. 
+For every method, benchmarking requires testing different parameter combinations for each available data file. 
+As little as 10 data files processed with 4 parameters with 3 values each would already result in over a 100 workflows, 
+posing the challenges for method maintenance and interpretation of benchmarking results. 
+A substantial effort will be needed to handle such a number of workflows in case of any changes/edits or error messages. 
+Furthermore, generating a workflow for every method-data-parameter combination is redundant and will rapidly increase in complexity as the number of 
+methods and data increases.
+Given that the planned workflow commands will allow us to modify parameters for workflow template execution, 
+we aspire to rely on the new functionality `renku run --map` to search the whole parameter space,
```
Panaetius, line 37: Actually, we envision a command `renku workflow loop <workflow>` that does a repeat execution with different values for parameters, `renku run` itself will only take a single value

almutlue: Thanks a lot, renku workflow loop seems exactly what we were looking for.

Panaetius (approved):

PR #2120, opened 2021-05-31
feat(core): extended template variable functionality

### LINKED ISSUE #2034: Extend variables support in templates used by the init command
Currently, templates define variables without specifying any metadata.
To take advantage of the jinja templating tool, it would be nice to support type validation, enumerations and default values as specified here SwissDataScienceCenter/renku-project-template#109

The new properties of a variable would be:
- `name`: variable name, currently taken from the key. It must be unique (check when validating the template)
- `description`: a short description for the variable intended use, currently taken from the value.
- `type`: variable type. We should support at least `string`, `enum`, `number`, `boolean`. The validation should be done at project creation (check the specified values is valid/allowed when validating the template). Requiring the specific type is probably the easiest solution, but I would relax the constrain to accept also stringy numbers (`"12"` instead of `12`), as well as `0` / `1`, `"0"` / `"1"` or `"true"` / `"false"` (any capitalization) for booleans.
- `enum`: an array of allowed values. This is mandatory when the type is "enum" and can't be specified when it is any other type (check when validating the template).
- `default`: optional default value. It must have a valid type (check when validating the template).

The renku-core endpoint `templates.read_manifest` should change accordingly, providing the variables as an array of objects with the same properties specified above.

The primary goal is to support enumerations. Any alternative solution is welcome

Panaetius:
Adds support for default values and typing (string, enum, boolean, number) to template variables.

https://dev.renku.ch/gitlab/renku-qa/core-it-template-variable-test-project contains an example of this.

closes #2034

/deploy


rokroskar:
cc @lorenzo-cavazzi

lorenzo-cavazzi:
Great to see this taking form!

I was messing around a bit, and I've noticed that `--description` doesn't work well when I provide a specific source. In other words, `renku init -d` worked as expected showing me the description, while `renku init -s <source> -d` initialize the project ignoring the `-d` switch.

Should `--description` also show the expected type (if any)?

Panaetius:
@lorenzo-cavazzi Oh actually that was a bug that has been around for a long time it seems, `-d` didn't work with a template repo that only has 1 template (In the case of 1 template, that one is used and the user is not prompted to pick a template, so the `-d` code was bypassed).

Not really a part of this ticket but I fixed it so you can test :slightly_smiling_face:

lorenzo-cavazzi:
This works well in the cli :+1: 
Could you add the `/deploy` string in the first post? I'd like to quickly try it on the service

lorenzo-cavazzi:
Sorry for taking a bit longer for the review. I expected the core service to be compatible with the UI since no breaking changes have been introduced, but it crashes when creating a new project so I'm trying to go deeper into the problem.

In the meanwhile, I've noticed the CLI doesn't ask for a value when the parameter has a default. The only way to change it is by providing it through the console.
Would It be a better UX asking anyway for a value and using the default only if the user doesn't enter anything?

Screenshot_20210604_115315

Funny fact: the cli shows the parameters in different orders every time. It's not a big deal, but I found it a bit unusual (I typed `fake description`  a few times before realizing it was asking me for the numeric variable first, since the first 3-4 times I got asked for the description first)

lorenzo-cavazzi:
This looks good to me! I tested a bit with https://dev.renku.ch/gitlab/lorenzo.cavazzi.tech/template-variables

It contains a breaking change for the UI since the `templates.read_manifest` API returns an object for each variable instead of a string, causing the UI to crash with all the templates.
Since that seems a good solution, I would try to have a compatible version of the APIs but I would wait to merge this until the UI catches up.

rokroskar:
looks like https://github.com/SwissDataScienceCenter/renku-ui/pull/1579 is ready to go so should we merge this as well @Panaetius ?

Panaetius:
@rokroskar yes I think that makes sense

rokroskar:
Is it a breaking change or can it be merged independent of the UI PR?

Panaetius:
It should be backwards compatible (otherwise tests would fail as current templates don't contain this yet).

renku/core/commands/init.py
```diff
@@ -153,18 +161,118 @@ def select_template_from_manifest(
...
+        if "default_value" in template_variables[key]:
+            default_value = template_variables[key]["default_value"]
+
+            if not validate_template_variable_value(key, template_variables[key], default_value)[0]:
+                raise errors.InvalidTemplateError(
+                    f"Default value {default_value} for template variable {key} does not match variable type."
+                )
+
+            metadata[key] = default_value
+            continue
```
mohammad-sdsc, line 273: This won't prompt for parameters with a default value. Shouldn't we prompt for a value and use the default value when user enters nothing?

tests/cli/test_init.py
```diff
@@ -322,6 +322,128 @@ def test_init_remote(isolated_runner, project_init):
...
+
+    metadata = json.loads(project.template_metadata)
+    assert metadata["bool_var"]
+    assert metadata["enum_var"] == "ask again"
+    assert metadata["description"] == "some description"
+    assert metadata["number_val"] == 70.12
+
+    shutil.rmtree(new_project)
+
+    # create project using wrong number
```
mohammad-sdsc, line 364: It's more readable if we split this test into multiple smaller tests.

mohammad-sdsc (commented):
Looks very good! I've made some comments. I also notice that `--describe` gets ignored when passed without `-l`. This is a bit confusing. Perhaps we should rename it to `--verbose` or make also list and describe templates. This of course should be done in a new story.

mohammad-sdsc (approved):
Thank you!

PR #2122, opened 2021-06-02
feat(service): improve formatting for migrationscheck response

### LINKED ISSUE #2048: Return more finegrained Dockerfile update information in service
Currently, the service just returns `docker_update_possible` True/False.

But `False` can mean 3 things:
- The Dockerfile is up to date
- The Dockerfile doesn't contain an `ARG ....` line
- There is no Dockerfile

We should differentiate these cases and return finegrained information in the service, so the Ui can show different messages to the user depending on the case.

Panaetius:
closes #2048 
/deploy extra-values=core.debug=true


vfried:
Hi, thanks for your changes.... i think this will improve the way we deal with things in the frontend and make the code a bit better to understand.

I have a question,the JSON is currently like this:

```
{
  "result": {
    "project_renku_version": "0.12.2.dev0+dirty",
    "dockerfile_renku_status": {
      "automated_dockerfile_update": false,
      "newer_renku_available": null,
      "latest_renku_version": "0.15.2.dev32+g434705f2+dirty",
      "dockerfile_renku_version": null
    },
    "template_status": {
      "project_template_version": "66d4f56959466cca06c2c5fdbf63c9eeda5c728f",
      "template_id": "python-minimal",
      "latest_template_version": "66d4f56959466cca06c2c5fdbf63c9eeda5c728f",
      "template_ref": "template_updates",
      "template_source": "https://github.com/SwissDataScienceCenter/renku-project-template",
      "newer_template_available": false,
      "automated_template_update": true
    },
    "core_renku_version": "0.15.2.dev32+g434705f2+dirty",
    "core_compatibility_status": {
      "current_metadata_version": "8",
      "migration_required": false,
      "project_metadata_version": "8"
    },
    "project_supported": true
  }
}
```

I thought that project_supported, core_renku_version and project_renku_version were related to "core_compatibility_status" shouldn't we put them inside that object?

I think in the UI we should make 3 separated cards inside the status page...
image

I will write what i would do in the UI here to make sure we are on the same page, let me know if something is wrong:

**- Renku Core** 
---> if migration_required && project_suported we recommend UI migration,
---> if migration_required && !project_suported we recommend manual migration
---> else we display a sign that says things are up to date
_??? shall we do something with the !migration_required && !project_suported?_

**- Renku inside Dockerfile**
---> if newer_renku_available && automated_docker_update we recomend UI migration
---> if newer_renku_available && !automated_docker_update we recomend manual migration
---> else all is OK

**- Template**
---> if newer_template_available && automated_template_update we recommend ui migration
---> if newer_template_available && !automated_template_update we recommend manual migration
---> else all is OK

**Are these three things related or are they fully independent???**

Panaetius:
> I thought that project_supported, core_renku_version and project_renku_version were related to "core_compatibility_status" shouldn't we put them inside that object?

`project_supported` is somewhat related, `core_renku_version` and `project_renku_version` aren't.
`project_supported` depends on whether we are dealing with a renku project or not (i.e. if it has a `.renku/metadata.yml`) and on whether the metadata version of the project is higher than the service metadata version.
`core_renku_version` and `project_renku_version` only say what version each is, but renku will happily work with them if they have the same metadata version, so whether core is compatible with it or not is not influenced by them.

> if migration_required && !project_supported we recommend manual migration
this can't happen, `migration_required` depends in the project being supported to be true.

> ??? shall we do something with the !migration_required && !project_supported?
this means it's either not a renku-project or it was made/modified with a newer version of renku. Nothing the service can do until it gets updated with a new release to match the version.

>  if newer_template_available && !automated_template_update we recommend manual migration
I don't think there can be such a thing as manual migration for templates. In the case above, a user could migrate the template with a normal `cache.migrate` request with the `force_template_update` field set to `true`, so it's done for the user but `manual` in the sense that it has to be forced and that the template author hasn't marked their template as being updatable (Which could be an oversight or set on purpose).

> Are these three things related or are they fully independent???
They're fully independent except for the template containing a `Dockerfile`. It'd first update the Dockerfile with the Dockerfile from the template and then do the Dockerfile migration. But I don't think this matters for our use-case.

> "force_template_update": false, ???
This is only useful if the template doesn't allow automatic updates but we want to force an update.

> "migrate_project": true, ??
Oh, I just noticed this, this comes from the porcelain changes Sam did. It allows to on the fly migrate a project before executing a command (So you could set it to true on e.g. `dataset.create` and it'd migrate before creating the dataset). I think this is a bad thing to do, since it can affect performance and it could end up pushing a migration without the user being aware of it. Setting it on `cache.migrate` would execute migrations twice. I think this shouldn't exist on this endpoint at all, but it is part of the base class that's used for all endpoint request schemas :disappointed: 
So you should always set `migrate_project` and `force_template_update` to false.

renku/core/commands/migrate.py
```diff
@@ -45,38 +47,25 @@ def migrations_check():
...
+        "project_renku_version": latest_version,
+        "core_compatibility_status": _metadata_migration_check(client),
+        "dockerfile_renku_status": _dockerfile_migration_check(client),
+        "template_status": _template_migration_check(client),
+    }
 
 
 def migrations_versions():
     """Return a command to get source and destination migration versions."""
-    return Command().command(_migrations_versions).lock_project().with_database()
```
mohammad-sdsc, line 79: Isn't database required to read project's metadata?

mohammad-sdsc (changes requested):
Thanks Ralf! This looks pretty good.

mohammad-sdsc:
Isn't `core_renku_version` the same as `dockerfile_renku_status.latest_renku_version`? Is there a reason to have redundant info here? Also `project_renku_version` comes from the commit agent which should be newer than `dockerfile_renku_status.dockerfile_renku_version`; do we need both of them here?

mohammad-sdsc:
API version is currently done like `/v0.9/` but according to the RFC it should be like `/0.9/`.

renku/core/commands/migrate.py
Panaetius: no, it falls back to reading `metadata.yml` otherwise migration wouldn't work.

Panaetius:
> Isn't `core_renku_version` the same as `dockerfile_renku_status.latest_renku_version`? Is there a reason to have redundant info here? Also `project_renku_version` comes from the commit agent which should be newer than `dockerfile_renku_status.dockerfile_renku_version`; do we need both of them here?

Yes it is duplicate, UI requested `core_renku_version` and `dockerfile_renku_status.latest_renku_version` like this. I don't think it matters having it duplicated and if this makes implementing logic on there side easier, why not.

`dockerfile_renku_status.dockerfile_renku_version` is relevant for updating the Dockerfile (Can also be None). `project_renku_version` is irrelevant imo, maybe nice to have to present to a user. But if a user manually modifies Dockerfile, they could be disparate (And the project probably broken). I think the UI just uses `project_renku_version` to say "Updating from 0.16.2 to 1.0.0" even though what is actually done is updating from 8 to 9, but users don't know what "8" or "9" means. 

From our point of view it doesn't matter but I think it's useful for the UI to show to users (to hide some complexity).

mohammad-sdsc (approved):

PR #2145, opened 2021-06-08
feat(core): add renku version as a project template parameter

### LINKED ISSUE #2025: feat(core): make renku version a templatable parameter
The current version of renku should be passed as `{{renku_version}}` parameter to project templates on project initialization.

This allows the Dockerfile in the template to have the renku version as `ARG RENKU_VERSION={{renku_version}}` so we don't need top keep updating it in the template repos.

Also update the documentation of template parameters accordingly.

vigsterkr:
# Description

Fixes #2025


rokroskar:
nice! This should be followed up with https://github.com/SwissDataScienceCenter/renku-project-template/issues/98 asap.

Question: could we add a flag to `renku init` override the version? This could be useful if I have a newer version of renku locally but I want to push my project to a renkulab instance that runs an older version in the core service.

vigsterkr:
@rokroskar sure!
imo this should do the trick
```renku init --parameter "renku_version"="1.2.3"```

**Review by Panaetius — COMMENTED** [review]
Also needs to be added to https://github.com/SwissDataScienceCenter/renku/blob/master/docs/user/templates.rst#create-a-template-repository

I wonder if we should name it `__renku_version__` to be consistent with the other renku-defined parameters? @rokroskar

**Comment by vigsterkr** [comment]
@Panaetius yeah that was the initial idea, but thought it'd a bit 'clunky' with all the underscores if we wanna support the feature that @rokroskar mentioned. unless of course instead of `--parameter "renku_version":"<VERSION>"` we would go with a dedicated cmd line param for `renku init`, e.g. `renku init --version <VERSION>`

**Comment by rokroskar** [comment]
I was thinking a dedicated flag - but it's also likely that this won't be used very often so maybe leaving it with underscores and together with `--parameter` is enough, as you suggested @vigsterkr. 

Another thing that comes to mind in this vein is that perhaps we should warn the user if the version that the project will be initialized with is different than what is the default for the template - but this should be done in a follow-up.

File: .github/workflows/test_deploy.yml [hunk 1]
```diff
@@ -69,6 +69,7 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install coveralls setuptools wheel
         python -m pip install .[all]
+        python setup.py --version
```
vigsterkr, line 72: @Panaetius hope u dont mind that i've added this for the docs building job.

Panaetius: No I think that can be useful

Panaetius (approved):

PR #2146, opened 2021-06-09
fix(cli): prevent --template-ref from being set without --template-source in renku init

### LINKED ISSUE #2023: Specifying `-r` without a source template should fail on init
This should either fail or try to actually initialize from the reference provided:

```
renku init -r 73db625 test-init
```

The `.renku/metadata.yml` file shows the correct ref even though this is not the ref that was used to initialize the project (whatever the version of the template that is hard-coded in renku-python was used instead).

Panaetius:
closes #2023


vigsterkr (commented):
LGTM!

vigsterkr (dismissed):

vigsterkr (approved):

PR #2159, opened 2021-06-10
chore: Fix coveralls

Panaetius:
Fixes Coveralls coverage report.

Makes coveralls run in parallel and adds a final step at the end to aggregate the results.

Adds integration tests to coveralls.


rokroskar (commented):
Should we be restricting the scopes on the github token for all of the coveralls steps?

Panaetius:
The other steps don't use the github token, only the coveralls token. It's only this weird endpoint for `--finish` that requires a github token

rokroskar (approved):
Thanks!! 🎉

PR #2160, opened 2021-06-10
fix(core): do not remove unused template parameters

vigsterkr:
# Description

Do not remove from the template parameters - just warn user - that are not explicitly specified in metadata. moreover, internal parameters like `__parameter__` should not be exposed to the user.

### DIFF
--- a/renku/core/commands/init.py
+++ b/renku/core/commands/init.py
@@ -18,6 +18,7 @@
 """Project initialization logic."""
 
 import json
+import re
 import tempfile
 from collections import OrderedDict, namedtuple
 from pathlib import Path
@@ -37,7 +38,7 @@ from renku.core.management.config import RENKU_HOME
 from renku.core.management.repository import INIT_APPEND_FILES, INIT_KEEP_FILES
 from renku.core.models.tabulate import tabulate
 from renku.core.utils import communication
-from renku.version import __version__ as renku_version
+from renku.version import __version__, is_release
 
 TEMPLATE_MANIFEST = "manifest.yaml"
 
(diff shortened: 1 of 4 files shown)


vigsterkr:
@rokroskar just to clarify, in case of dirty version (i.e. when working with a developer branch of renku) the default value of `__renku_version__` for template parameter should be
 a) unset, i.e. rely on the template's default value
 b) `0.15.2.dev9+g0e9b1328+dirty` -> `0.15.2`

rokroskar:
it should be unset because afaik the dev versions are one bugfix release ahead of the last stable release. So in your example 0.15.2 doesn't actually exist yet. You could (I suppose) downgrade to 0.15.1 but it's probably safer to just go with the template default. Not sure tbh - what do you think?

vigsterkr:
mmm indeed, i'd go with unset renku version and basically rely on the default value of the parameter in the template; and anyways the user can always override this value, if the default is not recent enough.

vigsterkr:
> it should be unset because afaik the dev versions are one bugfix release ahead of the last stable release. So in your example 0.15.2 doesn't actually exist yet. You could (I suppose) downgrade to 0.15.1 but it's probably safer to just go with the template default. Not sure tbh - what do you think?

one last thing: shall we validate the user input for `__renku_version__`, i.e. that it's actually a proper semver/release. personally i'd go w/o validation and leave it up to the user to make sure that it's a valid release that they are providing when setting `--parameter __renku_version=<VERSION>`

rokroskar (approved):
Works as advertised. Thanks! 🚀

PR #2161, opened 2021-06-10
feat(core): new metadata persistent layer

### LINKED ISSUE #2108: Implement custom persistent.Persistent backend
We should implement a custom backend that allows us to store `persistent.Persistent` objects.

Features it should support:
- storing under different roots (i.e. having named collections with different names that will be stored in different paths)
- logic to generate IDs for the objects (can we hash the json_ld @id or do we want custom incremental ids?)
- Cache objects in memory for performance
- Walk object trees and store child object, support deduplication

## API

```
from renku.core.storage import Storage

storage = Storage(root=path) #initializes the storage, could be handled similar to `Communication` so it can be set on CommandBuilder and accessed everywhere.
storage["plans"] = BTree() #adds an object to root storage
storage["plans"][plan_id] = Plan() 
storage.commit() #saves all outstanding changes on the root storage (and other tracked objects)

```

Looking at https://github.com/zopefoundation/ZODB/blob/master/src/ZODB/Connection.py I think it makes sense to implement the `get`, `add`, `root`, `register`, `setstate`, `commit` methods, albeit in a very simplified way compared to the ZODB implementation.

We can add transaction support at a later point if we want. We don't need to follow their interface or we could leave some parameters `None` if they might be needed later. Drop as much ZODB relevant stuff as possible.

## Implementation

ZODB uses https://github.com/zopefoundation/ZODB/blob/1fb097b41cae4ca863c8a3664414c9ec0e204393/src/ZODB/serialize.py#L432 to call their custom pickle implementation that deduplicates objects in an object graph and returns pickled data.

We probably want something similar that recursively stores objects as json using the `__reduce__()` method and the objects `_p_oid` id property, adding id references if we encounter a nested `persistent.Persistent` object (and storing them separately), otherwise just recursively including the `__reduce__()` output in the top object. 

In ZODB, a lot of this stuff happens in the zodbpickle C code, so not really something we can reuse. But the C-code calls [thos](https://github.com/zopefoundation/ZODB/blob/22d1405d413a0c210ecbf43575d6bb15b34ecdc9/src/ZODB/serialize.py#L339) to create references. 
So what happens is:
- We ask the Storage to store some object
- this object gets added to a stack
- while the stack is not empty:
  * serialize the object and it's properties, recursively, using __reduce__() and adding type information in case of classes
  * if one of the children is `persistent.Persistent`, replace it with an entry like [type, oid] in some special format and add the object itself to the stack (generating a new id if the object doesn't have one yet).
  * write the serialized data to the store in a location based on its oid
I think we can copy this approach, just not with pickle.
Also pay attention to [persistent.wref](https://github.com/zopefoundation/persistent/blob/6b500fc3f517fb566e8d0b85b756ddf49758c8d8/persistent/wref.py), we probably need to support that for this to work with zope stuff.

It should be implemented in a way so it's easy to swap out `json.dumps` for something else, like `msgpack.packb`. This allows us to easily do performance tests.

A class can then be loaded back similar to how pickle does it:
```
obj = cls.__new__(cls)
obj.__dict__.update(attributes)
```
Note that this bypasses the class' `__init__`. We might go a different route if we want to, but this is probably very efficient.
So we get the data from json as a dict, read `cls` to see what to instantiate, then instantiate with the data. Child objects that are [type, oid] references get loaded as GHOST objects (see below), which just have oid and the Storage/DataManager set.

When loading a nested `persistent.Persistent` object, it should be initialized as a [ghost object](https://persistent.readthedocs.io/en/latest/glossary.html#term-ghost). This is `persistent`s way of doing proxies and lazy-loading objects. See [here](https://github.com/zopefoundation/persistent/blob/6b500fc3f517fb566e8d0b85b756ddf49758c8d8/persistent/persistence.py#L389) for the mechanism (p_activate is called when accessing an object's properties), which calls `setstate` on the DataManager` to instantiate the object. [here](https://github.com/zopefoundation/ZODB/blob/22d1405d413a0c210ecbf43575d6bb15b34ecdc9/src/ZODB/Connection.py#L760) is how that loads it from the database based on `_p_oid`.

For storage, `_p_oid` should be a hex string, then we can have a directory structure similar to git, like `oid[0:2]/oid[2:4]/oid.json`. maybe it makes sense if each of our `persistent.Persistent` classes needs a special `_storage_name_` class attribute that then gets used to determine where to store it, making the path  `.renku/_storage_name_/oid[0:2]/oid[2:4]/oid.json` (with a generic folder for classes that don't have this property, e.g. `BTree`s. Probably using something like sha256 of the @id of an object makes sense for us, even though ZODB uses incremental ids (just counting up from 0x000000000000). That's also more thread safe.

mohammad-sdsc:
# Description

Adds a metadata database that stores ZODB's Persistent objects. There are some root classes (`Activity`, `Entity`, and `Plan` atm) that include objects of these types. Each object is serialized to its own file.

Fixes #2108


renku/core/incubation/database.py
```diff
@@ -0,0 +1,466 @@
...
+    def _create_root_type(root, root_type):
+        if root_type in root:
+            return
+
+        object = BTrees.OOBTree.BTree()
+        object._p_oid = root_type.encode("ascii")
+        root[root_type] = object
+
+    @staticmethod
+    def new_oid():
```
Panaetius, line 85: This probably also doesn't need to be a static method

renku/core/incubation/database.py
```diff
@@ -0,0 +1,466 @@
...
+        """Generate a random oid."""
+        return f"{uuid4().hex}{uuid4().hex}".encode("ascii")
+
+    @staticmethod
+    def generate_oid(object) -> bytes:
```
Panaetius, line 90: Same as above, all the places this is used have a reference to the database already, so it doesn't need to be a static method.

renku/core/incubation/database.py
```diff
@@ -0,0 +1,466 @@
...
+        try:
+            del self._cache[object._p_oid]
+        except KeyError:
+            pass
+        else:
+            type_name = type(object).__name__
+            if type_name in self._root_type_names:
+                root_object = self.root[type_name]
+                del root_object[object._p_oid.decode("ascii")]
+                root_object._p_changed = 1
```
Panaetius, line 231: Does deleting an object from a BTree not set _p_changed ?

renku/core/incubation/database.py
```diff
@@ -0,0 +1,466 @@
...
+
+    def store(self, oid: bytes, data: Union[Dict, List]):
+        """Store object."""
+        assert isinstance(oid, bytes)
+
+        filename = oid.decode("ascii").lower()
+        compressed = len(filename) >= MIN_COMPRESSED_FILENAME_LENGTH
+
+        if compressed:
+            path = self.path / filename[0:2] / filename[2:4] / filename
```
Panaetius, line 257: :+1:

renku/core/incubation/database.py
```diff
@@ -0,0 +1,466 @@
...
+            path.parent.mkdir(parents=True, exist_ok=True)
+            open_func = open  # TODO: Change this to gzip.open for the final version
+        else:
+            path = self.path / filename
+            open_func = open
+
+        with open_func(path, "w") as file:
+            json.dump(data, file, ensure_ascii=False, sort_keys=True, indent=2)
```
Panaetius, line 265: It'd be nice if `Storage` just had two method stubs
```
def load(path):
    raise NotImplementedException()

def dump(data, path):
    raise NotImplementedException()
```

And then we can have a child class `JsonStorage`, `GZipStorage` (Which could just wrap any storage), `MsgPackStorage` that implement these functions. Then we can easily do comparisons and it's a bit less tighly coupled.

renku/core/models/entity.py
```diff
@@ -0,0 +1,121 @@
...
+
+        return entity
+
+
+class Collection(Entity):
+    """Represent a directory with files."""
+
+    def __init__(self, id: str, checksum: str, path: Union[Path, str], members: List[Entity]):
+        super().__init__(id, checksum, path)
+        self.members: List[Entity] = members
```
Panaetius, line 94: do we need to switch lists on `Persistent` classes over to `PersistentList` ?

renku/core/incubation/database.py
```diff
@@ -0,0 +1,466 @@
...
+            if not object._p_changed and object._p_serial != NEW:
+                continue
+
+            object_type = type(object).__name__
+            if object_type in self._root_type_names:
+                self._create_root_type(self.root, object_type)
+                oid = oid.decode("ascii")
+                root_object = self.root[object_type]
+                root_object[oid] = object
+                root_object._p_changed = 1
```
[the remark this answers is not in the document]
mohammad-sdsc: Old habit of not trusting SW engineers.

renku/core/incubation/database.py
mohammad-sdsc: This does not use any member from `Database` so does not need to be a normal method. Moreover, we can call it like this from outside to generate an oid if needed.

renku/core/models/entity.py
mohammad-sdsc: I believe for lists it does not matter since they don't call `__hash__` or `__eq__` when an object is added. For dict and set we probably need to check.

renku/core/incubation/database.py
mohammad-sdsc: For some reason I saw the other one first. Same as below.

renku/core/incubation/database.py
```diff
@@ -0,0 +1,470 @@
...
+        self._reader.set_ghost_state(object, data)
+
+        object._p_serial = PERSISTED
+
+    def register(self, object):
+        """See persistent.interfaces.IPersistentDataManager::register."""
+        self._registered_objects.append(object)
+
+    @classmethod
+    def from_path(cls, path: Union[str, Path]) -> "Database":
```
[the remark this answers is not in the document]
mohammad-sdsc: Actually, `Database` used to get a `Storage` in its constructor which I've changed it to create the storage itself. On a second thought, that approach was cleaner in case we want to create a test database with a test storage.

renku/core/models/entity.py
Panaetius: I was more thinking of 
```
Note that mutating a non-persistent attribute of a persistent object such as a dict or list will not cause the containing object to be changed. Instead you can either explicitly control the state as described below, or use a PersistentList or PersistentMapping.
```
from https://persistent.readthedocs.io/en/latest/using.html#associating-an-object-with-a-data-manager

So as I understand it, adding a member wouldn't cause the `_p_changed` of the `Collection` to be updated.

renku/core/incubation/database.py
mohammad-sdsc: I've changed it to be like this with a difference that it accepts filename instead of full path.

renku/core/models/entity.py
mohammad-sdsc: Actually, if we change a Collection then its checksum changes and we need to create a new object with update `oid` and then replace the current object with it. In that case, we don't need to change the current object.

renku/core/incubation/graph.py
```diff
@@ -113,6 +113,9 @@ def process_datasets(commit):
     elif client.has_graph_files() or client.has_datasets_provenance():
         raise errors.OperationError("Graph metadata exists. Use ``--force`` to regenerate it.")
 
+    # database = Database.from_path(path=client.database_path)
```
Panaetius, line 116: I guess we can completely remove this. I remember I had to add it because I got some errors otherwise but that seems to be fixed

renku/core/management/repository.py
```diff
@@ -504,17 +504,11 @@ def update_graphs(self, activity: Union[ProcessRun, WorkflowRun], database: Data
...
-                database.add(activity.association.plan)
-
-            database.commit()
-        except AssertionError:
-            pass
+        for activity in activity_collection.activities:
+            database.get("activities").add(activity)
+            database.get("plans").add(activity.association.plan)
 
+        database.commit()
```
Panaetius, line 511: This reminds me, I never checked if we need the `database.commit()` here, since the command with `with_database(write=True)` would do it at the end of the command.

renku/core/management/repository.py
```diff
@@ -531,16 +525,14 @@ def initialize_graph(self, database: Database):
...
 
         from renku.core.models.provenance.activity import Activity
         from renku.core.models.workflow.plan import Plan
 
         database.add_index(name="activities", value_type=Activity, attribute="id")
         database.add_index(name="plans", value_type=Plan, attribute="id")
 
-        # # NOTE: Access dataset's root to make sure that it is created in case there are no commits
-        _ = database.root
+        database.commit()
```
Panaetius, line 535: Same as above

renku/core/incubation/database.py
```diff
@@ -0,0 +1,671 @@
...
+        if values:
+            assert isinstance(values, list), f"Value in IndexList is not a list: {values}"
+            for value in values:
+                if value._p_oid == object._p_oid:  # Object is already indexed
+                    return
+            values.append(object)
+            self._entries._p_changed = True
+            # TODO: Do we need self._p_changed = True
+        else:
+            self._entries[key] = [object]
```
Panaetius, line 426: If we used `PersistentList` (https://persistent.readthedocs.io/en/latest/api/collections.html) we wouldn't have to set _p_changed on append manually, I think.

Panaetius (approved):

Panaetius (dismissed):

Panaetius (approved):

Panaetius (approved):

PR #2169, opened 2021-06-11
fix(core): externalize version_template of setup.py

vigsterkr:
# Description

scm_version's version template value has moved to a template file.
This avoids unnecessary escaping required in future patch of the template


Panaetius (approved):
Thank you!

PR #2173, opened 2021-06-14
chore: update image builds; use slim

rokroskar:
* Change from alpine to slim to take advantage of pre-built wheels. 
* Use a two-stage build for the service to reduce image size
* Remove the `debug` group since it's not needed because code isn't running in editable mode

/deploy

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,3 +5,4 @@
 !*.rst
 !entrypoint-svc.sh
 !.git
+!.gitignore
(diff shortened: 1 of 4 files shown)


rokroskar:
Note that the acceptance tests are failing because of https://github.com/SwissDataScienceCenter/renku-project-template/issues/112

Panaetius (approved):
Looks great, thank you!

PR #2175, opened 2021-06-16
chore: fix image build and test in ci

rokroskar:
Fixes the image build; also adds image building to the CI on pushes so broken builds don't sneak by in the future.

### DIFF
--- a/.github/workflows/test_deploy.yml
+++ b/.github/workflows/test_deploy.yml
@@ -558,18 +558,21 @@ jobs:
         name: release-url
         path: release-url.txt
 
-  push-image:
+  build-images:
     runs-on: ubuntu-latest
     needs: [docs-linux, test-linux-cli, test-linux-core, test-linux-service, test-macos-cli, test-macos-core, test-macos-service]
-    if: "startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/master'"
     steps:
     - uses: actions/checkout@v2
+    - name: build images
+      run: make docker-tag
     - name: Docker Login
+      if: "startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/master'"
       uses: Azure/docker-login@v1
       with:
         username: ${{ secrets.DOCKER_USERNAME }}
         password: ${{ secrets.DOCKER_PASSWORD }}
-    - name: build and push image
+    - name: push images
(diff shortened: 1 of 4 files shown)


.github/workflows/test_deploy.yml
```diff
@@ -558,18 +558,21 @@ jobs:
...
     runs-on: ubuntu-latest
-    needs: [docs-linux, test-linux-cli, test-linux-core, test-linux-service, test-macos-cli, test-macos-core, test-macos-service]
-    if: "startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/master'"
+    # needs: [docs-linux, test-linux-cli, test-linux-core, test-linux-service, test-macos-cli, test-macos-core, test-macos-service]
     steps:
     - uses: actions/checkout@v2
+    - name: build images
+      run: make docker-tag
     - name: Docker Login
+      if: "startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/master'"
```
vigsterkr, line 569: nitpicking: in case we change the default branch from `master` to `main` or something else this requires a patch

.github/workflows/test_deploy.yml
```diff
@@ -558,18 +558,21 @@ jobs:
...
       uses: Azure/docker-login@v1
       with:
         username: ${{ secrets.DOCKER_USERNAME }}
         password: ${{ secrets.DOCKER_PASSWORD }}
-    - name: build and push image
+    - name: push images
+      if: "startsWith(github.ref, 'refs/tags/') || github.ref == 'refs/heads/master'"
```
vigsterkr, line 575: same as above. i'm wondering if it would be possible to make this at least a workflow parameter, or actually check whether the github.ref == `git symbolic-ref refs/remotes/origin/HEAD`.
but this is just a really minor details

vigsterkr (commented):
super minor - nitpicking - comments

.github/workflows/test_deploy.yml
rokroskar: good idea - I'd like to make this more robust... but won't your suggestion always be false?

rokroskar: btw we use this pattern elsewhere - I'd say we should make an issue to fix it in general and maybe also propagate that fix to other component repos

vigsterkr (approved):

Panaetius (approved):

PR #2179, opened 2021-06-17
fix(core): version template lint

vigsterkr:
# Description

With the current version template the generated `version.py` fails passing linting constraints. this patch is addressing those warnings.


Panaetius (approved):
Thanks!

PR #2181, opened 2021-06-17
feat(core): new dataset provenance

### LINKED ISSUE #2119: Change dataset provenance to store datasets in storage
We should have two indices, one for the HEAD of each dataset chain, and one for all the revisions of dataset. The dataset code needs to update both when a dataset is modified.

so storage["datasets"] would contain all the current datasets while storage["dataset_history"] would contain all revisions of a dataset.

mohammad-sdsc:
# Description

Adds two indexes to the new persistent layer: `datasets` and `datasets-provenance`. The first index is a mapping from dataset names to data objects for all current datasets in a project. The second index is a mapping from dataset id to dataset object; this index contains current datasets and all previous versions of all datasets.

Fixes #2119


renku/core/management/datasets.py
```diff
@@ -126,42 +128,23 @@ def renku_pointers_path(self):
...
-        if not self.has_datasets_provenance():
-            return
-        if not self._datasets_provenance:
-            self._datasets_provenance = DatasetProvenance.from_json(self.datasets_provenance_path)
-
-        return self._datasets_provenance
-
-    def update_datasets_provenance(self, dataset, remove=False):
+    @inject.autoparams()
+    def update_datasets_provenance(
```
Panaetius, line 132: Would it be cleaner to have two methods, `add_or_update_dataset_provenance` and `remove_dataset_provenance`?

It would also be cleaner to just call the methods on `DatasetProvenance` directly and inject that where needed, since this method in LocalClient doesn't really contain any additional logic (especially since `commit_database` isn't set to `True` anywhere). In the future I'd like to rely on `LocalClient` less and have the dedicated classes do more of the heavy lifting.
If we inject `DatasetProvenance` using the `binder.bind_to_constructor(Class, lambda: Class(*args))` way of injection, it would only initialise `DatasetProvenance` if code that uses it (well, requests it as an injection) is called and then it would be a singleton, so we could just inject this, `ProvenanceGraph` and `DependencyGraph` this way in the `with_database` CommandBuilder without performance overhead, and it'd save performance since we wouldn't have to do `from_database` more than once.

renku/core/management/datasets.py
```diff
@@ -1488,7 +1462,7 @@ def _check_url(url):
         if not is_git:
             # NOTE: Check if the url is a redirect.
             url = requests.head(url, allow_redirects=True).url
-            u = parse.urlparse(url)
+            _ = parse.urlparse(url)
```
Panaetius, line 1465: I know I wrote this, but now I'm confused, does this line actually do anything other than raise an uncaught error if the redirect url is not a valid url? I feel like we could remove this line? Looks like a copy&paste error by me

renku/core/models/dataset.py
```diff
@@ -0,0 +1,765 @@
...
+    @classmethod
+    def from_database(cls, database: Database) -> "DatasetsProvenance":
+        """Return an instance from a metadata database."""
+        datasets = database["datasets"]
+        provenance = database["datasets-provenance"]
+
+        return DatasetsProvenance(datasets=datasets, provenance=provenance)
+
+    def get_by_id(self, id: str) -> Optional[Dataset]:
+        """Return a dataset by its name."""
```
Panaetius, line 569:
```suggestion
        """Return a dataset by its id."""
```

Panaetius (approved):

PR #2185, opened 2021-06-17
chore(service): default to never deleting job results

rokroskar:
Sets the TTL of RQ job results in the non-cron queues to forever. This way we don't lose information about e.g. completed dataset jobs. We could get around this by having a counter of some sort in redis, but for now this seems to do the trick. See [these metrics](https://grafana.dev.renku.ch/d/uu0muZHMz/redis-queue-metrics-dashboard?orgId=1&var-DS_PROMETHEUS=Prometheus&var-job=kubernetes-pods&var-namespace=rok&from=now-1h&to=now)

/deploy


rokroskar:
closing for now - instead of storing results forever we should see if we can modify the metrics exporter to report something like an instantaneous value that can get aggregated self-consistently.

PR #2187, opened 2021-06-18
fix(core): CWL to work with filenames with spaces

### LINKED ISSUE #2182: change CWL to work with filenames with spaces
We need to specify the `relax_path_checks` option (see https://github.com/common-workflow-language/cwltool/blob/main/cwltool/command_line_tool.py#L364) when running cwl so paths with spaces work properly.

vigsterkr:
# Description

Fixes #2182

### DIFF
--- a/renku/core/commands/cwl_runner.py
+++ b/renku/core/commands/cwl_runner.py
@@ -56,7 +56,7 @@ def execute(client, output_file, output_paths=None):
     runtime_context = RuntimeContext(
         kwargs={"rm_tmpdir": False, "move_outputs": "leave", "preserve_entire_environment": True}
     )
-    loading_context = LoadingContext(kwargs={"construct_tool_object": construct_tool_object})
+    loading_context = LoadingContext(kwargs={"construct_tool_object": construct_tool_object, "relax_path_checks": True})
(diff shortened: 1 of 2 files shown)


renku/core/commands/cwl_runner.py
```diff
@@ -56,7 +56,7 @@ def addLocation(d):
     runtime_context = RuntimeContext(
         kwargs={"rm_tmpdir": False, "move_outputs": "leave", "preserve_entire_environment": True}
     )
-    loading_context = LoadingContext(kwargs={"construct_tool_object": construct_tool_object})
+    loading_context = LoadingContext(kwargs={"construct_tool_object": construct_tool_object, "relax_path_checks": True})
```
vigsterkr, line 59: @Panaetius should this be actually hardcoded here, or rather make it default `True` in `LoadingContext`?

Panaetius: This way is fine. LoadingContext is not one of our classes anyways :slightly_smiling_face:

Panaetius (approved):
This is perfect, glad it actually was as easy as adding this flag :+1:

PR #2191, opened 2021-06-21
chore: Update swagger authorization docs

Panaetius:


rokroskar (dismissed):
thanks a lot!! 🥇

rokroskar:
As a side note - we should figure out where to put this in the main renku docs

rokroskar (approved):

PR #2203, opened 2021-06-30
chore: update renku action paths

rokroskar:
Renku github actions have been moved to https://github.com/SwissDataScienceCenter/renku-actions so we need to update the references. 

/deploy


Panaetius (approved):
Thank you!

PR #2204, opened 2021-07-01
feat(docs): new design of docs

vfried:
I am almost sure that this is what we need to apply the new sphinx template to the renku docs.

https://github.com/SwissDataScienceCenter/renku-sphinx-theme/pull/4

Could someone check? 😄  🙏


Panaetius (approved):
Looks good to me

PR #2205, opened 2021-07-01
feat(core): migration for new metadata

### LINKED ISSUE #2111: Make classes persistent for new storage
Change all our relevant classes (e.g. Activity, Dataset, Plan, Entity, but not Usage/Generation) to inherit from `persistent.Persistent`. 

Replace python `weakref` references with `persistent.WeakRef`.

Add `BTree`s to the `Storage` `root` object where our objects can be stored (by id). This should be initialized when creating a storage (In the CommandBuilder for storage) if it doesn't exist already.

### LINKED ISSUE #2114: Add migration for old metadata to Storage
Write a migration that loads the old jsonld metadata (and walks git history) and adds the objects to `Storage`.

This has to be done for datasets, activities, entities, plans (and probably Project?).
This should also move from old metadata format to new metadata format.

mohammad-sdsc:
# Description
Migration for converting old workflow metadata to the new metadata and creating datasets' provenance. It stores them in the new persistent storage. 

Fixes #2111 
Fixes #2114


renku/core/management/migrations/m_0009__new_metadata_storage.py
```diff
@@ -0,0 +1,295 @@
...
+            communication.warn(f"Cannot process commit '{commit.hexsha}' - Exception: {traceback.format_exc()}")
+
+        # NOTE: We need to commit in case we want to resume because we delete processed files
+        database.commit()
+
+
+def _process_workflows(
+    client: LocalClient, commit: Commit, database: Database, dependency_graph: DependencyGraph, remove: bool
+):
+    def replace_agents(activity):
```
Panaetius, line 105: A docstring describing why this function exists/what its use is would be nice. I don't think it's immediately apparent, especially to someone new.

renku/core/management/migrations/m_0009__new_metadata_storage.py
```diff
@@ -0,0 +1,295 @@
...
+        if not path.startswith(".renku/workflow") or not path.endswith(".yaml"):
+            continue
+
+        if not (client.path / path).exists():
+            communication.warn(f"Workflow file does not exists: '{path}'")
+            continue
+
+        workflow = Activity.from_yaml(path=path, client=client)
+
+        activity_collection = ActivityCollection.from_activity(workflow, dependency_graph)
```
Panaetius, line 138: Do we still need the class `ActivityCollection`? It seems that we could just replace it with a plain function (with mostly the same code as `from_activity` that returns a list.

renku/core/management/migrations/m_0009__new_metadata_storage.py
```diff
@@ -0,0 +1,295 @@
...
+            _process_datasets(client=client, commit=commit, datasets_provenance=datasets_provenance)
+        except errors.MigrationError:
+            communication.echo("")
+            communication.warn(f"Cannot process commit '{commit.hexsha}' - Migration failed: {traceback.format_exc()}")
+        except Exception:
+            communication.echo("")
+            communication.warn(f"Cannot process commit '{commit.hexsha}' - Exception: {traceback.format_exc()}")
+
+        # NOTE: We need to commit in case we want to resume because we delete processed files
+        database.commit()
```
Panaetius, line 99: Do we also need some code to delete `.renku/datasets/` once they're migrated if `remove` is set?

renku/core/models/provenance/activity.py
```diff
@@ -122,13 +131,15 @@ def __init__(
 
     @classmethod
     @inject.params(client="LocalClient")
-    def from_process_run(cls, process_run: ProcessRun, plan: Plan, client, order: Optional[int] = None):
+    def from_process_run(
+        cls, process_run: ProcessRun, plan: Plan, rerun_plan: Plan, client, order: Optional[int] = None
```
Panaetius, line 135: `rerun_plan` is not used

renku/core/models/workflow/dependency_graph.py
```diff
@@ -88,12 +88,12 @@ def add(self, plan: Plan) -> Plan:
 
     def _find_similar_plan(self, plan: Plan) -> Optional[Plan]:
         """Search for a similar plan and return it."""
-        for p in self._plans:
+        for p in self._plans.values():
```
Panaetius, line 91: `is_similar_to` would be a good candidate for a zc.relation catalog.

Panaetius (changes requested):

renku/core/incubation/graph.py
```diff
@@ -63,72 +60,19 @@
...
-    commits = list(client.repo.iter_commits(paths=[f"{client.workflow_path}/*.yaml", ".renku/datasets/*/*.yml"]))
-    n_commits = len(commits)
-    commits = reversed(commits)
-
-    if force:
-        client.remove_graph_files()
-    elif client.has_graph_files():
+    try:
+        generate_new_metadata(client=client, force=force, remove=False)
+    except errors.OperationError:
```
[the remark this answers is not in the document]
mohammad-sdsc: We will remove `renku graph generate` in future and always force-generate so this won't be an issue. The verbose message does not make sense in `generate_new_metadata` since users don't interact with it directly and `--force` won't have a meaning there.

renku/core/management/migrations/m_0009__new_metadata_storage.py
mohammad-sdsc: This should be done when we are done removing old datasets' metadata from the project (#2206).

renku/core/management/migrations/m_0009__new_metadata_storage.py
mohammad-sdsc: We still rely on it to update the new metadata when a `run/update/rerun` command is executed. I guess once we move to the new workflow metadata and use them directly for execution we can remove `ActivityCollection` and also remove it from here. We might want to keep it, however, to store `GroupedRun` executions.

Panaetius (dismissed):

Panaetius (approved):

PR #2216, opened 2021-07-09
chore: remove release action

Panaetius:
removes the release action as we create releases manually


vigsterkr (approved):
🚀

mohammad-sdsc:
Does it make sense to remove Python `3.7` env from tests after merging to master? Tests already passed before merging to master for that version and there should be no need to rerun the same test on the same exact code:
```
echo "::set-output name=matrix::{\"python-version\": [\"3.6\", \"3.7\", \"3.8\"]}"
```

mohammad-sdsc (approved):
Thank you!

PR #2217, opened 2021-07-09
feat(core): Implement workflow list/edit/export commands

vigsterkr:
# Description

Implement the workflow commands using the new storage backend and based on the new workflow rfc specs


renku/core/commands/workflow.py
```diff
@@ -118,3 +128,70 @@ def _create_workflow(output_file, revision, paths, client: LocalClient):
...
+    if description:
+        workflow.description = description
+
+    for name, value in set_params.items():
+        for param in workflow.parameters:
+            if param.name == name:
+                param.default_value = value
+                break
+        else:
+            # TODO: what's the position
```
vigsterkr, line 172: @Panaetius based on the RFC specs `--set` adds a new parameter in case if it doesn't exist. should the optional `position` argument for `CommandParameter` be None or rather set it to be the last parameter, based on the current parameters in the parameters list?

Panaetius: `position` should be `None` if it's not in the parameter list for the shell command.

Panaetius:
Could you change the title of the PR? since these show up in the changelog and not all new workflow commands are implemented yet, it might be misleading. Something like "feat(core): add renku workflow list/remove/edit commands using new metadata" maybe?

renku/cli/workflow.py
```diff
@@ -151,25 +162,82 @@ def rename(old, new, force):
...
+@click.option(
+    "--map", "map_params", metavar="<parameter>=<parameter or expression>", help="New description of the workflow"
+)
+@click.option(
+    "--rename-param",
+    "rename_params",
+    multiple=True,
+    metavar='<parameter>="name"',
+    callback=_kv_to_dict,
+    help="New description of the workflow",
```
Panaetius, line 196:
```suggestion
    help="New name for parameter",
```

renku/cli/workflow.py
```diff
@@ -151,25 +162,82 @@ def rename(old, new, force):
...
+@click.option(
+    "--set",
+    "set_params",
+    multiple=True,
+    metavar="<parameter>=<value>",
+    callback=_kv_to_dict,
+    help="Set default <value> for a <parameter>/add new parameter",
+)
+@click.option(
+    "--map", "map_params", metavar="<parameter>=<parameter or expression>", help="New description of the workflow"
```
Panaetius, line 188:
```suggestion
    "--map", "map_params", metavar="<parameter>=<parameter or expression>", help="New mapping on the workflow"
```

renku/core/commands/workflow.py
```diff
@@ -41,23 +44,19 @@ def _deref(ref):
...
+def _list_workflows(database: Database, format: str, columns: List[str]):
     """List or manage workflows with subcommands."""
-    from renku.core.models.refs import LinkReference
+    workflows = unique_workflow(database)
 
-    names = defaultdict(list)
-    for ref in LinkReference.iter_items(common_path="workflows"):
-        names[ref.reference.name].append(ref.name)
+    if format not in WORKFLOW_FORMATS:
+        raise errors.UsageError(f'Provided format "{format}" is not supported.')
```
Panaetius, line 52: maybe print valid formats in the error as well so a user knows what to change.

renku/core/commands/workflow.py
```diff
@@ -87,34 +86,123 @@ def rename_workflow_command():
...
 
 @inject.autoparams()
-def _remove_workflow(name, client: LocalClient):
-    """Remove the remote named <name>."""
-    from renku.core.models.refs import LinkReference
+def _remove_workflow(name, force, database: Database):
+    """Remove the workflow named <name>."""
+    now = datetime.utcnow()
+    workflows = unique_workflow(database)
+    not_found_text = f'The specified workflow is "{name}" is not an active workflow.'
```
Panaetius, line 93: Should be declare closer to the exception that is raised (on the line above it or in the `ParameterError(...)` itself

renku/cli/workflow.py
```diff
@@ -151,25 +162,82 @@ def rename(old, new, force):
...
+        set_params=set_params,
+        map_params=set_params,
+        rename_params=rename_params,
+        describe_params=describe_params,
+    )
+
+
+@workflow.command()
+@click.argument("workflow_name", metavar="<name or uuid>")
+@click.option(
```
Panaetius, line 221: I know it's not in the spec, an oversight on my part, but we probably want to allow a user to specify a values file (like `--values values.yml`) as in the `renku workflow execute` spec, so they can export with value overrides. You should be able to just use the method I wrote in the other PR `apply_run_values` and pass it as a dict after loading from yaml.

Panaetius (changes requested):
Starting to look really nice :)

renku/core/commands/view_model/composite_plan.py
```diff
@@ -71,13 +71,15 @@ def __init__(
...
         name: str,
+        derived_from: str,
         mappings: List[ParameterMappingViewModel],
         links: List[ParameterLinkViewModel],
         steps: List[StepViewModel],
         description: Optional[str] = None,
     ):
         self.id = id
         self.name = name
+        self.derived_from = derived_from
```
Panaetius, line 82: Is this used in `renku.cli`? I don't see a place where it is displayed. Since we currently treat immutability as a behind-the-scenes thing, it probably shouldn't be on the viewmodel.

renku/core/commands/workflow.py
```diff
@@ -48,98 +49,69 @@ def _deref(ref):
...
-    graph = Graph()
-    outputs = graph.build(paths=paths, revision=revision)
-
-    workflow = graph.as_workflow(outputs=outputs)
-
-    if output_file:
-        output_file = Path(output_file)
+def _remove_workflow(name, force, database: Database):
+    """Remove the workflow named <name>."""
+    workflows = unique_workflow(database)
```
Panaetius, line 89: unique_workflow won't be needed since the Gateways take care of it.

but the Gateway probably needs to be extended to deal with deleted workflows.

renku/core/commands/workflow.py
```diff
@@ -236,3 +208,108 @@ def compose_workflow_command():
...
+        else:
+            workflow.parameters.append(
+                CommandParameter(default_value=value, id=CommandParameter.generate_id(plan_id=workflow.id), name=name)
+            )
+
+    if len(map_params) and isinstance(workflow, CompositePlan):
+        workflow.set_mappings_from_strings(map_params)
+
+    for name, new_name in rename_params.items():
+        for param in workflow.parameters:
```
Panaetius, line 261: We should probably check if a user tries to rename a param that doesn't exist and raise an error

renku/core/commands/workflow.py
```diff
@@ -236,3 +208,108 @@ def compose_workflow_command():
...
+            if param.name == name:
+                param.name = new_name
+                break
+
+    for name, descirption in describe_params.items():
+        for param in workflow.parameters:
+            if param.name == name:
```
Panaetius, line 268: Also raise an error if a param doesn't exist

renku/core/commands/workflow.py
```diff
@@ -236,3 +208,108 @@ def compose_workflow_command():
...
+                param.default_value = value
+                break
+        else:
+            workflow.parameters.append(
+                CommandParameter(default_value=value, id=CommandParameter.generate_id(plan_id=workflow.id), name=name)
+            )
+
+    if len(map_params) and isinstance(workflow, CompositePlan):
+        workflow.set_mappings_from_strings(map_params)
```
Panaetius, line 258: I think the design doc also mentions being able to delete a mapping by leaving the right hand side empty. I think this still needs to be implemented.

renku/core/models/workflow/converters/__init__.py
```diff
@@ -16,3 +16,28 @@
...
+
+from abc import ABCMeta, abstractmethod
+from pathlib import Path
+from typing import List, Optional, Tuple, Union
+
+from renku.core.models.workflow.composite_plan import CompositePlan
+from renku.core.models.workflow.plan import Plan
+
+
+class WorkflowConverter(metaclass=ABCMeta):
```
Panaetius, line 30: Not that we really discussed this convention (I jsut added it in my PR), but I think it makes sense to have interfaces start with I (`IWorkflowConverter`) so it's clear they're interfaces (Since python doesn't support proper interfaces).

renku/core/models/workflow/converters/cwl.py
```diff
@@ -18,15 +18,23 @@
 """Converter for workflows to cwl."""
 
 import os
+import re
```
Panaetius, line 21: We should move this file, maybe to somewhere in management, since it's not really a domain model but more business logic

renku/core/models/workflow/converters/cwl.py
```diff
@@ -18,15 +18,23 @@
...
 import tempfile
 from collections import defaultdict
 from pathlib import Path
+from typing import Optional, Union
 from uuid import uuid4
 
 import cwlgen
 
+from renku.core import errors
+from renku.core.management.workflow.concrete_execution_graph import ExecutionGraph
```
Panaetius, line 31: We should move this file, maybe to somewhere in management, since it's not really a domain model but more business logic.

Also so a model doesn't have a dependency on a management class (dependencies should go the other way)

renku/core/models/workflow/converters/cwl.py
```diff
@@ -425,3 +433,356 @@ def _convert_argument(argument):
...
+            return (
+                cwlgen.CommandOutputParameter(
+                    "output_{}".format(output.mapped_to.stream_type),
+                    param_type=output.mapped_to.stream_type,
+                    streamable=False,
+                ),
+                None,
+            )
+
+        # type_ = "Directory" if isinstance(entity, Collection) else "File"
```
Panaetius, line 753: Is this needed or what is the state of this line?

Panaetius (changes requested):
This looks really nice, glad to see things progressing. 

You'll probably have to deal with quite a few merge conflicts, let me know if I can help.

vigsterkr:
@Panaetius just rebased over latest master and refactored it. i'll address now your comments. i'll ping u when things are done. thnx heaps for the review!

renku/core/commands/workflow.py
```diff
@@ -45,100 +49,69 @@ def _deref(ref):
...
-    for path in client.workflow_path.glob("*.yaml"):
-        communication.echo(
-            "{path}: {names}".format(path=path.name, names=", ".join(_deref(name) for name in names[path.name]))
-        )
-
-
-def list_workflows_command():
-    """Command to list or manage workflows with subcommands."""
-    return Command().command(_list_workflows).require_migration().with_database()
+def _to_view(workflow: AbstractPlan) -> Union[CompositePlanViewModel, PlanViewModel]:
```
vigsterkr, line 52: @Panaetius should we move this utility function to somewhere else?

renku/core/commands/workflow.py
```diff
@@ -45,100 +49,69 @@ def _deref(ref):
...
+    return PlanViewModel.from_plan(workflow)
 
 
 @inject.autoparams()
-def _set_workflow_name(name, path, force, client: LocalClient):
-    """Sets the <name> for remote <path>."""
-    from renku.core.models.refs import LinkReference
-
-    LinkReference.create(client=client, name=_ref(name), force=force).set_reference(path)
+def _find_workflow(name_or_id: str, plan_gateway: IPlanGateway) -> AbstractPlan:
```
vigsterkr, line 59: @Panaetius same as above. one could argue that this is not the best place to define this utility function

renku/core/commands/workflow.py
Panaetius: maybe next to the view models?

renku/core/errors.py
```diff
@@ -505,6 +505,14 @@ def __init__(self, existing_mappings: List[str]):
         )
 
 
+class MappingNotFoundError(RenkuException):
+    """Raised when a parameter mapping exists already."""
```
Panaetius, line 509: one of these two lines is wrong :slightly_smiling_face:

Panaetius (changes requested):
Almost ready to merge, just a copy&paste error :)

vigsterkr:
@Panaetius 🚀 IMO (finally!) ready for merge

Panaetius (approved):

PR #2225, opened 2021-07-21
chore: change tests to output stacktrace on failed renku runner.invok…

Panaetius:
replaces all 
``` 
result = runner.invoke(...)
assert result.exit_code[, result.output]
```

occurences in tests with 
```
result = runner.invoke(...)
assert result.exit_code, format_result_exception(result)
```

with outputs a nicely formatted stack trace of whatever exception happened along with the output of the command


lorenzo-cavazzi (approved):
Great to see the whole stack trace printed when a test fails, it simplifies identifying errors on the fly :+1: 
Spoiler alert: it may become _a little_ verbose when too many fail at once :grin:

vigsterkr (approved):
🚀

PR #2228, opened 2021-07-25
refactor(core): split requests schemas

### LINKED ISSUE #2224: Split `RepositoryContext` into individual schemas by topic/concern
currently on the service, requests are split into 3 parts, a Request schema, a Response schema and a Control that executes the request.

For instance the `cache.migrate` endpoint here https://github.com/SwissDataScienceCenter/renku-python/blob/master/renku/service/views/cache.py#L156 uses the `MigrateProjectCtrl` controller  which has `ProjectMigrateRequest` ( https://github.com/SwissDataScienceCenter/renku-python/blob/master/renku/service/serializers/cache.py#L213 ) as its Request serializer.

A lot of these serializers (in https://github.com/SwissDataScienceCenter/renku-python/tree/master/renku/service/serializers ) are based on `RepositoryContext` (in https://github.com/SwissDataScienceCenter/renku-python/blob/master/renku/service/serializers/common.py#L30 ), which is shared by almost all requests. 

There are some issues with his, for instance this context has a field `migrate_project` which tells a controller to first migrate the project before executing the action. But on the `cache.migrate` endpoint, this would mean "migrate the project before you migrate the project", which doesn't make sense.

On the other hand, there's e.g. https://github.com/SwissDataScienceCenter/renku-python/blob/94a781b006308229cb5f5447a3a72dd7db58ab14/renku/service/serializers/cache.py#L102 which has some of the same fields as `RepositoryContext`, so it's duplicated there.

We should split the schemas up so that each schema has a coherent concern ("Cloning a project to cache", "migrating the project on the fly", "setting the commit information for a write action", "executing an action async" etc.), which we can then compose as needed using multiple inheritance.

Ideally we also shouldn't have a deeply nested hierarchy of Schemas, and keep it as flat as possible, to not complicate things.

lorenzo-cavazzi:
I've tried to break up some of the big schemas used as request serializers for the core service APIs (e.g. `RepositoryContext`) and to find other similarities wherever possible.

I initially tried to go a bit deeper with the changes but I found some difficulties and I decided to stick to an easier solution with changes only to the serializers.
Let me know if this is in line with what is described in #2224 or if this PR should contain deeper changes.

P.S. There are some inline comments starting with `# ?` with ideas for possible improvements (like early validation on parameters). Let me know if you think they may be useful. I'll remove the comments before merging.

fix #2224


renku/service/serializers/cache.py
```diff
@@ -210,20 +207,30 @@ class ProjectListResponseRPC(JsonRPCResponse):
...
-class ProjectMigrateRequest(RepositoryContext):
+class ProjectMigrateRequest(AsyncSchema, CommitSchema, LocalRepositorySchema, RemoteRepositorySchema):
     """Request schema for project migrate."""
 
     force_template_update = fields.Boolean(default=False)
     skip_template_update = fields.Boolean(default=False)
     skip_docker_update = fields.Boolean(default=False)
     skip_migrations = fields.Boolean(default=False)
 
+    # ! temporary -- should be removed but there are tests to update
```
Panaetius, line 218: Did you update the tests?

renku/service/serializers/cache.py
```diff
@@ -210,20 +207,30 @@ class ProjectListResponseRPC(JsonRPCResponse):
...
+    migrate_project = fields.Boolean(default=False, missing=False)
+
     @pre_load()
     def default_commit_message(self, data, **kwargs):
         """Set default commit message."""
+        # ? Should we move this to MigrateProjectCtrl?
```
Panaetius, line 224: I think this makes sense, I guess you'll do it in your new PR for commit messages.

renku/service/serializers/common.py
```diff
@@ -16,31 +16,107 @@
...
+    project_id = fields.String(description="Reference to access the project in the local cache.")
+
+
+class RemoteRepositorySchema(Schema):
+    """Schema for tracking a remote repository."""
+
+    git_url = fields.String(description="Remote git repository url.")
+    branch = fields.String(description="Remote git branch.")
+
+    # ? Add early validation?
```
Panaetius, line 40: Adding this makes sense

renku/service/serializers/common.py
```diff
@@ -16,31 +16,107 @@
...
+    )
 
-    git_url = fields.String()
-    project_id = fields.String()
 
-    ref = fields.String()
-    commit_message = fields.String()
-    client_extras = fields.String()
+class ArchiveSchema(Schema):
+    """Schema for adding a commit at the end of the operation."""
```
Panaetius, line 75: Copy-paste error?

renku/service/serializers/datasets.py
```diff
@@ -16,24 +16,51 @@
...
+    LocalRepositorySchema,
+    MigrateSchema,
+    RemoteRepositorySchema,
+    RenkuSyncSchema,
+)
 from renku.service.serializers.rpc import JsonRPCResponse
 
 
+class DatasetRefSchema(Schema):
+    """Schema for dataset name."""
```
Panaetius, line 38: Copy paste error

Panaetius (changes requested):
Looks great already, thank you. I replied to some of the code comments and there were some minor issues

renku/service/serializers/cache.py
lorenzo-cavazzi: There is this one I'm not sure how to fix https://github.com/SwissDataScienceCenter/renku-python/blob/master/tests/service/jobs/test_project.py#L26-L51
Removing the `migrate_project` makes it fail in the `delayed_ctrl_job` function. I can't understand the goal of the test (the description is probably a wrong copy/paste from another test).
Is it testing the behavior of `"migrate_project": True` and should I use another serializer then `ProjectMigrateRequest` ?

Panaetius (approved):
Thank you!

Panaetius (approved):

PR #2229, opened 2021-07-26
test: align maximum repetitions to 5

lorenzo-cavazzi:
Tests are re-run multiple times when they fail thanks to the `@flaky` fixture.
The number is currently high for some tests (up to 30 repetitions) but that doesn't seem to provide an added value.

This PR aligns all of them to a maximum of 5 repetitions.


lorenzo-cavazzi:
The last commits introduce a `@retry_failed` decorator to be used in place of `@flaky(...)`.
That should simplify changing common `flaky` settings at once and prevent using different numbers.

The standard is still 5 repetitions, with the possibility to increase it to 20 using the `extended` argument of the decorator -- currently used only for `test_config_write_concurrency`.

tests/utils.py
```diff
@@ -149,3 +151,25 @@ def with_dataset(client, name=None, commit_database=False):
...
+        @flaky(max_runs=limit, min_passes=1)
+        @wraps(fn)
+        def wrapper(*args, **kwargs):
+            return fn(*args, **kwargs)
+
+        return wrapper
+
+    if fn:
+        return decorate(fn)
+    return decorate
```
Panaetius, line 175: I don't think we need 3 level nesting here
```
@flaky(max_runs=5, min_passes=1)
def my_function(...):
    [...]
```

is just syntactic sugar for 
```
my_function = flaky(max_runs=5, min_passes=1)(my_function)
```

so something like
```
def retry_failed(fn=None, extended: bool = False):
    """
    Decorator to run flaky with same number of max and min repetitions across all tests.
    :param extended: allow more repetitions than usual.
    """
    @wraps(func)
    def wrapper(*args, **kwargs):
        func = flaky(max_runs=5, min_passes=1)(func)
        return func(*args, **kwargs)
    return wrapper
```

Should do the trick

lorenzo-cavazzi: I think that's doesn't properly handle passing variables to the decorator.
With a solution like that, `@retry_failed` works but `@retry_failed(extended=True)` doesn't.

tests/utils.py
```diff
@@ -149,3 +151,25 @@ def with_dataset(client, name=None, commit_database=False):
     if commit_database:
         client.get_database().register(dataset)
         client.get_database().commit()
+
+
+def retry_failed(fn=None, extended: bool = False):
```
mohammad-sdsc, line 156: Should we replace `extended` with `max_runs=5`? This would allow us to have more than two possible values if needed in future.

lorenzo-cavazzi: The original motivation for switching to this fixture was the possibility to handle the number of repetitions in a single place (e.g. switching to 1 rep when testing locally).
I would start with 2 categories (for standard vs occasionally failing tests) without the possibility to set a specific number of repetitions, and make this more flexible later if we need it.

A simpler solution would be using the original decorator and create constants to pass to `flaky(max_runs)`

Panaetius (approved):

PR #2232, opened 2021-07-27
feat(dataset): refactor DatasetTag

### LINKED ISSUE #2210: Improve handling of dataset tags
Currently, `DatasetTag` is stored as part of the dataset's metadata and it uses commit sha to refer to an exact dataset version.
We are moving away from using commit shas and we have a metadata database that can return objects by their ids. We should change this class to use datasets' ids instead of commit shas. We should also store it in the metadata database as an independent entity (and not as part of Dataset's metadata).

**Acceptance Criteria**
* Use dataset ids when tagging a dataset version
* Store `DatasetTag` as a Persistent object
* A migration to convert old `DatasetTag`s
* Make sure all tag-related commands work (e.g. `tag`, `ls-tags`, `export`, ...)

mohammad-sdsc:
# Description

Refactor DatasetTag to be an independent entity that points to a dataset identifier instead of a commit.

Fixes #2210


renku/core/management/datasets.py
```diff
@@ -835,28 +835,15 @@ def _fetch_lfs_files(repo_path, paths):
...
-        #
-        # commits = self.repo.iter_commits(paths=paths, max_count=max_results)
-        #
-        # return commits
-        return [self.repo.head.commit]
-
-    def add_dataset_tag(self, dataset, tag, description="", force=False):
+    @staticmethod
+    @inject.autoparams()
+    def add_dataset_tag(
```
Panaetius, line 840: In the spirit of having less functionality on LocalClient, moving this and the remove_dataset_tag code to a dedicated file in `renku/core/management/dataset` would be nice.

renku/core/commands/dataset.py
```diff
@@ -799,31 +786,15 @@ def _filter(
...
-    force=False,
-):
-    try:
-        client.add_dataset_tag(dataset, tag, description, force)
-    except ValueError as e:
-        raise ParameterError(e)
-    else:
-        if update_provenance:
-            datasets_provenance.add_or_update(dataset)
+def _tag_dataset_helper(dataset, tag, description, client: LocalClient, force=False):
```
Panaetius, line 796: We probably don't need this helper if it just forwards the call

renku/core/commands/dataset.py
```diff
@@ -833,16 +804,10 @@ def tag_dataset():
 
 
 @inject.autoparams()
-def _remove_dataset_tags(name, tags, client: LocalClient, datasets_provenance: DatasetsProvenance):
+def _remove_dataset_tags(name, tags, client: LocalClient):
```
Panaetius, line 807: Also don't need this if we have a dedicated business logic class that does this.

renku/core/commands/providers/dataverse.py
```diff
@@ -424,7 +425,8 @@ def export(self, publish, client=None, **kwargs):
                     path = (client.path / file.entity.path).relative_to(self.dataset.data_dir)
                 except ValueError:
                     path = Path(file.entity.path)
-                deposition.upload_file(full_path=client.path / file.entity.path, path_in_dataset=path)
+                filepath = get_content(repo=client.repo, path=file.entity.path, checksum=file.entity.checksum)
```
Panaetius, line 428: nice!

renku/core/utils/git.py
```diff
@@ -136,6 +138,43 @@ def get_object_hash_from_submodules() -> Optional[str]:
...
+                    )
+                except subprocess.CalledProcessError:
+                    pass
+                else:
+                    return True
+
+    with tempfile.NamedTemporaryFile(mode="w+b", delete=False) as temp_file:
+        try:
+            subprocess.run(
+                ["git", "cat-file", "--filters", "--path", str(path), checksum],
```
Panaetius, line 166: Do we need `-e` flag here? 
```
Exit with zero status if <object> exists and is a valid object. If <object> is of an invalid format exit with non-zero and emits an error on stderr.
```

Also what does this do with LFS files? And how does it handle symlinks?

Panaetius (changes requested):
Good work. Will have quite a few conflicts with my finalize metadata PR, due to changes to DatasetsProvenance object

renku/core/management/datasets.py
mohammad-sdsc: I moved these to `renku/core/management/dataset/tag`. I believe our namespaces are getting too long; maybe we can drop the `management` part later.

renku/core/commands/dataset.py
mohammad-sdsc: This is still checking if the dataset exists and raises an error. It's also in par with other commands where we have a private method to do some validations and then forward a call to core business logic. So, I'll keep it for now.

renku/core/utils/git.py
mohammad-sdsc: `-e` only checks if the object exits and won't print its content. `--filters` takes care of LFS and complains if object is invalid.
This won't work with symlinks and I couldn't find a workaround for it. However, we don't have symlinks in the project except for external files. AFAIK, we don't have a clear policy on how to handle external files on export/import. So, I'll create a story for this.

renku/core/management/datasets.py
Panaetius: I agree. Though maybe we could also move things up a level and "remove" core, that'd help keeping separate things separate in regards to dependencies.

Also `management` isn't a nice name for it. Clean Architecture calls it `use_cases` but I don't like that, either. At least as a developer I'd think it contains documents and diagrams if I saw a folder with that name. `business_logic` is too long. Maybe `controllers`

Panaetius (dismissed):

Panaetius (approved):

PR #2234, opened 2021-07-29
feat(service): align commit messages made by the service

### LINKED ISSUE #2152: Improve commit messagges from the core service
While using the new `config` APIs, I've noticed the commit messages created by the core service are a bit cryptic. Specifically, I see:
```
gunicorn renku.service.entrypoint:app --preload -b 0.0.0.0:8080 --timeout 600 --workers 4 --worke...
```
For a user, it may be unclear what happened.
It would be nice to have something similar to other APIs, where the message includes the cli equivalent command, like `templates.create_project`
```
service: renku init -n "project-config" -s "https://github.com/SwissDataScienceCenter/renku-proje...
```

**Proposal**
Having a standard for the commit messages would be great. For the APIs, it could be something like:
* First line: scope (E.G. `renku service - update config`, `renku service - create project`)
* Second and following lines: cli equivalent command, or other relevant information to re-run the command where the command is not meaningful (E.G. `renku init -n "project-config" -s "https://github.com [...]"`, `renku config set [...]`). Do not cut the renku command, or do that only when the number of chars is very high -- currently, it's a bit low

**Update**
Scope of the issue:
- [ ] Align the commit messages coming from the service. Ideally, preserve the CLI equivalent message with a prefix to indicate it's coming from the service. This may not be possible for some commands not having a 1:1 CLI equivalent.
Info: generally keep the possibility to provide a custom `commit_message` wherever reasonable, but prevent setting a custom message early (e.g. https://github.com/SwissDataScienceCenter/renku-python/blob/master/renku/service/serializers/datasets.py#L40-L43)
- [x] Evaluate the possibility to extend the commit messages to prevent cutting out important information.
Info: the length is cut after 100 chars https://github.com/SwissDataScienceCenter/renku-python/blob/master/renku/core/utils/scm.py#L54-L75
Switching to `cut=False` to have multi-line messages could be a solution, but we may need another "message body hard limit" variable to prevent some commands from failing due to the excessive length (e.g. `renku run` or `renku dataset add *`). Picking a reasonable default length may be hard, being system-dependent.
P.S. users may be happy to see again the CLI command in the commit: https://renku.discourse.group/t/how-to-see-the-full-command-again/262 . Answer there in case of any significant change.

lorenzo-cavazzi:
_INFO: this requires #2239 to be merged first. Otherwise, most dataset APIs fail._

This PR fixes the first part of #2152 by aligning the commit messages created by the service (see screenshot below).
If we want, we could further extend the messages to include more details since the second part of #2152 has already been addressed -- though I'm not sure that adds much.

I initially decided to store the commit messages in `self.ctx["commit_message"]`, but since the APIs don't expose that field anymore, I'm not sure if it would make sense to move it to `self.commit_message` instead.

In [this test project](https://dev.renku.ch/gitlab/lorenzo.cavazzi.tech/test-renku-python-2234-rebased/-/commits/master) I triggered a few APIs from the UI to show the commits (ignore the `flask run` tests). Once #2239 is merged, it should be testable from the PR deployment https://renku-ci-rp-2234.dev.renku.ch

Screenshot_20210809_175331

/deploy
fix #2152


renku/service/controllers/cache_migrate_project.py
```diff
@@ -71,7 +72,7 @@ def __init__(self, cache, user_data, request_data, migrate_project=False, skip_l
         self.skip_template_update = self.ctx.get("skip_template_update", False)
         self.skip_docker_update = self.ctx.get("skip_docker_update", False)
         self.skip_migrations = self.ctx.get("skip_migrations", False)
-        self.commit_message = self.ctx.get("commit_message", None)
+        self.commit_message = "{0} migrate to latest version for renku {1}".format(MESSAGE_PREFIX, __version__)
```
Panaetius, line 75: use an f-string

renku/service/controllers/config_set.py
```diff
@@ -33,6 +34,10 @@ class SetConfigCtrl(ServiceCtrl, RenkuOpSyncMixin):
     def __init__(self, cache, user_data, request_data, migrate_project=False):
         """Construct controller."""
         self.ctx = SetConfigCtrl.REQUEST_SERIALIZER.load(request_data)
+        self.ctx["commit_message"] = "{0} config set {1}".format(
```
Panaetius, line 37: Also f-string (the ", ".join(...) will have to be done in the line above)

renku/service/controllers/datasets_add_file.py
```diff
@@ -41,9 +42,7 @@ class DatasetsAddFileCtrl(ServiceCtrl, RenkuOpSyncMixin):
     def __init__(self, cache, user_data, request_data, migrate_project=False):
         """Construct a datasets add controller."""
         self.ctx = DatasetsAddFileCtrl.REQUEST_SERIALIZER.load(request_data)
-
-        if not self.ctx["commit_message"]:
-            self.ctx["commit_message"] = "service: dataset add {0}".format(self.ctx["name"])
+        self.ctx["commit_message"] = "{0} dataset add {1}".format(MESSAGE_PREFIX, self.ctx["name"])
```
Panaetius, line 45: f-string

renku/service/controllers/datasets_create.py
```diff
@@ -34,9 +34,7 @@ class DatasetsCreateCtrl(ServiceCtrl, RenkuOpSyncMixin):
     def __init__(self, cache, user_data, request_data, migrate_project=False):
         """Construct a datasets create controller."""
         self.ctx = DatasetsCreateCtrl.REQUEST_SERIALIZER.load(request_data)
-
-        if self.ctx.get("commit_message") is None:
-            self.ctx["commit_message"] = "service: dataset create {0}".format(self.ctx["name"])
+        self.ctx["commit_message"] = "{0} dataset create {1}".format(MESSAGE_PREFIX, self.ctx["name"])
```
Panaetius, line 37: f-string

renku/service/controllers/datasets_edit.py
```diff
@@ -34,9 +34,7 @@ class DatasetsEditCtrl(ServiceCtrl, RenkuOpSyncMixin):
     def __init__(self, cache, user_data, request_data, migrate_project=False):
         """Construct a datasets edit list controller."""
         self.ctx = DatasetsEditCtrl.REQUEST_SERIALIZER.load(request_data)
-
-        if self.ctx.get("commit_message") is None:
-            self.ctx["commit_message"] = "service: dataset edit {0}".format(self.ctx["name"])
+        self.ctx["commit_message"] = "{0} dataset edit {1}".format(MESSAGE_PREFIX, self.ctx["name"])
```
Panaetius, line 37: f-string

renku/service/controllers/datasets_import.py
```diff
@@ -36,9 +37,7 @@ class DatasetsImportCtrl(ServiceCtrl, RenkuOpSyncMixin):
     def __init__(self, cache, user_data, request_data, migrate_project=False):
         """Construct a datasets import controller."""
         self.ctx = DatasetsImportCtrl.REQUEST_SERIALIZER.load(request_data)
-
-        if self.ctx.get("commit_message") is None:
-            self.ctx["commit_message"] = "service: dataset import of {0}".format(self.ctx["dataset_uri"])
+        self.ctx["commit_message"] = "{0} dataset import of {1}".format(MESSAGE_PREFIX, self.ctx["dataset_uri"])
```
Panaetius, line 40: f-string

Panaetius (changes requested):
Looks great, thank you!

We decided a while back to switch to f-string instead of the `{0}` style syntax and we're changing it to that whenever we encounter an old style format, so there's a few places this needs to be changed. I marked some of them but there's lots :sweat_smile: 

Otherwise ready to merge

lorenzo-cavazzi:
Ouch, sorry for that.
I changed to f-strings wherever I used the `.format` notation.

Panaetius:
by the way, "improvement" isn't a tag that works in PR headers, I think. The valid ones are listed in https://github.com/SwissDataScienceCenter/renku-python/blob/master/CONTRIBUTING.rst#commit-message-guidelines (though having improvement would be nice, I think).

Not sure if the changelog generator would swallow it.

rokroskar:
I think this falls under "chore"... :)

lorenzo-cavazzi:
Oh ok, I'll stick to those listed in the readme. I'm using a plugin based on Commitizen and it lists a few others like "ci", "build", "improvement".

Not sure this is a `chore` since the change is user-facing. Shouldn't it be a `feat`? The linked docs say it's either `fix` or `feat` if the commit will end up in a changelog.

Panaetius:
feat makes sense.

Ready to merge in my view, except for the failing integration test

lorenzo-cavazzi:
I fixed the tests

PR #2238, opened 2021-08-03
improvement: allow longer commit messages

lorenzo-cavazzi:
### Changes

This PR extends the commit messages length. By default, they are now wrapped and trimmed.

Old message:
```
Author: Lorenzo Cavazzi <lorenzo.cavazzi.tech@gmail.com>
Date:   Mon Aug 2 17:32:07 2021 +0200

    renku init --name my test project --template-source=https://github.com/SwissDataScienceCenter/ren...
```
New message:
```
Author: Lorenzo Cavazzi <lorenzo.cavazzi.tech@gmail.com>
Date:   Tue Aug 3 09:18:50 2021 +0200

    renku init --name my test project
            --template-source=https://github.com/SwissDataScienceCenter/renku-project-template
            --template-id=python-minimal --parameter description=Lorem ipsum dolor sit amet, consectetur
            adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim
            veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. --force

```

### Info

It's hard to find reliable information about the maximum size of commit messages without being flooded by `72 vs 80 vs 100 line length` discussions.
Generally speaking, git documentation doesn't mention any limit. Doing empiric tests show that the limit is very high on most systems, enough to not care about the actual limit.

Digging a bit in the code, the commit message `len` field is a `size_t`, and the upper limit is guaranteed to be an unsigned integer type of at least 16 bit.
We could set a very conservative cut-off limit to 65535 (or round it to 65k). In practice, it's much higher on basically any system -- (almost) for sure on any system that would run renku. Otoh, I'm not sure anyone would dig into commit messages longer than 65k chars, so we are not losing much in adding that limit.

P.S. Here are 2 questions on StackOverflow containing useful references for the above statements.
https://stackoverflow.com/questions/9733757
https://stackoverflow.com/questions/2550774

re #2152


renku/core/utils/scm.py
```diff
@@ -51,25 +51,35 @@ def git_unicode_unescape(s: str, encoding: str = "utf-8") -> str:
...
+    :param body_length: maximum body length before cut. 0 for infinite.
+    :return: message wrapped and trimmed.
+    """
+
+    if body_length and len(message) > body_length:
+        message = message[: body_length - 3] + "..."
 
-    if cut:
-        return message[: max_length - 3] + "..."
+    if not line_length or len(message) <= line_length:
```
Panaetius, line 66: if `line_length` is allowed to be None, the type above should be `Optional[int]`

tests/core/utils/test_utils.py
```diff
@@ -34,3 +35,36 @@ def test_hostname():
...
+    # The message is trimmed when it's longer than the max body length
+    assert len(shorten_message(long_message, 0, body_length=max_line)) < len(long_message)
+    assert len(shorten_message(short_message, 0, body_length=max_line)) == len(short_message)
+    assert len(shorten_message(long_message, 0, body_length=0)) == len(long_message)
+    assert len(shorten_message(long_message, 0, body_length=len(long_message))) == len(long_message)
+
+    # The the message is wrapped correctly
+    assert len(shorten_message(long_message, max_line, body_length=max_line * 2)) < len(long_message)
+    assert len(shorten_message(short_message, max_line, body_length=max_line * 2)) == len(short_message)
+    assert shorten_message(long_message, max_line, body_length=max_line * 2).count("\n") == 1
```
Panaetius, line 70: Should we add a test where `shorten_message(msg, ...).replace('\n', ' ')[:body_length - 3] == msg[:body_length - 3]` (or something similar) that makes sure the content of the message is preserved when wrapping it to multiple lines?

Panaetius (changes requested):
This looks really nice, thank you! I have two suggestions but nothing related to functionality.

renku/core/utils/scm.py
lorenzo-cavazzi: I think `line_length != 0` (or `<1`) and keeping the `[int]` type may be better in this case.

lorenzo-cavazzi:
I added a couple of extra tests, and now the function also throws an error in case of negative length -- since I'm already validating the types of the parameters, I guess it makes sense to check for nonsense values.

Panaetius (approved):
👍

PR #2243, opened 2021-08-05
fix(service): handle uploaded images in datasets.edit

### LINKED ISSUE #2240: datasets.edit endpoint doesn't support uploaded images
on `datasets.create` we have https://github.com/SwissDataScienceCenter/renku-python/blob/master/renku/service/controllers/datasets_create.py#L48 but this piece of code is missing on dataset edit, so editing a dataset with uploaded images doesn't work.

We should also have the `_handle_uploaded_images` logic on the `DatasetsEditCtrl`

mohammad-sdsc:
Fixes #2240

### DIFF
--- a/renku/service/controllers/datasets_edit.py
+++ b/renku/service/controllers/datasets_edit.py
@@ -45,8 +45,16 @@ class DatasetsEditCtrl(ServiceCtrl, RenkuOpSyncMixin):
         """Controller operation context."""
         return self.ctx
 
+    def _handle_uploaded_images(self):
+        """Handles uploaded or relative dataset images."""
+        for img in self.ctx.get("images", []):
+            if img.get("file_id"):
+                file = self.cache.get_file(self.user, img.pop("file_id"))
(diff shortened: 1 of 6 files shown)


renku/service/controllers/datasets_edit.py
```diff
@@ -45,8 +45,16 @@ def context(self):
         """Controller operation context."""
         return self.ctx
 
+    def _handle_uploaded_images(self):
```
Panaetius, line 48: should we move this method to `renku.service.utils` somewhere and call it like `handle_uploaded_images(self.ctx.get("images", []), self.cache, self.user)` in the controls, so there's no duplication?

It not a lot of lines of code, but still a bit cleaner

Panaetius (changes requested):
Nicely done with the new fixture, much cleaner.

Panaetius (approved):

Panaetius (approved):

PR #2245, opened 2021-08-06
feat(core): fix autocommit LFS files in pre-commit hook

### LINKED ISSUE #2230: bug: Git LFS autocommit fails
**Describe the bug**
With `autocommit_lfs=true`, I pushed a few large files (>100kb) but those files were not correctly pushed to Git LFS.

**Link to project**
https://renkulab.io/projects/tao.sun/test-renku-lfs

**To Reproduce**
1. `renku config set renku.autocommit_lfs true; git push`
2. create a few large files, i.e. `large1.txt`, `large2.txt`, `large3.txt`
3. `git add -A; git commit -m "Add large files"; git push`
4. `cat .gitattributes` - it was not correctly generated
5. `git lfs ls-files` - no files were added to Git LFS

**Expected behavior**
Large files above the threshold should be pushed to Git LFS.

**Execution output**
```bash
$ cat .gitattributes
large1.txt
large2.txt
large3.txt filter=lfs diff=lfs merge=lfs -text

$ git lfs ls-files
# no output
```
if you start a new session (i.e. clone the project again), you will get:
```
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   large3.txt

no changes added to commit (use "git add" and/or "git commit -a")

$ git restore large3.txt
Encountered 1 file(s) that should have been pointers, but weren't:
        large3.txt
```

**Run environment:**
```bash
# renkulab, Jupyterlab, Firefox
$ renku --version
0.16.0
$ python --version
Python 3.9.5
```

mohammad-sdsc:
# Description

Fixes handling of arrays in the pre-commit hook. Users need to do `renku githooks uninstall` and `renku githooks install` in the existing repos after updating the renku version.

Fixes #2230


Panaetius:
Since users need to take a manual action to fix this, we should have some way to inform them of this. 

I seems pip itself doesn't support anything like that. 
And implementing some mechanism on our end to show a warning the first time a command is run after a user has upgrade renku is probably too much effort. Though it might be useful to have in general, not just for this?
But if we just write it in the changelog, I'm sure it'll get lost for most users

Panaetius (approved):

mohammad-sdsc:
> Since users need to take a manual action to fix this, we should have some way to inform them of this.
> 
> I seems pip itself doesn't support anything like that.
> And implementing some mechanism on our end to show a warning the first time a command is run after a user has upgrade renku is probably too much effort. Though it might be useful to have in general, not just for this?
> But if we just write it in the changelog, I'm sure it'll get lost for most users

We have `renku doctor` check for githooks which warns users if their project's githooks is outdated. Maybe we can somehow include the githooks in the migrations. That would required a proper design.

Panaetius:
Yes it's probably best done in a followup issue.

Something to let users know of important changes after an update might be nice (not just for hooks).

PR #2246, opened 2021-08-06
chore: add telepresence instructions and script

lorenzo-cavazzi:
This adds a `start-telepresence.sh` script to set up telepresence to override on the `<name>-renku-core` pod of a dev deployment.
Once connected, the service can be bootstrapped locally without any further setup to side components (like Redis). The projects are cached locally.

There are a couple of limitations:

- `--no-reload`: that's required if we want to be able to attach the debugger. Otherwise, invoking `ptvsd.enable_attach()` multiple times makes the application crash. The drawback is that the service is not hot-reloaded as soon as a file is modified.
This can be omitted if we are no interested in attaching a debugger (remove both `--no-reload` and `DEBUG_MODE=true`).

- `securityContext.runAsUser: 1000`: telepresence requires `runAsUser` ([reference](https://github.com/telepresenceio/telepresence/issues/875#issuecomment-447872305)) and it fails with `runAsGroups`, though it seems the very latest versions have solved this ([reference](https://github.com/telepresenceio/telepresence/issues/1051))


helm-chart/renku-core/templates/deployment.yaml
```diff
@@ -45,7 +45,7 @@ spec:
           {{ if .Values.debug }}
           args: ["api", "--debug"]
           securityContext:
-            runAsGroup: 2000
+            runAsUser: 1000
```
Panaetius, line 48: The group here was to enable modifying files when having VSCode attached. basically root and the debug group (2000) have write access. Then in a normal deployment, it'll run as the `shuhitsu` user which has no write access, so there's no way for a potential attacker to modify files. But when running in debug mode, the `runAsGroup: 2000` runs in the debug group which has write access in the pod's filesystem.

So I think the `runAsGroup` is still needed unless I'm missing something

Panaetius (changes requested):
Great script and thank you for updating the README!

Just had two minor comments, otherwise good to go

helm-chart/renku-core/templates/deployment.yaml
lorenzo-cavazzi: With telepresence, the writing permissions are not needed anymore since you run the local code instead of the remote one, so any edit is also local.
I couldn't replace the `<namespace>-renku-core` pod when the security context was set to `runAsGroup: 2000`, this is why I switched to the `runAsUser: 1000`.
If you prefer to keep the previous solution, I can try something with the lastest telepresence version. This comments from 5 days ago mention that something changed with the latest version and now telepresence should support `runAsGroup` https://github.com/telepresenceio/telepresence/issues/1051#issuecomment-893661183

Panaetius: Being able to edit files in the deployed pod can be nice for debugging purposes, so if we can have both, that'd be nice. Though if it's either/or, I'd prefer telepresence. 

If we can't have both, we should at least remove the README part that mentions mounting the pod filesystem in VS Code.

lorenzo-cavazzi: I have tried to:
- Find a solution to preserve the `runAsGroup` -- according to [this issue](https://gitmemory.com/issue/telepresenceio/telepresence/1051/502175707) it's a matter of permission, but I couldn't get around it.
Screenshot_20210810_164225
- Use telepresence 2+, which is a [completely overhauled product](https://www.telepresence.io/announcing-telepresence-2/) and supports `runAsGroup` out of the box (since it uses sidecar instead of replacing the pod), but it has different permissions [currently not enabled in our dev deployment](https://swiss-data-science.slack.com/archives/G67N59QL8/p1628602423035800).

I could either try to explore better the telepresence 2 solution -- that will likely require a bit of time.
Or we could add an extra env variable to switch between `runAsUser` and `runAsGroup`, though both are desirable in the dev/ci environments.

PR #2263, opened 2021-08-11
feat(core): use toil-cwl-runner for running workflows

vigsterkr:
# Description

Switch to use `toil-cwl-runner` for executing workflows


vigsterkr:
closing in favour of #2273

PR #2268, opened 2021-08-16
test: make config write concurrency test deterministic

lorenzo-cavazzi:
The `test_config_write_concurrency` test used to fail often because the operation to test was very quick and trying to start many at the same time wasn't always enough.

The new test now monkey patches the write operation to introduce a slowdown, making the result deterministic.

P.S. I tried another solution using [site](https://docs.python.org/3/library/site.html) but I couldn't get a working test in a reasonable amount of time. Also, that solution was going to be more complicated and harder to debug should the test fail in the future due to some code change.


Panaetius (approved):
This is great, thank you! I didn't even consider threads!

PR #2272, opened 2021-08-17
feat(cli): add renku graph export command

### LINKED ISSUE #2271: Add renku graph export command
Add a `renku graph export` command that exports JSON-LD for use by TG

Panaetius:
adds `renku graph export` command.

Supports `--revision <rev>` to only produce triples for changes made in `rev` (Defaults to `HEAD`)
Supports  `--full` to dump the full graph
Supports `--workflows-only` to only produce triples for activities/plans
Supports `--strict` to validate produced json-ld

closes #2271


renku/core/metadata/database.py
```diff
@@ -751,7 +751,7 @@ def _deserialize_helper(self, data, create=True):
                 if "@reference" in data and data["@reference"]:  # A reference
                     assert create, f"Cannot deserialize a reference without creating an instance {data}"
                     new_object = self._database.get_cached(oid)
-                    if new_object:
+                    if new_object is not None:
```
vigsterkr, line 754: just out of curiosity: what's the difference? between `if obj` and `if obj is not None`, apart from semantics? (sorry for the noise/not being fluent enough in python)

renku/core/models/provenance/activity.py
```diff
@@ -167,39 +170,37 @@ def from_plan(
...
             parameters=parameter_values,
+            project_id=project_gateway.get_project().id,
             started_at_time=started_at_time,
             ended_at_time=ended_at_time,
             annotations=annotations,
         )
 
+        pm = get_plugin_manager()
+
+        plugin_annotations = pm.hook.activity_annotations(activity=activity)
```
vigsterkr, line 191: 🚀 will address this change in MLS

vigsterkr (approved):
💃

PR #2296, opened 2021-08-31
feat(api): re-add api datasets commands using new database backend

### LINKED ISSUE #2259: fix renku.api to work with new metadata and database
Some `renku.api` functionality currently doesn't work (mainly project and dataset related), the tests for this are currently skipped.

This might need some more work to make the api work with the new database architecture as the api just uses LocalClient at the moment but doesn't deal with the DB injections that are needed. Maybe switching to also using command_builder with dedicated view models makes sense here.

Panaetius:
closes #2259

### DIFF
--- a/renku/api/models/dataset.py
+++ b/renku/api/models/dataset.py
@@ -42,7 +42,8 @@ property:
 from operator import attrgetter
 
 from renku.api.models.project import ensure_project_context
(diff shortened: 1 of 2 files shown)


mohammad-sdsc (approved):
This was quick 😉! Thanks Ralf!

PR #2303, opened 2021-09-01
feat: update core service error handling

### LINKED ISSUE #2100: improve JSON-RPC error codes
Some of the error codes returned by the renku core service could be improved. For example, from the JSON-RPC 2.0 spec:

* code -32000 --> generic server error
* code -32601 --> method does not exist

We use -32000 for a very specific server error - to indicate that something is wrong with git repository access. We also use it to indicate a generic HTTP server error. 

We use -32601 to indicate that the request headers were wrong (typically because authentication headers are missing) but this error code is for trying to use non-existent methods. We should probably use -32602 (invalid parameters) in this case. 

In addition, since JSON-RPC allows a lot of flexibility to define custom errors, we should understand where the most frequent failures are occurring and provide more informative errors - at the moment, all that is returned is `-32000 - internal error`. 

This is a follow-up to #2056.

lorenzo-cavazzi:
This PR attempts to improve error handling in the core service.
It's doesn't cover all the errors yet, it serves as a POC to verify the solution viability.

The main changes are:
- Different error response format: the `code` field is still there but it's expanded to assign different codes to different types of errors. Instead of a single message (previously `reason`) there are 2 fields targeting the user and the developer, 2 optional URL to documentation, and a link to the generated Sentry issue.
- Documentation for each type of error is automatically generated. Verbose description of the error classes will give further context to the developer encountering the error, without adding too much text in the `devMessage` error field.
- Sentry is handled in a single place, and the sentry errors have been enriched with user data and other useful metadata. We may want to exclude some errors .

The overall idea is to keep the decorators to handle common errors (I'm open to different approaches there :smile: ) , and narrow down to a more specific error on the single endpoint when that provides a better feedback to users/devs.
An example is `renku/service/controllers/templates_read_manifest.py`, where `ErrorUserTemplateNotFound` is a more specific error than `ErrorUserRepoNotFound`.

I was a bit undecided on how to properly create the service error classes. I've decided not to inherit from `Exception` but I'm not sure it's a good idea.

If the approach is good, I will expand this to cover the other existing errors. We can iterate on this by covering more common errors when the need arises.

/deploy
fix #2100


lorenzo-cavazzi:
Here is a sample of the auto-generated documentation.
The base link (currently renkupython.readthedocs.io) should probably come from an OS variable.
Checking the documentation page could be done by manually creating the HTML files with 
```
sphinx-build -qnNW docs docs/_build/html
```

Screenshot_20210903_144102

renku/service/views/__init__.py
```diff
@@ -18,9 +18,26 @@
 """Renku service views."""
 from flask import current_app
 
+from renku.service.config import SVC_ERROR_GENERIC
 from renku.service.serializers.rpc import JsonRPCResponse
 
 
+def error_response_new(serviceError):
```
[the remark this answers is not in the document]
lorenzo-cavazzi: The idea is to remove the previous one, the `_new` is temporary but I will get rid of it when the PR is finished

renku/service/entrypoint.py
```diff
@@ -47,7 +53,10 @@
 
 logging.basicConfig(level=os.getenv("SERVICE_LOG_LEVEL", "WARNING"))
 
-if os.getenv("SENTRY_DSN"):
+
+HAS_SENTRY = True if os.getenv("SENTRY_DSN") else False
```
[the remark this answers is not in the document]
lorenzo-cavazzi: I'll change this either here or on the other PR, depending on which will be merged first

renku/service/errors.py
```diff
@@ -16,7 +16,328 @@
...
+
+
+class ServiceError(Exception):
+    """Error handler for service errors.
+
+    Set the documentation reference for developers willing to explore the base code.
+    Handle Sentry, adding user data (when available) and error details.
+    """
+
+    devReference = f"{DOCS_URL_BASE}{DOCS_URL_ERRORS}" + "{class_name}"
```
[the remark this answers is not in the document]
lorenzo-cavazzi: Actually, the `"{class_name}"` is _not_ an f-string on purpose here. It's used only to provide meaningful autogenerated documentation since `self.__class__.__name__` doesn't work there.
The proper class name is set in the `__init__` function.

Screenshot_20220210_163910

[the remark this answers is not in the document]
Panaetius: Ah right, I missed that

renku/service/errors.py
```diff
@@ -16,7 +16,328 @@
...
+        Either provide an `errorObject` taken from the error classes in renku.service.errors,
+        or a set of the remaining arguments. They are mutually exclusive.
+        """
+        super().__init__()
+        self.exception = exception
+        self.devReference = f"{DOCS_URL_BASE}{DOCS_URL_ERRORS}{self.__class__.__name__}"
+
+        if errorObject is None:
+            if code:
+                self.code = code
```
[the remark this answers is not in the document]
lorenzo-cavazzi: I see your point. Otoh, this class is (almost) always used as a base class from which more specific errors can inherit, and some properties are already specified there. They would be overwritten by the local `__init__` if I don't test for them to be non-null (e.g. `self.code` from `ErrorUserRepoUrlInvalid` would be lost when invoking `super().__init__()`).

Perhaps should I _always_ include the local values from the sub-classes? In the case of `ErrorUserRepoUrlInvalid`, that would be:
```
super().__init__(
    devMessage=self.devMessage.format(error_message=error_message),
    code=self.code,
    userMessage=self.userMessage,
    exception=exception
)
```

It looks a bit verbose to me.
Maybe there is a better solution.

renku/service/errors.py
```diff
@@ -16,7 +16,328 @@
...
+
+    code = SVC_ERROR_PROGRAMMING + 20
+    userMessage = "Fatal error occured while processing a git operation on the repository."
+    devMessage = "Unexpected git error. Git error message: {error_message}"
+
+    def __init__(self, exception=None, error_message=ERROR_NOT_AVAILABLE):
+        super().__init__(devMessage=self.devMessage.format(error_message=error_message), exception=exception)
+
+
+# ! FIX THIS
```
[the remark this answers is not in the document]
lorenzo-cavazzi: Ah sorry, it's a leftover :facepalm:

lorenzo-cavazzi:
Thank you @Panaetius for the early review!
I addressed most of the comments, there is one tricky point on the `__init__` function of the `ServiceError` class (see inline).

There are a couple of other (minor) changes I included in the last commit since I was polishing a bit the code.
The Sentry errors should also contain a few extra details, and the reference error be more meaningful.

Screenshot_20220210_151252

I still need to handle a bunch of errors in this PR. If you think I'm on the right way, I'll continue with those :slightly_smiling_face:

Panaetius:
Looks really nice already!

lorenzo-cavazzi:
@Panaetius This is finally ready for a review.

I addressed all the points from your first review, apart from the one not marked as solved. My suggestion there is probably a bit verbose, perhaps you have a better solution to overcome the problem with auto docs.

renku/service/errors.py
```diff
@@ -16,53 +16,675 @@
...
+            devMessage=self.devMessage.format(field=field),
+            exception=exception,
+        )
+
+
+class UserTemplateInvalidError(ServiceError):
+    """The provided URL doesn't lead to a valid template repository."""
+
+    code = SVC_ERROR_USER + 101
+    userMessage = "The target repository is not a valid RenkuLab template repository."
```
Panaetius, line 265:
```suggestion
    userMessage = "The target repository is not a valid Renku template repository."
```

renku/service/errors.py
```diff
@@ -16,53 +16,675 @@
...
+
+    def __init__(self, exception=None):
+        super().__init__(exception=exception)
+
+
+class UserDatasetsMultipleImagesError(ServiceError):
+    """Multiple images have been set for the same dataset."""
+
+    code = SVC_ERROR_USER + 130
+    userMessage = "Cannot set multiple images for the dataset."
```
Panaetius, line 305: This one is slightly off. You can set more than one images for a dataset (i.e. first one being the thumbnail and then some other images, maybe charts or something?). But each image needs a position(int), so they're ordered. And you can't have two images with the same position. So `[{url: img1.png, position: 1},{url: img2.png, position:1}]` gives this error, but `[{url: img1.png, position: 1},{url: img2.png, position:2}]` is perfectly fine.

renku/service/errors.py
```diff
@@ -16,53 +16,675 @@
...
+
+    def __init__(self, exception=None):
+        super().__init__(exception=exception)
+
+
+class UserDatasetsUnreachableImageError(ServiceError):
+    """Dataset image not reachable."""
+
+    code = SVC_ERROR_USER + 131
+    userMessage = "Cannot get the dataset image."
```
Panaetius, line 316:
```suggestion
    userMessage = "Cannot download the dataset image."
```
I think this is clearer, as it's downloading the image into the project that fails.

renku/service/errors.py
```diff
@@ -16,53 +16,675 @@
...
+    devMessage = "The dataset image is no reachable."
```
Panaetius, line 317:
```suggestion
    devMessage = "The remote dataset image is no reachable."
```
maybe?

renku/service/views/apispec.py
```diff
@@ -39,7 +39,7 @@
 gitlab_token_scheme = {"type": "apiKey", "name": "Authorization", "in": "header"}
 
 TOP_LEVEL_DESCRIPTION = """
-This is the API specification of the renku core service. The API follows the
+This is the API specification of the renku core service. The API loosely follows the
```
Panaetius, line 42: I honestly don't think we follow JSON-RPC at all. I wonder if we should just remove that.

renku/service/views/error_handlers.py
```diff
@@ -0,0 +1,403 @@
...
+def handle_migration_errors(f):
+    """Wrapper which handles migrations exceptions."""
+    # noqa
+    @wraps(f)
+    def decorated_function(*args, **kwargs):
+        """Represents decorated function."""
+        try:
+            return f(*args, **kwargs)
+        except (TemplateUpdateError, DockerfileUpdateError, MigrationError) as e:
+            raise UserOutdatedProjectError(e)
```
Panaetius, line 386: This means a migration failed. Not that a migration is required. So telling they user they should migrate when they just tried this and it failed might not be the best thing to do :smile:

renku/service/views/error_handlers.py
```diff
@@ -0,0 +1,403 @@
...
+            return f(*args, **kwargs)
+        except ValidationError as e:
+            if getattr(e, "field_name", None) == "_schema":
+                error_message = str(e)
+                if match_schema("Project name", error_message):
+                    raise UserProjectCreationError(e, "project name must contain at least a valid character")
+                elif match_schema("git_url", error_message):
+                    raise ProgramProjectCreationError(e, "git_url is invalid")
+            raise
+        except KeyError as e:
```
Panaetius, line 265: This is pretty bad, to be honest (not on your part, in general), whatever we decorate here should handle the KeyError and reraise it as a more descriptive error. It shouldn't be up to the caller to figure out what could have happened. We should create a follow-up issue to clean this up

tests/service/views/test_cache_views.py
```diff
@@ -67,7 +67,7 @@ def test_list_upload_files_all_no_auth(svc_client):
     assert 200 == response.status_code
 
     assert {"error"} == set(response.json.keys())
-    assert INVALID_HEADERS_ERROR_CODE == response.json["error"]["code"]
+    assert UserAnonymousError().code == response.json["error"]["code"]
```
Panaetius, line 70: I don't think you need to instantiate the error to get the class variable, it should work without brackets.

```suggestion
    assert UserAnonymousError.code == response.json["error"]["code"]
```

Panaetius (changes requested):
Very nice, mostly I found some typos and minor things. This definitely makes all our lives easier in the future!

And now you're the expert on everything that is wrong with renku-python :stuck_out_tongue_closed_eyes:

renku/service/views/apispec.py
lorenzo-cavazzi: Right, I guess this only applies to the response format. That still *loosely* follows JSON-RPC 2.0.
I'll remove this sentence and rephrase a bit the content below. Let me know if you think it's better, or if this section should be overhauled.

renku/service/views/error_handlers.py
lorenzo-cavazzi: Ah right! I reused an error originally meant to supersede `MigrationRequired`, not `MigrationError` :facepalm:

tests/service/views/test_cache_views.py
lorenzo-cavazzi: Oh... you are right... let me adjust the 51 occurrences of `().code` :sweat_smile:

renku/service/views/error_handlers.py
lorenzo-cavazzi: Yeah, this is a mix of different errors. In general, a good strategy could be trying to catch all the errors we can identify (especially if they are likely to be determined by wrong data provided by users), and let the others trigger a more generic `ProgramSomthingError` since it's always possible the client is passing wrong keys in creative ways.
If that sounds like a reasonable strategy, I'll open a follow-up issue with a few details.

renku/service/entrypoint.py
```diff
@@ -116,38 +170,29 @@ def after_request(response):
...
+    reloadable = os.environ.get("DEBUG_RELOAD", "false") == "true"
+    if reloadable:
+        app.logger.debug("Debug mode - reloadable:")
+        app.logger.debug("Changes to the code will be instantly applied, but you won't be able to attach a debugger.")
+    else:
+        ptvsd.enable_attach()
+        app.logger.debug("Debug mode - no reload:")
+        app.logger.debug(
+            "The service allows attaching external debuggers, but changes to the code won't be instantly"
+            "applied. You may need to manually restart the service whenever required."
```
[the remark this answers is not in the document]
lorenzo-cavazzi: Ouch!
I was using this together with telepresence to go through the code in CI deployments. I think it's very useful to debug.
Is there an alternative way to attach to the service for debugging purposes? I guess `ptvsd.enable_attach()` is the missing piece after #2648 , but it's very possible I'm not aware of an alternative.

lorenzo-cavazzi:
Thanks for the review!
I should have addressed all the points, except the one on `ptvsd.enable_attach()`. I assume there is already an alternative to that. If not, this PR doesn't seem the place to address it anyway, so I'll remove the code (keeping it here for now since I use it for debugging while implementing the UI counterpart).

There are a couple of non-resolved conversations where I asked for additional info or confirmation of my tentative solution.

renku/service/entrypoint.py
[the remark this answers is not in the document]
Panaetius: hmm, shouldn't it be possible using e.g. the VSCode debugger + Telepresence, without anything in code? With the correct launch configuration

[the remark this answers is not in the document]
lorenzo-cavazzi: You are right, there is a way to work without `ptvsd.enable_attach()`, although starting telepresence in the python debugger terminal inside VScode it's a bit clunky.
Anyway, that is not absolutely necessary for me to debug, so I can remove it from here :+1:

renku/service/views/error_handlers.py
lorenzo-cavazzi: #2729

PR #2304, opened 2021-09-01
feat(core): make update work with new storage

### LINKED ISSUE #2257: implement renku update with new database
`renku update` currently doesn't work (it still uses old graph building or provenance graph). We have to implement it again with the new Database and ActivityGateway. 

Reenable the `renku update` tests to verify this works as before.

mohammad-sdsc:
# Description

`renku update` with the new storage/metadata. It has a `PATHS` parameter which limits the update to the specified paths.

Fixes #2257


renku/core/commands/update.py
```diff
@@ -46,96 +51,192 @@ def update_workflows():
...
-    wf, path = CWLConverter.convert(workflow, client.path)
-    # Don't compute paths if storage is disabled.
-    if client.check_external_storage():
-        # Make sure all inputs are pulled from a storage.
-        paths_ = (i.consumes.path for i in workflow.inputs)
-        client.pull_paths_from_storage(*paths_)
+    activities = set(starting_activities)
+    while activities:
+        activity = activities.pop()
+        child_activities = activity_gateway.get_downstream_activities(activity, max_depth=1)
```
Panaetius, line 92: I think we can just get everything from the catalog directly instead of repeatedly querying it.

renku/core/metadata/gateway/activity_gateway.py
```diff
@@ -62,6 +62,17 @@ def get_downstream_activities(self, activity: Activity, max_depth=None) -> Set[A
 
         return downstream
 
+    def get_downstream_activity_chains(self, activity: Activity) -> List[Tuple[Activity, ...]]:
+        """Get a list of tuples of all downstream paths of this activity."""
+        database = self.database_dispatcher.current_database
+
+        activity_catalog = database["activity-catalog"]
+        tok = activity_catalog.tokenizeQuery
+        downstream_chains = activity_catalog.findRelationChains(tok(upstream=activity))
```
Panaetius, line 71:
```
downstream_chains = activity_catalog.findRelationChains(tok(upstream=zc.relation.catalog.any(*activities)))
```

might be faster?

renku/core/commands/update.py
```diff
@@ -46,96 +52,211 @@ def update_workflows():
...
-#     default_parameters[name] = parameter
-
-# run.run_parameters = list(default_parameters.values())
+    database = database_dispatcher.current_database
+
+    activities = []
+
+    for plan in plans:
+        # NOTE: Update plans are copies of Plan objects. We need to use the original Plan objects to avoid duplicates.
+        original_plan = database["plans"].get(plan.id)
```
Panaetius, line 172: We shouldn't access the database directly and instead go through the interfaces, to make sure the code is properly insulated from implementation details of the database. So `IPlanGateway.get_by_id(plan.id)`

renku/core/commands/update.py
```diff
@@ -46,96 +52,211 @@ def update_workflows():
...
+        activity = Activity.from_plan(plan=original_plan, started_at_time=started_at_time, ended_at_time=ended_at_time)
+        activity_gateway.add(activity)
+        activities.append(activity)
+
+    if len(activities) > 1:
+        activity_collection = ActivityCollection(activities=activities)
+        activity_gateway.add_activity_collection(activity_collection)
```
Panaetius, line 179: It looks like the `"activity-collections"` only exists to enable some tests. it is a bit odd to me to store something in users' repositories that is only used for our tests, essentially littering the database with unneeded data.

I guess we could have a `TestingActivityGateway(ActivityGateway)` in `tests/` and inject that for tests, to only have this index when testing code?

renku/core/management/workflow/concrete_execution_graph.py
```diff
@@ -141,6 +141,9 @@ def workflow_graph(self):
                     workflow_graph.add_node(node)
                 continue
 
+            if not next(self.graph.predecessors(node), None):
```
Panaetius, line 144: It might only make a difference on very large repositories with complex workflows, but 
```
intermediate_predecessor = next(self.graph.predecessors(node), None)
if not intermediate_predecessor:
    continue
[...]
source = next(self.graph.predecessors(intermediate_predecessor), None)
```
would only have to calculate the predecessor once.

renku/core/models/workflow/composite_plan.py
```diff
@@ -45,38 +45,31 @@ class CompositePlan(AbstractPlan):
...
         description: str = None,
         id: str,
         invalidated_at: datetime = None,
         keywords: List[str] = None,
+        links: List[ParameterLink] = None,
+        mappings: List[ParameterMapping] = None,
         name: str,
-        derived_from: str = None,
-        plans: List[Union["CompositePlan", Plan]] = None,
+        plans: List[Plan] = None,
```
Panaetius, line 56: Should it be `AbstractPlan` ? A `CompositePlan` created by a user could contain a `CompositePlan`

Panaetius (changes requested):
Looks really nice now. It's great seeing things finally come together!

renku/core/utils/git.py
```diff
@@ -225,3 +228,18 @@ def get_previous_commit_from_submodules() -> Optional[Commit]:
 def get_path(url: str):
     """Return path part of a url."""
     return urllib.parse.urlparse(url).path
+
+
+def get_modified_entities(entities, repo: Repo):
```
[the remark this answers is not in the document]
mohammad-sdsc: Good point!

renku/core/commands/update.py
```diff
@@ -46,96 +52,211 @@ def update_workflows():
...
+
+
+def _get_downstream_activities(
+    starting_activities: Generator[Activity, None, None], activity_gateway: IActivityGateway, paths: List[str]
+) -> Set[Activity]:
+    """Return an ordered list of activities so that an activities comes before all its downstream activities."""
+    all_activities = defaultdict(set)
+
+    def have_identical_inputs_and_outputs(activity1, activity2):
+        return sorted(u.entity.path for u in activity1.usages) == sorted(
```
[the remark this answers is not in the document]
mohammad-sdsc: I believe there is not clear way of comparing activities. The idea here was to cover cases like `cat A A B > C` and `cat A B B > C`.

renku/core/management/workflow/concrete_execution_graph.py
mohammad-sdsc: It always helps with the code's readability.

renku/core/models/workflow/composite_plan.py
mohammad-sdsc: 👍
I made it as it was before (`Union["CompositePlan", Plan]`) which should also help with ide type linter.

renku/core/commands/update.py
mohammad-sdsc: `ActivityCollection` is to mark that these activities have been executed together as a result of an update or a rerun. So, it's not just for testing. I was not sure what other metadata we need to include here (specifically if we need a link in the `Activity` to its `ActivityCollection` if any). What do you think?

renku/core/commands/update.py
```diff
@@ -46,96 +52,211 @@ def update_workflows():
...
-    # input_paths = {node.path for node in graph.nodes} - output_paths
+    for activity in starting_activities:
+        downstream_chains = activity_gateway.get_downstream_activity_chains(activity)
 
-    # # Store the generated workflow used for updating paths.
-    # workflow = graph.as_workflow(input_paths=input_paths, output_paths=output_paths, outputs=outputs)
+        if paths:
+            # NOTE: Add the activity to check if it also matches the condition
+            downstream_chains.append((activity,))
+            downstream_chains = [c for c in downstream_chains if any(g.entity.path in paths for g in c[-1].generations)]
```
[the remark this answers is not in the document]
mohammad-sdsc: Yes, they can. I've fixed this and added a test for it.

renku/core/commands/update.py
Panaetius: a link from ActivityCollection to Activity makes sense to me. Best to discuss it with the KG team as well

Panaetius (approved):
Thank you!

PR #2306, opened 2021-09-02
chore: remove python 3.6 support, add python 3.9 support

Panaetius:
Phase out support for python 3.6 as it nears its end of life this december.


mohammad-sdsc (approved):
👍

PR #2310, opened 2021-09-06
feat(core): add custom dataset metadata

### LINKED ISSUE #2288: Add support for custom annotations in Dataset
When creating/editing a Dataset, it should be possible to pass custom json-ld as a file to the command, which will be included as an oa:Annotation in the metadata (similar to the workflow annotations just without plugin)

Panaetius:
closes #2288


mohammad-sdsc (approved):
LGTM, thank you!

PR #2313, opened 2021-09-06
feat(core): allow adding custom metadata to projects

### LINKED ISSUE #2289: Add support for custom annotations in Project
When creating/editing a Project, it should be possible to pass custom json-ld as a file to the command, which will be included as an oa:Annotation in the metadata (similar to the workflow annotations just without plugin)

Panaetius:
closes #2289


mohammad-sdsc (approved):
LGTM, thanks!

PR #2316, opened 2021-09-07
feat(cli): add renku workflow inputs/outputs

### LINKED ISSUE #2261: implement renku show with new storage backend
`renku show` (sub)commands are currently not implemented with the new database backend and should be reimplemented.

The main command that is important is `renku show outputs`, which is used by the pre-commit hook to check if any generated output files have been modified by the user.

`renku show inputs` is second priority, with `renku show siblings` being lowest priority.

The `--revision` flag for these commands doesn't make sense anymore. It never shows the status at a revision, instead if would just limit the `paths` argument to paths touched in commits from revision..HEAD, which still just looking at the metadata in HEAD. So I think this can be left out.

Panaetius:
closes #2261


renku/core/commands/workflow.py
```diff
@@ -301,3 +302,73 @@ def _export_workflow(
...
+        path = Path(p).resolve().relative_to(client.path)
+        if path.is_dir():
+            target_dirs.append(str(path))
+        else:
+            target_files.add(path)
+
+    result = set()
+
+    for target_file in target_files:
+        if target_file in files or any(str(target_file).startswith(d) for d in dirs):
```
mohammad-sdsc, line 335: It's safer to check `any(Path(d) in Path(target_file).parents for d in dirs)` since `startswith` can be true for partial matches.

renku/core/commands/workflow.py
```diff
@@ -301,3 +302,73 @@ def _export_workflow(
...
+            result.add(str(target_file))
+
+    for target_dir in target_dirs:
+        if target_dir in dirs or any((str(f).startswith(target_dir) for f in files)):
```
mohammad-sdsc, line 339: same as the previous comment.

renku/cli/workflow.py
```diff
@@ -534,3 +555,41 @@ def export(workflow_name, format, output, values):
...
+
+    <PATHS>    Limit results to these paths.
+    """
+    result = workflow_inputs_command().build().execute(paths=paths)
+
+    input_paths = result.output
+
+    click.echo("\n".join(input_paths))
+
+    ctx.exit(0 if not paths or len(input_paths) == len(paths) else 1)
```
mohammad-sdsc, line 574: Do we need a similar check as the `output` to cover directories?

renku/data/pre-commit.sh
```diff
@@ -39,34 +39,6 @@ if [ ${#MODIFIED_FILES[@]} -ne 0 ] || [ ${#ADDED_FILES[@]} -ne 0 ]; then
 fi
 
 if [ ${#MODIFIED_FILES[@]} -ne 0 ] ; then
-  MODIFIED_OUTPUTS=$(renku show outputs "${MODIFIED_FILES[@]}")
```
mohammad-sdsc, line 42: Should we create a followup story for this?

tests/cli/test_workflow.py
```diff
@@ -295,3 +295,29 @@ def _get_plan_id(output):
...
+def test_workflow_show_outputs_with_directory(runner, client, run):
+    """Output files in directory are not shown as separate outputs."""
+    base_sh = ["bash", "-c", 'DIR="$0"; mkdir -p "$DIR"; ' 'for x in "$@"; do touch "$DIR/$x"; done']
+
+    assert 0 == run(args=["run"] + base_sh + ["output", "foo", "bar"])
+    assert (client.path / "output" / "foo").exists()
+    assert (client.path / "output" / "bar").exists()
+
+    cmd = ["workflow", "outputs"]
+    # result = runner.invoke(cli, cmd)
```
mohammad-sdsc, line 309: Should these lines be uncommented?

mohammad-sdsc (commented):
Thanks Ralf! Please see the comments.

tests/cli/test_workflow.py
Panaetius: right, left-over from debugging

renku/data/pre-commit.sh
Panaetius: I tested it and modifying outputs didn't seem to break anything, so I don't think we need a followup story

renku/cli/workflow.py
Panaetius: This is how the original show inputs worked. I honestly don't know why it differs from output.

renku/core/commands/workflow.py
```diff
@@ -301,3 +302,73 @@ def _export_workflow(
...
+        if path.is_dir():
+            dirs.append(str(path))
+        else:
+            files.add(path)
+
+    target_dirs = []
+    target_files = set()
+
+    for p in target_paths:
+        path = Path(p).resolve().relative_to(client.path)
```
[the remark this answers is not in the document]
Panaetius: It's not entities. I use the keys of the `activities-by-generation` index, which are strings.

mohammad-sdsc (approved):
Thank you!

PR #2328, opened 2021-09-14
fix(core): fix migration of workflow metadata

Panaetius:
Fixes migration of workflows where a `ProcessRun` links to a Plan that has subprocesses. Will create an `Activity` per subprocess (So possibly multiple `Activity`s per `ProcessRun`), splitting Inputs/Outputs to the activities they belong to


mohammad-sdsc (approved):
LGTM, thank you!

PR #2339, opened 2021-09-16
feat(svc): add support for template images

### LINKED ISSUE #2311: Support template icons

The POC team asked to support associating icons with project templates. Those would be used by the UI when showing the available project templates from a templates repository (i.e. listed in the manifest.yaml file).

Considering the discussion here SwissDataScienceCenter/renku-ui/issues/1467 , we would like to get those images as a base64 encoded string (other formats are welcome), exposed as a new `icon` field in the payload provided for each template.
The image should be part of the templates repository, and the manifest file should allow specifying an `icon` field with the local path of the image.
Encoding should either scale down big images (that should be an icon, so nothing too big/detailed), or allow a maximum size.

We could allow specifying a URL instead of including the image in the repository. That would simplify the implementation of this issue, but we would have no control of the image (more details on SwissDataScienceCenter/renku-ui/issues/1467 ).

Panaetius:
closes #2311 

needs https://github.com/SwissDataScienceCenter/renku-project-template/pull/129


renku/service/controllers/templates_read_manifest.py
```diff
@@ -43,7 +48,27 @@ def context(self):
...
+        template_manifest, template_folder, _, _ = fetch_template(self.ctx["git_url"], self.ctx["ref"])
+
+        # NOTE: convert icons to base64
+        for template in template_manifest:
+            if "icon" not in template:
+                continue
+
+            # NOTE: prevent path traversal attack
+            icon_path = template_folder / (
+                (template_folder / template["icon"]).resolve().relative_to(template_folder.resolve())
```
mohammad-sdsc, line 62: Shouldn't we resolve `template_folder` first?

mohammad-sdsc (commented):
Looks very good! I've made a comment.

renku/service/controllers/templates_read_manifest.py
Panaetius: Actually it is already resolved in `fetch_template`, I've removed the `.resolve()` from template_folder

mohammad-sdsc (approved):
🚀

PR #2340, opened 2021-09-16
fix(core): add missing project properties to SHACL file

Panaetius:
adds renku:hasDataset, renku:hasPlan, renku:hasActivity to Project SHACL shape


mohammad-sdsc (approved):
LGTM, thank you!

PR #2345, opened 2021-09-21
fix(core): change project-id to be based on project slug instead of name

### LINKED ISSUE #2335: create project IRI based on project slug
Currently we URL encode the project name to create the project @id, like https://renku-ci-gr-738.dev.renku.ch/projects/jakub.chrobasik/test%25202021-09-14%252015-59-30. But for downstream applications, using the project slug would be better, as it can be directly resolved in the browser, e.g. https://renku-ci-gr-738.dev.renku.ch/projects/jakub.chrobasik/test-2021-09-14-15-59-30

We should switch the project id generation to calculate and use the project slug instead. renku.service already has some code to calculate the slug, we should move that to renku.core.

Panaetius:
closes #2335


vigsterkr (dismissed):
🚀

vigsterkr (approved):
🚀

PR #2349, opened 2021-09-21
feat(cli): add a flag to fail on migration errors

### LINKED ISSUE #2337: Allow migrations to fail on errors
When migrating a Renku project to a newer version, all errors are caught and reported and the migration continues. While this is convenient for the end user, it also hide errors in our tests when a project migration is needed.

**Acceptance Criteria**
* Add a `--strict` flag to the migration command that make it to fail on errors. This should be `False` by default.
* Modify tests to use this flag to make failures explicit

mohammad-sdsc:
# Description

Add a `--strict` hidden flag to allow migrations to fail in tests. Some refactoring. I also removed one of the old projects that we use for migration test; the problem was that datasets' JSON-LD metadata generated by renku is not valid at `v0.5.0` and this project cannot be migrated without error. We have another project from `v0.5.1` that tests the same migration.

Fixes #2337


Panaetius (approved):
Love the refactoring with migration type! Looks great 👍

PR #2350, opened 2021-09-22
fix(core): pin cwltool to lower version as newer version is broken

Panaetius:
Pins cwltool to lower version until https://github.com/common-workflow-language/cwltool/issues/1525 is fixed


vigsterkr (approved):

setup.py
```diff
@@ -162,7 +162,7 @@ def run(self):
     "click>=7.0,<8.0.2",
     "cryptography>=3.4.1,<3.5",
     "cwlgen>=0.4.0,<=0.4.2",
-    "cwltool>=3.1.20210816212154,<3.2",
+    "cwltool>=3.1.20210816212154,<3.1.20210921111717",
```
mr-c, line 165: Might have been better to just exclude the problematic release with `!=3.1.20210921111717` :-)

Panaetius: Good point, thanks for the tip!

PR #2355, opened 2021-09-22
feat(core): add position to mapped input/output streams of a workflow

Panaetius:
adds position to inputs/outputs that were mapped to stdin/stdout/stderr


renku/core/management/workflow/plan_factory.py
```diff
@@ -395,6 +397,12 @@ def add_command_output(
 
         mapped_stream = self.get_stream_mapping_for_value(default_value)
 
+        if mapped_stream and not position:
```
mohammad-sdsc, line 400: Does it matter if we check `not position` instead of `position is None`?

mohammad-sdsc (dismissed):
LGTM, made a single comment. Feel free to ignore if not relevant.

renku/core/management/workflow/plan_factory.py
Panaetius: it shouldn't, position always starts at 1. but `is None` is cleaner, I'll change it

mohammad-sdsc (approved):
Thank you!

PR #2356, opened 2021-09-22
fix(dataset): same_as and DatasetFile id corrections

### LINKED ISSUE #2336: Errors when migrating datasets to the new metadata
* `DatasetFile` should receive a new id when a file is deleted. Currently, we keep the same id and mark it as deleted.
* `sameAs` field for imported renku datasets should be fixed by removing dashes from the id. We removed dashes from datasets' ids and we need to reflect that change for the imported datasets.

mohammad-sdsc:
# Description

Fix id for `DatasetFile` when removing a file. Fix `sameAs` during migration by removing dashed from the id.

Fixes #2336


Panaetius (approved):
LGTM, thank you

PR #2358, opened 2021-09-23
feat(cli): add renku log command

### LINKED ISSUE #2132: Add `renku workflow history` command
Add a `renku workflow history` command, as per the 001 RFC. Output should be similar to a `git log` (A list of executions by time) and it should allow filtering by workflow name/id as well as per input/output file.

Should support human readable (tabular?) output and JSON output.

Panaetius:
closes #2132


mohammad-sdsc (approved):
LGTM, thank you!

PR #2361, opened 2021-09-23
fix(cli): actually flatten 'json-ld' output and remove 'json-ld-graph'

Panaetius:
In old old old renku, `json-ld` format would output nested data and `json-ld-graph` would output flattened data. Since we now always output flattened data, removes `json-ld-graph` format and fixes `json-ld` to properly flatten the graph.


mohammad-sdsc (approved):
LGTM, thank you!

PR #2362, opened 2021-09-23
fix(core): parse key when overriding params in workflow execute

vigsterkr:
# Description

`workflow execute` `--set` argument's key should be parsed into a nested dictionary.


Panaetius (changes requested):
Looks great but needs a test

vigsterkr:
since we dont consider dot as a protected character for workflow naming some extra gymnastics was required to resolve cli `--set` params for composite plans.

PR #2363, opened 2021-09-23
refactor(core): add an abstraction layer for git

### LINKED ISSUE #2253: externalise dependency on git
we currently depend on gitpython directly in many places, but ideally git would just be an implementation detail.

We should have an interface that defines certain common operations that we use git for that then has a git specific implementation that gets injected as a dependency, so our code only depends on the interface.

As such, the interface should have methods for "git add", "git commit", "git push" directly, but should potentially combine those to units that make sense from our code (e.g. add and commit in a single method called `persist_file_changes`), in a way that it could still be used with a different implementation of the interface, like if we wanted to use subversion as a backend (not that we'd ever do this...).

mohammad-sdsc:
# Description

- Adds a `Repository` class as an access layer to all git operations.
- Moves most of the git functionality to `git` utility file.
- Refactors `request` library usage.
- Reorganizes parts of the code to speed up imports.
- Splits integration tests to service/non-service to speed up tests.

Fixes #2253 

# TODO (for followup stories):
- Use dependency injection for `Repository`
- Refactor all core models so that they don't deal with git
- Increase abstraction level of git usage so that we deal with use-cases and not git commands (TBD)


renku/core/metadata/repository.py
```diff
@@ -0,0 +1,1192 @@
...
+
+        return Commit.from_commit(self._repository, self._repository.head.commit)
+
+    def checkout(self, reference: Union["Branch", str]):
+        """Check-out a specific reference."""
+        self.run_git_command("checkout", reference)
+
+    def clean(self):
+        """Remove all untracked files."""
+        self.run_git_command("clean", "-xdff")
```
Panaetius, line 197: Is the double `f` here a typo? I don't see anything in the git clean command that would suggest it does something special

renku/core/metadata/repository.py
```diff
@@ -0,0 +1,1192 @@
...
+        force: bool = False,
+    ):
+        """Remove paths from repository or index."""
+        self.run_git_command("rm", "--", *paths, cached=index, ignore_unmatch=not_exists_ok, r=recursive, force=force)
+
+    def reset(
+        self, reference: Union["Branch", "Commit", "Reference", str] = None, hard: bool = False, soft: bool = False
+    ):
+        """Reset a git repository to a given reference."""
+        if hard:
```
Panaetius, line 259: I don't see a place where `soft` is used, I think we can just remove it?

renku/core/metadata/repository.py
```diff
@@ -0,0 +1,1192 @@
...
+            for submodule in self.submodules:
+                try:
+                    Path(absolute_path).relative_to(submodule.path)
+                except ValueError:
+                    continue
+                else:
+                    return submodule.get_object_hash(path=absolute_path, revision="HEAD")
+
+        try:
+            return self.run_git_command("rev-parse", f"{revision}:{path}")
```
Panaetius, line 442: I had to also change this in https://github.com/SwissDataScienceCenter/renku-python/pull/2426/files#diff-1c9cff6463dbad55fd4faaa72d921d9edb9c042365e2efa432b533c83d6bdae6R134 so it works without a revision, using `git hash-object`. We could call the `hash_object()` below if no revision is passed.

renku/core/metadata/repository.py
```diff
@@ -0,0 +1,1192 @@
...
+        try:
+            self._repository: git.Repo = _create_repository(path, search_parent_directories=False)
+        except errors.GitError:
+            # NOTE: Submodule directory doesn't exists yet, so, we ignore the error
+            pass
+
+    @classmethod
+    def from_submodule(cls, parent: git.Repo, submodule: Optional[git.Submodule]) -> Optional["Submodule"]:
+        """Create an instance from a git submodule."""
+        if submodule is None:
```
Panaetius, line 589: Is submodule ever actually None? I didn't see any obvious place where it would be None

Panaetius (changes requested):
This is awesome! I'm amazed at how many only slightly related improvements/fixes you also managed to sneak in! Thanks :rocket:

renku/core/management/datasets.py
```diff
@@ -1030,9 +992,9 @@ def update_external_files(self, records: List[DynamicProxy]):
         if not updated_files_paths:
             return
 
-        add_to_git(self.repo.git, *updated_files_paths, force=True)
-        self.repo.git.add(self.renku_pointers_path, force=True)
-        self.repo.index.commit("renku dataset: updated {} external files".format(len(updated_files_paths)))
+        self.repository.add(*updated_files_paths, force=True)
+        self.repository.add(self.renku_pointers_path, force=True)
+        self.repository.commit(f"renku dataset: updated {len(updated_files_paths)} external files")
```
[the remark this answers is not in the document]
mohammad-sdsc: No, this PR should not make any changes to the underlying functionalities so, it should be ok to just remove the commit from here.

renku/core/metadata/repository.py
mohammad-sdsc: Haha, I had to double check too! The doc says: `Git will refuse to modify untracked nested git repositories (directories with a .git subdirectory) unless a second -f is given.` (`git version 2.24.3 (Apple Git-128)`).

renku/core/metadata/repository.py
mohammad-sdsc: We can use `HEAD` if no revision is passed. Is there a reason for not doing this? The two methods should return the same results if the object is not modified. However, `hash-object` reads the file and calculate the hash while `rev-parse` reads it from the git metadata. Also, if the file is modified, `rev-parse` returns the hash of the original content.

Panaetius: HEAD only works for committed files. With the change I mentioned above that we don't do unnecessary commits, when we do a `renku run` we need to create the entities for the activity for files that are maybe not yet committed (as they'll be committed only once the whole command is finished).

Panaetius (approved):

PR #2368, opened 2021-09-24
fix(dataset): fix shacl shape for datasets

mohammad-sdsc:
Removes `schema:location` from `schema:PublicationEvent` SHACL shape. Removes `sh:lessThanOrEquals` constraint from Dataset's `schema:dateCreated`.


Panaetius (approved):
LGTM, thank you

PR #2373, opened 2021-09-27
fix(dataset): unset date_created after import

mohammad-sdsc:
# Description

Unset `schema:dateCreated` for imported datasets. Remove microseconds from datetime fields so that the exported format will be for example like `2021-09-16T12:13:35+00:00` (and not `2021-09-16T12:13:35.540364+00:00`)


vigsterkr (approved):
LGTM

PR #2381, opened 2021-10-01
fix(core): pin git-lfs on macos to 2.13.3

vigsterkr:
# Description

this is a temporary fix until a proper fix is done for `git-lfs` v3.0.0 major release


Panaetius (approved):
LGTM, thanks!

PR #2382, opened 2021-10-01
fix(service): fix git config getting included in service images

### LINKED ISSUE #2269: Fix copying all .git data in core service Docker build
https://github.com/SwissDataScienceCenter/renku-python/blame/78fad8967660bb973d72e2d544dcd7978b4ea260/Dockerfile.svc#L17 copies over the whole `.git` folder into the Docker image. This includes `.git/config` which could contain secrets, which we don't want in the image.

We should docker ignore the config file (or if it is needed, sanitize it) so it can't leak secrets by accident

Panaetius:
closes #2269

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,3 +6,4 @@
 !entrypoint-svc.sh
 !.git
 !.gitignore
+.git/config


vigsterkr (approved):
LGTM

rokroskar (approved):
thanks for the well thought-out fix! 😆

PR #2383, opened 2021-10-01
chore: add nix derivations

rokroskar:
This PR adds a nix derivation for `renku-python` to the repo. To use it, first [install nix](https://nixos.org/manual/nix/stable/#chap-installation) then try it with 

```
nix-shell -p 'callPackage (fetchTarball https://github.com/swissdatasciencecenter/renku-python/tarball/nix) { version = "0.16.1.dev"; }'
```

The command fetches this branch and builds an environment with the `renku` CLI and library inside. You may also check out the branch locally and just type

```
nix-shell
```

This installs renku in development/editable mode.


Panaetius:
Closing because it's probably quite outdated and not worth the effort to bring up to speed.

Issue #2378, opened 2021-09-29
Enable multiple simultaneous versions of the renku core service


rokroskar:
In order to support projects without migrating them by force, the core service should allow for multiple versions to be live simultaneously.

ableuler:
For specifying the version I'd probably go for a custom request header in this case.

What is less clear to me: would the client need to pick an exact version? In this case the core-service would probably need to advertise the supported versions in a publicly accessible endpoint. Or is it more like a default version that is used in absence of the custom header and if the answer to such a request is that the project is outdated, the client will try with the "old" version, where "old" evolves over time? The second case seems a bit ugly to me, but I'm wondering if it's more appropriate for this use case?

rokroskar:
No I agree - the core service needs to be able to advertise the supported versions. Taking the current situation as an example, it would return

```
{ supported_versions: [1.0, 0.16] }
```

The assumption is that the running version is the latest released patch for that minor version.

Panaetius:
But if we have multiple versions of the service deployed, how would there be an endpoint that can tell which those are?

Also what use would the service saying "I support version 1.0" be if knowing what version a project is on requires asking the service? Either way the service has to clone the project to get that information, so there's no difference between "Ask what version the service supports, then ask what version a project is on, then pick the service" vs. "Ask the service if it supports the project"

Panaetius:
Another thing to consider is that a solution along these lines only works for direct UI<-->RP interaction. if e.g. the endpoint for querying datasets on the KG changes (which renku uses for importing), and we make a new release to fix that, the old release wouldn't be able to import datasets, as it wouldn't know how to query the new endpoint. So depending on the situation, we might only be able to support the most recent version, or to only partially support a past version. I'm not sure how we could properly handle that

rokroskar:
I thought we agreed that the client would check the project for the version and try an API endpoint accordingly. I believe that was @ciyer's suggestion. 

Since there will need to be an API proxy in front, it would be its job to serve the endpoint detailing the available endpoints.

Panaetius:
Well, yes, but that works for pre-1.0.0 and post-1.0.0. It wouldn't work for the future, e.g. 1.0.0 vs 1.1.0.

with pre/post 1.0.0 there is a clear signal in the existence of `.renku/metadata.yml`, but since that is gone and the project version is stored in the db, this won't be possible afterwards.

ableuler:
[...]

PR #2422, opened 2021-10-15
chore: RFC-002 for core-service versioning

Panaetius:
RFC to specify changes for https://github.com/SwissDataScienceCenter/renku-python/issues/2378


mohammad-sdsc (approved):
🚀

PR #2425, opened 2021-10-20
feat(core): add renku workflow loop command

### LINKED ISSUE #2291: Add `renku workflow loop` command
Implement a `renku workflow loop` command, reusing as much as possible from `renku workflow execute`. Follow the spec in RFC-001

vigsterkr:
# Description

fix #2291

iter


renku/cli/workflow.py
```diff
@@ -1058,3 +1058,51 @@ def visualize(sources, columns, exclude_files, ascii, interactive, no_color, pag
...
+    "-p",
+    "--provider",
+    default="cwltool",
+    show_default=True,
+    type=click.Choice(Proxy(_available_workflow_providers), case_sensitive=False),
+    help="The workflow engine to use.",
+)
+@click.option("config", "-c", "--config", metavar="<config file>", help="YAML file containing config for the provider.")
+@click.argument("name_or_id", required=True)
+def loop(name_or_id, mapping, dry_run, provider, config):
```
Panaetius, line 1089: Command also needs to be documented at the top of this file (as that's what actually ends up on RTD). Ideally with an asciinema recording (i used `asciinema rec -i 5`, with PS1="\$ " in my .bash_profile, a terminal size of 55x102 (hxw) and asciicast2gif to create gifs, e.g. `asciicast2gif -w 102 -h 55 execute_plan.rec execute_plan.gif`)

Panaetius (changes requested):
This looks really nice, I think many of our users will love this! :rocket: 

Could you also add a short summary to https://github.com/SwissDataScienceCenter/renku-python/blob/1.0.0-release-notes/renku-release-notes-1.0.0.md?plain=1#L53 please?

rokroskar:
Could you add a nice PR description so we can ask others to comment on the UX?

vigsterkr:
> Could you add a nice PR description so we can ask others to comment on the UX?

the asciinema is good enough description? :)

vigsterkr:
> This looks really nice, I think many of our users will love this! 🚀
> 
> Could you also add a short summary to https://github.com/SwissDataScienceCenter/renku-python/blob/1.0.0-release-notes/renku-release-notes-1.0.0.md?plain=1#L53 please?

sorry i should have done a PR, but i've just directly pushed it in:
https://github.com/SwissDataScienceCenter/renku-python/commit/dcb9a7a20aebe2b1e2aef725b5f80887399dd0fe

rokroskar:
sorry to bring this up late in the game, but the name of this command is a bit weird. It's not really a loop - it's a... matrix? 

The docstring needs some help:

```
renku workflow loop -h                                                                                                                                                                                                               ~/Projects/renku-demos/cli-demo
Usage: renku workflow loop [OPTIONS] NAME_OR_ID

  Generate and execute a set of workflows based on provided set of parameters.

Options:
  --mapping <file>            YAML file containing parameter mappings to be
                              used.
  -n, --dry-run               Print the generated plans with their parameters
                              instead of executing.  [default: False]
  -p, --provider [cwltool]    The workflow engine to use.  [default: cwltool]
  -c, --config <config file>  YAML file containing config for the provider.
  -h, --help                  Show this message and exit.
```

why can't I specify a parameter range on the command-line?

vigsterkr:
yeah it has somewhat changed from it's original design but the loop is short and somewhat covers what happens: loops through the possible values. `renku workflow matrix`... i'm not so sure if that's better than loop :>

regarding setting params from CLI args: the team discussed this and we felt that for this command having `--set param1=[LONG LIST OF VALUES] --set param2=[LONG LIST OF VALUES]...` would be rather cumbersome from the UX perspective, hence we opted for cfg file only as an input.

vigsterkr:
but maybe @SwissDataScienceCenter/poc has some insights about this as well: 
> regarding setting params from CLI args: the team discussed this and we felt that for this command having `--set param1=[LONG LIST OF VALUES] --set param2=[LONG LIST OF VALUES]...` would be rather cumbersome from the UX perspective, hence we opted for cfg file only as an input.

rokroskar:
Right, I seem to recall that conversations... Still it seems like at some point it would be nice to have the option so you can quickly try it instead of having to make a file and commit it etc. That can be done in a future iteration though.

Panaetius:
> yeah it has somewhat changed from it's original design but the loop is short and somewhat covers what happens: loops through the possible values. `renku workflow matrix`... i'm not so sure if that's better than loop :>

`renku workflow execute-many-times-but-with-different-parameters` ?

If we had support for some evaluation metric, I guess it'd be a grid search.
Going with CWL lingo, it'd be [`renku workflow scatter`](https://www.commonwl.org/user_guide/23-scatter-workflow/index.html). Other options that come to mind would be `distribute` (this might be confusing though) or `iterate`

vigsterkr:
> Right, I seem to recall that conversation... Still it seems like at some point it would be nice to have the option so you can quickly try it instead of having to make a file and commit it etc. That can be done in a future iteration though.

it's pretty easy to add this CLI args feature now, the question is should it be `--set`? i feel set is not really what we are looking for... maaaaybe `--map` ?

Panaetius:
Could you add a delay at the end of the gif so it doesn't repeat immediately? You can use `convert loop.gif '(' -clone 0 ')' -swap 0 +delete '(' +clone -set delay 500 ')' +swap +delete loop.delay.gif` with image-magick to set a 5s delay

gavin-k-lee:
re: cfg file vs `--set-param=[LONGLIST]` - I am personally in favour of a cfg file. I often have fat-finger errors when running things at the command line and since `renku run` (and its derivatives) does things which aren't easily undoable, I'd find it easier to calmly check the cfg file. I think if one wanted to run a matrix of 10 different `renku run`s a file to commit would not be a big hurdle.

rokroskar:
I think we absolutely need the cfg file - but I'd like to have the option to quickly try something out via the command line.

gavin-k-lee:
I'm just testing it out and I guess the flow is this: 1) create an initial renku run, with one set of parameters; 2) iterate that plan using another set of parameters. Is there a strong reason to have to do step 1) first? Would it make sense to simply define a config file and iterate without doing the base case? Or do you necessarily need some metadata from the original run?

rokroskar:
hmm that's a good question @gavin-k-lee - I could see using it both ways, but in personal experience it's very easy to shoot off 1000 jobs that fail because of a typo somewhere... so it's always safer to have one execution that you know for a fact works. I guess this way we're implicitly pushing people to do that.

vigsterkr:
> I'm just testing it out and I guess the flow is this: 1) create an initial renku run, with one set of parameters; 2) iterate that plan using another set of parameters. Is there a strong reason to have to do step 1) first? Would it make sense to simply define a config file and iterate without doing the base case? Or do you necessarily need some metadata from the original run?

well currently the iterate really is like the semantics of execute, i.e. in a previous run it has been already identified what's the input, output, parameters. note if there's no such info available this would need to look like something along the renku run... and there comes a lot of question how would you hijack the execution coz your script doesn't know anything about list of params... if you see what i mean. so yeah with the current architecture of renku workflow iterate we rely on the fact that the workflow has been already executed once, i.e. on the metadata of that workflow

gavin-k-lee:
Ok gotcha. I think that flow is fine but it might be worthwhile to stress that an existing plan needs to exist (in order to re-execute).

renku/cli/workflow.py
```diff
@@ -111,6 +111,55 @@
...
+could be used. This sub-command is basically conducting a 'grid search'-like
+execution of a Plan, with parameter-sets provided by the user.
+
+.. code-block:: console
+
+    $ renku workflow iterate --map parameter-1=[1,2,3] \
+            --map parameter-2=[10,20] my-run
+
+The set of possible values for a parameter can be given by ``--map`` command
+line argument or or by specifying them in a values YAML file and passing that
```
Panaetius, line 131:
```suggestion
line argument or by specifying them in a values YAML file and passing that
```

renku/cli/workflow.py
```diff
@@ -111,6 +111,55 @@
...
+using ``--mapping``.
+
+By default ``renku workflow iterate`` will execute all the combination of the
+given parameters' list of possible values. Sometimes it is desired that instead
+of all the combination of possible values, a specific tuple of values are
+executed. This could be done by marking the parameters that should be bind
```
Panaetius, line 137:
```suggestion
executed. This could be done by marking the parameters that should be bound
```

renku/cli/workflow.py
```diff
@@ -111,6 +111,55 @@
...
+together with the ``@tag`` suffix in their names.
+
+.. code-block:: console
+
+    $ renku workflow iterate --map parameter-1@tag1=[1,2,3] \
+            --map parameter-1@tag1=[10,5,30] my-run
+
+This will result in only three distinct execution of the ``my-run`` Plan,
+with the following parameter combinations: ``[(1,10), (2,5), (3,30)]``. It is
+important to note, that parameters that have the same tag, should have the same
```
Panaetius, line 147:
```suggestion
important to note that parameters that have the same tag, should have the same
```

renku/cli/workflow.py
```diff
@@ -111,6 +111,55 @@
...
+number of possible values, i.e. the values list should have the same length.
+
+There's a special template variable for parameter values ``{iter_index}``, which
+can be used to mark each iteration's index in a value of a parameter. The template
+variable is going to be substituted with the iteration index (0, 1, 2, ...).
+
+.. code-block:: console
+
+    $ renku workflow iterate --map parameter-1@tag1=[10,20,30] \
```
Panaetius, line 156:
```suggestion
    $ renku workflow iterate --map parameter-1=[10,20,30] \
```
tag is not relevant in this example, right?

tests/cli/test_workflow.py
```diff
@@ -785,3 +786,75 @@ def test_workflow_compose_execute(runner, project, run_shell, client):
...
+                "input-2": ["environment.yml", "Dockerfile", "requirements.txt"],
+                "n-1": ["3", "4"],
+                "output-3": "output_{iter_index}",
+            },
+            6,
+        ),
+        (
+            "head -n 1 Dockerfile > output",
+            {
+                "input-2@tag1": ["environment.yml", "Dockerfile", "requirements.txt"],
```
Panaetius, line 809: I think one last case we need to test is mixing different tags, like `tag1` and `tag2` in the same command.

Panaetius (changes requested):

mohammad-sdsc (approved):
Thank you!

Panaetius (approved):

Issue #338, opened 2018-10-31
provide an overview of renku commands in git history with option to undo


rokroskar:
**Is your feature request related to a problem? Please describe.**

Undoing the result of a `renku` command can be mind-boggling for new users. 

**Describe the solution you'd like**

We should provide functionality for 

a) seeing a listing of the `renku` commands that have been run (i.e. a selective `git log`
b) reverting a particular `renku` action, with the caveat that we need to show and somehow handle the dependencies of such a revert. 

**Describe alternatives you've considered**

Document manual rebase/revert

jirikuncar:
@ciyer wrote in #440

**Is your feature request related to a problem? Please describe.**
Undoing the effect of renku commands in the repository is tedious. It would nice if this were simpler.

**Describe the solution you'd like**
Add a command `renku undo [--force]`:

Undo the last renku operation

- if working dir is dirty, ask user to stash changes or clean
- if the last commit was not a renku operation, fail with a message that undo can only undo renku operations
- if the last commit has been pushed to its tracking branch (assuming this is easy to detect), fail unless the force flag is turned on.
- if all sanity checks pass, reset HEAD to HEAD~1 and clean working dir. 

**Describe alternatives you've considered**
Do the same operation by hand.

PR #2426, opened 2021-10-20
feat(cli): add renku rollback command

### LINKED ISSUE #440: renku undo command
Undoing the effect of renku commands in the repository is tedious. It would nice if this were simpler.

Add a command `renku undo [--force]`:

Undo the last renku operation

- if working dir is dirty, ask user to stash changes or clean
- if the last commit was not a renku operation, fail with a message that undo can only undo renku operations
- if the last commit has been pushed to its tracking branch (assuming this is easy to detect), fail unless the force flag is turned on.
- if all sanity checks pass, reset HEAD to HEAD~1 and clean working dir. 

Do the same operation by hand.

### LINKED ISSUE #2351: add a `renku rollback` command
With new metadata, triples generator can properly support `git reset` in repositories (and maybe even history rewrites).

This allows users to undo changes that they did without us having to deal with `invalidatedBy` triples and keeping metadata consistent.

Since some of our users might be scared of doing `git reset` and holding there hand a bit could be helpful, we should implement a `renku rollback` command that lists commits (maybe differentiating between renku and non-renku commits), allows a user to pick one (maybe show 10 last commits with numbers 1-10 and also a `more` option to show the next 10 commits ?), then show the user the changes that would be made (Additions/Modifications/Deletions), and if they confirm, do a `git reset` under the hood.

Panaetius:
closes #2351
closes #440

adds a `renku rollback` command. The user gets presented with a (paginated) history of renku actions and can pick a point in time to return to, they are then shown all entities and files that would be changed by this action and, if they confirm, the repo is reset to that state.

Also removes the need for most renku commands to create multiple commits (with the exception of migrations).

Can be tested in https://renkulab.io/projects/ralf.grubenmann/test-rollback

Command to run: `renku rollback`


renku/cli/rollback.py
```diff
@@ -0,0 +1,86 @@
...
+    Files:
+
+    Files that would be restored:
+            blabla
+
+    Files that would be removed:
+            asdasf
+
+    Proceed? [y/N]: y
+
```
rokroskar, line 63: this is the old text

one nit: I noticed in the new version that there isn't any indentation under `Metadata:`, `Files:` etc. I think the indentation is nice to be able to parse the information

rokroskar:
This looks good (apologies for taking a while to re-review)!

rokroskar (approved):
🔥 ♻️

rokroskar:
This actually also (at least partially) closes #440 and #338

renku/core/commands/rollback.py
```diff
@@ -0,0 +1,294 @@
...
+from renku.core.management.command_builder import inject
+from renku.core.management.command_builder.command import Command
+from renku.core.management.interface.client_dispatcher import IClientDispatcher
+from renku.core.management.interface.database_dispatcher import IDatabaseDispatcher
+from renku.core.models.dataset import Dataset
+from renku.core.models.provenance.activity import Activity
+from renku.core.models.workflow.plan import AbstractPlan
+from renku.core.utils import communication
+
+CHECKPOINTS_PER_PAGE = 50
```
vigsterkr, line 34: why define twice this static var? see in `renku/cli/rollback.py` above `CHECKPOINTS_PER_PAGE = 10`

vigsterkr (approved):
:shipit:

PR #2436, opened 2021-10-28
fix(core): fix workflow graph generation and CompositePlan view

Panaetius:
Changes printing of CompositePlan to show the full path of parameters in parameter links

There was an issue with removing overridden activities in graph generation if the activities shared more than one generation. Only the first generation would be checked and then later on the activity wouldn't be considered as overridden, as it didn't share all generations with the latter activity (since only the first generation was considered)


mohammad-sdsc (approved):
LGTM, thanks!

PR #2440, opened 2021-11-01
chore: bump pyshacl to 0.17.2

Panaetius:
pin pyshacl


rokroskar:
I think the Github API has changed slightly - when the PR description is empty, they now return `null` rather than an empty string. This made our action that checks for the presence of `/deploy` in the PR description body fail (hence the tests were failing). I added a string to the description and now it's green.

rokroskar (approved):

PR #2443, opened 2021-11-02
feat(service): improve migrations_check performance

### LINKED ISSUE #2432: Optimize check_migrations endpoint to only clone what's necessary
Currently, check_migrations clones the whole repo. To improve speed, we should selectively clone files using the `git archive` approach as per https://github.com/SwissDataScienceCenter/renku-python/blob/a318ed8ad6705f5d3c66a1eed4645a799a2dc6c4/design/002-core-service-api-versioning/002-core-service-api-versioning.md#detecting-if-a-project-is-supported

On top of https://github.com/SwissDataScienceCenter/renku-python/pull/2122

Panaetius:
closes #2432 

the `git archive` approach only works with git+ssh URLs, and we only get HTTPS in the service (Plus OAUTH2 doesn't work with SSH), so I opted for using the gitlab API to get the files.


renku/service/gateways/gitlab_api_provider.py
```diff
@@ -0,0 +1,76 @@
...
+            full_path = target_folder / path
+            tmp_file = tempfile.NamedTemporaryFile(delete=False)
+            tmp_filename = None
+
+            try:
+                project.files.raw(file_path=path, ref=ref, streamed=True, action=tmp_file.write)
+                tmp_filename = tmp_file.name
+                tmp_file.close()
+
+                full_path.parents[0].mkdir(parents=True, exist_ok=True)
```
mohammad-sdsc, line 65: `full_path.parents[0]` -> `full_path.parent`

mohammad-sdsc: Or `target_folder.mkdir(exist_ok)` outside the for block.

renku/service/gateways/gitlab_api_provider.py
```diff
@@ -0,0 +1,76 @@
...
+
+        git_data = GitURL.parse(remote)
+        gl = gitlab.Gitlab(git_data.instance_url, private_token=token)
+        project = gl.projects.get(f"{git_data.owner}/{git_data.name}")
+
+        result_paths = []
+
+        for path in paths:
+            full_path = target_folder / path
+            tmp_file = tempfile.NamedTemporaryFile(delete=False)
```
mohammad-sdsc, line 57: Do we need a tempfile? Couldn't we just use normal files:
```
try:
    with open(full_path, 'w') as f:
        ... action=f.write ...
except gitlab.GitlabGetError:
   ...
   renku.core.utils.os.delete_file(full_path)
```

renku/service/gateways/gitlab_api_provider.py
```diff
@@ -0,0 +1,76 @@
...
+            try:
+                project.files.raw(file_path=path, ref=ref, streamed=True, action=tmp_file.write)
+                tmp_filename = tmp_file.name
+                tmp_file.close()
+
+                full_path.parents[0].mkdir(parents=True, exist_ok=True)
+                shutil.copy2(tmp_filename, full_path)
+
+                result_paths.append(full_path)
+            except gitlab.GitlabGetError:
```
mohammad-sdsc, line 69: We probably need to close the file here: `tmp_file.close()`

mohammad-sdsc (dismissed):

mohammad-sdsc (commented):
Sorry, I approved this by mistake and can't find how to dismiss it :|.

renku/service/gateways/gitlab_api_provider.py
```diff
@@ -0,0 +1,68 @@
...
+            full_path.parent.mkdir(parents=True, exist_ok=True)
+
+            try:
+                with open(full_path, "w") as f:
+                    project.files.raw(file_path=path, ref=ref, streamed=True, action=f.write)
+
+                result_paths.append(full_path)
+            except gitlab.GitlabGetError:
+                if fail_on_missing:
+                    raise errors.NotFound(path)
```
mohammad-sdsc, line 66: Shouldn't we delete `full_path` here in case it can't be found in the repo? I guess having empty `root` and `project` files might cause some errors.

renku/service/gateways/gitlab_api_provider.py
```diff
@@ -0,0 +1,68 @@
...
+
+                continue
```
mohammad-sdsc, line 68: This is not needed.

renku/service/gateways/gitlab_api_provider.py
Panaetius: good point

mohammad-sdsc (dismissed):
Thank you!

mohammad-sdsc (dismissed):

mohammad-sdsc (approved):
🚀

PR #2444, opened 2021-11-02
fix(core): add dummy metadata.yml for backwards compatibility

Panaetius:
Removal of `metadata.yml` removed forward incompatibility of old versions.

This PR changes behavior to add a dummy `metadata.yml` that makes old versions incompatible with 1.0.0


vigsterkr (approved):
:shipit:

vigsterkr (approved):

vigsterkr (approved):

PR #2448, opened 2021-11-04
fix(cli): fix renku run called with absolute path to executable outside repo

### LINKED ISSUE #2445: PlanFactory considers files outside of the project as potential inputs
On https://github.com/SwissDataScienceCenter/renku-python/blob/master/renku/core/management/workflow/plan_factory.py#L119 we check if the first part of a command line passed to `renku run` is an existing file, and if so, treat it as an input (instead of as a command). This is to enable tracking e.g. `script.sh` as an input in `renku run script.sh`.

If we do `renku run bash`, since `bash` does not exist in the project, `bash` would be correctly identified as a command, not an input. But if we do `renku run /bin/bash`, the code in question will see that `/bin/bash` exist, try to treat it as an input and then try and pull it from lfs, which obviously fails.

We should change the check to only consider files inside the project as inputs.

Panaetius:
closes #2445


renku/core/management/workflow/plan_factory.py
```diff
@@ -125,15 +126,15 @@ def split_command_and_args(self):
...
+        while args and re.match(self._RE_SUBCOMMAND, args[0]) and not self.is_existing_path(args[0])[0]:
             cmd.append(args.pop(0))
 
         return cmd, args
 
     def is_existing_path(self, candidate, ignore=None):
         """Return a path instance if it exists in current directory."""
         if ignore and candidate in ignore:
-            return
+            return None, False
```
mohammad-sdsc, line 137: Should we create a new functions that just does this check since most of the time we only use the first returned value?

mohammad-sdsc (commented):
Made a single comment, looks good otherwise.

renku/core/management/workflow/plan_factory.py
Panaetius: I had that at one point, but then it meant doing the check twice, once in `is_existing_path` and once in `is_relative_to_client` or whatever it'd be called. 

Come to think of it, right now the `is_existing_path` does more than the name implies. Probably should be properly refactored a bit to make it more clear what it actually does. `"""Return a path instance if it exists in current directory."""` is definitely not what this does 🙂 It's more `resolve_path()` or something like that if I read it correctly.

Feel free to change it if you want

mohammad-sdsc: Will do!

vigsterkr (approved):
🚢

PR #2452, opened 2021-11-05
fix(core): fix creation of output folders on rerun/update

Panaetius:
Fixes marking existing folders with `create_folder=True` so they are created before rerunning workflows.


mohammad-sdsc (approved):
LGTM, thank you!

PR #2459, opened 2021-11-09
fix(core): encoding format for output

vigsterkr:
# Description

in one scenario (generated output from script) the `CommandOutput`'s `encoding_format` parameter is not set currently in the metadata that will cause errors when trying to run with CWL backend (current default).

This bug has been found by @gavin-k-lee see: https://renkulab.io/gitlab/lee.gavin.k/debug-workflow repo for the workflows


mohammad-sdsc (approved):
LTGM, thanks!

PR #2462, opened 2021-11-11
feat(core): add toil provider

### LINKED ISSUE #2391: Add workflow execute toil plugin
Add a toil plugin to run workflows in HPC clusters (when logged in on the cluster)

vigsterkr:
# Description

This PR introduces a toil based provider as well as refactors the IWorkflowProvider, so that providers are actually executing a DAG of workflows instead of a (Composite)Plan.

fixes #2391


renku/core/management/workflow/providers/toil.py
```diff
@@ -0,0 +1,244 @@
...
+
+        # construct cmd
+        cmd = []
+
+        if self._workflow.command:
+            cmd.extend(self._workflow.command.split(" "))
+
+        arguments = itertools.chain(self._workflow.inputs, self._workflow.outputs, self._workflow.parameters)
+
+        arguments = filter(lambda x: x.position and (not hasattr(x, "mapped_to") or x.mapped_to is None), arguments)
```
Panaetius, line 104:
```suggestion
        arguments = filter(lambda x: x.position and not getattr(x, "mapped_to", None), arguments)
```

renku/core/management/workflow/providers/toil.py
```diff
@@ -0,0 +1,244 @@
...
+
+        outputs = list()
+        try:
+            with Toil(options) as toil:
+                start_wfs = next(nx.topological_generations(dag))
+                rootJob = Job.wrapJobFn(initialize_jobs, start_wfs, basedir, dag)
+
+                job_outputs = toil.start(rootJob)
+
+                num_outputs = functools.reduce(lambda a, b: a + len(b.values()), job_outputs, 0)
```
Panaetius, line 224:
```suggestion
                num_outputs = sum(map(len, job_outputs))
```
?

setup.py
```diff
@@ -139,7 +138,7 @@ def run(self):
         "sphinx-rtd-theme>=0.5.0,<1.1",
         "sphinxcontrib-spelling==7.*",
     ],
-    "runner": ["cwlref-runner==1.0"],
+    "runner": ["cwlref-runner==1.0", "toil==5.5.0"],
```
Panaetius, line 141: Should we add this to install requires instead? 
Or if we want to make it optional, add a "toil" extras so one can do `pip install renku[toil]` for renku with toil?

Panaetius (changes requested):
This looks really nice, must have been a pain to implement! Thank you!

renku/core/management/workflow/providers/toil.py
vigsterkr: does `map(..., job_outputs)` iterates through the values of `job_outputs` instead of the `(k,v)` tuple?

setup.py
vigsterkr: yeah actually i was thinking of the latter, i.e. toil extras... but basically i would rename the `runner` then to `cwltool` or `cwl` instead of `runner`

renku/core/management/workflow/providers/toil.py
Panaetius: oh right, I missed the `.values()`. then `num_outputs = sum(map(lambda x: len(x.values()), job_outputs))`

in any case, I think using `sum()` makes the intention more clear.

setup.py
Panaetius: I don't think we need cwlref-runner. I've removed the whole `runner` extras in the poetry PR.
All it does is install an executable `cwl-runner` that references `cwltool` so a user can do e.g. `cwl-runner workflow.cwl`. But we don't use it anywhere in code (it used to be used in some CI a long time ago), so I see no reason keeping it.

vigsterkr: @Panaetius i've added `toil` extras as an optional extra, i.e. it's not part of `all`, hence toil based execution/rerun/update tests are not being tested on CI. should we add to one of the CI jobs toil as well, so that way its being tested?

Panaetius: I'd just add it to all as well. The cache is based on `all` and all tests should use the same cache, so that makes sense (and users don't usually do `pip install .[all]`)

Panaetius: And don't forget to update the lock file with `poetry.lock`

vigsterkr: `poetry lock --no-update` updates quite a lot of dependencies. any other suggested way to update the `poetry.lock` or it's fine :)

Panaetius: I think that's just toil and all its subdependencies.

Panaetius (approved):

Panaetius (approved):

Panaetius (approved):
:rocket:

PR #2468, opened 2021-11-16
feat(service): multiple versions deployment

### LINKED ISSUE #2435: Change helm chart to deploy multiple versions side by side
We currently have the renku-core deployment in https://github.com/SwissDataScienceCenter/renku-python/blob/master/helm-chart/renku-core/templates/deployment.yaml
We need to change this to have multiple deployments with different renku-core versions each, to deploy multiple versions of renku-python side-by-side.

In addition, we need a ConfigMap that enumerates all the deployed (metadata-)versions in a renku instance.

See https://github.com/SwissDataScienceCenter/renku-python/blob/a318ed8ad6705f5d3c66a1eed4645a799a2dc6c4/design/002-core-service-api-versioning/002-core-service-api-versioning.md#deployment for more details.

mohammad-sdsc:
# Description

Different versions are defined in `helm-chart/renku-core/values.yaml::versions` and a service/deployment pair is created for each version.
An Nginx reverse-proxy is added that routes request to each of these services. I also serves a list of all deployment on `/renku/versions` endpoint.

This can be tested on https://mohammad.dev.renku.ch/api/renku/version, https://mohammad.dev.renku.ch/api/renku/8/version.

Fixes #2435

/deploy renku=448c938d4d4a4b263857fba8ff2d402f72ad9fba


helm-chart/renku-core/templates/configmap.yaml
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    http {
+      upstream v8 {
+        server {{ include "renku-core.fullname" . }}-v8;
+      }
+
+      upstream v9 {
+        server {{ include "renku-core.fullname" . }}-v9;
+      }
```
Panaetius, line 21: I think we can template this with `range` based on the values as well

helm-chart/renku-core/templates/configmap.yaml
```diff
@@ -0,0 +1,59 @@
...
+          proxy_set_header Host $host;
+          proxy_pass http://v8;
+        }
+
+        location /renku/9 {
+          rewrite /renku/9/(.*) /renku/$1  break;
+          proxy_set_header Host $host;
+          proxy_pass http://v9;
+        }
+
```
Panaetius, line 48: Same here, can be templated

helm-chart/renku-core/templates/deployment.yaml
```diff
@@ -106,113 +111,150 @@ spec:
...
         {{- toYaml . | nindent 8 }}
     {{- end }}
-    {{- with .Values.tolerations }}
+    {{- with $.Values.tolerations }}
       tolerations:
         {{- toYaml . | nindent 8 }}
     {{- end }}
+{{ end }}
+---
+apiVersion: apps/v1
```
Panaetius, line 226: Should we split this file in two, one for the core service, one for NginX?

helm-chart/renku-core/values.schema.json
```diff
@@ -209,4 +208,4 @@
     ],
     "title": "Values",
     "type": "object"
-  }
\ No newline at end of file
+  }
```
Panaetius, line 211: We should add `versions` (and its properties) to the schema as well so the values file gets properly validated by helm

Panaetius (changes requested):
Thank you, this is really nice. There are still a few optimizations of the chart, but the approach/structure is great!

helm-chart/renku-core/templates/deployment.yaml
```diff
@@ -51,48 +56,48 @@ spec:
           {{ end }}
           env:
             - name: REDIS_HOST
-              value: {{ include "call-nested" (list . "redis" "redis.fullname") }}-master
+              value: {{ include "call-nested" (list $ "redis" "redis.fullname") }}-master
             - name: REDIS_PORT
               value: "6379"
             - name: REDIS_DATABASE
               value: "0"
```
ableuler, line 63: Are all versions of the service using the same redis db (ie for example by prefixing "their" keys)?

helm-chart/renku-core/templates/configmap.yaml
```diff
@@ -0,0 +1,55 @@
...
+        location /renku/{{ $version.prefix }} {
+          rewrite /renku/{{ $version.prefix }}/(.*) /renku/$1  break;
+          proxy_set_header Host $host;
+          proxy_pass http://{{ $version.name }};
+        }
+        {{- end }}
+
+        location /renku {
+          proxy_set_header Host $host;
+          proxy_pass http://v9;
```
ableuler, line 45: I think this should not be hardcoded. Either you can just use the first element in the versions array or add another chart value to specify the default version.

helm-chart/renku-core/templates/deployment-nginx.yaml
```diff
@@ -0,0 +1,35 @@
...
+    matchLabels:
+      app.kubernetes.io/name: {{ include "renku-core.fullname" $ }}-nginx
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: {{ include "renku-core.fullname" $ }}-nginx
+    spec:
+      containers:
+        - name: nginx
```
ableuler, line 18: Please add a [security context](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#securitycontext-v1-core) here and give nginx only what's absolutely needed. Ideally this will at least include 
`allowPrivilegeEscalation = false`
`capabilites = "drop all"`
`runAsNonRoot = false`
I _think_ this should be possible (after swapping to a different port, see below), but I'm not 100% sure.

helm-chart/renku-core/templates/deployment-nginx.yaml
```diff
@@ -0,0 +1,35 @@
...
+          image: nginx:alpine
+          imagePullPolicy: IfNotPresent
+          ports:
+          - containerPort: 80
```
ableuler, line 22: Listening on port 80 needs some privileges that we'd like to restrict (see above). Just move it to some port higher than 1000 and switch the target port in the k8s service spec

helm-chart/renku-core/templates/deployment-nginx.yaml
```diff
@@ -0,0 +1,35 @@
...
+          volumeMounts:
+            - name: nginx-volume
+              readOnly: true
+              mountPath: /etc/nginx/nginx.conf
+              subPath: nginx.conf
+            - name: nginx-volume
+              readOnly: true
+              mountPath: /usr/share/nginx/html/deployments.json
```
ableuler, line 30: maybe call it `versions.json`? 
(a matter of taste more than anything else)

helm-chart/renku-core/templates/service-nginx.yaml
```diff
@@ -0,0 +1,15 @@
...
+  labels:
+{{ include "renku-core.labels" $ | indent 4 }}
+spec:
+  type: ClusterIP
+  selector:
+    app.kubernetes.io/name: {{ include "renku-core.fullname" $ }}-nginx
+  ports:
+  - protocol: TCP
+    port: 80
+    targetPort: 80
```
ableuler, line 14: Using some unprivileged port here would be nice.

helm-chart/renku-core/templates/configmap.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+      server {
+        listen 80;
+
+        server_name {{ include "renku-core.fullname" . }};
+
+        sendfile on;
+        tcp_nopush on;
+
+        location /renku/deployments {
```
ableuler, line 29: maybe `/renku/versions`?

helm-chart/renku-core/templates/deployment.yaml
```diff
@@ -1,48 +1,53 @@
...
     {{- end }}
       volumes:
         - name: shared-volume
           emptyDir: {}
       containers:
-      {{ if .Values.metrics.enabled }}
-        - name: {{ .Chart.Name}}-rqmetrics
-          image: "{{ .Values.metrics.image.repository }}:{{ .Values.metrics.image.tag }}"
-          imagePullPolicy: {{ .Values.metrics.image.pullPolicy }}
+      {{ if $.Values.metrics.enabled }}
```
ableuler, line 39: It would be nice to also apply a more restrictive security context here (same as for the new nginx deployment.yaml). This should be straight forward as the core service is already listening on port 8080.

ableuler (changes requested):
This looks really good to me. As far as I can tell the NGINX config has everything it needs, however I wouldn't call myself a nginx expert either.

Most of my comments are related to security standards of running these services. This is an area that we'll need to improve all across our services very soon, so I think it makes sense to address them now for the core-service as you're already working on it.

helm-chart/renku-core/templates/deployment.yaml
mohammad-sdsc: Right. They don't prefix anything because a valid project only exists in one of those deployments and so there is only one redis key for it at any moment.

helm-chart/renku-core/templates/deployment-nginx.yaml
mohammad-sdsc: If you meant `runAsNonRoot = true` (false is the default) then it can't be run with the default nginx image since it runs as root.

ableuler: yes, that's indeed what I meant, sorry. Can you try with https://hub.docker.com/r/bitnami/nginx
Usually all the images published by bitnami run as non-root.

helm-chart/renku-core/templates/deployment.yaml
```diff
@@ -1,98 +1,109 @@
...
-        - name: {{ .Chart.Name }}
-          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
-          imagePullPolicy: {{ .Values.image.pullPolicy }}
-          {{ if .Values.debug }}
-          args: ["api", "--debug"]
+        - name: {{ $.Chart.Name }}
+          image: "{{ $version.image.repository }}:{{ $version.image.tag }}"
+          imagePullPolicy: {{ $version.image.pullPolicy }}
           securityContext:
+            {{- if $.Values.debug }}
```
ableuler, line 51: Why only if debug is set to true? I'd either always run as user 1000 or add the `runAsNonRoot = true` here as well.

helm-chart/renku-core/templates/deployment-nginx.yaml
mohammad-sdsc: It worked well. Some configuration changes were needed since this image uses a different config path.

helm-chart/renku-core/templates/deployment.yaml
mohammad-sdsc: The library that we use to interface with redis supports namespaces. We decided to add namespaces here to avoid surprises in future. There is still one redis instance but its keys are namespaced.

helm-chart/renku-core/templates/deployment.yaml
mohammad-sdsc: Setting DROP ALL causes some issues here and the service won't work as expected. I couldn't find what are the exact required capabilities since there is no logs/error messages. I removed DROP ALL from the core service.

Panaetius (approved):

helm-chart/renku-core/templates/deployment.yaml
ableuler: That's ok, for the moment making sure that containers run as non-root is the main concern that I have.

helm-chart/renku-core/templates/deployment.yaml
ableuler: 👍

ableuler (approved):

PR #2471, opened 2021-11-17
chore: move to pyproject and poetry

### LINKED ISSUE #2405: Switch to pyproject.toml
As per https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/?#fallback-behaviour setup.py is already just fallback behavior and it looks like support for `pyproject.toml` in pip is pretty mature right now. We've already encountered issues with https://github.com/SwissDataScienceCenter/renku-python/pull/2404 and I think in the future there is a high chance of further issues as `pyproject.toml` becomes the standard and support for `setuptools` slowly becomes obsolete.

As such we should 
- update our `pyproject.toml` to contain all the relevant information from our setup.py, with most likely poetry as the build system
- add config that's outside setup.py to the pyproject.toml (e.g. pytest.ini config)  where possible
- have a transitional phase where we still have setup.py, by automatically converting from poetry to setup.py using something like Dephell in a Github action.

And alternative would be to add all settings to pyproject.toml but still use setuptools as the build system (see https://setuptools.pypa.io/en/latest/build_meta.html#how-to-use-it ). This is the a bit more conservative approach.

Benefits of switching to Poetry for me are faster dependency resolution and more control over pinning packages (similar to how pipenv does it without having all the baggage pipenv has).

Panaetius:
closes #2405 

Moves all configuration to pyproject.toml
Moves building/dependency management to poetry
Gets rid of setup.py
Updates Github Actions to work with poetry

Also:
Adds caching of python dependencies between runs in the github actions, to speed things up a bit (and reduce bandwidth)
Reduces exponential backoff in our `requests` wrapper so the worst case is not 17 minutes but 6seconds.
Formatter the github actions to have consistent indent
Added some missing __init__.py (poetry threw warnings when building)

Note:
version.py.tmpl is gone, `poetry-dynamic-versioning` patches version.py directly. `0.0.0` is just a placeholder. The way it works is that before a `poetry build` or `poetry install`, it patches version.py with the correct version, and after in un-patches it again. This has the downside that the version locally is always 0.0.0, only renku installed in the env and published packages have the version actually set. I didn't notice any issues with running tests or commands with this, so I think it's fine.


Panaetius:
For an example of the cache:
Without cache: https://github.com/SwissDataScienceCenter/renku-python/runs/4264883858?check_suite_focus=true
With cache: https://github.com/SwissDataScienceCenter/renku-python/runs/4265611395?check_suite_focus=true#step:5:10 (rerun of the previous execution)

vigsterkr (approved):
:shipit:

vigsterkr (approved):

vigsterkr (approved):
🚢 🚀

PR #2481, opened 2021-11-24
feat(core): error-resilience in workflow migrations

mohammad-sdsc:
# Description

Fix an issue with workflow migrations. Add error-resilience.


Panaetius (approved):
Thank you for looking into this!

PR #2482, opened 2021-11-25
fix(core): improve renku status performance

Panaetius:
Better early filtering for overridden activities.

Get object-hashes in batches.


renku/core/metadata/repository.py
```diff
@@ -423,6 +425,65 @@ def get_content_from_submodules():
 
         raise errors.ExportError(f"File not found in the repository: '{revision}/{checksum}:{path}'")
 
+    def get_object_hashes(self, paths: List[Union[Path, str]], revision: str = None) -> Dict[str, str]:
```
mohammad-sdsc, line 428: The returned dictionary from this function does not have consistent keys: For modified objects the key is an absolute path while for the rest it's a relative path.
I believe it should return return the passed input values as keys so that we can reliably get a hash value from it (converting Path to str is ok). Ideally, if a path doesn't exists in the repo then it's hash should be set to None in the returned dict.

renku/core/metadata/repository.py
```diff
@@ -423,6 +425,65 @@ def get_content_from_submodules():
...
+
+        submodule_paths = defaultdict(list)
+        main_repo_paths = []
+
+        if self.submodules:
+            for absolute_path in absolute_paths:
+                for submodule in self.submodules:
+                    try:
+                        Path(absolute_path).relative_to(submodule.path)
+                        submodule_paths[submodule].append(absolute_path)
```
mohammad-sdsc, line 461: I would put this in the `else` block to make the intention of this `try/catch` more clear. Also we can `break` once we find a submodule.

renku/core/metadata/repository.py
```diff
@@ -518,10 +579,69 @@ def get_global_configuration(writable: bool = False) -> "Configuration":
         return Configuration(repository=None, writable=writable)
 
     @staticmethod
-    def hash_object(path: Union[Path, str]) -> str:
+    def list_paths_in_revision(
+        paths: List[Union[Path, str]] = None, revision: str = "HEAD", working_dir="."
+    ) -> List[str]:
+        """List all paths that exist in a revision.
+
+        Warning: ls-tree is sensitive to the current working directory, make sure to set working_dir to repo root.
```
mohammad-sdsc, line 587: It's perhaps cleaner to calculate relative paths to the working_dir in. this function instead of relying on the caller to pass the correct values.

renku/core/metadata/repository.py
```diff
@@ -423,6 +425,65 @@ def get_content_from_submodules():
...
+                        submodule_paths[submodule].append(absolute_path)
+                    except ValueError:
+                        main_repo_paths.append(os.path.relpath(absolute_path, start=self.path))
+        else:
+            main_repo_paths = map(lambda p: os.path.relpath(p, start=self.path), absolute_paths)
+
+        if main_repo_paths:
+            main_repo_paths = Repository.list_paths_in_revision(
+                main_repo_paths, revision=revision, working_dir=self.path
+            )
```
mohammad-sdsc, line 470: The reason to call `list_paths_in_revision` is to make sure that paths exists in the repo (to make rev-parse not failing later), right? If so, can we. use a better name for it to make this more. clear?

renku/core/metadata/repository.py
```diff
@@ -518,10 +579,69 @@ def get_global_configuration(writable: bool = False) -> "Configuration":
...
+                existing_dirs = git.Git().ls_tree(r=revision, name_only=True, d=True).splitlines()
+                return existing_dirs + existing_files
+            except git.GitCommandError as e:
+                raise errors.GitCommandError(
+                    message=f"Git command failed: {str(e)}",
+                    command=e.command,
+                    stdout=e.stdout,
+                    stderr=e.stderr,
+                    status=e.status,
+                ) from e
```
mohammad-sdsc, line 634: We should use inner functions to remove redundant calls to ls_tree. and its error handling.

renku/core/metadata/repository.py
mohammad-sdsc: This function is a bit difficult to follow. Can you please use nested/inner functions to make it more readable?

renku/core/utils/metadata.py
```diff
@@ -101,6 +107,42 @@ def get_modified_activities(
...
+            if outputs.issubset(k):
+                subset_of.append((k, a))
+            elif outputs.issuperset(k):
+                superset_of.append((k, a))
+
+        if not subset_of and not superset_of:
+            relevant_activities[outputs] = activity
+            continue
+
+        subset_of = [(k, a) for k, a in relevant_activities.items() if outputs.issubset(k) or outputs.issuperset(k)]
```
mohammad-sdsc, line 130: This is not needed I believe since we already calculated the correct `subset_of` before.

renku/core/utils/metadata.py
```diff
@@ -101,6 +107,42 @@ def get_modified_activities(
...
+            continue
+
+        older_subsets = [k for k, s in superset_of if activity.ended_at_time > s.ended_at_time]
+
+        for older_subset in older_subsets:
+            del relevant_activities[older_subset]
+
+        relevant_activities[outputs] = activity
+
+    return relevant_activities.values()
```
mohammad-sdsc, line 143: Should we convert this to list?

mohammad-sdsc (changes requested):
Thanks for looking into this! I've made a few comments.

renku/core/metadata/repository.py
Panaetius: I don't think there are redundant calls, there are two calls each for with/without paths, showing directories and showing files, so 4 distinct calls. Moving that to an inner function doesn't really make it more readable.

But the error handling can just be wrapped around the whole function body.

mohammad-sdsc (approved):
LGTM, thank you!

PR #2483, opened 2021-11-26
chore: fix IT due to deleted requirements-devel.txt

Panaetius:
Some test actually depended on requirements-devel.txt, a file that was otherwise irrelevant. I switched it to use poetry.lock instead.


mohammad-sdsc (approved):
Thank you!

PR #2484, opened 2021-11-29
chore: update SHACL publish workflow token

Panaetius:
removes SHACL token in favor of organisation level token.


mohammad-sdsc (approved):
Thank you!

PR #2488, opened 2021-11-29
fix(core): various migration issues

mohammad-sdsc:
-  Various fixes in migrations
- Print warning messages only once for a pair of path and revision


Panaetius (approved):
Thank you, looks great!

PR #2489, opened 2021-12-01
fix(core): prevent catalog from creating tons on separate files

Panaetius:
Prevents OOBucket, OOSet, OOTreeSet and BTree.Length from being stored in separate files.

Reduces files generated in renku metadata by a factor of ~15 (depends on the project)


vigsterkr (approved):
🚀

PR #2508, opened 2021-12-03
fix(core): add missing encodig_format when migrating a project

vigsterkr:
# Description

when migrating a project `CommandInput` and `CommandOutput` should always have the `encoding_format` defined.


renku/core/management/migrations/m_0009__new_metadata_storage.py
```diff
@@ -226,6 +232,10 @@ def extract_run_uuid(run_id: str) -> str:
 
     plan_id = Plan.generate_id(uuid=uuid)
 
+    def get_mime_type(path_str: str) -> List[str]:
+        path = Path(path_str)
+        return [DIRECTORY_MIME_TYPE] if path.exists() and path.is_dir() else ["application/octet-stream"]
```
Panaetius, line 237: since e.g. `output.produces` is either `v9.Entity` or `v9.Collection`, we could fall back to using that if not `path.exists()`.

Or use it directly without `path.is_dir()` (in case it used to be a directory but is a file now)

renku/core/management/migrations/m_0009__new_metadata_storage.py
```diff
@@ -226,6 +232,9 @@ def extract_run_uuid(run_id: str) -> str:
 
     plan_id = Plan.generate_id(uuid=uuid)
 
+    def get_mime_type(entity: Union[Entity, Collection]) -> List[str]:
+        return [DIRECTORY_MIME_TYPE] if isinstance(entity, Collection) else ["application/octet-stream"]
```
Panaetius, line 236: Should be `renku.core.management.migrations.models.v9.Collection` and `renku.core.management.migrations.models.v9.Entity` here. `Collection`/`Entity` as imported at the top are the new ones from `renku.core.models.entity`

Panaetius (approved):
Thank you!

PR #2509, opened 2021-12-03
refactor(core): replace pkg_resources with importlib

### LINKED ISSUE #2417: replace pkg_resource with importlib_metadata
We have many places where we use `pkg_resources`, which is old and according to https://setuptools.pypa.io/en/latest/pkg_resources.html#package-discovery-and-resource-access-using-pkg-resources should not be used anymore. Plus pkg_resources has an overhead of around 0.3s load time, so it slows everything down.

`importlib.metadata` has been backported to python 3.7, so we should switch all usages of `pkg_resources` over to this.

See https://github.com/SwissDataScienceCenter/renku-python/pull/2416/files#diff-83de7e0c43272bbb49c2340d90259ba63ae088c348fefc4858863728472e500fR101-R104 for a usage example for both python 3.7 and 3.8+ support.

Panaetius:
closes #2417 

Note: importlib_resources is sued over importlib.resources in Python 3.8 as well as the 3.8 version isn't compatible with the >=3.9 version but the backport importlib_resources is.


pyproject.toml
```diff
@@ -20,7 +20,7 @@
 
 [tool.poetry]
 name = "renku"
-version = "0.0.0" # placeholder, see poetry-dynamic-versioning
+version = "1.0.0.dev3+g2833b272" # placeholder, see poetry-dynamic-versioning
```
mohammad-sdsc, line 23: Should we keep `0.0.0` format or it's not important?

mohammad-sdsc (dismissed):
LGTM, just made a single comment, feel free to ignore.

pyproject.toml
Panaetius: Well it's not important but cleaner if it's 0.0.0, good catch :)

mohammad-sdsc (approved):
Thanks!

PR #2510, opened 2021-12-03
fix(core): preserve dataset ids for KG migrations

### LINKED ISSUE #2511: Inconsistent dataset ids when migrating a project in KG
**Describe the bug**
In migration to metadata `v9`, we always modify datasets' identifiers to have correct dataset mutation chain for old projects. This works fine when users migrate their projects.
However, when KG wants to index an old project, migrating to `v9` will create new dataset ids which don't exists in the project and therefore project's metadata and its metadata in KG are not consistent.

mohammad-sdsc:
# Description

Adds an `--preserve-identifiers` flag to the `migrate` command which keeps original datasets' identifiers when migrating. This option is intended only for KG during indexing to keep the identifiers consistent with the project's metadata.

Fixes #2511


Panaetius (approved):
Thank you!

PR #2515, opened 2021-12-07
fix(core): import the component in database.get_attribute 

vigsterkr:
# Description

`get_attribute` needs a module import for getting the attribute of the module. make sure that the module gets imported before getting the attribute

### DIFF
--- a/renku/core/metadata/database.py
+++ b/renku/core/metadata/database.py
@@ -53,6 +53,13 @@ NEW = z64  # NOTE: Do not change this value since this is the default when a Per
 PERSISTED = b"1" * 8
 
 
+def _is_module_allowed(module_name: str, type_name: str):
+    """Checks whether it is allowed to import from the given module for security purposes."""
+
+    if module_name not in ["BTrees", "builtins", "datetime", "persistent", "renku", "zc", "zope"]:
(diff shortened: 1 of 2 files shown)


renku/core/metadata/database.py
```diff
@@ -80,9 +86,25 @@ def get_class(type_name: Optional[str]) -> Optional[type]:
...
+
+    module_name = _module_name(object)
+    root_module_name = module_name.split(".")[0]
+
     for component in components:
+        module_name = _module_name(object)
+        if not hasattr(object, component) and f"{module_name}.{component}" not in sys.modules:
+            try:
+                _is_module_allowed(root_module_name, object.__name__)
+                __import__(f"{module_name}.{component}")
```
Panaetius, line 104: Looking at https://docs.python.org/3/library/importlib.html#importlib.__import__ it says `Programmatic importing of modules should use import_module() instead of this function.`

The main difference is that `importlib.import_module` returns the imported thing, so we can use it directly and save the `getattr()` call below

Panaetius (approved):
Thanks for looking into this. Just a minor comment, feel free to ignore

Panaetius (approved):

PR #2521, opened 2021-12-08
refactor(graph): move JSON-LD schemas to their own namespace

### LINKED ISSUE #2256: move JSON-LD schemas out of models namespace
The JSON-LD schemas are an implementation detail for the view layer and there's no reason for them to be in the `renku.core.models` namespace. Right now they're specific to the CLI entirely (for `renku log` ) and as such they should live as a `renku.core.commands` view model specific to the CLI.

mohammad-sdsc:
Fixes #2256


vigsterkr (approved):
🚢

PR #2522, opened 2021-12-09
fix(core): recurse input of workflows when using toil provider

vigsterkr:
# Description

This fixes the execution of a workflow using toil, when the input of a workflow is a directory that contains directories as children.

### DIFF
--- a/renku/core/management/workflow/providers/toil.py
+++ b/renku/core/management/workflow/providers/toil.py
@@ -77,19 +77,23 @@ class AbstractToilJob(Job):
         for p in self._parents_promise:
             parent_inputs.update(p)
 
+        def _read_input(input: str, file_metadata):
+            input_path = Path(input)
+
+            if isinstance(file_metadata, dict):
+                input_path.mkdir(parents=True, exist_ok=True)
+                for path, file_id in file_metadata.items():
+                    _read_input(path, file_id)
+            elif not input_path.exists():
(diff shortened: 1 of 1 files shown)


renku/core/management/workflow/providers/toil.py
```diff
@@ -77,19 +77,23 @@ def run(self, storage):
...
+            input_path = Path(input)
+
             if isinstance(file_metadata, dict):
-                input_path.mkdir(exist_ok=True)
+                input_path.mkdir(parents=True, exist_ok=True)
                 for path, file_id in file_metadata.items():
-                    storage.readGlobalFile(file_id, userPath=path)
-            else:
+                    _read_input(path, file_id)
+            elif not input_path.exists():
```
mohammad-sdsc, line 87: This won't `readGlobalFile` if the `input_path` exists; is this the correct behavior?

mohammad-sdsc (commented):
LGTM, just made two minor comments.

renku/core/management/workflow/providers/toil.py
vigsterkr: yes that's intentional. `readGlobalFile` has no option for overwrite, i.e. it throws an exception if `userPath` already exists; this avoid pulling the given file from the global storage if it's already available.

mohammad-sdsc (approved):
Thank you!

PR #2526, opened 2021-12-10
fix(core): preserve dateCreated when migrating datasets

mohammad-sdsc:
# Description

During a migration, dataset's creation date was set to current date/time instead of the original creation date. This PR fixes this issue and also removed some unused code.


vigsterkr (approved):
🚀

PR #2527, opened 2021-12-13
fix(core): disable interpolation when loading/storing renku config

### LINKED ISSUE #2486: handle value escaping in renku.ini
We use Python ConfigParser to deal with `.renku/renku.ini`, which supports string interpolation https://docs.python.org/3.7/library/configparser.html#interpolation-of-values

While this is technically a feature, it can cause issues for users, e.g.
```
lab = /my%2furl
```
is a validly encoded URL, but ConfigParser will try to interpolate the `%2furl` which fails. To work properly, the percent would have to be escaped (`%%`).

We could turn off interpolation like `ConfigParser(interpolation=None)` which should work on all supported Python versions, at the cost of losing this feature (which no one uses to my knowledge, but might be useful if people were aware of it).

Alternatively, we'd have to make sure `renku config set` correctly escapes values and maybe add a `renku doctor` check for this.

mohammad-sdsc:
# Description

Interpolation is a non-standard INI feature seems to be only implemented by `ConfigParser`. `GitPython` does not use it either. This PR disables this feature.

Fixes #2486

### DIFF
--- a/renku/core/management/config.py
+++ b/renku/core/management/config.py
@@ -83,7 +83,8 @@ class ConfigManagerMixin:
         except ImportError:
             import importlib.resources as importlib_resources
 
-        config = configparser.ConfigParser()
+        # NOTE: Use RawConfigParser because ConfigParser does non-standard INI interpolation of some values
+        config = configparser.RawConfigParser()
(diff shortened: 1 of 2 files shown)


vigsterkr (approved):
🚢

Issue #2394, opened 2021-10-06
Launch Jupyter/Rstudio session with Renku-CLI Locally


olevski:
**Is your feature request related to a problem? Please describe.**
This is not related to a direct problem / bug. But it occurred to me when a user asked "How do I use renku locally?". It is not very easy to answer this question - especially if the user is not very familiar with docker. So I think it would be nice if we had a command called `renku session` in the CLI that will build the docker image from a Renku project locally and run it with docker - mounting the local folder (i.e. git repo) into the docker container and even potentially opening the browser to the link where the session is running. This would give an almost identical experience between using renku on renkulab and locally.

The only problem is that support for windows with this will probably be a pain.

**Describe the solution you'd like**
Add a command in the Renku cli that starts a session locally - assuming that the user has docker installed.

A few subcommands could be used:
- `renku session start` - starts a session from the current renku project
- `renku session stop <session-name>` - stops a renku session that is running
- `renku session list` - list all running renku sessions on the machine for any project

**Describe alternatives you've considered**
Not sure there are any alternatives.

**Additional context**
It would be nice to have the same/similar UX between using renku locally and online.

PR #2529, opened 2021-12-15
feat(core): design of renku session

### LINKED ISSUE #1991: CLI command for launching an interactive session locally
https://github.com/SwissDataScienceCenter/renku/pull/2043 adds some documentation for running renkulab images locally. The commands required to get a nice running session are a bit heavy and incomprehensible to most users - it would be helpful to have some CLI functionality to make this simpler for users. 

## A possible solution

Two commands:

Return the image name:commit for the current state of the project:

```
renku session image 

registry.<renku-host>/<namespace>/<project>:<commit-sha>
```

Start/stop a local docker container with the correct image for the current commit; if the image doesn't exist, build it first. If the session for this commit is already running, connect to it. 

```
renku session start/stop
```

One open question is what happens to the container logs - my opinion is that they should not be sent to the console by default but perhaps be available via another wrapper command.

vigsterkr:
# Description

Fixes #1991 #2394

session

### DIFF
--- a/design/003-interactive-session/003-interactive-session.md
+++ b/design/003-interactive-session/003-interactive-session.md
@@ -38,23 +38,24 @@ that's going to be used for running the interactive session. In case the docker 
 user is going to be prompted whether the user wants to build the docker image, as in some cases it can be
 time consuming.
 
-By using the `--remote [endpoint]` flag an interactive session will be spawned on the specified endpoint. It requires
-that the user has authenticated herself previously with the `renku login` command and aquired the authentication
-token for the endpoint.	If the endpoint is not defined the command will try to assume the endpoint set in the project's
-configuration if available.
+With the `--provider <provider_type>` flag a user can specify session provider to be used for starting
+an interactive session. By default two different providers are shipped with Renku CLI:
+ - docker
+ - renkulab
+The provider specific configuration values can be specified by using the `--config <config.yaml>` flag.
 
 ##### Detailed Parameter Description
 
 ```
 renku session start [OPTIONS]
 
---image <image_name>	override the docker image to be used for the interactive session.
---remote [endpoint]	start a remote interactive session using the JWT aquired for the
-			specific endpoint with renku login.
---cpu <num of cpus>	Specify how much of the available CPU resources a container can use.
---disk <disk space>	Specify the amount of disk space for the container.
---gpu <device>		Specify the GPU allocated for the container.
---memory <max memory>	The maximum amount of memory the container can use.
+--image <image_name>			Override the docker image to be used for the interactive session.
+-p|--provider [docker|renkulab] 	Use the specified backend for starting an interactive session.
+-c|--config   				YAML file containing configuration for the provider.
+--cpu <num of cpus>			Specify how much of the available CPU resources a container can use.
+--disk <disk space>			Specify the amount of disk space for the container.
+--gpu <device>				Specify the GPU allocated for the container.
+--memory <max memory>			The maximum amount of memory the container can use.
 
 ```
 
@@ -85,8 +86,8 @@ detailed information regarding the network port mappings.
 ```
 renku session list [OPTIONS]
 
---remote       List the remote interactive sessions currently running.
-               By default only the local interactive sessions are listed.
+-p|--provider [docker|renkulab] 	Use the specified backend for listing the interactive sessions.
+-c|--config   				YAML file containing configuration for the provider.
 
 ```
 
--- a/design/003-interactive-session/003-interactive-session.md
+++ b/design/003-interactive-session/003-interactive-session.md
@@ -1,24 +1,24 @@
 - Start Date: 2021-12-02
 - Status: Proposed
 
-# Run local interactive sessions
+# Run interactive sessions
 
 ## Summary
 
 Propose a new sub-group command `renku session` that is responsible for
-managing local interactive sessions for a project.
+managing interactive sessions, both local and remote, for a project.
 
 ## Motivation
 
 Renkulab.io provides the option to run interactive sessions of a project. Among other things the session
 contains a jupyter notebook.
 
-The motivation behind this RFC is to provide a simple renku CLI command that basically mimics this, i.e. the
-option to start an interactive session of a project locally.
+The motivation behind this RFC is to provide a simple renku CLI sub-command that basically allows the
+starting and stopping of these interactive sessions.
 
 ## Design Detail
 
-The proposed sub-commands of `renku session`, namely to run interactive sessions locally, heavily relies on
+The proposed sub-commands of `renku session`, namely to run interactive sessions. The local session heavily relies on
 [docker](https://www.docker.com/). Docker is one of the most popular container engines, that eases the development
 and deployment of applications, by off-loading the installation of the runtime environment dependencies of the
 application.
@@ -38,12 +38,19 @@ that's going to be used for running the interactive session. In case the docker 
 user is going to be prompted whether the user wants to build the docker image, as in some cases it can be
 time consuming.
 
+By using the `--remote [endpoint]` flag an interactive session will be spawned on the specified endpoint. It requires
+that the user has authenticated herself previously with the `renku login` command and aquired the authentication
+token for the endpoint.	If the endpoint is not defined the command will try to assume the endpoint set in the project's
+configuration if available.
+
 ##### Detailed Parameter Description
 
 ```
 renku session start [OPTIONS]
 
 --image <image_name>	override the docker image to be used for the interactive session.
+--remote [endpoint]	start a remote interactive session using the JWT aquired for the
+			specific endpoint with renku login.
 
 ```
 
@@ -58,7 +65,7 @@ One can shut down all the currently running interactive sessions at once with th
 ```
 renku session stop [OPTIONS] <name>
 
-<name>
+<name>      Name of the session that the user wants to stop.
 
 --all       Stops all the running containers.
 
@@ -68,6 +75,16 @@ renku session stop [OPTIONS] <name>
 
 The command to list all the currently running local interactive sessions, with the detailed information regarding
 the network port mappings.
+
+##### Detailed Parameter Description
+
+```
+renku session list [OPTIONS]
+
+--remote       List the remote interactive sessions currently running.
+               By default only the local interactive sessions are listed.
+
+```
 
 ## Drawbacks
 
--- a/design/003-interactive-session/003-interactive-session.md
+++ b/design/003-interactive-session/003-interactive-session.md
@@ -11,7 +11,7 @@ managing interactive sessions, both local and remote, for a project.
 ## Motivation
 
 Renkulab.io provides the option to run interactive sessions of a project. Among other things the session
-contains a jupyter notebook.
+contains a jupyter server.
 
 The motivation behind this RFC is to provide a simple renku CLI sub-command that basically allows the
 starting and stopping of these interactive sessions.
--- a/design/003-interactive-session/003-interactive-session.md
+++ b/design/003-interactive-session/003-interactive-session.md
@@ -66,7 +66,7 @@ renku session stop [OPTIONS] <name>
 
 #### renku session list
 
-The command to list all the currently running local interactive sessions, with the detail information regarding
+The command to list all the currently running local interactive sessions, with the detailed information regarding
 the network port mappings.
 
 ## Drawbacks
--- a/design/003-interactive-session/003-interactive-session.md
+++ b/design/003-interactive-session/003-interactive-session.md
@@ -77,7 +77,7 @@ renku session stop [OPTIONS] <name>
 
 #### renku session list
 
-The command to list all the currently running local interactive sessions, with the detailed information regarding
+The command to list all the currently running interactive sessions, with the detailed information regarding
 the network port mappings.
 
 ##### Detailed Parameter Description
@@ -89,6 +89,13 @@ renku session list [OPTIONS]
                By default only the local interactive sessions are listed.
 
 ```
+
+#### renku session open
+
+`renku session open <my_session>` opens the given interactive session in the user's default browser.
+
+Note, it is possible that the user is running renku on a remote machine via SSH. In this case this command
+might fail, unless for example X-Fowarding is not enabled.
 
 ## Drawbacks
 
--- a/design/003-interactive-session/003-interactive-session.md
+++ b/design/003-interactive-session/003-interactive-session.md
@@ -77,8 +77,8 @@ renku session stop [OPTIONS] <name>
 
 #### renku session list
 
-The command to list all the currently running interactive sessions, with the detailed information regarding
-the network port mappings.
+The command to list all the currently running interactive sessions for the given project, with the
+detailed information regarding the network port mappings.
 
 ##### Detailed Parameter Description
 
--- a/design/003-interactive-session/003-interactive-session.md
+++ b/design/003-interactive-session/003-interactive-session.md
@@ -1,0 +1,94 @@
+- Start Date: 2021-12-02
+- Status: Proposed
+
+# Run local interactive sessions
+
+## Summary
+
+Propose a new sub-group command `renku session` that is responsible for
+managing local interactive sessions for a project.
+
+## Motivation
+
+Renkulab.io provides the option to run interactive sessions of a project. Among other things the session
+contains a jupyter notebook.
+
+The motivation behind this RFC is to provide a simple renku CLI command that basically mimics this, i.e. the
+option to start an interactive session of a project locally.
+
+## Design Detail
+
+The proposed sub-commands of `renku session`, namely to run interactive sessions locally, heavily relies on
+[docker](https://www.docker.com/). Docker is one of the most popular container engines, that eases the development
+and deployment of applications, by off-loading the installation of the runtime environment dependencies of the
+application.
+
+In order to start a local interactive session, just as in case of renkulab.io, it is required to provide
+a `Dockerfile`. This is by default provided by `renku init` when creating a renku project, but if the user
+is using a custom project template it might not be present.
+
+### Proposed Commands
+
+#### renku session start
+
+A command to start an interactive session.
+
+By default, the `Dockerfile` present in the given project is going to be used for building the docker image
+that's going to be used for running the interactive session. In case the docker image is not available the
+user is going to be prompted whether the user wants to build the docker image, as in some cases it can be
+time consuming.
+
+##### Detailed Parameter Description
+
+```
+renku session start [OPTIONS]
+
+--image <image_name>	override the docker image to be used for the interactive session.
+
+```
+
+#### renku session stop
+
+`renku session stop my_session` simply stops the given interactive session.
+
+One can shut down all the currently running interactive sessions at once with the `renku session stop --all`.
+
+##### Detailed Parameter Description
+
+```
+renku session stop [OPTIONS] <name>
+
+<name>
+
+--all       Stops all the running containers.
+
+```
+
+#### renku session list
+
+The command to list all the currently running local interactive sessions, with the detail information regarding
+the network port mappings.
+
+## Drawbacks
+
+One of the major drawback of the current design is the dependency on docker. Namely, to only support local
+interactive session if docker is available. Many HPC environments do not support docker, but they support
+other container formats, like [runC](https://github.com/opencontainers/runc). As a long term goal of this
+effort it would be good to consider support for different container engines, hence the design and implementation
+of the initial `renku session` should take this into consideration.
+
+## Rationale and Alternatives
+
+> Why is this design the best in the space of possible designs?
+
+> What other designs have been considered and what is the rationale for not choosing them?
+
+> What is the impact of not doing this?
+
+## Unresolved questions
+
+> What parts of the design do you expect to resolve through the RFC process before this gets merged?
+
+> What parts of the design do you expect to resolve through the implementation of this feature before stabilisation?
+
+> What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
--- a/design/003-interactive-session/003-interactive-session.md
+++ b/design/003-interactive-session/003-interactive-session.md
@@ -51,6 +51,10 @@ renku session start [OPTIONS]
 --image <image_name>	override the docker image to be used for the interactive session.
 --remote [endpoint]	start a remote interactive session using the JWT aquired for the
 			specific endpoint with renku login.
+--cpu <num of cpus>	Specify how much of the available CPU resources a container can use.
+--disk <disk space>	Specify the amount of disk space for the container.
+--gpu <device>		Specify the GPU allocated for the container.
+--memory <max memory>	The maximum amount of memory the container can use.
 
 ```
 
--- a/renku/core/management/session/docker.py
+++ b/renku/core/management/session/docker.py
@@ -19,6 +19,7 @@
 
 from pathlib import Path
 from typing import List, Optional, Tuple
+from uuid import uuid4
 
 import docker
 from yaspin import yaspin
@@ -53,11 +54,11 @@ class DockerSessionProvider(ISessionProvider):
         return f"{remote['owner']}/{remote['name']}"
 
     @staticmethod
-    def _get_jupyter_urls(ports, jupyter_port=8888):
+    def _get_jupyter_urls(ports, auth_token, jupyter_port=8888):
         port_key = f"{jupyter_port}/tcp"
         if port_key not in ports:
             return None
-        return map(lambda x: f'http://{x["HostIp"]}:{x["HostPort"]}/', ports[port_key])
+        return map(lambda x: f'http://{x["HostIp"]}:{x["HostPort"]}/?token={auth_token}', ports[port_key])
 
     def _get_docker_containers(self, client):
         return self.docker_client().containers.list(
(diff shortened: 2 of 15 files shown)


vigsterkr:
@SwissDataScienceCenter/poc this is an initial implementation of supporting a local interactive session. any feedback is more than welcome.

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,94 @@
...
+### Proposed Commands
+
+#### renku session start
+
+A command to start an interactive session.
+
+By default, the `Dockerfile` present in the given project is going to be used for building the docker image
+that's going to be used for running the interactive session. In case the docker image is not available the
+user is going to be prompted whether the user wants to build the docker image, as in some cases it can be
+time consuming.
```
rokroskar, line 39: What do we do in the case of a private project? We _should_ be able to get access to the hosted gitlab registry with the token we get from `renku login` but I'm not 100% sure. This has to be tested/investigated since I believe it's a JWT token which is replaced by an oauth token by the gateway - but I don't know if this works for the registry endpoints or not. It might require changes to the gateway.

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,94 @@
...
+a `Dockerfile`. This is by default provided by `renku init` when creating a renku project, but if the user
+is using a custom project template it might not be present.
+
+### Proposed Commands
+
+#### renku session start
+
+A command to start an interactive session.
+
+By default, the `Dockerfile` present in the given project is going to be used for building the docker image
```
rokroskar, line 36: I would say by default we look for the image on the hosted instance (if we can identify one). That would be quicker than building in most cases. We can identify the registry from the origin remote. 

There is a lot of potential for failure here so we need to think about the error handling with user-friendly suggestions for what to do next. E.g. 

* check local docker images --> none exists, check remote
* remote image lookup fails because image isn't there and no local image exists --> prompt to build one
* remote image lookup fails (private project and no token) --> suggest to login 
* remote image lookup fails, (private project with token) --> suggest image build
* image build fails, no docker --> link to docker install / documentation
* image build fails, docker working but problem in Dockerfile --> suggest a base image? or 🤷 

A related question is how do we actually identify which image we need? Simply by the commit sha? When the runtime doesn't change for many commits, we could just check which commit we need (i.e. when did the runtime dependencies change last - though we might not know all the paths in the repo that affect the runtime). Probably the safest is to go with the commit sha even if it results in too many image builds.

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,94 @@
...
+that's going to be used for running the interactive session. In case the docker image is not available the
+user is going to be prompted whether the user wants to build the docker image, as in some cases it can be
+time consuming.
+
+##### Detailed Parameter Description
+
+```
+renku session start [OPTIONS]
+
+--image <image_name>	override the docker image to be used for the interactive session.
```
rokroskar, line 46: 👍

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,94 @@
...
+
+<name>
+
+--all       Stops all the running containers.
+
+```
+
+#### renku session list
+
+The command to list all the currently running local interactive sessions, with the detail information regarding
```
rokroskar, line 69:
```suggestion
The command to list all the currently running local interactive sessions, with the detailed information regarding
```

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,94 @@
...
+the network port mappings.
+
+## Drawbacks
+
+One of the major drawback of the current design is the dependency on docker. Namely, to only support local
+interactive session if docker is available. Many HPC environments do not support docker, but they support
```
rokroskar, line 75: I think considering other runtimes is definitely a valid concern but with regards to HPC specifically, there are a lot of other hurdles (port-forwarding to compute nodes, for example). We might want to consider that a separate case altogether.

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,94 @@
...
+
+By default, the `Dockerfile` present in the given project is going to be used for building the docker image
+that's going to be used for running the interactive session. In case the docker image is not available the
+user is going to be prompted whether the user wants to build the docker image, as in some cases it can be
+time consuming.
+
+##### Detailed Parameter Description
+
+```
+renku session start [OPTIONS]
```
rokroskar, line 44: If we want to consider the more general case of launching remote sessions as well from the CLI, we might add a `--remote` flag here.

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,94 @@
...
+the network port mappings.
+
+## Drawbacks
+
+One of the major drawback of the current design is the dependency on docker. Namely, to only support local
+interactive session if docker is available. Many HPC environments do not support docker, but they support
+other container formats, like [runC](https://github.com/opencontainers/runc). As a long term goal of this
+effort it would be good to consider support for different container engines, hence the design and implementation
+of the initial `renku session` should take this into consideration.
+
```
rokroskar, line 79: One general question - the idea here is to mount the local repo directory into the container. Our containers typically run as user 1000 with gid 1000 - what happens with the file ownership outside the container?

rokroskar (commented):
Nice! I really like this, it's very simple and clean. There are a couple of outstanding questions, with the error handling and UX around suggesting alternative paths in case of failure being a big one here, I think, since we have to assume we are dealing with users with very little docker experience.

design/003-interactive-session/003-interactive-session.md
vigsterkr: while implementing this functionality couple of questions came up:
 - remote notebook sessions have various additional parameters that are in case of local sessions might not have need to be defined (of course they could be), namely the resource based parameters:
```
  "serverOptions": {
    "cpu_request": 0.25,
    "defaultUrl": "/lab",
    "disk_request": "1G",
    "gpu_request": 0,
    "lfs_auto_fetch": false,
    "mem_request": "1G"
  }
``` 
in case of a remote session, i believe these parameters should be exposed to the `docker session start --remote`, so that a user can override the default values (in case of a local session these params are not necessarily needed to be passed to docker, although it's possible). but since by design (or at least atm) the remote starting of a session  should be just a flag of `docker session start`, i believe that they should be exposed.

 * if we do support `--remote` from the CLI then we should support for all the subcommands, i.e. `start, stop, list`.      
     * should we include the remote sessions for `renku workflow ls` by default? or should it be the exact same behaviour like in case of starting a session (i.e. local by default)
     * in case of `docker session stop --all` should be both for local and remote sessions?

design/003-interactive-session/003-interactive-session.md
vigsterkr: by using commit hash when the runtime env doesn't change the caching mechanism (at least in case of docker) will circumvent the unnecessary rebuilds. by default the code-base itself (of the project) is mounted into the running container, so it shouldn't affect the image building at all.

design/003-interactive-session/003-interactive-session.md
rokroskar: imho this should not be implemented now, it was just a suggestion to add to the RFC to be implemented in the future, precisely because of the issues you raise, i.e. it requires a design all on its own.

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,115 @@
...
+By default, the `Dockerfile` present in the given project is going to be used for building the docker image
+that's going to be used for running the interactive session. In case the docker image is not available the
+user is going to be prompted whether the user wants to build the docker image, as in some cases it can be
+time consuming.
+
+By using the `--remote [endpoint]` flag an interactive session will be spawned on the specified endpoint. It requires
+that the user has authenticated herself previously with the `renku login` command and aquired the authentication
+token for the endpoint.	If the endpoint is not defined the command will try to assume the endpoint set in the project's
+configuration if available.
```
rokroskar, line 44: or guess it from the remote?

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,115 @@
...
+
+<name>      Name of the session that the user wants to stop.
+
+--all       Stops all the running containers.
+
+```
+
+#### renku session list
+
+The command to list all the currently running local interactive sessions, with the detailed information regarding
```
rokroskar, line 80: is this listing the sessions for this project or renku sessions in general? Should there be a corresponding `renku session open/connect` command that will automatically open the browser with the correct URL?

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,115 @@
...
+the network port mappings.
+
+##### Detailed Parameter Description
+
+```
+renku session list [OPTIONS]
+
+--remote       List the remote interactive sessions currently running.
```
rokroskar, line 88: do you have to specify the remote endpoint here or is it assumed/guessed? Maybe it can be an optional argument?

design/003-interactive-session/003-interactive-session.md
rokroskar: @vigsterkr ☝️ -- any thoughts on this?

rokroskar (changes requested):
This is looking great! Just a few clarifying comments. Sorry it took me a while 🙈

design/003-interactive-session/003-interactive-session.md
vigsterkr: you mean from the `git remote`? in that case i'm wondering in case of multiple remote which one should be used? `origin` or `upstream`?

design/003-interactive-session/003-interactive-session.md
vigsterkr: the listing would only be only done for the `active project` (meaning based on the cwd). 
but yeah you are right i should add an example of the `renku session list`; currently this is my idea:

```
ID         STATUS    URL
---------  -------   -------
deadbeef   running   http://localhost:31337
deeeeeed   running   https://renkulab.io/projects/vigsterkr/renku-mls-demo/sessions/show/vigsterkr--renku-2dmls-2ddemo-7c289dbd
```

this in case of even a quite basic terminal URLs can be opened directly.

or you reckon, it would be still nice to have an explicit command for opening, like `renku session open/connect`?

design/003-interactive-session/003-interactive-session.md
vigsterkr: the idea here is that by default only local sessions are listed. for remote session listing the user should explicitly define the `--remote` flag (this is an optional flag)

design/003-interactive-session/003-interactive-session.md
rokroskar: I guess origin

design/003-interactive-session/003-interactive-session.md
rokroskar: Yeah I wouldn't think of this except I really like (and often use) the `hub pr show` command from the github cli which opens up the current pr in the browser. Super useful to not have to copy/paste!

design/003-interactive-session/003-interactive-session.md
rokroskar: Yes but in `session start` the user specifies a remote

rokroskar: Maybe `--remote` should be a flag on `session` and apply to all subcommands

vigsterkr: > Maybe `--remote` should be a flag on `session` and apply to all subcommands

yeah this should work, although it's a bit 'redudant'/useless in case of `renku session stop <id>` as the <ID> is a unique key for a running session, regardless whether it's remote or local

design/003-interactive-session/003-interactive-session.md
vigsterkr: hence my comment about "this in case of even a quite basic terminal URLs can be opened directly", see CMD+left-click in case of iTerm or simply just click in case of alacritty or CTRL+left-click in case of Gnome TERM

vigsterkr: but none the less adding such a comment like `renku session connect <ID>` shouldn't be too complicated. note as in case of `stop`, this sub-command does not require a `--remote` flag as it's unique as well.

design/003-interactive-session/003-interactive-session.md
rokroskar: oh I see - where are the session ids stored?

vigsterkr: These IDs are actually provided by the backend itself. so for example now in case of the local interactive session `renku session list` lists all the running docker containers that are using the image provided by the project itself, i.e. the `Dockerfile`. as stated above these docker images are named the following way: `renku_client.remote['owner']:renku_client.remote['name']:commit_sha` (e.g. `vigsterkr/renku-workflow-demo:9f3d367`).

Note in the meanwhile we've added support to override the default docker image to be used (the one generated based on the project's `Dockerfile`) when starting a session (`renku session start --image renku/renkulab-py:3.9-0ae1458`). This will break the above logic - i.e. listing all the sessions started from the currently active project. But, in case of docker this can be easily fixed by using `--label` when starting a container/session, simply use a the project specific label when creating a container. - will fix this now ASAP.

in case of a remote session this should be fairly easy - as based on the swagger of the notebook api - one can simply query the running servers based on `namespace`, `project` etc.

vigsterkr: in other words currently I think there's no need to actually store these running session IDs ourselves.

rokroskar: > in case of a remote session this should be fairly easy

yes, definitely. But then you need to know which remote to look at. 

Also with regards to local sessions, should we add some more reliable metadata to the container? Or name it in a way that it's identifiable beyond just using the image as a proxy?

vigsterkr: i've added a simple label in order to support custom images for session:
```
"renku_project": renku_client.remote['owner']:renku_client.remote['name']
```

so for example `"renku_project": "vigsterkr/renku-workflow-demo"`

vigsterkr:
actually today i was thinking a bit more about this whole local vs remote story and realised that this is not the best approach, i.e. it shouldn't be distinguished like this. rather `renku session` sub-command should just have a support for various container engines---as plugins---and just as in case of workflow execute provider one would choose which one to use when running the command, for example:

```
renku session start --engine singularity
```

by default renku would provide two different plugins: `docker` and `renkulab`.  all the others would be simply plugins as usual (e.g. `singularity` or `youki`).
Note i think the other problem of the current `--remote` design is that a `docker` can be remote as well! it all depends on the env variables settings.

maybe the `--engine` flag is not the best name. maybe we should be consistent and use `--provider` just as in case of workflows... none the less this would in a way unify the whole `renku session` across various container backends and 'locations'

rokroskar:
I think `--provider` makes a lot of sense! I could imagine something like this for e.g. HPC clusters which would have very specific needs but it would be self-consistent if we name them all "providers". So I guess the RFC doc should then define the interface?

vigsterkr:
> I think `--provider` makes a lot of sense! I could imagine something like this for e.g. HPC clusters which would have very specific needs but it would be self-consistent if we name them all "providers". So I guess the RFC doc should then define the interface?

refactored, updated the RFC. now docker based interactive sessions are implemented as a plugin.

i'll add the `ISessionProvider` interface definition to the RFC ASAP

vigsterkr:
mmm seems a good side-effect of this refactor that the `docker` plugin won't even show up if docker engine is not installed 🙃 just need to do some extra exception handling (see macos ci failure)

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Summary
+
+Propose a new sub-group command `renku session` that is responsible for
+managing interactive sessions, both local and remote, for a project.
+
+## Motivation
+
+Renkulab.io provides the option to run interactive sessions of a project. Among other things the session
+contains a jupyter notebook.
```
Panaetius, line 14: `contains a jupyter server`?

renku/core/management/session/docker.py
```diff
@@ -0,0 +1,178 @@
...
+import docker
+from yaspin import yaspin
+
+from renku.core import errors
+from renku.core.models.session import ISessionProvider
+from renku.core.plugins import hookimpl
+from renku.core.utils import communication
+
+
+class DockerSessionProvider(ISessionProvider):
```
Panaetius, line 33: This class could use more typing information. Return types of methods, that `client` is `LocalClient`, `commit_sha` is `str` etc. Other classes in this PR as well.

olevski (commented):
@vigsterkr I tried to install this and run it on my mac. With a project I created with `renku init`. I am getting this message:
```
AttributeError: 'ImageNotFound' object has no attribute 'msg'
```
It also happens if I specify a wrong not-real image with `--image`

If I specify a real image then things work.

I know this is a WIP but just thought I would point this out.

Also see my comment about getting the default url from the project. I think that would be really useful.

olevski:
Sorry for spamming. I noticed I was a few commits back on this branch when I was testing. A small clarification on the errors I am seeing.

- `renku session start` results in:
```
AttributeError: 'APIError' object has no attribute 'msg'
```

- `renku session start --image fdsfdsf` results in:
```
AttributeError: 'ImageNotFound' object has no attribute 'msg'
```

vigsterkr:
> Sorry for spamming. I noticed I was a few commits back on this branch when I was testing. A small clarification on the errors I am seeing.
> 
>     * `renku session start` results in:
> 
> 
> ```
> AttributeError: 'APIError' object has no attribute 'msg'
> ```
> 
>     * `renku session start --image fdsfdsf` results in:
> 
> 
> ```
> AttributeError: 'ImageNotFound' object has no attribute 'msg'
> ```

@olevski nw, added a fix for those comments

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,123 @@
...
+By default, the `Dockerfile` present in the given project is going to be used for building the docker image
+that's going to be used for running the interactive session. In case the docker image is not available the
+user is going to be prompted whether the user wants to build the docker image, as in some cases it can be
+time consuming.
+
+With the `--provider <provider_type>` flag a user can specify session provider to be used for starting
+an interactive session. By default two different providers are shipped with Renku CLI:
+ - docker
+ - renkulab
+The provider specific configuration values can be specified by using the `--config <config.yaml>` flag.
```
rokroskar, line 45: I should be able to persist them also in the project settings?

design/003-interactive-session/003-interactive-session.md
rokroskar: >There is a lot of potential for failure here so we need to think about the error handling with user-friendly suggestions for what to do next. E.g.
check local docker images --> none exists, check remote
remote image lookup fails because image isn't there and no local image exists --> prompt to build one
remote image lookup fails (private project and no token) --> suggest to login
remote image lookup fails, (private project with token) --> suggest image build
image build fails, no docker --> link to docker install / documentation
image build fails, docker working but problem in Dockerfile --> suggest a base image? or 🤷

we should describe this logic or do you think we should always build and never fetch from the remote registry?

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,123 @@
...
+renku session list [OPTIONS]
+
+-p|--provider [docker|renkulab] 	Use the specified backend for listing the interactive sessions.
+-c|--config   				YAML file containing configuration for the provider.
+
+```
+
+#### renku session open
+
+`renku session open <my_session>` opens the given interactive session in the user's default browser.
```
rokroskar, line 96: don't we need a `provider` here as well? Or is the provider going to be encoded in the session id returned by `session list`?

rokroskar (changes requested):
Some additional comments on the RFC. 

Also, could we separate the implementation from the RFC? The PR discussion is muddled by the fact that both things are getting addressed at once imho and it would be easier for posterity if they were separate.

design/003-interactive-session/003-interactive-session.md
vigsterkr: this is in line with the workflow execute provider logic.... of course these values could be stored anywhere... i dont have any strong opinion about this.

design/003-interactive-session/003-interactive-session.md
vigsterkr: no, coz the session id is unique so if there's such a session then it'll open it.

vigsterkr:
regarding the logic of image fetching/building: since now it is a plugin based, this is strictly up to the plugin what and how to do things. the most this RFC can do is to kind of write a suggestion how it should be handled, but honestly the plugin will do whatever it wants to do... note i'm not so sure what happens in case of notebook api when an image is defined...

design/003-interactive-session/003-interactive-session.md
vigsterkr: regarding the logic of image fetching/building: since now it is a plugin based solution, this is strictly up to the plugin what and how to do things. the most this RFC can do is to kind of write a suggestion how it should be handled, but honestly the plugin will do whatever it wants to do... note i'm not so sure what happens in case of notebook api when an image is defined...

design/003-interactive-session/003-interactive-session.md
vigsterkr: deprected... this needs to be addressed in https://github.com/SwissDataScienceCenter/renku-python/issues/2635 but to be honest trying to do this with the git remote is prone to be error.... unless the plugin goes through all the git remote urls and tries to access the notebook api... but for that a `ping` api would be good for the notebook api

design/003-interactive-session/003-interactive-session.md
vigsterkr: although i'm not so sure why would one want to store `youki`, `singularity` etc config values on a project level...these these key-value pairs meant for actually specifying some config value for the provider plugin, that i see more like a global setting not a project level setting. none the less, they could be persisted anywhere...

renku/core/management/session/docker.py
```diff
@@ -0,0 +1,188 @@
...
+                    )
+
+                    with yaspin(text="Building docker image"):
+                        _ = self.docker_client().images.build(path=str(client.docker_path.parent), tag=image_name)
+
+            auth_token = uuid4().hex
+            project_config = client.load_config()
+            default_url = project_config.get("interactive", "default_url")
+            container = self.docker_client().containers.run(
+                image_name,
```
olevski, line 139: @vigsterkr it just occurred to me that there you could also have the docker image pinned in the the config in a project. So it may be a good idea to check and use that if it is defined.

It is under `interactive.image`.

vigsterkr: @olevski do you have a public example somewhere on renkulab that has that set?

olevski: Also if this complicates things I think it is ok to address it in a followup PR/improvement. I know there is a lot of discussions on how to handle images/building/etc.

vigsterkr: i mean if the renkulab interactive session is following that logic, then so should the local docker based one.

design/003-interactive-session/003-interactive-session.md
rokroskar: But you fetch session IDs from the provider - are you going to check all providers for the id?

vigsterkr: see https://github.com/SwissDataScienceCenter/renku-python/pull/2529/files#diff-9eec9990e65005652022b90320c5717aabc1ea27e3fbdb25441e9bca96a09bc9R64
i.e. it doesn't have `firstresult=True` defined so by default all plugins are going to be executed. so indeed it does go through over all providers. note it's highly unlikely that there's gonna be more than 3-4 providers at once, and this seemed to be a good tradeoff between having the user to be forced to specify a `provider` vs not (and then have all the providers checked).

vigsterkr: and i dont imagine having like more than a dozen active sessions altogether.
`640kb should be enough for everyone` (c)

renku/core/management/session/docker.py
olevski: Yeah here: https://dev.renku.ch/projects/tasko.olevski/test-pinned-image-1

vigsterkr: @olevski wait just to be clear: in this case when you set `image = renku/renkulab-py:3.9-0.10.3` in the `renku.ini` this will set the `RENKU_BASE_IMAGE` or this is purely setting the image that is going to be used in the interactive session. just because if the latter, then all the dependencies that are defined in `requirements.txt` of the given project is gonna be ignored.

olevski: @vigsterkr yeah if the image is pinned (i.e. if the image field is set in renku.ini) then the dockerfile does not have to be built. When you launch a session on renkulab.io with a pinned image the notebook service immediately tries to pull the pinned image and if that is not available it fails. Similarly if the image is pinned the UI does not check if any of the CI pipelines for the project (that usually build the images) have completed or not. It is assumed that the pinned image is ready to go. So yeah with a pinned image all the stuff in `requirements.txt` and even in the Dockerfile is ignored.

vigsterkr:
closing in favour of #2687 #2688

PR #2531, opened 2021-12-15
feat(service): add project.lock_status endpoint

### LINKED ISSUE #2507: add a `project.lock_status` endpoint to the service
Currently, long running operations like cache.migrate lock the project to other requests, raising an exception of a lock couldn't be acquired withing 15 seconds.

We should add a `project.lock_status` endpoint that can let clients know if a project is currently (write-)locked, returning a boolean.

The code for this could be something like
```
if "project_id" in self.context:
    project = self.cache.get_project(self.user, self.context["project_id"])
elif "git_url" in self.context:
    project = Project.get(
        (Project.user_id == self.user_data["user_id"]) & (Project.git_url == self.context["git_url"])
    )
else:
    raise Exception()

locked = True

try:
    with project.read_lock(timeout=0):
        locked=False
except (portalocker.LockException, portalocker.AlreadyLocked):
    locked = True

return locked
```

and of course without using the `RenkuOperationMixin`, similar to how renku.service.views.jobs doesn't use the mixin.

### LINKED ISSUE #2532: Make service API version 1.0 the default
Currently, API version `0.9` is the default version for renku core service when a client doesn't specify a version (e.g. `/renku/9/cache.migrations_check` calls `/renku/9/0.9/cache.migrations_check`). Since UI is using the new API version, we can make version `1.0` as the default.

mohammad-sdsc:
# Description

- Adds a `project.lock_status` endpoint which returns whether a project is locked for writing or not.
- Make API version `1.0` the default version.

**NOTE**: When a project is not in the cache or the project_id/git_url is not found, this will NOT raise an error and just returns false (i.e. not locked).

Fixes #2507 
Fixes #2532 

/deploy renku-ui=1.0-next


lorenzo-cavazzi (approved):
Thanks, it works well from the client side :+1:

vigsterkr (approved):
🚢

PR #2534, opened 2021-12-16
fix(core): set project when exporting graph

### LINKED ISSUE #2533: Some activities are missing from the project when exporting graph
**To Reproduce**
1. `renku clone git@renkulab.io:covid-19/covid-19-public-data.git`
2. `git checkout 7acd9507`
3. `renku graph export --full --strict`
4. Project has 658 activities but only 32 of them are listed under Project's `hasActivity`

 - renku-python `v1.0.1`

**Additional context**
`project_id` is not set during migration for `Activity`, therefore, only activities that are created after the migration are exported.

mohammad-sdsc:
# Description

Sets `project_id` for Activities when exporting graph.

Fixes #2533


vigsterkr (approved):
:shipit:

PR #2537, opened 2021-12-20
feat(core): ignore quotation mark in git user/email config

### LINKED ISSUE #2539: Quotation mark around user's email in git config results in invalid ids
**Describe the bug**
When user's git configuration include quotation marks, gitpython returns the value as it is. For example, when reading `email = "john.doe@example.com"`, the return value is `"john.doe@example.com"`; vanilla git removes the quotation marks and returns `john.doe@example.com` in this case.
This causes ids like `emailto:\"john.doe@example.com\"` which are not valid.

**To Reproduce**
1. `renku dataset create mydata`
2. Edit `~/.gitconfig` and put user's email in quotation mark
3. `renku dataset add /some/data`
4. See the invalid ids in `renku graph export`'s output

mohammad-sdsc:
# Description

Removes quotation marks from user's name/email git config.

Fixes #2539


vigsterkr (approved):
🚢

PR #2541, opened 2021-12-22
fix(service): disable migration check optimization for anonymous sessions

mohammad-sdsc:
# Description
Disable optimized `/cache.migrations_check` for now.

/deploy renku=1.0-next renku-graph=cli-v1 renku-ui=1507b-migrations-giturl #notest


Panaetius:
Maybe for anonymous users we should just go the slow route instead of disabling it (`execute_op` instead of `_fast_op_without_cache`)

Panaetius (dismissed):
Thank you!

Panaetius:
That's weird. So the token sent in the JWT with that user does not have access to that repo through gitlab API, even though the repo is owned by that user.

lorenzo-cavazzi:
> That's weird. So the token sent in the JWT with that user does not have access to that repo through gitlab API, even though the repo is owned by that user.

That seems to be an issue also on public projects, where an invalid token _shouldn't_ be a problem .

Anyway, I modified the initial message to trigger a CI deployment.
You can verify the error here by trying to access as non logged user (it works) and then as logged user (currently it doesn't work, and the the `GET /cache.migrations_check` endpoint return the following error message `{"error": {"code": -32603, "reason": "internal error"}}`)
https://renku-ci-rp-2541.dev.renku.ch/projects/lorenzo.cavazzi.tech/project-with-core-1-0-1/overview/status

Panaetius (approved):
Thank you!

PR #2551, opened 2021-12-29
chore(tests): cache network requests

### LINKED ISSUE #2506: Cache external requests for integration tests
To speed up our tests (and not hammer external services as much) we should record web requests to be replayed in tests.

We could use something like https://pypi.org/project/pytest-recording/ which allows annotating tests with `@pytest.mark.vcr`.

Then we need to add a github actions cache (akin to https://github.com/SwissDataScienceCenter/renku-python/blob/master/.github/workflows/test_deploy.yml#L84 ) for our integration tests. 
The cache key should include:
- The current week
- the hash of the `tests/` directory
- a flag set at the top of the tests to manually trigger a cache miss
- (optionally)the hash of pyproject.toml. Not sure if this is beneficial (like it requests library changes)

Here we should call pytest with `--record-mode=once` if the cache was hit and `--record-mode=all` if the cache was a miss.

This should significantly speed up tests and prevent us from being rate-limited.

mohammad-sdsc:
# Description

Caches network requests using cache Github action.  Cache key includes the job's id to differentiate caches for different jobs.

Fixes #2506


vigsterkr (approved):
:shipit:

PR #2552, opened 2022-01-04
chore: add lock package and fix poetry building platform whl

### LINKED ISSUE #2502: create renku pypi package with fixed dependencies
With the introduction of poetry, we can create a second package (maybe `renku-stable` ?) that installs from the lock file instead of from pyproject.toml

We should add a Github action that automatically builds and pushes this on release.

This way, with that version installed, dependencies can't change out from under users (at the cost of more restrictive versions/compatibility with user envs).

We should also switch the renkulab Docker to use this instead of the normal image. Same for the core service.

Panaetius:
closes #2502


vigsterkr (approved):
🚀

vigsterkr (approved):
🚀

PR #2554, opened 2022-01-04
feat(core): support forward compatibility of datasets

### LINKED ISSUE #2516: make Database support fully compatible Schema evolution
Our `Database` implementation is already backward compatible to some extent (missing properties can be added as class variables with default value and will be included on write, without a migration).

To reduce the need for migrations even more, we should also make it forward compatible. This entails:
- properties that are found in data but don't exist on the model shouldn't cause an error
- those properties should be added to some hidden property on the respective instance, e.g. `_p_unknown_properties`
- when saving an object, the `_p_unknown_properties` have to be written as well, so they're not lost when using an older version of renku to work on newer metadata.

This should greatly reduce migrations and forced-upgrades needed, making things easier for us and for users. This allows new fields to be introduced, old fields to be deleted (ignored). We might still need migrations in some cases, but reducing their number is nice to have.

https://docs.confluent.io/platform/current/schema-registry/avro.html details the concept of schema evolution quite nicely (although it's for the Avro format)

Panaetius:
Our database already supported backwards compatibility (add properties at the class level) and forward compatibility (we're schemaless so additional properties are loaded just fine even if they don't exist in code).

The main issue was that newer properties were lost when editing datasets/plans, as we didn't do a deep copy on editing, causing fields from newer schemas to be lost. Now they are ported/copied to the derived objects properly.

closes #2516

### DIFF
--- a/renku/core/models/dataset.py
+++ b/renku/core/models/dataset.py
@@ -396,28 +396,18 @@ class Dataset(Persistent):
 
     def copy(self) -> "Dataset":
         """Return a clone of this dataset."""
-        return Dataset(
-            annotations=[a.copy() for a in self.annotations],
-            creators=self.creators.copy(),
-            dataset_files=[f.copy() for f in self.dataset_files],
-            date_created=self.date_created,
-            date_published=self.date_published,
-            date_removed=self.date_removed,
-            derived_from=self.derived_from,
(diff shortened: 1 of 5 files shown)


vigsterkr (approved):
:shipit:

PR #2555, opened 2022-01-05
fix(core): require click 8.x

vigsterkr:
# Description

`shell_complete` kwarg for `click` commands is only available in click 8.0.0 and later.


vigsterkr:
the integration test errors are unrelated.

Panaetius (approved):
Thank you!

PR #2556, opened 2022-01-06
fix(core): remove .git from CWL export

### LINKED ISSUE #2548: Do not copy `.git` directory during CWL reruns
`.git` directory is not required anymore when running CWL jobs: https://github.com/SwissDataScienceCenter/renku-python/blob/915c471a9c1ede97a40820eaa62a6a4c460ba165/renku/core/management/workflow/converters/cwl.py#L290. It should be removed because it can cause problems for large project where copying this directory takes a long time.

vigsterkr:
# Description

Fixes #2548


Panaetius (approved):
:shipit:

PR #2558, opened 2022-01-07
chore: fix caching of templates folder with Makefile

Panaetius:
templates weren't downloaded in all actions, so depending on which ran first, they might not be included in the cache.

Also, the Makefile has to be used as a cache key, as template version depends on it.


vigsterkr (approved):
🚢

PR #2562, opened 2022-01-12
fix(core): reintroduce shell completion command

### LINKED ISSUE #2530: Reintroduce shell completion command
#2512 removed `click_completion` i.e `renku --install-completion` command due to the fact that `click_completion` is not working with click 8.x anymore.

It would be still nice to have a command that basically provides the user the completion command for `$SHELL`, that even could be interpreted, so that way adding it to the shell config file something like  `renku --shell-integration` would activate the shell completion functionality

vigsterkr:
# Description

Fixes #2530


renku/cli/env.py
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Renku environment related commands."""
+
+import click
+
+from renku.core import errors
```
Panaetius, line 22: Should we import this down below as well, just to keep renku.cli imports clean for performance reasons

renku/cli/env.py
```diff
@@ -0,0 +1,41 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Renku environment related commands."""
+
+import click
+
+
+@click.command()
+@click.option("shell_completion", "--shell-completion", is_flag=True, help="Print shell completion command")
```
Panaetius, line 24: We should print a help text if no flag is passed.

You can use
```
@click.pass_context
def func(ctx, ...):
    [...]
    click.echo(ctx.parent.get_help())
```

Panaetius (changes requested):

vigsterkr:
note i have not updated the documentation with the new command as unfortunately (due to fish) it cannot be really unified. in case of bash/zsh one could do this:
```
eval "$(eval `renku env --shell-completion`)"
```

Panaetius (approved):

PR #2570, opened 2022-01-13
fix(core): fix workflow iterate on composite plan with links

### LINKED ISSUE #2571: Bad derived_from reference in plan and floats in CWL not supported
our method `_get_argument_type` for the cwltool provider only considers `str` and `int` as type, but this precludes using floats. We should add support for floats, as they're not that uncommon as parameters.

Additionally there is a small bug with `renku update` not working due to accessing `Plan.derived_from` as an URL object where actually it's a string.

Panaetius:
Closes #2571


vigsterkr (approved):
🚀

PR #2573, opened 2022-01-14
fix(service): fix cache.migrate not locking the project

### LINKED ISSUE #2545: Locking when migrating a project in the service
`/cache.migrate` endpoint does not lock a project during migrations. It must lock the project to avoid concurrent writes.

Panaetius:
closes #2545


mohammad-sdsc (dismissed):
LGTM, thank you!

mohammad-sdsc (approved):

PR #2577, opened 2022-01-14
docs: add YAML example for workflow execute and iterate

### LINKED ISSUE #2567: Add documentation/example of YAML file for workflow execute/iterate
https://renku.readthedocs.io/projects/renku-python/en/latest/reference/commands.html?highlight=iterate#executing-plans
and
https://renku.readthedocs.io/projects/renku-python/en/latest/reference/commands.html?highlight=iterate#iterate-plans

mention that setting values is possible using a yaml file also, but they don't give an example on what the structure of such a file might look like.

We should add a short example to each.

vigsterkr:
# Description

Fixes #2567

### DIFF
--- a/renku/cli/workflow.py
+++ b/renku/cli/workflow.py
@@ -125,14 +125,14 @@ the YAML should follow the this structure:
 
 .. code-block:: yaml
 
-learning_rate: 0.9
-dataset_input: dataset.csv
-chart_output: mychart.png
-myworkflow:
-    lr: 0.8
-    lookuptable: lookup.xml
-    myotherworkflow:
-        language: en
+    learning_rate: 0.9
+    dataset_input: dataset.csv
+    chart_output: mychart.png
+    myworkflow:
+        lr: 0.8
+        lookuptable: lookup.xml
+        myotherworkflow:
+            language: en
 
 Provider specific settings can be passed as file using the ``--config`` parameter.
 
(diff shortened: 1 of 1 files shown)


renku/cli/workflow.py
```diff
@@ -149,7 +162,13 @@
...
 The set of possible values for a parameter can be given by ``--map`` command
 line argument or by specifying them in a values YAML file and passing that
-using ``--mapping``.
+using ``--mapping``. Content of the mapping file for the above example
+should be:
+
+.. code-block:: yaml
+
+parameter-1: [1,2,3]
+parameter-2: [10,20]
```
Panaetius, line 171: For setting a value on a child workflow, would it be `child.parameter-1` or 
```
child:
  parameter-1
```
?

vigsterkr: yep. just as in case of `workflow execute`.

Panaetius: Both work? nice

vigsterkr: in case of execute there's a test for it using both yaml and cli interface: https://github.com/SwissDataScienceCenter/renku-python/blob/develop/tests/cli/test_workflow.py#L517

Panaetius (approved):

Panaetius (approved):

PR #2578, opened 2022-01-17
chore(ci): Update renku actions to v0.4.1

olevski:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action
/deploy


olevski:
The setup is as follows:

- if you do not provide `#persist` in the PR description (i.e. like  `/deploy #persist`) then the CI deployment is deleted after the acceptance tests complete
- if you provide `#persist` the deployment is kept until the PR is closed.
- when the PR is closed the namespace of the deployment is also deleted
- when the acceptance tests complete AND `#persist` is not used the CI deployment is deleted but the namespace is not deleted in k8s - this makes it possible to reuse the TLS secret and not hit the limits for requesting certs with letsencrypt - when the acceptance tests for the same PR are run many times
- artifacts from the acceptance tests are managed by the single (composite) test-renku action

Panaetius:
This can be closed in favor of https://github.com/SwissDataScienceCenter/renku-python/pull/2621 right?

olevski:
> This can be closed in favor of #2621 right?

Actually in favor of https://github.com/SwissDataScienceCenter/renku-python/pull/2622. But yeah https://github.com/SwissDataScienceCenter/renku-python/pull/2621 was a mistake essentially and https://github.com/SwissDataScienceCenter/renku-python/pull/2622 supersedes both this and the https://github.com/SwissDataScienceCenter/renku-python/pull/2621 PR.

PR #2581, opened 2022-01-17
fix(core): make activity ids deterministic in migration

### LINKED ISSUE #2580: Make activity ids deterministic in migration
Currently, we just generate new UUIDs when migrating ProcessRuns.

This means that repeat migrations of the same project (or on forks) result in the same node in the KG getting different IDs, which leads to problems further down the line for the KG team.

We should make the migration of ProcessRuns/Activities deterministic.

Panaetius:
closes #2580 

Activity ids are now deterministic sha1 hashes when migrating.

### DIFF
--- a/renku/core/management/migrations/m_0009__new_metadata_storage.py
+++ b/renku/core/management/migrations/m_0009__new_metadata_storage.py
@@ -22,6 +22,7 @@ import shutil
 import traceback
 import uuid
 from collections import defaultdict
+from hashlib import sha1
 from itertools import chain
 from pathlib import Path, PurePosixPath
 from typing import List, Optional, Union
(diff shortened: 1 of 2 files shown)


renku/core/management/migrations/m_0009__new_metadata_storage.py
```diff
@@ -372,6 +373,24 @@ def _process_workflows(client: LocalClient, activity_gateway: IActivityGateway,
...
 def _process_run_to_new_activity(process_run: old_schema.ProcessRun, client: LocalClient) -> List[Activity]:
     """Convert a ProcessRun to a new Activity."""
+
+    def generate_activity_id(process_run_id: str, suffix: str = None) -> str:
+        # https://localhost/activities/commit/b092dca8866c940dd4ff119b7d76d163a1be6a65
+        # https://localhost/activities/commit/b092dca8866c940dd4ff119b7d76d163a1be6a65/steps/step_1
+        parts = process_run_id.rstrip("/").rsplit("/", maxsplit=3)
+
+        if parts[-2] == "steps":
+            parts = parts[-3:-1]
```
mohammad-sdsc, line 383: Is there a reason not to use `parts[-3:]`? It already has the step number in it and there is no need for the suffix.

mohammad-sdsc (approved):
Thank you!

PR #2583, opened 2022-01-17
feat(cli): add explicit parameters to renku run

### LINKED ISSUE #2549: Allow defining explicit parameters in `renku run`
Renku considers a parameter passed to a user's script as input or output if a file/directory with that name exists in the current directory. We should add a flag to `renku run` to enable marking a command line argument as a workflow's parameter. See https://renku.discourse.group/t/renku-python-v1-0-1-output-path-recognized-as-output-instead-of-output-file/506 for a use case. A copy of the mentioned git repo is attached: [test-flow-master.tar.gz](https://github.com/SwissDataScienceCenter/renku-python/files/7789426/test-flow-master.tar.gz)

Panaetius:
closes #2549 

Adds a `--param` flag to `renku run` that can be used to explicitly define Parameters.

E.g.
```
$ renku run touch test
$ renku run --name my-plan --param test -- echo test > output
$ renku workflow show my-plan
```
Should show a parameter instead of an input.


renku/core/management/workflow/plan_factory.py
```diff
@@ -693,7 +718,7 @@ def get_indirect_outputs_path(client_path):
 def get_indirect_parameters_path(client_path):
     """Return path to file that contains indirect parameters list."""
     parent = _get_indirect_parent_path(client_path)
-    return parent / "parameters.yml"
+    return parent / "parameters..yml"
```
vigsterkr, line 721: i'm wondering about this change. i mean i see that in the unit test it's being used the very same way, just not sure why the double dots...

vigsterkr (commented):
LGTM, only the `parameters..yml` part is not clear to me.

renku/core/management/workflow/plan_factory.py
Panaetius: Ah it was originally `parameters.yml` but for inputs and outputs, it's .txt, so I'd changed to to .txt, then saw that actually it is a yaml file, so changed with back with a search&replace, which I seem to have botched 🙈 
Thanks for catching that!

vigsterkr (approved):
🚀

PR #2586, opened 2022-01-18
fix(core): move NodeJS requirement check to cwltool plugin

### LINKED ISSUE #2575: NodeJS is required with toil backend
NodeJS should only be required for the cwltool backend, but it is also imported when using `renku update -p toil`, causing unneccessary dependencies. This can be a problem / annoyance on university HPCs.

Steps to reproduce the behavior on a system without nodeJS:
1. `renku init`
2. `renku -S run --name demo-wf  -- echo "0,1,2,3" > demo.csv`
3. `renku update -p toil -a`

Instead of updating all files, this produces the following error:
```
Error: NodeJs could not be found on this system
Please install it, for details see https://nodejs.org/en/download/package-manager/
```

Environment:
- renku-python 1.0.1
- Python 3.9.7
- local terminal (EPFL's helvetios cluster)

vigsterkr:
# Description

Fixes #2575


Panaetius (changes requested):
We should also remove RequireNodeJs from command builder so we don't have dead code lying around.

vigsterkr:
due to this change currently the code related to `require_nodejs`, namely:
 - https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/core/management/command_builder/command.py#L394 
 - https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/core/management/command_builder/nodejs.py

becomes unused in the whole project. i left it there for possible future use, but maybe we wanna drop the whole thing?

vigsterkr:
> We should also remove RequireNodeJs from command builder so we don't have dead code lying around.

ok this was cross-posting 🙃

Panaetius (approved):

PR #2594, opened 2022-01-21
chore: update helm chart core image tags

Panaetius:
Updates the versions of core-service used in the chart values


Panaetius:
Closed in favor of https://github.com/SwissDataScienceCenter/renku-python/pull/2604

PR #2596, opened 2022-01-21
feat(core): pass parameters as env vars to scripts (and renku.api)

### LINKED ISSUE #2563: renku.api.Parameter does not work with value set using `-s` in workflow execute
when using e.g. `renku.api.Parameter("gamma", 1)` in code and then executing a workflow using that script like `renku workflow execute -s gamma=2 my-plan`, the original value of 1 is used instead of 2.

We should fix this so the values are applied correctly.

Panaetius:
closes #2563 

Parameters, Input, Outputs are now passed as `RENKU_ENV_{name}` environment variables. For parameters, this is picked up in `renku.api.Parameter` so a user can changes it's value.

Example `script.py`: 
```
from renku.api import Parameter
print(Parameter("test", "hello world"))
```

Run like 
```
$ renku run --name run1 -- python script.py > output
$ cat output 
hello world
$ renku workflow execute -p toil --set test=goodbye run1
$ cat output 
goodbye
```

Enabling renku.api.{Input,Output} to pick up modified values will be done in ticket


vigsterkr (approved):
🚢

PR #2598, opened 2022-01-21
fix(core): fix SHACL for Plan and CompositePlan

### LINKED ISSUE #2595: Update SHACL for Plan and CompositePlan
Currently we don't have `prov:invalidatedAtTime` on `Plan` nodes in our SHACL. Additionally, `CompositePlan` is missing completely (But `GroupedRun` is still there, which was the old name of `CompositePlan`).

We should update our SHACL to be valid.

Panaetius:
closes #2595

### DIFF
--- a/renku/data/shacl_shape.json
+++ b/renku/data/shacl_shape.json
@@ -889,7 +889,7 @@
                "@id": "rdf:type"
             }
          ],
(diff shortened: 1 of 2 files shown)


vigsterkr (approved):
:shipit:

PR #2600, opened 2022-01-24
fix(core): Execution graph linking of plans

### LINKED ISSUE #2569: `renku workflow iterate` doesn't work with setting intermediate files
Checkout https://github.com/Panaetius/renku-cli-1.0.0-presentation/tree/iterate-bug

It has a Composite Plan with two steps, `train` and `evaluation` which are linked like `train -->[model]-->evaluation`.

Trying to do a `renku workflow iterate` setting the intermediate model fails with various errors:

Command executed:
```
renku workflow iterate --provider toil --map gamma=[0.1,0.5,1.0,1.5,2.0] --map c=[0.0001,0.001,0.01,0.1,1,10] --map model=model_{iter_index} --map evaluation.result-5=output/results_{iter_index}.txt --map evaluation.matrix-6=outputs/confusion_matrix_{iter_index}.png pipeline
```

In toil:
```
Log from job "'SubprocessToilJob' evaluation kind-SubprocessToilJob/instance-inusoc1z v6" follows:
=========>
          File "/home/zenon/.pyenv/versions/3.7.8/envs/cli-demo/lib/python3.7/site-packages/toil/worker.py", line 402, in workerScript
            job._runner(jobGraph=None, jobStore=jobStore, fileStore=fileStore, defer=defer)
          File "/home/zenon/.pyenv/versions/3.7.8/envs/cli-demo/lib/python3.7/site-packages/toil/job.py", line 2362, in _runner
            returnValues = self._run(jobGraph=None, fileStore=fileStore)
          File "/home/zenon/.pyenv/versions/3.7.8/envs/cli-demo/lib/python3.7/site-packages/toil/job.py", line 2283, in _run
            return self.run(fileStore)
          File "/home/zenon/.pyenv/versions/3.7.8/envs/cli-demo/lib/python3.7/site-packages/renku/core/management/workflow/providers/toil.py", line 94, in run
            parent_inputs[i.actual_value] if i.actual_value in parent_inputs else self._input_files[i.actual_value]
        KeyError: 'model_20'
```

vigsterkr:
# Description

Fixes #2569

(especially) in case of a `renku workflow iterate`, a `CommandParameterBase` can share the same ID, but with different parameterization, i.e. different `actual_value`.  Hence `Plan::find_parameter` should only be true if both of the case matches (not only the ID)

### DIFF
--- a/renku/core/models/workflow/plan.py
+++ b/renku/core/models/workflow/plan.py
@@ -210,7 +210,10 @@ class Plan(AbstractPlan):
 
     def find_parameter(self, parameter: CommandParameterBase) -> bool:
         """Find if a parameter exists on this plan."""
-        return any(parameter.id == p.id for p in self.inputs + self.outputs + self.parameters)
+        return any(
+            parameter.id == p.id and parameter.actual_value == p.actual_value
+            for p in self.inputs + self.outputs + self.parameters
+        )
 
     def get_parameter_path(self, parameter: CommandParameterBase):
         """Get the path to a parameter inside this plan."""


Panaetius (approved):
Good catch! Such a small change for something so difficult to debug/figure out...

Thanks for looking into this!

PR #2602, opened 2022-01-24
fix(service): use separate queues for multi core service deployment

### LINKED ISSUE #2601: RQ queues aren't namespace by deployment
Currently, RQ queues like `delayed.ctrl.DatasetsCreateCtrl` have the same name in all versioned deployments of the core service.

This can cause e.g. version v8 to pick up jobs enqueued by v9, which leads to failures if the behavior changed between versions.

We should prefix work queues with the version of the deployment to make them distinct.

Panaetius:
Changes rq workqueues to be namespaced by deployment

Closes #2601 

/deploy renku=1.0-next renku-graph=development renku-ui=0000-fix-add-dataset


mohammad-sdsc (approved):
LGTM, thanks for taking care of this!

PR #2603, opened 2022-01-25
fix(core): replace cwlgen with cwl-utils

### LINKED ISSUE #2592: Replace cwlgen with cwl-utils
We currently use https://pypi.org/project/cwlgen/ to create CWL files, which has been deprecated.

We should switch to using https://github.com/common-workflow-language/cwl-utils .

Migrations notes can be found directly in the cwlgen project description.

vigsterkr:
# Description

Fixes #2592

### DIFF
--- a/renku/core/management/workflow/converters/cwl.py
+++ b/renku/core/management/workflow/converters/cwl.py
@@ -336,7 +336,9 @@ class CWLExporter(IWorkflowConverter):
         return cwl.CommandInputParameter(
             id=CWLExporter._sanitize_id(parameter.id),
             type=type_,
-            inputBinding=cwl.CommandLineBinding(position=parameter.position, prefix=prefix, separate=separate),
+            inputBinding=cwl.CommandLineBinding(position=parameter.position, prefix=prefix, separate=separate)
+            if parameter.position or prefix
+            else None,
             default=value,
         )
 
@@ -364,7 +366,9 @@ class CWLExporter(IWorkflowConverter):
         return cwl.CommandInputParameter(
             id=sanitized_id,
             type=type_,
-            inputBinding=cwl.CommandLineBinding(position=position, prefix=prefix, separate=separate),
+            inputBinding=cwl.CommandLineBinding(position=position, prefix=prefix, separate=separate)
+            if position or prefix
+            else None,
             default={"location": (basedir / input.actual_value).resolve().as_uri(), "class": type_},
         )
 
--- a/renku/core/management/workflow/converters/cwl.py
+++ b/renku/core/management/workflow/converters/cwl.py
@@ -112,7 +112,9 @@ class CWLExporter(IWorkflowConverter):
                 workflow, tmpdir, basedir, filename=filename, output_format=output_format
             )
         else:
-            path = CWLExporter._convert_step(workflow, tmpdir, basedir, filename=filename, output_format=output_format)
+            _, path = CWLExporter._convert_step(
+                workflow, tmpdir, basedir, filename=filename, output_format=output_format
+            )
 
         return path.read_text()
 
@@ -143,10 +145,10 @@ class CWLExporter(IWorkflowConverter):
         import networkx as nx
 
         for i, wf in enumerate(nx.topological_sort(graph.workflow_graph)):
-            path = CWLExporter._convert_step(
+            cwl_workflow, path = CWLExporter._convert_step(
                 workflow=wf, tmpdir=tmpdir, basedir=basedir, filename=None, output_format=output_format
             )
-            step = WorkflowStep("step_{}".format(i), str(path))
+            step = WorkflowStep(in_=[], out=[], run=str(path), id="step_{}".format(i))
 
             for input in wf.inputs:
                 input_path = input.actual_value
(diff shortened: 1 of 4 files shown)


renku/core/management/workflow/converters/cwl.py
```diff
@@ -260,16 +262,12 @@ def _convert_step(
...
-                    )
+                    cwl.Dirent(entry='$({"listing": [], "class": "Directory"})', entryname=path, writable=True)
                 )
                 dirents.append(path)
                 jsrequirement = True
 
-            environment_variables.append(
-                cwlgen.EnvVarRequirement.EnvironmentDef(f"{RENKU_ENV_PREFIX}{output_.name}", output_.actual_value)
-            )
+            environment_variables.append(cwl.EnvironmentDef(f"{RENKU_ENV_PREFIX}{output_.name}", output_.actual_value))
```
vigsterkr, line 270: @SwissDataScienceCenter/renku-devs i wonder why these env var definitions? same goes for inputs and parameters (see below).

Panaetius: I think you just tagged the whole renku team ;)  We'd be `renku-python-maintainers`

They can be used by scripts/commands and are also how `renku.api.Parameter` picks up values set with `--set` (and in the future, `renku.api.Output` and `renku.api.Input` as well), so people can properly use those in their scripts.

vigsterkr: \o/ yey, for some reason github was only offering this in it's auto-completion :< sorry people

vigsterkr: i'm not so sure what you mean by "properly use those in their scripts". imo passing config values via env vars is a very bad practice :(

Panaetius: so I have a script train.py
```
from renku.api import Parameter
learning_rate = Parameter("learning_rate", 0.1)
train_model(learning_rate)
```

And I do `renku run my-run -- python script.py`. Then `learning_rate` gets detected as a parameter, but doesn't show up on the CLI.

So if later I do `renku workflow execute --set learning_rate=50 my-run`, before that didn't do anything. We need a way for `Parameter` to pick up the modified value. But it can't be passed on the CLI, since the script doesn't have `argparse` or similar.

So the two way to do this is to write the value to a file and have `Parameter()` pick it up, or to set it as an ENV var and have it pick it up there. When I did that change, I opted for ENV vars as then we don't need to worry about littering the file system, permissions, path issues, parallelism (The env vars are per process, a file in a fixed path wouldn't be) etc.

renku/core/management/workflow/converters/cwl.py
```diff
@@ -367,10 +362,15 @@ def _convert_input(input: CommandInput, basedir: Path):
...
-            input_binding=cwlgen.CommandLineBinding(position=position, prefix=prefix, separate=separate),
+            input_binding = cwl.CommandLineBinding(position=position, prefix=prefix, separate=separate)
+
+        if position:
+            input_binding = cwl.CommandLineBinding(position=position, prefix=prefix, separate=separate)
+
+        return cwl.CommandInputParameter(
+            id=sanitized_id,
+            type=type_,
+            inputBinding=input_binding,
```
vigsterkr, line 373: note, without this workaround the value of inputBinding in the generated CWL will be `{}` in case `position` and `prefix` are actually `None` that will cause an erroneous execution of the workflow with `cwltool` 🤷

renku/core/management/workflow/converters/cwl.py
```diff
@@ -361,10 +362,15 @@ def _convert_input(input: CommandInput, basedir: Path):
                 prefix = prefix[:-1]
                 separate = True
 
+            input_binding = cwl.CommandLineBinding(position=position, prefix=prefix, separate=separate)
+
+        if position:
```
Panaetius, line 367: Should this be `elif`?

renku/core/management/workflow/converters/cwl.py
Panaetius: `inputBinding=cwl.CommandLineBinding(position=position, prefix=prefix, separate=separate) if position or prefix else None,` would also work, no?

renku/core/management/workflow/converters/cwl.py
vigsterkr: that was my first ideas as well, but felt cleaner like this. tbh no hard preference about this at all

renku/core/management/workflow/converters/cwl.py
vigsterkr: yeah, that should fly as well.

vigsterkr:
mmm idk yet what's the reason for the failing migration tests (`tests/cli/test_migrate.py`). i've ran those locally and all of them passes.

Panaetius:
> mmm idk yet what's the reason for the failing migration tests (`tests/cli/test_migrate.py`). i've ran those locally and all of them pass.

image
:wink:

Panaetius (approved):

Panaetius (approved):

Panaetius (approved):
thank you!

PR #2606, opened 2022-01-26
chore: fix pip install in publish action

Panaetius:
The publish action was still using `-e` which doesn't work with poetry (until they release version 1.2). This removes -e


vigsterkr (approved):
🚀

PR #2609, opened 2022-01-28
hotfix(service): v1.0.4 unlimited file size in nginx config

mohammad-sdsc:
# Description

Reverse-proxy NginX in multi core-service deployment has a limited packet size which prevents uploading large files when adding files to datasets in the UI.
This PR makes it accept unlimited messages. We should revert this back once chunked upload is implemented in the service (https://github.com/SwissDataScienceCenter/renku-python/issues/910). 

/deploy renku=1.0-next renku-graph=development renku-ui=master

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,8 +18,17 @@
 Changes
 =======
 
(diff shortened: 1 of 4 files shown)


ciyer (approved):
This fixes the problem; thanks a lot! 🎉

PR #2612, opened 2022-01-30
feat(dataset): use posix mv semantics when adding with destination

### LINKED ISSUE #1765: semantics of "destination" on "renku dataset add --destination"
I was under the impression that the semantics of what `destination` meant in 

```
renku dataset add --destination <destination> <dataset> <url>
```

were the same as they are for posix filesystem commands like `cp` and `mv`: If `destination` exists and is a directory, the file from `<url>` is added to this directory. If `destination` doesn't exist, it is taken to be the filename of whatever is downloaded from `<url>`. This is not the way the current `renku dataset add` command behaves - it _always_ creates a directory and adds the data downloaded from `<url>` to that directory.

mohammad-sdsc:
# Description
Adding data to a dataset with passing a destination using `--destination` flag, acts similar to posix `cp` and `mv` commands: Added file is renamed it destination doesn't exists or it's a file and it's copied to the destination otherwise.

Fixes #1765 

# Examples:
```
renku dataset create my-data
renku dataset add my-data https://file-examples-com.github.io/uploads/2017/02/file-sample_100kB.doc -d sample.doc
renku dataset add /some/directory/* -d new-directory
```


renku/core/management/datasets.py
```diff
@@ -476,59 +476,72 @@ def is_protected_path(self, path):
 
         return False
 
-    def _add_from_local(self, dataset, path, external, destination):
+    def _add_from_local(self, dataset, path, external, destination) -> List[Dict]:
         """Add a file or directory from a local filesystem."""
-        src = Path(os.path.abspath(path))
+        action = "symlink" if external else "copy"
```
Panaetius, line 481: Not something you need to change here, but `action` seems to be a good candidate for using an Enum

renku/core/management/datasets.py
```diff
@@ -476,59 +476,72 @@ def is_protected_path(self, path):
...
-                        raise errors.ProtectedFiles([src])
+            if not source_root.exists():
+                raise errors.ParameterError(f"Cannot find source file: {path}")
+            if source_root.is_dir() and destination_exists and not destination_is_dir:
+                raise errors.ParameterError(f"Cannot copy directory '{path}' to non-directory '{destination}'")
 
-            if path_in_repo:
-                return [{"path": path_in_repo, "source": path_in_repo, "parent": self}]
+            name = source_root.name if destination_exists and destination_is_dir else ""
+            return destination / name
```
Panaetius, line 505:
```suggestion
            return destination / source_root.name if destination_exists and destination_is_dir else destination
```
Seems cleaner to me, composing a Path with an empty string seems weird to me.

Panaetius (changes requested):

renku/core/utils/os.py
```diff
@@ -59,22 +68,23 @@ def get_relative_paths(base: Union[Path, str], paths: List[Union[Path, str]]) ->
     return relative_paths
 
 
+def get_files_recursively(path: Path) -> Generator[Path, None, None]:
```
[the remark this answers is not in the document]
mohammad-sdsc: I renamed it to `get_files`

Panaetius (approved):

PR #2620, opened 2022-02-01
fix(core): Wipe out .renku directory when initializing

### LINKED ISSUE #2543: Wipe out .renku directory when initializing
`renku init --force` keeps the content of `.renku` directory and changes only the files that are in the template. We should delete this directory all together.

vigsterkr:
# Description

Fixes #2543

### DIFF
--- a/renku/core/management/repository.py
+++ b/renku/core/management/repository.py
@@ -290,6 +290,8 @@ class RepositoryApiMixin(GitCore):
 
         # initialize repo and set user data
         path = self.path.absolute()
+        if force and (path / ".renku").exists():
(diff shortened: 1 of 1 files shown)


renku/core/management/repository.py
```diff
@@ -290,6 +290,8 @@ def init_repository(self, force=False, user=None, initial_branch=None):
 
         # initialize repo and set user data
         path = self.path.absolute()
+        if force and (path / ".renku").exists():
```
Panaetius, line 293: maybe use `RENKU_HOME` here so it's not a magic string

Panaetius (approved):
:rocket:

PR #2623, opened 2022-02-02
chore(ci): update to renku actions 0.5.0

olevski:
# Description

This updates the CI workflows to use the latest actions. It also enhances how the CI deployments are started and deleted.

The setup is as follows:

- if you do not provide `#persist` in the PR description when you specify `/deploy` then the CI deployment is deleted after the acceptance tests complete
- if you provide `#persist` the deployment is kept until the PR is closed.
- when the PR is closed the namespace of the deployment is also deleted
- when the acceptance tests complete AND `#persist` is not used the CI deployment is deleted but the namespace is not deleted in k8s - this makes it possible to reuse the TLS secret and not hit the limits for requesting certs with letsencrypt when the acceptance tests for the same PR are run many times
- artifacts from the acceptance tests are managed by the single (composite) test-renku action

## Type of change

chore

/deploy

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -30,7 +30,7 @@ jobs:
       extra-values: ${{ steps.deploy-comment.outputs.extra-values}}
     steps:
       - id: deploy-comment
-        uses: SwissDataScienceCenter/renku-actions/check-pr-description@v0.3.2
+        uses: SwissDataScienceCenter/renku-actions/check-pr-description@v0.5.0
         with:
           string: /deploy
           pr_ref: ${{ github.event.number }}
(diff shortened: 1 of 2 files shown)


Panaetius (approved):
Wait, `Unable to resolve action `SwissDataScienceCenter/deploy-renku@v0.5.0`, repository not found` something seems wrong here

olevski:
@Panaetius just me not being able to copy/paste right. Should be fine now.

Panaetius (dismissed):

olevski:
@Panaetius sorry for the back and forth. Everything is good now. Even the acceptance tests fully pass - in addition to running as expected with the new action.

Panaetius (approved):

PR #2625, opened 2022-02-02
test(service): fix cleanup of temporary IT branches

Panaetius:
The fixture has code to cleanup temporary branches, but in 3/4 tests, this wasn't actually called. I've now added it to teardown.


vigsterkr (approved):
🚀

PR #2626, opened 2022-02-03
docs: add placeholder for mls plugin gallery

gavin-k-lee:
# Description

Adds a gallery placeholder for Renku plug-ins and mentions the MLS plugin specifically. Discourse post placeholder should be updated when it is uploaded.

## Type of change

`docs`

### DIFF
--- a/docs/reference/plugins.rst
+++ b/docs/reference/plugins.rst
@@ -16,6 +16,20 @@
     limitations under the License.
 
 
+Plugin Gallery
+==============
+
+Here you will find a list of Renku compatible plugins. A list of supported types of plugins is listed in the Plugin Support section.
+
+Renku-MLS plugin
+----------------
+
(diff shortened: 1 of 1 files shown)


ksanao (approved):

docs/reference/plugins.rst
```diff
@@ -16,6 +16,20 @@
     limitations under the License.
 
 
+Plugin Gallery
```
vigsterkr, line 19: as previously stated i'm strongly against aggregating in a static document dynamically created plugins. i can list here as well my reasons regarding this if it's required.

vigsterkr (commented):
thanks for this @gavin-k-lee. for details see my comments about this topic on slack in the thread.

docs/reference/plugins.rst
gavin-k-lee: thanks viktor - we will close this for now.

PR #2628, opened 2022-02-04
build: update sentry values

lorenzo-cavazzi:
Align sentry values with the other renku repositories.

Reference: https://swiss-data-science.slack.com/archives/G67N59QL8/p1643901194490749

/deploy


renku/cli/exception_handler.py
```diff
@@ -68,11 +68,10 @@
 
 _BUG = click.style("Ahhhhhhhh! You have found a bug. 🐞\n\n", fg="red", bold=True)
 
-HAS_SENTRY = None
-SENTRY_DSN = os.getenv("SENTRY_DSN")
+HAS_SENTRY = os.getenv("SENTRY_ENABLED").lower() == "true"
```
Panaetius, line 71:
```suggestion
HAS_SENTRY = os.getenv("SENTRY_ENABLED", default="false").lower() == "true"
```
otherwise this fails if `SENTRY_ENABLED` is None (`None.lower()` doesn't work).

Pay attention, this happened here: https://github.com/SwissDataScienceCenter/renku-python/runs/5068558729?check_suite_focus=true#step:10:76 where building the cheatsheet didn't fully work because of this exception. Since the cheatsheet changed due to this bug, it was updated, so you should undo that change/commit (I think it'll automatically fix that once the bug is fixed, making a new cheatsheet commit?)

While you're at it, could you change https://github.com/SwissDataScienceCenter/renku-python/blob/develop/.github/workflows/test_deploy.yml#L160 to fail on warnings? I'm a bit surprised that sphinx build actually passed here...

renku/service/entrypoint.py
```diff
@@ -47,7 +47,7 @@
 
 logging.basicConfig(level=os.getenv("SERVICE_LOG_LEVEL", "WARNING"))
 
-if os.getenv("SENTRY_DSN"):
+if os.getenv("SENTRY_ENABLED").lower() == "true":
```
Panaetius, line 50: Same as above

renku/service/worker.py
```diff
@@ -27,7 +27,7 @@
 from renku.service.jobs.queues import QUEUES, WorkerQueues
 from renku.service.logger import DEPLOYMENT_LOG_LEVEL, worker_log
 
-if os.getenv("SENTRY_DSN"):
+if os.getenv("SENTRY_ENABLED").lower() == "true":
```
Panaetius, line 30: Same as above

Panaetius (changes requested):

renku/service/entrypoint.py
rokroskar: the service has a `config.py` is that not used anymore? That way you can make sure to do it right in one place and use the proper boolean from `config` elsewhere and avoid the problem in the future.

Panaetius: Right, good point. I forgot about that one

renku/cli/exception_handler.py
lorenzo-cavazzi: I added the `-W` flag to make it fail on warnings, and it seems to work.
Hopefully, that's only on the last-1 run of the actions as soon as you read this :grin: 

Screenshot_20220208_174039

lorenzo-cavazzi:
Thanks for the review @Panaetius and the suggestion @rokroskar 
I should have fixed those problems, even though I'm optimistically writing this sentence before the tests have actually passed :crossed_fingers:

Panaetius (approved):

PR #2645, opened 2022-02-10
fix(service): bump Pillow to 9.0+ to fix security vulnerability

Panaetius:
Closes https://github.com/SwissDataScienceCenter/renku-python/security/dependabot/6
Closes https://github.com/SwissDataScienceCenter/renku-python/security/dependabot/5
Closes https://github.com/SwissDataScienceCenter/renku-python/security/dependabot/4


vigsterkr (approved):
🔐 🚀

PR #2676, opened 2022-02-11
fix(service): remove json requirement from project.lock_status

### LINKED ISSUE #2675: Wrong content type for `project.lock_status` in service
**Describe the bug**
Calling `project.lock_status` endpoint requires setting a `Content-Type: application/json` header while it uses query string parameters and there is no need for json.

mohammad-sdsc:
Fixes #2675 

/deploy #persist

### DIFF
--- a/tests/service/views/test_project_views.py
+++ b/tests/service/views/test_project_views.py
@@ -119,7 +119,9 @@ def test_get_lock_status_unlocked(svc_client_setup):
     """Test getting lock status for an unlocked project."""
     svc_client, headers, project_id, _, _ = svc_client_setup
 
-    response = svc_client.get("/1.0/project.lock_status", query_string={"project_id": project_id}, headers=headers)
+    response = svc_client.get(
+        "/1.0/project.lock_status", query_string={"project_id": project_id}, headers=headers, content_type="text/xml"
+    )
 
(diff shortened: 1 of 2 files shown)


tests/service/views/test_project_views.py
```diff
@@ -119,7 +119,9 @@ def test_get_lock_status_unlocked(svc_client_setup):
     """Test getting lock status for an unlocked project."""
     svc_client, headers, project_id, _, _ = svc_client_setup
 
-    response = svc_client.get("/1.0/project.lock_status", query_string={"project_id": project_id}, headers=headers)
+    response = svc_client.get(
+        "/1.0/project.lock_status", query_string={"project_id": project_id}, headers=headers, content_type="text/xml"
```
Panaetius, line 123: Is `content_type="text/xml"` to test that it works with any content_type? Shouldn't it work without setting a content type?

mohammad-sdsc: It is for having a failing test. The default content type set by the flask test client is `application/json` and this test won't fail if the endpoint has `@accepts_json` decorator.

Panaetius: ah I see

Panaetius (approved):

PR #2677, opened 2022-02-11
build(deps): update dependencies

Panaetius:
Update dependencies to newest version.


vigsterkr (approved):
📦 🚀

Panaetius:
Sorry, Dependabot just doesn't want to stop. I miss the times when it was broken :cry:

vigsterkr (dismissed):

vigsterkr:
mmm the `cwltool==3.1.20220210171524` update went through it seems :( unfortunately since it generates a warning on the stdout, this results a pytest failure for all `tests/cli/test_workflow.py` tests. either i need to refactor those tests, or (which i suggested in the cwltool update PR) would rather keep the pinned version of `cwltool` that's coming from `toil==5.6.0`

vigsterkr (approved):
🚀

PR #2680, opened 2022-02-12
fix(core): silence cwltool version warning of toil

vigsterkr:
# Description

this patch silences the `toil` warning for the version difference of the `cwltool`

### DIFF
--- a/renku/core/management/workflow/providers/toil.py
+++ b/renku/core/management/workflow/providers/toil.py
@@ -19,6 +19,7 @@
 
 import functools
 import itertools
+import logging
 import os
 import shutil
 import uuid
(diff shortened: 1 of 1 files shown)


renku/core/management/workflow/providers/toil.py
```diff
@@ -45,6 +45,13 @@
 from renku.core.plugins import hookimpl
 from renku.core.plugins.provider import RENKU_ENV_PREFIX
 
+# silence cwltool version warning
+logger = logging.getLogger("toil.cwl")
+logger.disabled = True
+from toil.leader import FailedJobsException  # noqa
+
+logger.disabled = False
```
Panaetius, line 53: Do we need to set disabled twice?

Panaetius (commented):
Would this prevent other messages from cwl to be shown?

renku/core/management/workflow/providers/toil.py
vigsterkr: well i though that maybe having it on again would be not a bad idea :) but i'll close this as pinning of `cwltool` fixes the whole problem that this supposed to be solving temporarily

PR #2681, opened 2022-02-14
feat(service): expose warnings and errors on cache.migrate endpoint

### LINKED ISSUE #2618: return warnings in core-service cache.migrate
Currently we only return `communication.messages` as `result.messages` in the `cache.migrate` endpoint. We should also return `communication.warnings` as `result.warnings` so the UI can show warnings to the user.

Panaetius:
close #2618


vigsterkr (dismissed):
:shipit:

vigsterkr (dismissed):
🚀

vigsterkr (approved):
🎸

vigsterkr (approved):

PR #2686, opened 2022-02-14
fix(dataset): unset wasDerivedFrom for imported datasets

### LINKED ISSUE #2685: Imported renku datasets have `prov:wasDerivedFrom` set
**Describe the bug**
`prov:wasDerivedFrom` is used to chain different versions of a dataset in a project. A freshly imported dataset should not have this field.

**To Reproduce**
1. Renku CLI v1.0.5
2. `renku dataset import https://dev.renku.ch/projects/renku-test-projects/test-subgroup-1/dataset-import-in-sub-groups/datasets/remote-dataset --yes --name remote-dataset`
2. Look at the dataset's metadata (find the oid: `cat .renku/metadata/datasets` then `zstdcat path/to/metadata | jq`)
3. `derived_from` is not null

mohammad-alisafaee:
Fixes #2685 
Fixes IT

### DIFF
--- a/renku/cli/dataset.py
+++ b/renku/cli/dataset.py
@@ -212,8 +212,7 @@ them.
 
 You can use ``--destination`` or ``-d`` flag to set the location where the new
 data is copied to. This location be will under the dataset's data directory and
-will be created if does not exists. You will get an error message if the
-destination exists and is a file.
+will be created if does not exists.
 
 .. code-block:: console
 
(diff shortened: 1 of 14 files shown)


Panaetius (approved):
Thank you!

Some Zenodo and Dataverse tests still failing, and two datasets on renku dev not found, but we can look into that separately

mohammad-sdsc:
> Some Zenodo and Dataverse tests still failing, and two datasets on renku dev not found, but we can look into that separately

All renku failures are related to the KG issue not returning the latest version of a dataset. The Zenodo one fails because (I guess) it was restricted before and it's now public. I replaced it with another one.

Panaetius (approved):

PR #2687, opened 2022-02-15
docs: add renku session RFC

vigsterkr:
# Description
 
add the `renku session` sub-command RFC

### DIFF
--- a/design/003-interactive-session/003-interactive-session.md
+++ b/design/003-interactive-session/003-interactive-session.md
@@ -1,0 +1,123 @@
+- Start Date: 2021-12-02
+- Status: Proposed
+
+# Run interactive sessions
+
+## Summary
+
+Propose a new sub-group command `renku session` that is responsible for
+managing interactive sessions, both local and remote, for a project.
+
+## Motivation
+
+Renkulab.io provides the option to run interactive sessions of a project. Among other things the session
+contains a jupyter server.
+
+The motivation behind this RFC is to provide a simple renku CLI sub-command that basically allows the
+starting and stopping of these interactive sessions.
+
+## Design Detail
+
+The proposed sub-commands of `renku session`, namely to run interactive sessions. The local session heavily relies on
+[docker](https://www.docker.com/). Docker is one of the most popular container engines, that eases the development
+and deployment of applications, by off-loading the installation of the runtime environment dependencies of the
+application.
+
+In order to start a local interactive session, just as in case of renkulab.io, it is required to provide
+a `Dockerfile`. This is by default provided by `renku init` when creating a renku project, but if the user
+is using a custom project template it might not be present.
+
+### Proposed Commands
+
+#### renku session start
+
+A command to start an interactive session.
+
+By default, the `Dockerfile` present in the given project is going to be used for building the docker image
+that's going to be used for running the interactive session. In case the docker image is not available the
+user is going to be prompted whether the user wants to build the docker image, as in some cases it can be
+time consuming.
+
+With the `--provider <provider_type>` flag a user can specify session provider to be used for starting
+an interactive session. By default two different providers are shipped with Renku CLI:
+ - docker
+ - renkulab
+The provider specific configuration values can be specified by using the `--config <config.yaml>` flag.
+
+##### Detailed Parameter Description
+
+```
+renku session start [OPTIONS]
+
+--image <image_name>			Override the docker image to be used for the interactive session.
+-p|--provider [docker|renkulab] 	Use the specified backend for starting an interactive session.
+-c|--config   				YAML file containing configuration for the provider.
+--cpu <num of cpus>			Specify how much of the available CPU resources a container can use.
+--disk <disk space>			Specify the amount of disk space for the container.
+--gpu <device>				Specify the GPU allocated for the container.
+--memory <max memory>			The maximum amount of memory the container can use.
+
+```
+
+#### renku session stop
+
+`renku session stop my_session` simply stops the given interactive session.
+
+One can shut down all the currently running interactive sessions at once with the `renku session stop --all`.
+
+##### Detailed Parameter Description
+
+```
+renku session stop [OPTIONS] <name>
+
+<name>      Name of the session that the user wants to stop.
+
+--all       Stops all the running containers.
+
+```
+
+#### renku session list
+
+The command to list all the currently running interactive sessions for the given project, with the
+detailed information regarding the network port mappings.
+
+##### Detailed Parameter Description
+
+```
+renku session list [OPTIONS]
+
+-p|--provider [docker|renkulab] 	Use the specified backend for listing the interactive sessions.
+-c|--config   				YAML file containing configuration for the provider.
+
+```
+
+#### renku session open
+
+`renku session open <my_session>` opens the given interactive session in the user's default browser.
+
+Note, it is possible that the user is running renku on a remote machine via SSH. In this case this command
+might fail, unless for example X-Fowarding is not enabled.
+
+## Drawbacks
+
+One of the major drawback of the current design is the dependency on docker. Namely, to only support local
+interactive session if docker is available. Many HPC environments do not support docker, but they support
+other container formats, like [runC](https://github.com/opencontainers/runc). As a long term goal of this
+effort it would be good to consider support for different container engines, hence the design and implementation
+of the initial `renku session` should take this into consideration.
+
+## Rationale and Alternatives
+
+> Why is this design the best in the space of possible designs?
+
+> What other designs have been considered and what is the rationale for not choosing them?
+
+> What is the impact of not doing this?
+
+## Unresolved questions
+
+> What parts of the design do you expect to resolve through the RFC process before this gets merged?
+
+> What parts of the design do you expect to resolve through the implementation of this feature before stabilisation?
+
+> What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?
--- a/design/003-interactive-session/003-interactive-session.md
+++ b/design/003-interactive-session/003-interactive-session.md
@@ -10,7 +10,7 @@ managing interactive sessions, both local and remote, for a project.
 
 ## Motivation
 
-Renkulab.io provides the option to run interactive sessions of a project. Among other things the session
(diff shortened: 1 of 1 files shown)


olevski (dismissed):
Looks good to me. Are the last two sections meant to be blank - ie just have questions with no answers?

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,123 @@
...
+# Run interactive sessions
+
+## Summary
+
+Propose a new sub-group command `renku session` that is responsible for
+managing interactive sessions, both local and remote, for a project.
+
+## Motivation
+
+Renkulab.io provides the option to run interactive sessions of a project. Among other things the session
```
rokroskar, line 13: Just want to make sure it's clear that this isn't limited to renkulab.io

```suggestion
A renkulab instance (such as Renkulab.io) provides the option to run interactive sessions of a project. Among other things the session
```

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,123 @@
...
+#### renku session stop
+
+`renku session stop my_session` simply stops the given interactive session.
+
+One can shut down all the currently running interactive sessions at once with the `renku session stop --all`.
+
+##### Detailed Parameter Description
+
+```
+renku session stop [OPTIONS] <name>
```
rokroskar, line 71: this needs a provider, otherwise you need to loop through all the possible providers to find the right session id

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,123 @@
...
+```
+renku session start [OPTIONS]
+
+--image <image_name>			Override the docker image to be used for the interactive session.
+-p|--provider [docker|renkulab] 	Use the specified backend for starting an interactive session.
+-c|--config   				YAML file containing configuration for the provider.
+--cpu <num of cpus>			Specify how much of the available CPU resources a container can use.
+--disk <disk space>			Specify the amount of disk space for the container.
+--gpu <device>				Specify the GPU allocated for the container.
+--memory <max memory>			The maximum amount of memory the container can use.
```
rokroskar, line 58: in case of docker, what are the defaults for cpu/memory? Do we take them from the project's configuration if it exists? If it does not exist, do we leave it blank (i.e. use all available resources)?

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,123 @@
...
+renku session list [OPTIONS]
+
+-p|--provider [docker|renkulab] 	Use the specified backend for listing the interactive sessions.
+-c|--config   				YAML file containing configuration for the provider.
+
+```
+
+#### renku session open
+
+`renku session open <my_session>` opens the given interactive session in the user's default browser.
```
rokroskar, line 96: needs `--provider` unless we encode the provider in the session id

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,123 @@
...
+
+Note, it is possible that the user is running renku on a remote machine via SSH. In this case this command
+might fail, unless for example X-Fowarding is not enabled.
+
+## Drawbacks
+
+One of the major drawback of the current design is the dependency on docker. Namely, to only support local
```
rokroskar, line 103: this is an excellent point - we have users that use podman - maybe we can consider what support for it would look like?

rokroskar (changes requested):
this is great, thanks @vigsterkr for separating it out from the implementation. I still don't fully understand how you intend to find the sessions in commands that require a session ID without knowing which provider to query. I guess it might not matter much in practice, but it seems unnecessary to fetch all lists of sessions from all providers every time. Either we should encode the provider name in the session id, e.g. 

`renkulab.io/<id>`
`docker/<id>`

or use a `--provider` flag in those commands as well. I prefer the first option because it's less typing for the user... but we have to come up with a good scheme.

vigsterkr:
> this is great, thanks @vigsterkr for separating it out from the implementation. I still don't fully understand how you intend to find the sessions in commands that require a session ID without knowing which provider to query. I guess it might not matter much in practice, but it seems unnecessary to fetch all lists of sessions from all providers every time. Either we should encode the provider name in the session id, e.g.
> 
> `renkulab.io/<id>` `docker/<id>`
> 
> or use a `--provider` flag in those commands as well. I prefer the first option because it's less typing for the user... but we have to come up with a good scheme.

the primary reason was to avoid the explicit mentioning of the provider in some cases (see when dealing with `session_id`) that to avoid unnecessary typing on the CLI from the user. `pluggy` by default calls all the registered plugin hooks, and of course there's a way around to do this... but i thought that the overhead on the backend side is less than the gained value of not typing things out. This week we talked out prefixing the `session_id` (mentioned above) with the provider, but to be honest then this adds additional string processing (prone to errors) on the backend side. in this case i'd rather just go with the explicit requiring the `-p/--provider <provider>` flag from the user. coz essentially it'll do the same thing on the end of the day (from `pluggy` perspective), but we'd just use the user provided `provider` instead of doing some string processing for extracting the provider type and the `session_id`

rokroskar:
> in this case i'd rather just go with the explicit requiring the -p/--provider <provider> flag from the user

That's fine - the default can be to cycle through all of them but having a consistent interface where one can specify a provider makes sense to me.

design/003-interactive-session/003-interactive-session.md
vigsterkr: it's the default of the provider...

vigsterkr: namely i dont see neither the requirement nor the point of trying to unify it in this level... each provider has it's own default values and they are going to be used unless these are set by the user...

rokroskar: right, but the user thinks they have set the default by setting it in the project's config

vigsterkr: it'll be used if it's set in the project's config... unless if it's not set. or maybe i'm misunderstanding here something...?

rokroskar: No that's fine, but it wasn't stated so I'm just making sure that's the intent.

vigsterkr: ah, i'll then just add a sentence in the command description that the resource related options are there to override the [project level resource settings](https://renku.readthedocs.io/en/latest/reference/templates.html?highlight=cpu_request#renku)

rokroskar (commented):
two small comments but then I think we're finally done :)

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,206 @@
...
+ - finding a specific container image. Both locally (cached) and as well in a remote repository.
+ - building a container image based on an image specs file.
+ - starting a container using a specific image.
+ - stopping a running container.
+ - get the URL of the interactive session of a running container.
+
+The implementation of this interface in detail:
+
+```python
+class ISessionProvider:
```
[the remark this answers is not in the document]
vigsterkr: definitely not. this is both in the code and the code follows the 'standards' within the whole codebase... i.e. if this then we need to change this everywhere (other interface definitions)

design/003-interactive-session/003-interactive-session.md
```diff
@@ -0,0 +1,206 @@
...
+## Drawbacks
+
+One of the major drawback of the current design is the dependency on docker. Namely, to only support local
+interactive session if docker is available. Many HPC environments do not support docker, but they support
+other container formats, like [runC](https://github.com/opencontainers/runc). As a long term goal of this
+effort, support for different container engines is essential. In order to allow extending the `renku session`
+command with different container engines a very simple API is defined.
+
+The interface requires that the container engine support the following functionalities:
+ - finding a specific container image. Both locally (cached) and as well in a remote repository.
```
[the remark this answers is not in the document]
vigsterkr: we are not dealing with authentication. this is a wrapper around working with the containers... the authentication shall be done from the container manager side.

design/003-interactive-session/003-interactive-session.md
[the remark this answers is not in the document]
rokroskar: oh is "Interactive" always abbreviated as "I"?

design/003-interactive-session/003-interactive-session.md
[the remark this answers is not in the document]
vigsterkr: if the user has previously done a `docker login` then the pull will be able to pull from a private repo... if that's not done it won't be able and then it will complain that it cannot find the image.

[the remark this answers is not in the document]
rokroskar: yeah but find image will fail if you try to look at a private registry and don't authenticate. So you're just saying the user will have to do that by some other means?

design/003-interactive-session/003-interactive-session.md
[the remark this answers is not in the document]
Panaetius: I is for interface. `IInteractiveSessionProvider`?

design/003-interactive-session/003-interactive-session.md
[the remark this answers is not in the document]
rokroskar: ok so in our case maybe `renku login` needs to handle the authentication with the gitlab registry...

design/003-interactive-session/003-interactive-session.md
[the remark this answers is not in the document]
rokroskar: haha oh ok. I misunderstood 😇 all good then

rokroskar (approved):

design/003-interactive-session/003-interactive-session.md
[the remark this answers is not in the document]
vigsterkr: no what i mean is that if you try that you'll see that you'll get an error from docker that says try to use `docker login` coz it couldn't find on the public repos the image that you are looking for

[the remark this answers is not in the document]
rokroskar: of course, but our users won't really know what to do with that probably. That's why I thought login could be one of the parts of the interface.

Panaetius (approved):

PR #2688, opened 2022-02-15
feat(core): renku session implementation

### LINKED ISSUE #1991: CLI command for launching an interactive session locally

vigsterkr:
# Description

Fixes #1991 #2394


rokroskar:
what happened to the nice gif? :)

olevski:
@vigsterkr I think I may be doing something wrong but this is what I did:
1. Install this PRs version of renku cli
2. `renku init`
3. `renku session start`

But I get the following error:
```
Error: Docker failed: 400 Client Error for http+docker://localhost/v1.41/images/None/None:52cb63b/json: Bad Request ("no such image: None/None:52cb63b: invalid reference format: repository name must be lowercase")
```

I guess it is expected that this is a real renku project that exists somewhere on renkulab or smth like that and if the project is simply a local project that is not pushed anywhere then things do not work.

Another thing I tested is passing the image as an option to the command. Then everything is fine except for the message that tells you where you can access the session. The message was blank - there is no link. If I do `renku session list` or `renku session open ...` then everything works and I can definitely access the session.
```../test_env/bin/renku session start --image renku/renkulab-py:python-3.8.8-0.11.0
The session for 'renku/renkulab-py:python-3.8.8-0.11.0' has been successfully started. It is available at:
```

The last thing I checked is fixing the image in the renku project config. In this case a simple `renku session start` worked exactly as expected.
```
renku config set interactive.image renku/renkulab-py:python-3.8.8-0.11.0`
```

renku/core/commands/session.py
```diff
@@ -0,0 +1,177 @@
...
+
+    providers = supported_session_providers()
+    if provider:
+        providers = list(filter(lambda x: x[1] == provider, providers))
+
+    if len(providers) == 0:
+        raise errors.ParameterError("No session provider is available!")
+
+    sessions_list = list(chain(*map(lambda x: x[0].session_list(config=config, project_name=project_name), providers)))
+    return SESSION_FORMATS[format](sessions_list)
```
Panaetius, line 64: In the DatasetsApiMixin refactor (and in general) I've started moving this around so the `FORMAT[format](something)` call happens in `renku.cli` not core, since it is rather cli-specific (service wouldn't use tabular format as a return so core shouldn't know about formats, as it's not a generic concept).

E.g. https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/cli/dataset.py#L554 and https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/core/management/dataset/dataset.py#L82

And all the logic should be in `renku.core.management.session` somewhere, this file should only contain the command wrapping functions, e.g. https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/core/commands/dataset.py#L44 to separate things better.

renku/core/commands/session.py
```diff
@@ -0,0 +1,177 @@
...
+    image_name: str = None,
+    cpu_request: Optional[float] = None,
+    mem_request: Optional[str] = None,
+    disk_request: Optional[str] = None,
+    gpu_request: Optional[str] = None,
+):
+    client = client_dispatcher.current_client
+
+    project_config = client.load_config()
+    pinned_image = project_config.get("interactive", "image", fallback=None)
```
Panaetius, line 86:
```suggestion
    pinned_image = client.get_value("interactive", "image")
```
Also for other places where config is used.

renku/core/management/session/docker.py
```diff
@@ -0,0 +1,197 @@
...
+
+class DockerSessionProvider(ISessionProvider):
+    """A docker based interactive session provider."""
+
+    JUPYTER_PORT = 8888
+
+    def __init__(self):
+        self._docker_client = None
+
+    def docker_client(self) -> docker.client.DockerClient:
```
Panaetius, line 40:
```suggestion
    @property
    def docker_client(self) -> docker.client.DockerClient:
```

renku/core/management/session/docker.py
```diff
@@ -0,0 +1,197 @@
...
+        self.docker_client().images.build(path=str(image_descriptor), tag=image_name)
+
+    def find_image(self, image_name: str, config: Optional[Dict[str, Any]]) -> bool:
+        """Find the given container image."""
+        try:
+            _ = self.docker_client().images.get(image_name)
+            return True
+        except docker.errors.ImageNotFound:
+            try:
+                _ = self.docker_client().images.pull(image_name)
```
Panaetius, line 70:
```suggestion
                _ = self.docker_client().images.pull(image_name)
                return True
```
?

renku/core/management/session/docker.py
```diff
@@ -0,0 +1,197 @@
...
+
+        try:
+            docker_is_running = self.docker_client().ping()
+            if not docker_is_running:
+                raise errors.DockerError(
+                    "Could not communicate with the docker instance. Please make sure it is running!"
+                )
+
+            auth_token = uuid4().hex
+            project_config = client.load_config()
```
Panaetius, line 124: Same here with the config

renku/core/models/session.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class ISessionProvider(metaclass=ABCMeta):
+    """Abstract class for a interactive session provider."""
+
+    @abstractmethod
+    def build_image(self, image_descriptor: Path, image_name: str, config: Optional[Dict[str, Any]]) -> Optional[str]:
+        """Builds the container image.
+
+        :param image_descriptor: Path to the container image descriptor file.
```
Panaetius, line 45: Might need to change docstring style depending on what we decide on in slack

Panaetius (changes requested):

renku/core/management/session/docker.py
vigsterkr: yeah i wanted this, but cannot recall why i had to resolve to use this... i'll change it locally and will report the whys

Panaetius (approved):

Panaetius (approved):

PR #2693, opened 2022-02-16
chore: limit concurrent dependabot PRs

Panaetius:
Limits number of open dependabot PRs to 1, so we don't have issues with Dependabot rebasing lots of PRs and using up github runners.


vigsterkr (approved):
:shipit:

PR #2695, opened 2022-02-16
chore: add action to update develop on tags

lorenzo-cavazzi:
# Description

Automatically merge `master` to `develop` after tagging a new release.
This will keep the tag in the history of the `develop` branch and simplify creating PRs on `master`, overcoming the limitations we have in merging directly to the protected `develop` branch.

### DIFF
--- a/.github/workflows/test_deploy.yml
+++ b/.github/workflows/test_deploy.yml
@@ -850,7 +850,7 @@ jobs:
           ref: 'develop'
       - name: merge changes from master
         run: |
-          git merge master -X theirs
+          git merge master
           git push
 
   build-images:
--- a/.github/workflows/test_deploy.yml
+++ b/.github/workflows/test_deploy.yml
@@ -840,6 +840,19 @@ jobs:
           PYPI_TOKEN: ${{ secrets.PYPI_TOKEN }}
         run: make publish-lock-package
 
+  update-develop-branch:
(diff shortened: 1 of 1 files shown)


.github/workflows/test_deploy.yml
```diff
@@ -840,6 +840,19 @@ jobs:
...
+    runs-on: ubuntu-latest
+    if: "startsWith(github.ref, 'refs/tags/')"
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          ref: 'develop'
+      - name: merge changes from master
+        run: |
+          git merge master -X theirs
```
Panaetius, line 853: I think we don't want `-X theirs`, better to fail on conflict than potentially lose work.

And when I tried locally, if there's no new commits on develop, there won't be conflicts.

lorenzo-cavazzi: I'm pretty sure I had conflicts on the (few) changes we introduce when merging to master, but we can try without and see how it goes

Panaetius: I tried it in two rounds. I did a 1.0.6 release like you, which unborked 1.0.5 and had conflicts. but then I added a dummy commit to develop, made a 1.0.7 release and got no conflicts.

Panaetius (approved):
Thank you!

PR #2638, opened 2022-02-09
chore: switch to global redis instance

Panaetius:
close #2519 

Shamelessly stolen from https://github.com/SwissDataScienceCenter/renku-gateway/pull/537

/deploy #persist

PR #2696, opened 2022-02-16
chore: switch to global redis instance

### LINKED ISSUE #2519: Use global redis instance in core service
We now have a global redis instance configured at the renku chart level. The core service should either just use that global instance or make it configurable wether it will deploy its own redis. 
See https://github.com/SwissDataScienceCenter/renku/pull/2332 and https://github.com/SwissDataScienceCenter/renku-gateway/pull/488 for context.

olevski:
close #2519 

Replaces #2638 

NOTE: This introduces breaking changes in the helm chart.

Need to publish a new version of renku cli 0.16.x to get renku-core to work with redis with a sentinel.

**Please add any comments on the helm chart and its organization in the parent PR in the renku repo at https://github.com/SwissDataScienceCenter/renku/pull/2483. That way we can all agree on the any potential changes to the helm chart.**

/deploy #persist


olevski:
I could not get this to work with passing the `extra values` to `/deploy` for some reason. So I just modified the `values.yaml` directly to use the v8 core version with the fixes for the redis sentinel.

The way to merge/approve this whole thing is:
1. Review, merge and release a fix for 0.16.x, see #2700 
2. Update values file to use the new release i.e. 0.16.5 for the v8 version of renku core
3. Test/review/merge this PR

helm-chart/renku-core/values.yaml
```diff
@@ -126,6 +123,6 @@ versions:
     nameOverride: ""
     fullnameOverride: ""
     image:
-      repository: renku/renku-core
-      tag: "v0.16.4"
+      repository: olevski90/renku-core
```
Panaetius, line 126: We should change this before we merge :wink:

Panaetius (changes requested):

PR #2699, opened 2022-02-17
fix(dataset): use correct destination when adding from url

mohammad-sdsc:
# Description

Fixes a remaining case from #1765


vigsterkr (approved):
🚀

PR #2700, opened 2022-02-18
fix: work with redis sentinel

olevski:
This is a PR that applies the same changes as #2696 to 0.16.4. It will result in a new version - 0.16.5.

NOTE: target branch is `0.16-lts` from there we can make a new release from 0.16.5.

### DIFF
--- a/helm-chart/renku-core/requirements.yaml
+++ b/helm-chart/renku-core/requirements.yaml
@@ -1,5 +1,1 @@
-dependencies:
-- name: redis
-  version: 10.7.11
-  repository: "https://charts.bitnami.com/bitnami"
-
+dependencies: []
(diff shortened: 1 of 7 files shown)


olevski:
This is deployed/tested in use over at https://github.com/SwissDataScienceCenter/renku-python/pull/2696

Panaetius:
I think the the base renku chart is not compatible anymore with 0.16.x, and acceptance tests changed anyways, so there's not really an easy way to deploy and test fixes to old versions.

1.0.x chart can deploy 0.16.x as part of its deployment, but then tests would run against the 1.0.x core-service, not 0.16.x

Panaetius (approved):

PR #2703, opened 2022-02-18
fix(core): avoid migration failure for invalid dataset names

### LINKED ISSUE #2702: migration fails for datasets with invalid name
**Describe the bug**
When importing datasets from old projects, migration fails if a dataset has an invalid name. The migration can be continued if we convert the dataset's name to have the correct format. 

**To Reproduce**
1. Renku CLI v1.0.5
2. `renku dataset import https://dev.renku.ch/datasets/428c36261c56463d8753336470cc6917/`

mohammad-sdsc:
Fixes #2702


Panaetius (approved):
Thank you

PR #2706, opened 2022-02-21
chore: improve CLI docker build

### LINKED ISSUE #2694: fix Dockerfile.cli build
Dockerfile.cli does not build anymore due to the Makefile not being copied over. We should fix this

rokroskar:
1. copy Makefile to the build container to avoid failing docker builds
2. run the CLI container as a non-root user
3. ensure there is a `-dirty` flag on the installed versions
4. move build dependencies to the builder 

closes #2694 

/deploy


Dockerfile.cli
```diff
@@ -1,14 +1,14 @@
-FROM python:3.7-slim as base
+FROM python:3.9-slim as base
 
 RUN apt-get update && \
-    apt-get install -y git make && \
+    apt-get install -y git make gcc python3-dev && \
```
vigsterkr, line 4: is there any specific reason for `gcc`? meaning why `gcc` and not `clang`? note `clang` usually compiles faster than `gcc`

vigsterkr: and a compiler is enough? maybe a broader `build-essential` would be better? but maybe that's an over-kill

vigsterkr (approved):
🚀 with some minor comments, that could be ignored - hence the approve

Dockerfile.cli
Panaetius: Seconding `build-essential`, it's basically make + gcc plus a few other useful things

rokroskar: "G" in "gnu" is for "generic"? 😆  No idk, I really don't have any reason other than `pip` seems to be looking for `gcc`:

```
#12 11.89   × python setup.py egg_info did not run successfully.
#12 11.89   │ exit code: 1
#12 11.89   ╰─> [3 lines of output]
#12 11.89       sys.argv ['/tmp/pip-install-nf1gluec/ruamel-yaml-clib_0963e9fa68ce45c8922f90bac4e98668/setup.py', 'egg_info', '--egg-base', '/tmp/pip-pip-egg-info-j01kvpnx']
#12 11.89       test compiling /tmp/tmp_ruamel_3v54wua7/test_ruamel_yaml.c -> test_ruamel_yaml compile error: /tmp/tmp_ruamel_3v54wua7/test_ruamel_yaml.c
#12 11.89       Exception: command 'gcc' failed: No such file or directory
#12 11.89       [end of output]
```

vigsterkr: mmm can you do an `echo $CC` plz?

vigsterkr: coz it's either a shitty package - where they presume that a C compiler is always named gcc in the PATH; note this can happen, seen it before - or the `CC` env var is forced to be set for gcc... coz by default one should use `cc` or `c++` which on an ubuntu/debian system is always pointing to `/etc/alternatives/cc` et.al. which is set automagically to the installed compiler

rokroskar: I set `CC=clang` and then get

```
#12 25.93       clang -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -I/usr/local/include/python3.9 -c api.c -o build/temp.linux-aarch64-3.9/api.o
#12 25.93       clang -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -I/usr/local/include/python3.9 -c dumper.c -o build/temp.linux-aarch64-3.9/dumper.o
#12 25.93       clang -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -I/usr/local/include/python3.9 -c emitter.c -o build/temp.linux-aarch64-3.9/emitter.o
#12 25.93       clang -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -I/usr/local/include/python3.9 -c loader.c -o build/temp.linux-aarch64-3.9/loader.o
#12 25.93       clang -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -I/usr/local/include/python3.9 -c parser.c -o build/temp.linux-aarch64-3.9/parser.o
#12 25.93       clang -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -I/usr/local/include/python3.9 -c reader.c -o build/temp.linux-aarch64-3.9/reader.o
#12 25.93       clang -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -I/usr/local/include/python3.9 -c scanner.c -o build/temp.linux-aarch64-3.9/scanner.o
#12 25.93       clang -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -I/usr/local/include/python3.9 -c writer.c -o build/temp.linux-aarch64-3.9/writer.o
#12 25.93       gcc -pthread -shared -Wl,--strip-all build/temp.linux-aarch64-3.9/_ruamel_yaml.o build/temp.linux-aarch64-3.9/api.o build/temp.linux-aarch64-3.9/dumper.o build/temp.linux-aarch64-3.9/emitter.o build/temp.linux-aarch64-3.9/loader.o build/temp.linux-aarch64-3.9/parser.o build/temp.linux-aarch64-3.9/reader.o build/temp.linux-aarch64-3.9/scanner.o build/temp.linux-aarch64-3.9/writer.o -L/usr/local/lib -o build/lib.linux-aarch64-3.9/_ruamel_yaml.cpython-39-aarch64-linux-gnu.so
#12 25.93       error: command 'gcc' failed: No such file or directory
#12 25.93       [end of output]
```

vigsterkr: mmm i wonder what defaults back to use gcc as a linker... 🤦 i've checked `ruamel-yaml`'s source and there's no hardcoding of `gcc` there, so this is really strange

rokroskar: apparently it's an open issue: https://github.com/pypa/setuptools/issues/2065

`gcc`it is for now then

vigsterkr: 🤦 python...

rokroskar: I suggest we leave it as-is for now - if you feel strongly about clang we can make an issue and watch for upstream changes that will allow us to do it

vigsterkr: as said above, i've approved this PR these were just suggestions that are good to have IMO, but not mandatory

rokroskar:
Sorry @vigsterkr I realized that the build dependencies were actually not being limited to the builder container so I decided to change it...

Dockerfile.cli
```diff
@@ -1,11 +1,14 @@
 FROM python:3.9-slim as base
 
 RUN apt-get update && \
-    apt-get install -y git make gcc python3-dev && \
+    apt-get install -y git git-lfs python3-dev && \
```
Panaetius, line 4: We should pin this to a version of git-lfs that works. 3.1.x definitely doesn't, 2.12.1 does (it's what our envs are pinned to)

vigsterkr (dismissed):
:shipit:

Dockerfile.cli
rokroskar: good point, thanks for the reminder

rokroskar: so the installed version is 

```
git lfs --version
git-lfs/2.13.2 
```

I guess it's ok if I pin to this?

Panaetius: Yes I think <3.x is ok

Panaetius (approved):

PR #2707, opened 2022-02-22
refactor(core): refactor/remove DatasetsApiMixin

### LINKED ISSUE #2227: Refactor datasets
There are some potential improvements after migration to the new persistent layer and metadata that would result in a cleaner code for datasets:
* ~Have data files and metadata in a single commit instead of committing data files first in a separate commit~
* Move `DatasetsApiMixin` to its own class/package and remove all `LocalClient` dependencies
* Separate interface and management logic from `commands/dataset.py` and `DatasetsApiMixin` to have a clear distinction between the two
* ~Use dependency injection everywhere~
* ~Remove redundant tests; convert integration tests to unit tests where ever possible~

Panaetius:
closes #2227 

Most functionality moved to renku/core/management/dataset/dataset.py

Dataset add is split off into renku/core/management/dataset/dataset_add.py, as it by itself made up ~40% of all dataset logic and having all that separated in one place makes sense to me.

`client.with_dataset` is not implemented as a `DatasetContext` contextmanager.

There's no more business logic in renku/core/commands/dataset.py

Providers are now in renku/core/management/dataset/providers/ instead of renku/core/commands/


renku/service/views/decorators.py
```diff
@@ -309,7 +309,7 @@ def decorated_function(*args, **kwargs):
             except (Exception, BaseException):
                 pass
             capture_exception(e)
-
+            breakpoint()
```
mohammad-sdsc, line 312: Debug leftover.

renku/core/management/dataset/dataset.py
```diff
@@ -0,0 +1,997 @@
+# -*- coding: utf-8 -*-
```
mohammad-sdsc, line 1: Some functions in this file can be private including `set_dataset_images`, most `update_*`, `filter_dataset_files`.

mohammad-sdsc (commented):
Thanks Ralf! This looks great. I've made some minor comments.

mohammad-sdsc (approved):
🚀

PR #2709, opened 2022-02-22
fix(core): downgrade git-lfs to fix CI

### LINKED ISSUE #2708: git-lfs path matching broken
Git lfs 3.1.x changed some things around in how they match paths, namely https://github.com/git-lfs/git-lfs/pull/4758 which breaks our `renku storage migrate`.

namely, `/.renku` only matches the folder `.renku` now, but not files inside it. Changing to `/.renku/**` should fix it.

Panaetius:
Git LFS 3.1.x breaks our CI due to various issues, this is just a small PR to downgrade LFS to a version that works.

Ubuntu `focal` has git-lfs version 2.9.x, but the `--above` flag that we use was introduced in 2.13.x, so we need to install it from the `jammy` package repo, which has 3.0.2.

### DIFF
--- a/.github/workflows/test_deploy.yml
+++ b/.github/workflows/test_deploy.yml
@@ -351,6 +351,9 @@ jobs:
         run: |
           sudo apt-get update -y
           sudo apt-get install -y libyaml-0-2 libyaml-dev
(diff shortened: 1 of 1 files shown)


vigsterkr (approved):
:shipit:

PR #2718, opened 2022-02-25
fix(service): prometheus annotations not properly disabled through chart

wesjdj:
# Description

Prometheus metrics annotations are applied to the core pod even if the Helm Values.metrics.enabled value has been set to false.

Fixes # (issue)

## Type of change

Set the Prometheus annotations to only apply to the core pod if the Helm metrics value has been set to true.

/deploy #persist extra-values=core.metrics.enabled=false


wesjdj:
I have tested that the opposite of the initial test (metrics enabled value set to true) does apply the Prometheus annotations to the core pods, as expected.

PR #2719, opened 2022-02-25
feat(core): color edges on a per-node basis

### LINKED ISSUE #2377: color edges in `renku workflow visualize` per node, not per edge
Currently, `renku workflow visualize` just assigns a color to each edge, making sure overlapping edges don't have the same color.

We should change this logic so edges belonging to an Activity all get the same color, still making sure we don't have overlapping edges of different Activities with the same color.

### LINKED ISSUE #2574: Show workflow names in renku workflow visualize
When running `renku workflow visualize`, the step boxes contain workflow commands by default.
In some cases, these commands are long and truncated, it can be better to see workflow names instead.
When using the `--interactive` option, we can see the full command`, but the workflow name is not available in the detail view either.

Making the workflow names visible in the details view would be very useful.

Optionally, it could be good to add a `--names` flag to `renku workflow visualize` to show names instead of commands on the steps boxes.

Panaetius:
closes #2377 
closes #2574 

Edges belonging to an activity are now colored the same color, with differing colors for overlapping edges, if possible.
Added Plan Name in details view of interactive mode, added `plan` column for `-c` argument, showing plan name in the activity boxes.

image

Can be tested on https://renkulab.io/gitlab/ralf.grubenmann/test_visualize_2719 run `renku workflow visualize out` or `renku workflow visualize -c plan out`

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -273,7 +273,7 @@ pycparser = "*"
 
 [[package]]
 name = "charset-normalizer"
-version = "2.0.11"
+version = "2.0.12"
 description = "The Real First Universal Charset Detector. Open, modern and actively maintained alternative to Chardet."
 category = "main"
 optional = false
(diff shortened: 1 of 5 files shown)


vigsterkr (approved):
🚀

vigsterkr (approved):

rokroskar:
😻

PR #2722, opened 2022-02-28
feat(core): use oauth 2.0 device auth grant

### LINKED ISSUE #2636: Use `Device Authorization Grant` flow for CLI login
The device authorization grant flow has become available in keycloak 13.0 and should make it possible for the CLI client to acquire a keycloak access token without any involvement from the gateway-auth component. Once a the CLI has acquired a keycloak access token through the device authorization grant flow, this token can be used to authenticate requests to the renku api in the same way as it is already the case. This would make the logic equivalent to the UI client and allow us to disentangle functionalities and clean-up the gateway.
See also https://www.keycloak.org/docs/latest/release_notes/index.html#keycloak-13-0-0

mohammad-sdsc:
# Description

Replace current custom CLI login flow implemented in the gateway with Device Auth Grant implemented in Keycloak.

Fixes #2636

### DIFF
--- a/renku/command/login.py
+++ b/renku/command/login.py
@@ -18,9 +18,8 @@
 """Logging in to a Renku deployment."""
 
 import os
-import sys
+import time
 import urllib
-import uuid
 import webbrowser
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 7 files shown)


mohammad-sdsc:
Note: This PR shouldn't be merged/released before https://github.com/SwissDataScienceCenter/renku/pull/2491 is deployed, or otherwise `renku login` won't work on the current deployments.

Panaetius (approved):
thanks!

PR #2726, opened 2022-02-28
feat(core): add check for invalid imported datasets

### LINKED ISSUE #2725: Cannot export graph for projects with invalid imported datasets
**Describe the bug**
Imported datasets can have both `schema:sameAs` and `prov:wasDerivedFrom` set due to https://github.com/SwissDataScienceCenter/renku-python/issues/2685. This causes `renku graph export` or `renku log -d` to fail because `prov:wasDerivedFrom` in those datasets don't point to a valid dataset in the project.

**To Reproduce**
same as #2685 

**Additional context**
We should fix the export/log logic to deal with these cases and add a `renku doctor` action to fix the metadata. A migration is not required.

mohammad-sdsc:
# Description

Adds a `--fix` option to the `renku doctor` command and implements a check/fix for datasets that have both `derived_from` and `same_as` set or are derived from themselves. Checks for invalid `derived_from` to avoid crashes in unfixed projects.

Fixes #2725

### DIFF
--- a/renku/core/commands/doctor.py
+++ b/renku/core/commands/doctor.py
@@ -16,6 +16,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Check your system and repository for potential problems."""
+
 import traceback
 
 from renku.core.commands.echo import ERROR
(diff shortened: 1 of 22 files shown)


renku/core/commands/doctor.py
```diff
@@ -18,6 +18,7 @@
 """Check your system and repository for potential problems."""
 import traceback
 
+from renku.cli.utils.callback import ClickCallback
```
Panaetius, line 21: communicator should be set in `renku.cli` (before calling .build()), `renku.core` shouldn't have a dependency on `renku.cli`

Panaetius (changes requested):
Thank you for the quick fix!

Panaetius (dismissed):
Thank you!

Panaetius (approved):
:rocket:

PR #2727, opened 2022-03-01
fix(core): check workflow execute inputs

### LINKED ISSUE #2639: rerun fails when the script has been moved
Taking one of the most simplest example, say we add a script `foo.py` that creates a file called `output.txt`

1. `renku run --name foo python foo.py`
2. `git mv foo.py bar.py`
3. `renku rerun output.txt`

step 3 will fail as we don't have anymore a file called `foo.py` in the project. we'll get a longer exception in case of cwl (toil is less verbose) and somewhere within that exception it's gonna be mentioned that the provider cannot find one of the inputs for the workflow.

i wonder whether:
 a) we at least should do a sanity check before executing the workflow whether the required inputs are present and in case something is missing mentioned this in a one-liner exception to the user
 b) we should actually detect from git changes a simple `git mv <src> <dst>` and automagically 'fix' the workflow

vigsterkr:
# Description

Fixes #2639

### DIFF
--- a/renku/core/commands/workflow.py
+++ b/renku/core/commands/workflow.py
@@ -462,10 +462,16 @@ def execute_workflow(
     """Execute a Run with/without subprocesses."""
     client = client_dispatcher.current_client
 
+    inputs = [i.actual_value for p in dag.nodes for i in p.inputs]
     # NOTE: Pull inputs from Git LFS or other storage backends
     if client.check_external_storage():
-        inputs = [i.actual_value for p in dag.nodes for i in p.inputs]
         client.pull_paths_from_storage(*inputs)
+
+    # check whether the none generated inputs of workflows are available
+    outputs = set([o.actual_value for p in dag.nodes for o in p.outputs])
+    for i in set(inputs) - outputs:
+        if not Path(i).exists():
+            raise errors.ParameterError(f"Input '{i}' for the workflow does not exists!")
 
     delete_indirect_files_list(client.path)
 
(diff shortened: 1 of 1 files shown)


renku/core/commands/workflow.py
```diff
@@ -462,11 +462,17 @@ def execute_workflow(
     """Execute a Run with/without subprocesses."""
     client = client_dispatcher.current_client
 
+    inputs = [i.actual_value for p in dag.nodes for i in p.inputs]
```
Panaetius, line 465:
```suggestion
    inputs = {i.actual_value for p in dag.nodes for i in p.inputs}
```

renku/core/commands/workflow.py
```diff
@@ -462,11 +462,17 @@ def execute_workflow(
...
     # NOTE: Pull inputs from Git LFS or other storage backends
     if client.check_external_storage():
-        inputs = [i.actual_value for p in dag.nodes for i in p.inputs]
         client.pull_paths_from_storage(*inputs)
 
+    # check whether the none generated inputs of workflows are available
+    outputs = set([o.actual_value for p in dag.nodes for o in p.outputs])
+    for i in set(inputs) - outputs:
```
Panaetius, line 472:
```suggestion
    outputs = {o.actual_value for p in dag.nodes for o in p.outputs}
    for i in inputs - outputs:
```

Panaetius (dismissed):
Thank you! I don't have a comment on anything functional, just some unneeded lists that could be sets directly.

Panaetius (approved):

PR #2728, opened 2022-03-01
feat(api): allow explicit input/output values to be set for workflow execution

### LINKED ISSUE #2597: Add support for renku.api.Input/Output to pick up modified values
In https://github.com/SwissDataScienceCenter/renku-python/pull/2596 the functionality to set `renku.api.Parameter` values on `renku workflow execute` with `--set` was added.

We should do the same for `renku.api.Input` and `renku.api.Output`. This entails forcing the user to set a name on these objects in addition to the path, and then looking them up from the env variables on run.

Panaetius:
closes #2597

### DIFF
--- a/renku/api/models/run.py
+++ b/renku/api/models/run.py
@@ -48,6 +48,7 @@ Once a Parameter is tracked like this, it can be set normally in commands like
 import re
 from os import PathLike, environ
 from pathlib import Path
+from typing import Union
 
 from renku.api.models.project import ensure_project_context
 from renku.core import errors
@@ -58,17 +59,30 @@ from renku.core.management.workflow.plan_factory import (
 )
 from renku.core.plugins.provider import RENKU_ENV_PREFIX
 
+name_pattern = re.compile("[a-zA-Z0-9-_]+")
+
 
 class _PathBase(PathLike):
     @ensure_project_context
-    def __init__(self, path, project):
-        self._path = Path(path)
+    def __init__(self, name: str, path: Union[str, Path], project):
+        if name and not name_pattern.match(name):
+            raise errors.ParameterError(
+                f"Name {name} contains illegal characters. Only characters, numbers, _ and - are allowed."
+            )
+        self.name = name
 
-        indirect_list_path = self._get_indirect_list_path(project.path)
-        indirect_list_path.parent.mkdir(exist_ok=True, parents=True)
-        with open(indirect_list_path, "a") as file_:
-            file_.write(str(path))
-            file_.write("\n")
+        env_value = environ.get(f"{RENKU_ENV_PREFIX}{name}", None)
+
+        if env_value:
+            self._path = Path(env_value)
+        else:
+            self._path = Path(path)
+
+            indirect_list_path = self._get_indirect_list_path(project.path)
+            indirect_list_path.parent.mkdir(exist_ok=True, parents=True)
+            with open(indirect_list_path, "a") as file_:
+                file_.write(f"{path}:::{name}")
+                file_.write("\n")
 
     @staticmethod
     def _get_indirect_list_path(project_path):
@@ -103,7 +117,7 @@ class Output(_PathBase):
 @ensure_project_context
 def parameter(name, value, project):
     """Store parameter's name and value."""
-    if not re.match("[a-zA-Z0-9-_]+", name):
+    if not name_pattern.match(name):
         raise errors.ParameterError(
             f"Name {name} contains illegal characters. Only characters, numbers, _ and - are allowed."
         )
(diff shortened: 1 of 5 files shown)


renku/api/models/run.py
```diff
@@ -58,17 +59,30 @@
...
 
+name_pattern = re.compile("[a-zA-Z0-9-_]+")
+
 
 class _PathBase(PathLike):
     @ensure_project_context
-    def __init__(self, path, project):
-        self._path = Path(path)
+    def __init__(self, name: str, path: Union[str, Path], project):
+        if name and not name_pattern.match(name):
```
mohammad-sdsc, line 68: If a name is not provided (or is empty) then thing can go wrong (e.g. multiple inputs will get the same value). So we should also check for that case. Perhaps it's better to put this check in a function since it's also used in another place.

renku/api/models/run.py
```diff
@@ -58,17 +59,30 @@
...
 from renku.core.plugins.provider import RENKU_ENV_PREFIX
 
+name_pattern = re.compile("[a-zA-Z0-9-_]+")
+
 
 class _PathBase(PathLike):
     @ensure_project_context
-    def __init__(self, path, project):
-        self._path = Path(path)
+    def __init__(self, name: str, path: Union[str, Path], project):
```
mohammad-sdsc, line 67: It should be possible to automatically generate names for Input/Output. I believe it has a better UX.

renku/api/models/run.py
```diff
@@ -58,17 +59,30 @@
...
 
-        indirect_list_path = self._get_indirect_list_path(project.path)
-        indirect_list_path.parent.mkdir(exist_ok=True, parents=True)
-        with open(indirect_list_path, "a") as file_:
-            file_.write(str(path))
-            file_.write("\n")
+            indirect_list_path = self._get_indirect_list_path(project.path)
+            indirect_list_path.parent.mkdir(exist_ok=True, parents=True)
+            with open(indirect_list_path, "a") as file_:
+                file_.write(f"{path}:::{name}")
```
mohammad-sdsc, line 84: I'd rather go with a common format (e.g. JSON) instead of this. The original use case for the indirect in/out was that users should write some paths to these files and then Renku reads and adds them to in/outs. So, users need to know about this format and it would be easier if the format is familiar.

renku/api/models/run.py
```diff
@@ -58,17 +59,30 @@
...
-    def __init__(self, path, project):
-        self._path = Path(path)
+    def __init__(self, name: str, path: Union[str, Path], project):
+        if name and not name_pattern.match(name):
+            raise errors.ParameterError(
+                f"Name {name} contains illegal characters. Only characters, numbers, _ and - are allowed."
+            )
+        self.name = name
+
+        env_value = environ.get(f"{RENKU_ENV_PREFIX}{name}", None)
```
mohammad-sdsc, line 74: What happens if we have multiple runs, executing in parallel (e.g. a `workflow iterate`)? I'd expect that the execution environments to be isolated, but do we know that for sure for all of the providers?

renku/api/models/run.py
mohammad-sdsc: We need to update our discourse post about the Renku Python API if we decide to force users to provide a name.

tests/cli/test_workflow.py
```diff
@@ -625,6 +625,144 @@ def test_workflow_execute_command_with_api_parameter_set(runner, run_shell, proj
     assert "goodbye\n" == output.read_text()
 
 
+@pytest.mark.parametrize("provider", available_workflow_providers())
+def test_workflow_execute_command_with_api_input_set(runner, run_shell, project, capsys, client, provider):
+    """Test executing a workflow with --set for a renku.api.Parameter."""
```
mohammad-sdsc, line 630: The docstrings for all these functions are the same.

mohammad-sdsc (changes requested):
Thanks Ralf! I've made a few comments/suggestions.

renku/cli/run.py
```diff
@@ -194,17 +194,22 @@
...
-files is expected to be the path to an input or output file within the
-project's directory. When the program is finished, Renku will look for
-existence of these two files and adds their content to the list of explicit
-inputs and outputs. Renku will then delete these two files.
+To address this issue, the program can dump a mapping of input and output files
+that it is accessing in ``inputs.yml`` and ``outputs.yml``. This YAML file
+should be of the format
+.. code-block:: YAML
+
+   - name: path
```
mohammad-sdsc, line 202: Since the file should be a YAML dict we should remove the `-`. I also would add another entry to it:
```
name1: path1
name2: path2
```

mohammad-sdsc (approved):

mohammad-sdsc (approved):
LGTM, thanks!

PR #2731, opened 2022-03-02
chore: consolidate dockerfiles

### LINKED ISSUE #2705: avoid code duplication between service and CLI dockerfiles
Currently `Dockerfile.cli` and `Dockerfile.svc` are almost identical - we don't need to maintain both and we should figure out a way to use a single one.

rokroskar:
Consolidates CLI and service Dockerfiles. Introduces a new build argument `BUILD_CORE_SERVICE` which, if set to a non-null value, causes the core service dependencies to be installed. 

/deploy

closes #2705


Panaetius (dismissed):

Panaetius (approved):
Thanks, looks good to me. Though maybe hold off on merging this until we made the new release, there's some more important PRs open :slightly_smiling_face:

rokroskar:
@Panaetius may this be merged now?

PR #2737, opened 2022-03-03
feat(core): add /version endpoint

### LINKED ISSUE #2593: Add /version endpoint
We are adding a public `GET /version` endpoint in every renku service.

The response format should be the following one:
```
{
    "name": string,
    "versions": [
        {
            "version": string,
            "data": object
        }
    ]
}
```

* "name" will be `renku-core`
* "version" should be a SemVer valid value, possibly the tag or some combination of last tag + commit short-sha if it's a dev version
* "data" is optional, and it can be any arbitrary valid JSON object. Since renku-core already has a `/version` endpoint providing additional data, this may be the place to move them. An alternative would be keeping these as separate endpoints, renaming the other one to avoid conflicts

mohammad-sdsc:
# Description

Renames `/versions` to `/version` and return renku CLI version information there. Renames `/version` to `/apiversion` in the core service.

Fixes #2593

### DIFF
--- a/helm-chart/renku-core/templates/configmap.yaml
+++ b/helm-chart/renku-core/templates/configmap.yaml
@@ -19,10 +19,10 @@ data:
       tcp_nopush on;
       client_max_body_size 0; # Required for uploading large files
 
-      location /renku/versions {
+      location /renku/version {
         root /;
         add_header Content-Type application/json;
-        try_files /usr/share/nginx/html/versions.json =404;
+        try_files /usr/share/nginx/html/version.json =404;
       }
 
       {{- range $version := .Values.versions }}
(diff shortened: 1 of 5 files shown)


Panaetius (approved):
Thank you!

mohammad-sdsc:
Merge this PR once https://github.com/SwissDataScienceCenter/renku-ui/issues/1714 is done to avoid breaking UI.

mohammad-sdsc:
No. It's because this endpoint is not part of the core service and therefore there is no Swagger doc for it. It is implemented in the reverse proxy that routes requests to different versions of the core service. The proxy serves a static JSON file for this endpoint which includes version information for different core service deployments. See `helm-chart/renku-core/templates/configmap.yaml` in the PR's changed file.

Panaetius (approved):

PR #2738, opened 2022-03-03
chore: update CI chart publish action

mohammad-sdsc:
# Description

This should allow chart publish action on `0.16-tls` branch to run successfully by:
- Using correct secret name in the chart publish action
- Removing unnecessary `renku` component update action


Panaetius (approved):
:rocket:

PR #2740, opened 2022-03-04
chore: fix poetry.lock check in CI

Panaetius:
Before the action checked the lock file but did not actually fail when it was out of date


rokroskar (dismissed):

Panaetius:
For some reason, `poetry lock` on my machine and in CI gives different results (triggering failure of the test), I'm not sure yet why.

E.g. https://github.com/SwissDataScienceCenter/renku-python/runs/5427498782?check_suite_focus=true#step:9:25

rokroskar:
did you sort out why there was a discrepancy?

mohammad-sdsc (approved):
🚀

PR #2741, opened 2022-03-08
fix(core): fix lfs migrate patterns

### LINKED ISSUE #2708: git-lfs path matching broken

### LINKED ISSUE #2710: git-lfs breaking change, --above does not work with --include/--exclude anymore
git-lfs introduced a change in https://github.com/git-lfs/git-lfs/pull/4771 v3.1.x that means that `--above` does not work together with `--include` and `--exclude` flags. We use both for lfs_threshold and .renkulfsignore. So we need to figure out a way to get this working again.

mohammad-sdsc:
# Description

Recent Git LFS versions don't support `--above` filter in `lfs migrate import` command. This PR fixes that issue. It also adds `/**` to all `.renkulfsignore` patterns to support matching subdirectories.

Fixes #2708 
Fixes #2710


renku/core/management/storage.py
```diff
@@ -432,27 +431,34 @@ def get_lfs_migrate_filters(self):
 
             if p.include:  # File ignored by LFS
                 excludes.append(pattern)
+                if pattern not in self.RENKU_PROTECTED_PATHS:
+                    excludes.append(f"{pattern.strip()}/**")
```
Panaetius, line 435: Nice solution, I didn't think of just adding the pattern with/without `/**`.

I guess if a user does `myfolder/*`, explicitly intending to only ignore top-level files in the directory, this would break that? But then I think we can live with that.

```suggestion
                    excludes.append(f"{pattern.strip().rstrip("/")}/**")
```
maybe? I'm not sure what happens if my patterns is `data/` and it turns to `data//**`?

mohammad-sdsc: > I guess if a user does myfolder/*, explicitly intending to only ignore top-level files in the directory, this would break that? But then I think we can live with that.
Good point! I've added a check for it.

Panaetius (approved):

PR #2746, opened 2022-03-09
feat(dataset): --dry-run for update

### LINKED ISSUE #2520: renku dataset update semantics
`renku dataset update` called without additional arguments will currently attempt to update _all_ datasets - it would be nice if it followed the same semantics as `renku update`, i.e. it should return the help text if called without any arguments and require the `--all` flag to update all datasets. ~~It would also be nice to have a "dry run" option to see which changes exist.~~

### LINKED ISSUE #2564: Add a --dry-run option to `renku dataset update`
As the title says, add a `--dry-run` flag to `renku dataset update` that shows what updates would be performed without actually performing them.

mohammad-sdsc:
# Description

WIP

Fixes #2564 
Fixes #2520


renku/cli/dataset.py
```diff
@@ -940,19 +943,66 @@ def import_(uri, name, extract, yes):
...
+            columns = {"path": ("path", None), "dataset": ("dataset.name", None), "external": ("external", None)}
+            return tabulate(collection=records, columns="path,dataset,external", columns_mapping=columns)
+
+        if not datasets and not dataset_files:
+            click.secho("Everything is up-to-date", fg=color.GREEN)
+            return
+
+        if datasets:
+            names = "\n\t".join(sorted([d.name for d in datasets]))
+            click.echo(f"The following imported datasets will be updated:\n\t{names}\n")
```
Panaetius, line 994:
```suggestion
            click.echo(f"The following imported datasets' metadata would be updated:\n\t{names}\n")
```

renku/cli/dataset.py
```diff
@@ -940,19 +943,66 @@ def import_(uri, name, extract, yes):
...
+
+        if not dataset_files:
+            return
+
+        files = [f for f in dataset_files if not f.deleted]
+        if files:
+            files = get_dataset_files(files)
+            click.echo(f"The following files will be updated:\n\n{files}\n")
```
Panaetius, line 1002:
```suggestion
            click.echo(f"The following files would be updated:\n\n{files}\n")
```

renku/cli/dataset.py
```diff
@@ -940,19 +943,66 @@ def import_(uri, name, extract, yes):
...
+
+        deleted_files = [f for f in dataset_files if f.deleted]
+        if deleted_files:
+            files = get_dataset_files(deleted_files)
+            message = " (pass '--delete' to remove them from datasets' metadata)" if not delete else ""
+            click.echo(f"The following files will be deleted{message}:\n\n{files}\n")
```
Panaetius, line 1008:
```suggestion
            click.echo(f"The following files would be deleted{message}:\n\n{files}\n")
```

Panaetius (changes requested):
We have quite a few places like 
```
        raise errors.ParameterError(
            "--include/--exclude is incompatible with datasets created by 'renku dataset import'"
        )
```
Where we refer to CLI flags in renku.core errors, but the errors themselves are the same type. So when we expose these things in the service, we'd have to parse the error message to create a suitable service error, which isn't that nice.

We might want to opt for a sort of middle ground approach, where we try to do as much cli-specific error handling in renku.cli and make errors in renku.core  independent of the cli, but only in places where it's easy to do. 

And then when we add a service endpoint for something, we can deal with the rest that's not so easy to port (like the case mentioned above that needs to check datasets names), in which case they should have a specific error instead of a generic one.

Panaetius (approved):

PR #2751, opened 2022-03-11
chore: change documentation over to google style format.

Panaetius:
Port over all existing sphinx-style documentation and adds docstrings for relevant namespaces such as `renku.api` and `renku.core.commands`.

Also did some cleanup, removing unused arguments and ascii.py which wasn't used.


mohammad-sdsc (commented):
Thanks Ralf! I've just made a few minor comments.

mohammad-sdsc (approved):
🚀

PR #2757, opened 2022-03-16
feat(dataset): warn if adding tracked files as external

### LINKED ISSUE #2075: Warn users when adding files from local repo to datasets as external files
Renku silently ignores `--external` flag when adding local files to a dataset. We should warn users about this situation.

**Acceptance Criteria**
- A warning with the list of local files that can't be added as external is printed. Those files are added as normal files.

mohammad-sdsc:
# Description

- Allows adding untracked files as external files to a dataset. We create a relative symlink to these files instead of absolute symlinks which are used for external files. This makes the project to work when it's moved to another location.
- Warns when adding files that are tracked as external

Fixes #2075


renku/core/management/dataset/dataset_add.py
```diff
@@ -191,14 +201,19 @@ def _process_urls(
...
                     )
+                    tracked_external_warnings.extend(warnings)
                 else:  # Remote URL
                     new_files = _add_from_url(client=client, url=url, destination=destination, extract=extract)
 
             files.extend(new_files)
 
+    if tracked_external_warnings:
+        message = "\n  ".join(tracked_external_warnings)
+        communication.warn(f"Warning: The following files cannot be added as external:\n  {message}")
```
Panaetius, line 215:
```suggestion
        message = "\n\t".join(tracked_external_warnings)
        communication.warn(f"Warning: The following files cannot be added as external:\n\t{message}")
```
I think in most places we use the Tab character for horizontal alignment, should probably also use it here for consistency.

Panaetius (changes requested):
Looks all good, thank you. Only a really minor formatting issue.

Panaetius (dismissed):

Panaetius (approved):

PR #2763, opened 2022-03-18
feat(core): consistent template versioning

### LINKED ISSUE #2714: Consistent template versioning
We use renku version for templates that are bundled with renku CLI while for templates in a git repository we use git tags as the version. This means that bundled templates (that also come from a git repository) will have two different versions depending on how we use them.
To make the versions consistent, we should use the git tag version for the bundled template as well.

mohammad-sdsc:
# Description

Use template repository's tag as version (instead of the Renku version) for templates bundled with Renku CLI.

Fixes #2714

### DIFF
--- a/renku/core/models/template.py
+++ b/renku/core/models/template.py
@@ -60,19 +60,26 @@ class TemplatesSource:
 
         return self.manifest.templates
 
-    @abstractmethod
     def is_update_available(self, id: str, reference: Optional[str], version: Optional[str]) -> Tuple[bool, str]:
-        """Return True if an update is available along with the latest version of a template."""
-        raise NotImplementedError
+        """Return True if an update is available along with the latest reference of a template."""
+        latest = self.get_latest_reference_and_version(id=id, reference=reference, version=version)
+        if not latest:
+            return False, reference
+
+        latest_reference, latest_version = latest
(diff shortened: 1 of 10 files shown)


renku/core/models/template.py
```diff
@@ -60,19 +60,26 @@ def templates(self) -> List["Template"]:
...
-        """Return True if an update is available along with the latest version of a template."""
-        raise NotImplementedError
+        """Return True if an update is available along with the latest reference of a template."""
+        latest = self.get_latest_reference_and_version(id=id, reference=reference, version=version)
+        if not latest:
+            return False, reference
+
+        latest_reference, latest_version = latest
+        update_available = latest_reference != reference or latest_version != version
+        return update_available, latest_reference
```
Panaetius, line 71: This would mean that if I install hypothetical version 1.3.0 and migrate, and then install hypothetical version 1.2.0 and run migration check, it'd say a new version is available and downgrade, right?

mohammad-sdsc: No, `EmbeddedTemplates::get_latest_reference_and_version` checks if project's template version is newer than renku's template version and returns the latest of the two, so, `1.3.0` will be the latest version/reference in that case and nothing will be changed.

Panaetius (approved):

PR #2764, opened 2022-03-18
chore: release 1.1.2

### LINKED ISSUE #2762: fix ParameterMapping in SHACL
In code and our ontology we have `renku:ParameterMapping` but in our SHACL we check for `renku:MappingParameter` (the two words swapped). We should fix the SHACL and make a hotfix for it that the KG/TG can use.

mohammad-sdsc:
Fixes #2762

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,17 @@
 Changes
 =======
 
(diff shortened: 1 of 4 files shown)


Panaetius (approved):
Thank you!

PR #2766, opened 2022-03-21
docs: add developer overview

### LINKED ISSUE #2495: add/improve documentation to get people started developing renku-python
We should split off the "Get started" section in Contributing.rst into its own file.
We should check that the instructions are correct and up to date and can be used to set up a working dev environment.
We should also document some common workflows like running tests and stylecheck locally, using black and isort for reformatting etc.

In addition we should have an architecture guide document that details the architecture (like namespaces and what goes where) and common patterns (CommandBuilder, Communication, injection, Gateways, the Database etc.)

Panaetius:
closes #2495 

Also adds many more pre-commit hooks to essentially do all the checks that `./run-tests.sh -s` does.


DEVELOPING.rst
```diff
@@ -0,0 +1,237 @@
+Developing
+==========
+
+.. _developing-reference:
+
+Architecture Overview
+---------------------
+
+The high-level architecture of `renku` follows a Hexogonal/Ports&Adapters
```
mohammad-sdsc, line 9: I'm not sure what single backquotes do in RST (it does nothing in GH; it makes the text italic in PyCharm). You need to replace \` with \`\` if you want to show them as code blocks: \`renku\` -> \`\`renku\`\`. This happens through the whole document (except when it is preceded by \:code\: which shows up as code block).

DEVELOPING.rst
```diff
@@ -0,0 +1,237 @@
...
+fields.
+
+Gateways
+--------
+
+To make code more readable, abstract away the database and keep things clean,
+we have various Gateway classes. These expose database (and other
+infrastructure) functionality in an easily understandable way.
+
+The consist of two parts, an interface definition in `renku.core.interface`
```
mohammad-sdsc, line 135: `The` -> `A gateway`!?

DEVELOPING.rst
```diff
@@ -0,0 +1,237 @@
...
+---------------
+
+We follow the git_flow_ model of branches for development, with `master` being
+the release branch and `develop` being the development branch.
+
+Release branches should be created off of master, have develop merged into them
+and then should be merged (not squash merged) back into master. A Github Action
+will then take care of merging master back into develop.
+
+.. _git_flow: (https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)
```
mohammad-sdsc, line 237: git_flow -> git-flow

mohammad-sdsc (commented):
Thanks Ralf! This looks great! I've made some comments regarding typos and formatting.

DEVELOPING.rst
Panaetius: It makes it italic or shows it as a code block. Actually it's not defined and implementation specific. 
From the spec:
> The rendering and meaning of interpreted text is domain- or application-dependent. It can be used for things like index entries or explicit descriptive markup (like program identifiers). 

For instance in our docstrings in code: 
\`\`renku\`\` -> autodoc would try and find a reference to `renku` and throw a warning if such a reference doesn't exist.
\`renku\` -> just treated as a literal, not used for cross referencing, not checked by the spell checker
renku -> normal text, might fail a spellcheck

So doing the whole docstrings thing a while back taught me that single backtick is safe. But of course that's only for code docstrings, and you're completely right that it doesn't make sense here.

DEVELOPING.rst
```diff
@@ -0,0 +1,238 @@
...
+We use the inject_ library for dependency injection. DI is set up by the
+Command Builder for each command call.
+
+When an inner layer in the architecture needs to interact with objects from an
+outer layer, since dependencies aren't allowed to go outwards, there needs to
+be an interface in the inner layer that is the basis for injection and an
+implementation in the outer layer that gets injected.
+
+
+.. _inject: (https://pypi.org/project/Inject/)
```
mohammad-sdsc, line 84: I've just noticed that this doesn't show up in GH or PyCharm. You need to remove parentheses for it to work correctly:
```
.. _inject: https://pypi.org/project/Inject/
```
It also happens in some other places and in `CONTRIBUTING.rst` as well.

mohammad-sdsc (commented):
Sorry, I've just noticed another issue with formatting.

DEVELOPING.rst
Panaetius: I'd copied that entry from CONTRIBUTING.rst and assumed it's fine :rofl: Went unnoticed for 2 years, good catch!

mohammad-sdsc (approved):

PR #2768, opened 2022-03-21
fix(core): fix migration of deleted datasets

### LINKED ISSUE #2755: Renku dataset import by keyword
Reusing datasets in multiple renku projects by importing them using `renku dataset import` is a very convenient way of interacting between projects. In our pipeline we import all datasets with a certain keyword (after some checks about the generator) into a specific project. 

As `renku dataset import` expects the dataset URI, we first need to identify the ID and project of all datasets with a specific keyword. 

For this we query the KG endpoints to show all datasets with our keyword, similar to:

`data_ids = curl -s "${KG_URL}/datasets?query=$KEY" | jq -r '.[] | .'\""identifier"\"''` 
(afterwards we check if the keywords of the resulting ids match with our keyword)

If a dataset has already been imported into another project these queries will result several dataset identifier for the same _"original"_ dataset. When we now import this dataset again, we want to import from the _"root/original"_ project, which it was generated in, because we rely on regular updates. 
If the datasets point towards the copy of a dataset within a third project, it might not get updated, when the _original_ dataset was changed. 

As it is hard to tell from the KG query which dataset was the original one and which is a copy of that dataset, we have a hard time to ensure to pick the original URI for import. 

This is why something like an new option or similar for `renku dataset import`  to import from the **"root/original" dataset** instead of the **dataset copy**, when using `renku dataset import` with a **dataset URI**, that points to an imported dataset.

Althought the `sameAs` field is helpful to identify parent/root datasets from KG query results, this will not work, if the parent has been updated, but not the child, yet.

Thanks a lot!
Almut

Panaetius:
Fixes migration of deleted datasets (they were previously not considered during migration).

fixes #2755

### DIFF
--- a/renku/core/management/migrations/m_0009__new_metadata_storage.py
+++ b/renku/core/management/migrations/m_0009__new_metadata_storage.py
@@ -622,6 +622,7 @@ def _process_datasets(
     changed_paths = [c.b_path for c in changes if not c.deleted]
     paths = [p for p in changed_paths if len(Path(p).parents) == 4]  # Exclude files that are not in the right place
     deleted_paths = [c.a_path for c in changes if c.deleted]
+    deleted_paths = [p for p in deleted_paths if len(Path(p).parents) == 4]
 
     datasets, deleted_datasets = _fetch_datasets(
         client=client, revision=commit.hexsha, paths=paths, deleted_paths=deleted_paths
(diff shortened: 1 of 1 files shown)


mohammad-sdsc (commented):
Thanks Ralf! I also would filter the `deleted_path` after this line https://github.com/SwissDataScienceCenter/renku-python/blob/4f2d6de3259ca5b35413b818335bdd57597c343d/renku/core/management/migrations/m_0009__new_metadata_storage.py#L624 by doing 

```deleted_paths = [p for p in deleted_paths if len(Path(p).parents) == 4]```

mohammad-sdsc (approved):
Thank you!

PR #2770, opened 2022-03-21
chore: backport Dockerfile consolidation

rokroskar:
Backport changes from #2731 to 0.16.x. 

/deploy


Makefile
```diff
@@ -37,7 +41,35 @@ docker-login:
...
+build-lock-package: download-templates
+	@echo "Building package"
+	poetry-dynamic-versioning
+	poetry run poetry-lock-package --build
+	sed -i --regexp-extended 's/^version\s=\s"[^"]+"/version = "0.0.0"/gm' pyproject.toml
+	sed -i --regexp-extended 's/^__version__\s=\s"[^"]+"/__version__ = "0.0.0"/gm' renku/version.py
+
+publish-lock-package: build-lock-package
+	@echo "Publishing package"
+	twine upload -u __token__ -p ${PYPI_TOKEN} dist/renku_lock*
```
Panaetius, line 75: 0.16 doesn't use poetry yet (it uses setuptools), so this shouldn't be ported

Dockerfile
```diff
@@ -22,18 +25,20 @@ COPY renku /code/renku/renku
 ARG CLEAN_INSTALL
 RUN if [ -n "${CLEAN_INSTALL}" ]; then git reset --hard ; fi
 
-RUN pip wheel --wheel-dir /wheels .[service] && \
-    pip install --no-index --no-warn-script-location --force --root=/pythonroot/ /wheels/*.whl && \
-    apt-get clean
+RUN make download-templates
```
Panaetius, line 28: 0.16 doesn't use poetry yet (it uses setuptools), so this shouldn't be ported

Panaetius (changes requested):
Thanks for looking into this! Just 0.16 doesn't use poetry :slightly_smiling_face:

Makefile
rokroskar: Ah yes... Will remove

rokroskar:
I triggered the acceptance test just to get this version of things up and running for a manual check.

rokroskar:
Sorry this took a while. In the end I deployed the acceptance tests with the current tip of `develop` in `renku-python` and `master` in the UI. This deploys a working version of the platform, see e.g. [this project](https://renku-ci-rp-2770.dev.renku.ch/projects/rokroskar/project-renku-0-16-2/overview/status). The acceptance tests that fail are because the CLI version can't be installed... which I guess is expected? Not sure.

Panaetius:
I thought acceptance tests can check out based on the git hash in the version. But it's been a while.

rokroskar:
> I thought acceptance tests can check out based on the git hash in the version. But it's been a while.

Sorry I'm not sure I follow - which git hash and which version? The problem here is that I don't want to deploy the chart from this branch - I need to just deploy this image as the v8 image but the chart should be from the `develop` or `master`.

Panaetius:
I remember if they encounter a version like `1.1.2.dev89+g5ea571fc` they see "oh +g" and then install from git with ref 5ea571fc instead of going through pypi.

rokroskar:
hmm I see what you mean. Was that implemented in the acceptance tests or just in the templates?

Panaetius:
https://github.com/SwissDataScienceCenter/renku/blob/master/acceptance-tests/src/test/scala/ch/renku/acceptancetests/workflows/CLIConfiguration.scala#L51
https://github.com/SwissDataScienceCenter/renku/blob/72f5fe4505b4d66329c172e5dff9b3d8f5726180/acceptance-tests/src/test/scala/ch/renku/acceptancetests/model/commons.scala#L60

Panaetius (approved):

Panaetius (approved):

.github/workflows/acceptance-tests.yml
```diff
@@ -54,17 +65,18 @@ jobs:
         RENKU_VALUES: ${{ secrets.CI_RENKU_VALUES }}
         RENKUBOT_KUBECONFIG: ${{ secrets.RENKUBOT_DEV_KUBECONFIG }}
         RENKUBOT_RANCHER_BEARER_TOKEN: ${{ secrets.RENKUBOT_RANCHER_BEARER_TOKEN }}
+        RANCHER_DEV_API_ENDPOINT: ${{ secrets.RANCHER_DEV_API_ENDPOINT }}
         RENKU_BOT_DEV_PASSWORD: ${{ secrets.RENKU_BOT_DEV_PASSWORD }}
         RENKU_ANONYMOUS_SESSIONS: true
         RENKU_TESTS_ENABLED: true
         TEST_ARTIFACTS_PATH: "tests-artifacts-${{ github.sha }}"
-        renku_core: "@${{ github.head_ref }}"
+        renku_core: "@develop"
```
lorenzo-cavazzi, line 73: Is this something temporary to test the CI deployment?

lorenzo-cavazzi (commented):
This seems to work well in the CI deployment :tada: 
Would love to see it merged for the next release. I had to redeploy renku a few times lately and without that fix the renku-core v8 fails without further manual interventions.
I noticed there is a reference to `@develop`, I was wondering if it was just for testing (see inline)

.github/workflows/acceptance-tests.yml
rokroskar: no this is on purpose - it deploys the current `develop` branch and plugs this branch in as a custom image (see a few lines below)

rokroskar: this branch cannot be deployed alone with the current deployment config and acceptance tests

rokroskar:
There is another acceptance test currently failing (modifying datasets) which I believe is because the currently released UI is incompatible with the renku-python `develop` branch... It passes if I set `ui` to `master`

Panaetius (approved):

PR #2773, opened 2022-03-22
fix(core): change graph export command to not print to stdout directly

### LINKED ISSUE #2759: Change `export_graph` to return JSON-LD data directly
Currently, `export_graph()` command prints directly to stdout and it takes a `format` parameter which is mostly relevant for the UI parts of renku-python. 

It should instead return a view model wrapping the data that exposes method for formatting, e.g. `as_jsonld()`, `as_nt()`, `as_rdf()` which can then be used in the calling code (cli or service). Or probably even cleaner, move all formatting code out of `renku.core` to something that is only shared by .cli and .service.

This also means that currently, the `/graph.export` endpoint on the service always returns `None`, which would be fixed by this change.

Also consider returning unindented, single-line data in the service and maybe a flag like `--pretty` to pretty-print/control indentation in the CLI.

Panaetius:
closes #2759


mohammad-sdsc (commented):
Thanks Ralf! LGTM, just a single typo.

mohammad-sdsc (approved):

PR #2774, opened 2022-03-22
feat(workflow): show created plan after renku run

### LINKED ISSUE #2732: Show generated Plan in `renku run`
as per the title, `renku run` currently outputs nothing after the run completes. We should instead pretty print the generated `Plan` so a user can double-check what was generated. This will also make it more consistent with the rest of the `renku workflow` commands that already print their result.

mohammad-sdsc:
# Description

Adds a `--verbose` flag to the `renku run` command that prints the generated plan after execution.

Fixes #2732

### DIFF
--- a/tests/cli/test_output_option.py
+++ b/tests/cli/test_output_option.py
@@ -21,7 +21,7 @@ import os
 from pathlib import Path
 
 from renku.cli import cli
-from tests.utils import write_and_commit_file
+from tests.utils import format_result_exception, write_and_commit_file
 
 
 def test_run_succeeds_normally(renku_cli, client, subdirectory):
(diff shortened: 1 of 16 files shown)


tests/cli/test_output_option.py
```diff
@@ -232,24 +233,29 @@ def test_explicit_inputs_can_be_in_inputs(renku_cli, client, subdirectory):
...
+    assert 0 == runner.invoke(cli, ["run", "--no-output", "mkdir", "foo"]).exit_code
+    assert 0 == runner.invoke(cli, ["run", "echo", "some changes"], stdout="foo/bar").exit_code
+    assert 0 == runner.invoke(cli, ["run", "echo", "cat foo/bar"], stdout="script.sh").exit_code
 
-    exit_code, _ = renku_cli("run", "--input", "foo/bar", "--input", "script.sh", "sh", "script.sh", stdout="output")
-    assert 0 == exit_code
+    result = runner.invoke(
+        cli, ["run", "--input", "foo/bar", "--input", "script.sh", "sh", "script.sh"], stdout="output"
+    )
+    assert 0 == result.exit_code
```
Panaetius, line 246: We should follow the `assert 0 == result.exit_code, format_result_exception(result)` pattern here

tests/cli/test_output_option.py
```diff
@@ -232,24 +233,29 @@ def test_explicit_inputs_can_be_in_inputs(renku_cli, client, subdirectory):
...
-    exit_code, _ = renku_cli("update", "--all")
-    assert 0 == exit_code
+    assert 0 == result.exit_code, result.output
+
+    result = runner.invoke(cli, ["status"])
+
+    assert 1 == result.exit_code
+
+    result = runner.invoke(cli, ["update", "--all"])
+    assert 0 == result.exit_code
```
Panaetius, line 258: Here also

tests/cli/test_run.py
```diff
@@ -207,3 +207,31 @@ def test_run_non_existing_command(runner, client):
 
     assert 2 == result.exit_code
     assert "Cannot execute command 'non-existing_command'" in result.output
+
+
+def test_run_prints_plan(split_runner, client):
+    """Test run shows the generated plan with --verbose."""
+    result = split_runner.invoke(cli, ["run", "--verbose", "--name", "echo-command", "--no-output", "echo", "data"])
+
+    assert 0 == result.exit_code
```
Panaetius, line 216: Also in this file.

Panaetius (changes requested):
Looks real good, I like the solution with the custom runner. Just a minor detail on checking exit code to make it easier to figure out problems in the future.

Panaetius (approved):

PR #2782, opened 2022-03-25
fix(core): add docker package as a hard requirement

vigsterkr:
# Description

add `docker` python package as a hard requirement.


mohammad-sdsc (approved):
🚀

Panaetius (approved):

PR #2785, opened 2022-03-28
chore: release 1.1.4

### LINKED ISSUE #2783: add cycle detection in `renku run`
We already have cycle detection in e.g. https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/core/management/workflow/activity.py#L208

But we should really do this during a `renku run`, after the run has been executed and the Plan/Activity generated, but before committing the database and finishing. 

If we detect a cycle, we should raise an error informing the user as such (also print out the equivalent of `[getattr(a, "id", a) for a in list(networkx.algorithms.cycles.simple_cycles(graph))[0]]` showing the cycle), to prevent cycles from being accidentally added to our metadata.

Panaetius:
closes #2783

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,17 @@
 Changes
 =======
 
+`1.1.4 <https://github.com/SwissDataScienceCenter/renku-python/compare/v1.1.3...v1.1.4>`__ (2022-03-28)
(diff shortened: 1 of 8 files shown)


mohammad-sdsc (approved):
LGTM, thank you!

PR #2786, opened 2022-03-28
chore: fix pydocstyle excludes

Panaetius:
adds missing ignores/excludes to new pre-commit yaml for pydocstyle


mohammad-sdsc (approved):
Thank you!

PR #2795, opened 2022-03-31
feat(workflow): docker container support for toil provider

### LINKED ISSUE #2712: expose docker container support by toil
Toil natively supports [docker containers](https://toil.readthedocs.io/en/latest/developingWorkflows/developing.html#using-docker-containers-in-toil). This should be exposed in our current toil provider plugin

mohammad-sdsc:
# Description

Runs workflows using Docker with the Toil provider when proper setting is present in the execution's config file.

Fixes #2712


vigsterkr (approved):
🚀 🚢

PR #2798, opened 2022-03-31
fix(core): correct CommandOutputs for outputs in subdirectories

### LINKED ISSUE #2777: CommandOutputs converted from CommandInputs don't set the `create_folder` flag correctly.
In https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/core/management/workflow/plan_factory.py#L454 we create a CommandOutput directly, instead of using `add_command_output()` like in https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/core/management/workflow/plan_factory.py#L471 . 
We don't set `create_folder` flag in this case (something `add_command_output()` would do, in which case on rerun/update/execute CWL does not copy the parent folder of the output and scripts will fail due to a missing folder to write to.

I think this is the direct cause of the issue seen in https://renku.discourse.group/t/renku-update-and-module-not-found/579/8 .

There probably isn't a way to fix Plans already created that have this issue after the fact.

mohammad-sdsc:
Fixes #2777

### DIFF
--- a/renku/core/management/workflow/plan_factory.py
+++ b/renku/core/management/workflow/plan_factory.py
@@ -410,6 +410,8 @@ class PlanFactory:
         postfix: Optional[str] = None,
         encoding_format: List[str] = None,
         name: Optional[str] = None,
+        id: str = None,
+        mapped_to: Optional[MappedIOStream] = None,
     ):
         """Create a CommandOutput."""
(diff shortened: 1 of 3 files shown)


Panaetius (approved):
Thank you!

Issue #2614, opened 2022-01-31
Change namespaces to better match architecture


Panaetius:
To better match the architecture, 
![image](https://user-images.githubusercontent.com/664486/151828639-0c144a3d-b8c9-4ba8-97af-8c77696140c2.png)

we should change the folder structure of the project like 
```
└── renku/
    ├── ui/
    │   ├── cli
    │   ├── api
    │   └── rpc  # previously service
    ├── infrastructure #gateway, database, git, plugin implementations
    ├── command # previously renku.core.commands
    ├── core   #for business logic/
    │   ├── dataset/
    │   │   └── view_model
    │   ├── workflow #more folder for different types of business logic/
    │   │   └── view_model
    │   ├── interface # interfaces for DB, gateways etc.
    │   ├── plugin # pluggy pluginmanager/
    │   │   └── hooks # plugin hook specs
    │   └── util
    ├── domain_model # previously renku.core.models
    └── data # same as before
```

PR #2800, opened 2022-04-04
chore: move cli, api and service to renku.ui.* namespace

Panaetius:
related to #2614 

Also updates all the version headers to the current year.


mohammad-sdsc (approved):
LGTM, thanks!

PR #2805, opened 2022-04-05
fix: fix backwards compatibility for new namespaces

### LINKED ISSUE #2804: fix backwards compatibility after move to new namespaces
In https://github.com/SwissDataScienceCenter/renku-python/issues/2614 we moved submodules around to match our architecture, leaving behind dummy placeholder files that map to old imports so backwards compatibility is kept.

In cases where a module had a `__all__` set, this broke backwards compatibility, as not everything in the new namespace was exposed under the old namespace.

Panaetius:
closes #2804


mohammad-sdsc (dismissed):
LGTM, thank you!

Panaetius:
@mohammad-sdsc I think we should disable the DeprecationWarning for `renku.core.models` and `renku.core.metadata`? Otherwise everyone will see warnings in renku commands on <1.1.x projects

mohammad-sdsc (approved):

PR #2808, opened 2022-04-05
chore(core): more SHACL validation

### LINKED ISSUE #2760: Add more validation in our SHACL
There have been a couple of cases where our SHACL file didn't catch errors that it could catch. We should extend it to do more strict checks.

Namely:
- Use regex rules to check id formats everywhere (already done in quite a few places), also for things that follow a strict format like e.g. SoftwareAgent name
- Where settings are mutually exclusive, add checks for that (same_as vs derived_from, date_created vs date_published etc.)
- Minimum/maximum number of child nodes
- Other cases?

mohammad-sdsc:
# Description

Adds more validation to our SHACL.
Removes outdated shapes.

Fixes #2760

### DIFF
--- a/renku/data/shacl_shape.json
+++ b/renku/data/shacl_shape.json
@@ -151,7 +151,7 @@
                      "sh:pattern": "http(s)?://[^/]+/persons/\\S+"
                   },
                   {
-                     "sh:pattern": "mailto:[^@\\s]+\\S+\\.\\S+"
+                     "sh:pattern": "mailto:[^@\\s]+@\\S+\\.\\S+"
                   }
                ]
             },
@@ -369,7 +369,7 @@
                      "sh:pattern": "http(s)?://[^/]+/persons/\\S+"
                   },
                   {
-                     "sh:pattern": "mailto:[^@]+\\S+\\.\\S+"
+                     "sh:pattern": "mailto:[^@\\s]+@\\S+\\.\\S+"
                   }
                ]
             },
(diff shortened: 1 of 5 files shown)


renku/data/shacl_shape.json
```diff
@@ -137,10 +141,19 @@
...
                },
-               "minCount": 1
+               "minCount": 1,
+               "maxCount": 1,
+               "or": [
+                  {
+                     "sh:pattern": "http(s)?://[^/]+/persons/\\S+"
+                  },
+                  {
+                     "sh:pattern": "mailto:[^@\\s]+\\S+\\.\\S+"
```
Panaetius, line 154:
```suggestion
                     "sh:pattern": "mailto:[^@\\s]+@\\S+\\.\\S+"
```
I think there is an @ missing in the regex. otherwise `ab.c` would be valid according to this, no?

renku/data/shacl_shape.json
```diff
@@ -339,7 +363,15 @@
...
                   "@id": "schema:Person"
                },
-               "minCount": 1
+               "minCount": 1,
+               "or": [
+                  {
+                     "sh:pattern": "http(s)?://[^/]+/persons/\\S+"
+                  },
+                  {
+                     "sh:pattern": "mailto:[^@]+\\S+\\.\\S+"
```
Panaetius, line 372:
```suggestion
                     "sh:pattern": "mailto:[^@\\s]+@\\S+\\.\\S+"
```
to be consistent?

Panaetius (changes requested):
Those are some great additions 👍  I think this'll make our code much more robust!

Panaetius (approved):

PR #2811, opened 2022-04-06
fix(core): fix SHACL shape for MappingParameter and add SHACL checks to more tests

### LINKED ISSUE #2761: add `renku graph export --full --strict` as part of more tests
There have been a few cases where SHACL validation was never run on relevant metadata on our side, and the SHACL is inconsistent with metadata that we produce.

Instead of creating a project with all possible combinations, we should instead add `renku graph export --full --strict` checks in existing tests.

Namely:
- datasets, both created, modified and added, with image and other metadata set
- Project created from template
- Workflows in all the various forms (Inputs/Outputs/Parameters/ParameterMappings/ParameterLinks/ParameterValue/Plan/CompositePlan)
- Custom metadata on Project/Dataset/Activity

There should be existing tests that set all these things in their temporary repositories, so simply adding a `assert 0 == runner.run(cli, ["graph", "export", "--full", "--strict"]).exit_code` style check to them should be enough.

Panaetius:
closes #2761

### DIFF
--- a/renku/data/shacl_shape.json
+++ b/renku/data/shacl_shape.json
@@ -1468,10 +1468,24 @@
                "maxCount": 1
             },
             {
-               "path": "schema:mapsTo",
-               "sh:class": {
-                  "@id": "renku:CommandParameterBase"
-               }
+               "path": "renku:mapsTo",
+               "or": [
+                  {
+                     "sh:class": {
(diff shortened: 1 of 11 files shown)


mohammad-sdsc (approved):
Thank you!

PR #2814, opened 2022-04-07
fix(core): tests for renku session subcommands

vigsterkr:
# Description

adds tests for `renku session` sub-commands

### DIFF
--- a/tests/core/plugins/test_session.py
+++ b/tests/core/plugins/test_session.py
@@ -1,0 +1,118 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2017-2022 - Swiss Data Science Center (SDSC)
+# A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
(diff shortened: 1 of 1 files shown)


tests/core/plugins/test_session.py
```diff
@@ -0,0 +1,118 @@
...
+    "parameters,result",
+    [
+        ({}, "0xdeadbeef"),
+        ({"image_name": "fixed_image"}, "0xdeadbeef"),
+        ({"image_name": "missing_image"}, ParameterError),
+    ],
+)
+@patch.object(DockerSessionProvider, "session_start", fake_start)
+@patch.object(DockerSessionProvider, "find_image", fake_find_image)
+@patch.object(DockerSessionProvider, "build_image", fake_build_image)
```
Panaetius, line 75: Could this be done in a single https://docs.python.org/3/library/unittest.mock.html#patch-multiple call ?

Panaetius (approved):

PR #2819, opened 2022-04-08
feat: handle migration errors from the template

### LINKED ISSUE #2769: change cache.check_migration and cache.migrate to gracefully handle if a template repository doesn't exist anymore
As part of check_migration/migrate we try and clone the template repository that was used to create a project, and if that repository or reference doesn't exist anymore, we raise an exception.

We should instead return that a template update is not available. And in the case of existing repo but missing reference, still try to allow updates to newer version (like if 2.3.4 tag was originally used but is now deleted, but 2.3.5 exists).

I don't think we need to inform the UI that this is the case, as we should be able to handle everything internally.

Original issue:
This happened because the branch of the template repo from which the project was initialized had been removed. We should handle such cases gracefully.

Sentry Issue: [RENKU-PYTHON-4C](https://sentry.dev.renku.ch/organizations/sentry/issues/465/?referrer=github_integration)

```
GitCommandError: Cmd('git') failed due to: exit code(1)
  cmdline: git checkout prep-0.3.1
  stderr: 'error: pathspec 'prep-0.3.1' did not match any file(s) known to git'
  File "renku/core/metadata/repository.py", line 1397, in _run_git_command
    return getattr(repository.git, command)(*args, **kwargs)
  File "git/cmd.py", line 638, in <lambda>
    return lambda *args, **kwargs: self._call_process(name, *args, **kwargs)
  File "git/cmd.py", line 1183, in _call_process
    return self.execute(call, **exec_kwargs)
  File "git/cmd.py", line 983, in execute
    raise GitCommandError(redacted_command, status, stderr_value, stdout_value)

GitCommandError: Git command failed: Cmd('git') failed due to: exit code(1)
  cmdline: git checkout prep-0.3.1
  stderr: 'error: pathspec 'prep-0.3.1' did not match any file(s) known to git'
  File "renku/core/utils/git.py", line 571, in clone_repository
    repository.checkout(checkout_revision)
  File "renku/core/metadata/repository.py", line 186, in checkout
    self.run_git_command("checkout", reference)
  File "renku/core/metadata/repository.py", line 263, in run_git_command
    return _run_git_command(self._repository, command, *args, **kwargs)
  File "renku/core/metadata/repository.py", line 1405, in _run_git_command
    ) from e

GitReferenceNotFoundError: Cannot checkout reference 'prep-0.3.1' in repository: https://github.com/swissdatasciencecenter/renku-project-template
  File "renku/core/management/template/template.py", line 427, in fetch
    repository = clone_repository(url=source, path=path, checkout_revision=reference, install_lfs=False)
  File "renku/core/utils/git.py", line 574, in clone_repository
    f"Cannot checkout reference '{checkout_revision}' in repository: {url}"

InvalidTemplateError: Cannot clone template repository from https://github.com/swissdatasciencecenter/renku-project-template
(17 additional frame(s) were not displayed)
...
  File "renku/core/commands/migrate.py", line 53, in _migrations_check
    "template_status": _template_migration_check(client),
  File "renku/core/commands/migrate.py", line 87, in _template_migration_check
    update_available, update_allowed, current_version, new_version = check_for_template_update(client)
  File "renku/core/management/template/usecase.py", line 72, in check_for_template_update
    templates_source = fetch_templates_source(source=metadata.source, reference=metadata.reference)
  File "renku/core/management/template/template.py", line 84, in fetch_templates_source
    else RepositoryTemplates.fetch(source, reference)
  File "renku/core/management/template/template.py", line 429, in fetch
    raise errors.InvalidTemplateError(f"Cannot clone template repository from {source}") from e
```

lorenzo-cavazzi:
/deploy renku=auto-update/renku-core-1.2.1 #persist

# Description

This PR improves the UX when a project's template cannot be checked out correctly. It adds new errors and some basic indications to the user in both the CLI and the service. The 2 new cases handled properly are:
- The template repository cannot be fetched (example: https://dev.renku.ch/gitlab/lorenzo.cavazzi.tech/template-is-gone)
- The repository can be fetched, but the original reference is gone (example: https://dev.renku.ch/gitlab/lorenzo.cavazzi.tech/template-missing-reference)

Screenshot_20220412_174611

fix #2769


renku/core/template/template.py
```diff
@@ -420,6 +420,10 @@ def fetch(cls, source: Optional[str], reference: Optional[str]) -> "RepositoryTe
         try:
             repository = clone_repository(url=source, path=path, checkout_revision=reference, install_lfs=False)
         except errors.GitError as e:
+            if "Cannot checkout reference" in str(e):
+                raise errors.TemplateMissingReferenceError(
+                    f"Cannot find the reference {reference} in the template repository from {source}"
```
mohammad-sdsc, line 425: `reference {reference} -> reference '{reference}'` so that it's more visible what the reference is.

renku/core/template/usecase.py
```diff
@@ -132,7 +132,15 @@ def update_template(
...
-    templates_source = fetch_templates_source(source=template_metadata.source, reference=template_metadata.reference)
+    try:
+        templates_source = fetch_templates_source(
+            source=template_metadata.source, reference=template_metadata.reference
+        )
+    except errors.TemplateMissingReferenceError as e:
+        message = f"{str(e)}; You can still manually update the template and set a difference reference."
+        raise errors.TemplateUpdateError(message)
+    except errors.InvalidTemplateError:
+        raise errors.TemplateUpdateError("Template cannot be fetched.")
```
mohammad-sdsc, line 143: What's the reason to convert these errors to `TemplateUpdateError`? I believe we don't need to catch them here.

renku/ui/service/views/error_handlers.py
```diff
@@ -375,8 +378,26 @@ def decorated_function(*args, **kwargs):
...
 
-def handle_migration_errors(f):
-    """Wrapper which handles migrations exceptions."""
+def handle_migration_read_errors(f):
+    """Wrapper which handles migrations read exceptions."""
+    # noqa
+    @wraps(f)
+    def decorated_function(*args, **kwargs):
+        """Represents decorated function."""
+        # NOTE: verify if this may better go in MigrationsCheckCtrl as try/except in to_response()
```
mohammad-sdsc, line 387: I don't get this comment. Is it a left-over TODO?

renku/ui/service/views/error_handlers.py
```diff
@@ -375,8 +378,26 @@ def decorated_function(*args, **kwargs):
...
+        except TemplateMissingReferenceError as e:
+            raise UserProjectTemplateReferenceError(e)
+        except (InvalidTemplateError, TemplateUpdateError) as e:
+            raise IntermittentProjectTemplateUnavailable(e)
+
+    return decorated_function
+
+
+@handle_migration_read_errors
+def handle_migration_write_errors(f):
```
mohammad-sdsc, line 399: Why do we need to distinguish between migration read/write errors. AFAIK, the difference is that in read we only check for migration; so, the same errors can happen in both cases.

mohammad-sdsc (commented):
Thanks Lorenzo! Please see my comments.

renku/core/template/usecase.py
lorenzo-cavazzi: The idea was to simplify catching/handling them in the service and in the CLI. For the latter, throwing a new error results in a clean error message, while not doing it means a long and confusing stack trace is printed (see screen below).
Is there a better way to handle this? Or maybe that should be done in the cli code?
Screenshot_20220413_134606

renku/ui/service/views/error_handlers.py
lorenzo-cavazzi: Ouch! Yes, that is a leftover, thanks for spotting it!

renku/ui/service/views/error_handlers.py
lorenzo-cavazzi: The write section handles a couple of extra errors, and those throw a more severe error (`Program` instead of `Intermittent`).
The point is that errors while reading are somewhat expected (e.g. template may be deleted, that's not a bug), while actually migrating the project should be invoked by the client only if the read errors didn't happen.
So a `DockerfileUpdateError` or a `MigrationError` caught on writing is likely to be a bug or severe error from which the user may not be able to recover in the UI.

mohammad-sdsc (approved):
Thank you!

PR #2820, opened 2022-04-08
chore(refactor): make entity mandatory on DatasetFile

### LINKED ISSUE #2792: make `entity` mandatory on `DatasetFile`
Currently, the `entity` property of `DatasetFile` is `Optional[Entity]` and the only case where it is `None` is in the Zenodo and Dataverse imports, where it later gets set when the files are downloaded.

But as it is optional, we would have to check everywhere in code that it is set before accessing its properties, make the code rather ugly.

We should refactor the Zenodo and Dataverse imports to use something else than `DatasetFile` (or use a dummy entity maybe?) and change the type of `entity` to just `Entity` and remove any checks for its existence.

There's also some places where this check is foregone and `# type: ignore` is used instead, these should be cleaned up as well.

mohammad-sdsc:
# Description

Makes `entity` non-Optional in `DatasetFile`: Uses `ProviderRecordSerializerApi::files_info` to pass file information to the import logic and doesn't set `files` on `Dataset` instances.

Refactoring:
* Adds a `ProviderRecordSerializerApi` interface to be inherited in the import providers.
* Removes `attr` in some places
* Providers return a `ProviderDataset` that is a sub-class of `Dataset` but doesn't allow accessing its `files` property.

Fixes #2792

### DIFF
--- a/renku/core/management/repository.py
+++ b/renku/core/management/repository.py
@@ -58,7 +58,7 @@ def _path_converter(path) -> Path:
 class PathMixin:
     """Define a default path attribute."""
 
-    path: Path = attr.ib(default=default_path, converter=_path_converter)
+    path: Path = attr.ib(default=default_path, converter=_path_converter)  # type: ignore
 
     @path.validator
     def _check_path(self, _, value):
--- a/renku/core/management/repository.py
+++ b/renku/core/management/repository.py
@@ -58,7 +58,7 @@ def _path_converter(path) -> Path:
 class PathMixin:
     """Define a default path attribute."""
 
-    path: Path = attr.ib(default=default_path, converter=_path_converter)  # type: ignore
+    path: Path = attr.ib(default=default_path, converter=_path_converter)
 
     @path.validator
     def _check_path(self, _, value):
--- a/renku/data/shacl_shape.json
+++ b/renku/data/shacl_shape.json
@@ -408,7 +408,7 @@
                   "@id": "xsd:string"
                },
                "maxCount": 1,
-               "sh:pattern": "\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\+\\d{2}:\\d{2}"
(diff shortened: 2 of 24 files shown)


renku/core/management/repository.py
```diff
@@ -58,7 +58,7 @@ def _path_converter(path) -> Path:
 class PathMixin:
     """Define a default path attribute."""
 
-    path: Path = attr.ib(default=default_path, converter=_path_converter)
+    path: Path = attr.ib(default=default_path, converter=_path_converter)  # type: ignore
```
Panaetius, line 61: I fought with this before (we really should get rid of attr, this PR is a great start).

no typing on `_path_converter(path):` caused mypy to complain, doing `def _path_converter(path:Union[str,Path]) -> Path:` caused it to complain as well, but just specifying the output type, but not the input git mypy to somehow infer the type correctly. But i think it just doesn't work well with attrs. I have `attrs==21.4.0` locally, I'm curious which version you have. That might play a role as well, Maybe we need to pin attrs to a higher version?

In any case, I'm fine with just ignoring, just wanted to give some background info on this.

renku/data/shacl_shape.json
```diff
@@ -408,7 +408,7 @@
                   "@id": "xsd:string"
                },
                "maxCount": 1,
-               "sh:pattern": "\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\+\\d{2}:\\d{2}"
+               "sh:pattern": "\\d{4}-\\d{2}-\\d{2}(T\\d{2}:\\d{2}:\\d{2}\\+\\d{2}:\\d{2})?"
```
Panaetius, line 411: getting yyyy-MM-dd with the current SHACL (without this fix) would cause triples generation to fail. Does this actually happen in the wild? If so, we'd probably need a bugfix release for this, no?

Panaetius (dismissed):
This is great, much cleaner than before. Thank you!

I had two minor comments but not sure if they require changes.

Panaetius (approved):

PR #2823, opened 2022-04-11
fix(core): fix Plan.invalidated_at datetime not being timezone aware

### LINKED ISSUE #2822: Plan.invalidated_at does not contain timezone info
Plan.invalidated_at is just set to UTC time, but without a timezone set on the datetime object. This causes the time to be exported in the wrong format when exporting to JSONLD

Panaetius:
closes #2822


mohammad-sdsc (approved):
LGTM, thank you!

PR #2824, opened 2022-04-11
chore: prevent priviledge escalation in containers

### LINKED ISSUE #2806: Run as non-root

Panaetius:
closes #2806 

/deploy


olevski (changes requested):
@Panaetius this is nice. But I did want to see what happens when you do `helm create test` and get a fresh new helm chart starter template created by helm. When helm does this it specifies a security context for the pod and containers in the values file.

So we should follow the same.

Also it is nice to set the `fsGroup` so that volumes are mounted with specific permissions.

I wasn't sure if this would work, but it does. I tested it here.
https://github.com/SwissDataScienceCenter/renku-python/tree/run-as-non-root-test

Feel free to merge this branch or just copy and paste. I even tested with the metrics enabled and things work.

Or here is a diff that may be more useful: https://github.com/SwissDataScienceCenter/renku-python/compare/feature/2806-non-root-containers...run-as-non-root-test

Panaetius:
> @Panaetius this is nice. But I did want to see what happens when you do `helm create test` and get a fresh new helm chart starter template created by helm. When helm does this it specifies a security context for the pod and containers in the values file.
> 
> So we should follow the same.
@olevski Can we set user/group 1000 in the securityContext? Because the rq-metrics image has a different user/group

olevski (approved):
🚢

PR #2830, opened 2022-04-11
fix(cli): inform user if a dataset wasn't found in dataset show

### LINKED ISSUE #2793: cli: renku dataset show my-dat
Using `renku dataset show <name>` with `<name>` that doesn't exist throws an exception instead of handling it gracefully. 

To reproduce:

```
renku init test
cd test
renku dataset show bla
```

Version:

```
renku --version
1.1.4
python --version                                                                                                                                                                                                                        
Python 3.9.9
```

Panaetius:
closes #2793


renku/core/dataset/dataset.py
```diff
@@ -735,6 +735,9 @@ def show_dataset(name):
         dict: JSON dictionary of dataset details.
     """
     dataset = DatasetsProvenance().get_by_name(name)
```
mohammad-sdsc, line 737: `DatasetProvenance::get_by_name` has a `strict` parameter which when set to `True` raises an `errors.DatasetNotFound` when dataset is not found. It's more consistent to use that parameter instead of repeating the logic here.

mohammad-sdsc (commented):
Please see my comment.

mohammad-sdsc (approved):
Thank you!

mohammad-sdsc (approved):

PR #2833, opened 2022-04-12
chore: update rqmetrics to 0.0.2

olevski:
I did some housekeeping on the rqmetrics side. Namely:
- created a release
- switched to poetry
- updated outdated requirements that had security vulnerabilities

I also tested this locally and it works. We should always use a version for this if the image pull policy is `IfNotPresent`


olevski:
My bad I will re-open this. I realize I started from master and not develop.

PR #2834, opened 2022-04-12
chore: update rqmetrics to 0.0.2

olevski:
I did some housekeeping on the rqmetrics side. Namely:

- created a release
- switched to poetry
- updated outdated requirements that had security vulnerabilities

I also tested this locally and it works. We should always use a version for this if the image pull policy is IfNotPresent

### DIFF
--- a/helm-chart/renku-core/values.yaml
+++ b/helm-chart/renku-core/values.yaml
@@ -72,7 +72,7 @@ metrics:
   enabled: false
   image:
     repository: renku/rqmetrics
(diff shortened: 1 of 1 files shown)


olevski:
p.s. The changes in the actual business logic of rqmetrics were pretty much zero. So I just merged things or committed straight to master in the rqmetrics repo. I can undo this and submit a proper PR if you would like in rqmetrics first if you would like.

Panaetius (approved):

PR #2840, opened 2022-04-14
feat(core): add a minimum version check to support breaking forward compatibility

### LINKED ISSUE #2810: add minimum_renku_version on project metadata and raise a warning
Sometimes we have changes that break forward compatibility of older versions without changing the metadata version. In these cases, a user doesn't need to run a migration but they do have to update their renku version to be compatible.

We should add a `minimum_renku_version` on project metadata that we can set in code (and update if a write command is run with a newer renku version). Probably a CommandBuilder to set this makes sense.

Then we should check this when running any renku command and warn the user if their renku version is lower than `project.minimum_renku_version`

Panaetius:
closes #2810 

can be tested with https://dev.renku.ch/gitlab/ralf.grubenmann/test_minimum_version (has a minimum version requirement if 999.9.9)


tests/core/commands/test_renku.py
```diff
@@ -17,11 +17,82 @@
...
+        project = Project(creator=Person(name="John Doe", email="jd@example.com"), name="testproject")
+        mocker.patch(
+            "renku.command.command_builder.database_dispatcher.Database.__getitem__", _mock_database_project(project)
+        )
+        mocker.patch("renku.version.__version__", "1.0.0")
+
+        try:
+            Command().command(lambda: "").with_database(write=False).build().execute()
+        except errors.MinimumVersionError:
+            assert False, "Shouldn't raise an exception"
```
mohammad-sdsc, line 58: Maybe use `pytest.fail` to be more explicit. There are more similar cases in this function.

renku/domain_model/project.py
```diff
@@ -36,6 +36,10 @@ class Project(persistent.Persistent):
 
     keywords: List[str] = list()
 
+    # NOTE: the minimum version of renku to needed to work with a project
+    # This should be bumped on metadata version changes and when we do non forward-compatible on-the-fly migrations
```
mohammad-sdsc, line 40: non -> not

renku/domain_model/project.py
```diff
@@ -36,6 +36,10 @@ class Project(persistent.Persistent):
...
+    minimum_renku_version: str = "1.2.0"
```
mohammad-sdsc, line 41: I suggest we put this in `renku.version` so that all version-related constants are in the same place.

mohammad-sdsc (commented):

mohammad-sdsc (dismissed):
🚀

PR #2845, opened 2022-04-14
feat(service): update template schema and errors

### LINKED ISSUE #2729: Handle KeyError better in the service error handler functions
This is a follow-up to #2303 (especially to [this conversation](https://github.com/SwissDataScienceCenter/renku-python/pull/2303/files/4ded6eec1a006a8323ca172f3221d03b691bf091#r813184365))

There are a number of different reasons for a `KeyError` to happen. We should handle this case better in the decorators in `renku/service/views/error_handlers.py` (or improve where the errors are originally raised, where possible).

In general, trying to catch all the errors we can identify (especially user-generated) and raise a more descriptive error is a good strategy. Everything un-catched could raise a  `ProgramSomethingError` explaining in the devMessage how the parameters were wrong. If something ends up as a Program error but it shouldn't, we can always handle the case as soon as we identify it.

lorenzo-cavazzi:
# Description

Update `ProjectTemplateRequest` to remove unnecessary derived fields and fix template errors and its handling function.
fix #2729

### DIFF
--- a/renku/ui/service/views/error_handlers.py
+++ b/renku/ui/service/views/error_handlers.py
@@ -251,25 +251,27 @@ def handle_templates_create_errors(f):
     def decorated_function(*args, **kwargs):
         """Represents decorated function."""
 
-        def match_schema(target, message):
-            if re.match(f".*{target}.*contains.*unsupported characters.*", message):
-                return True
-            return False
+        def get_schema_error_message(e):
+            if getattr(e, "messages", None) and e.messages.get("_schema"):
+                message = (
+                    "; ".join(e.messages.get("_schema"))
+                    if isinstance(e.messages.get("_schema"), list)
+                    else str(e.messages.get("_schema"))
+                )
+                return message
+            return None
 
         try:
             return f(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


renku/ui/service/views/error_handlers.py
```diff
@@ -251,25 +251,27 @@ def handle_templates_create_errors(f):
     def decorated_function(*args, **kwargs):
         """Represents decorated function."""
 
-        def match_schema(target, message):
-            if re.match(f".*{target}.*contains.*unsupported characters.*", message):
-                return True
-            return False
+        def get_schema_error_message(e):
+            if getattr(e, "messages", None) and e.messages.get("_schema"):
```
Panaetius, line 255:
```suggestion
            if isinstance(getattr(e, "messages", None), dict) and e.messages.get("_schema"):
```
to not risk "AttributeError: e.message has not attribute `get`"

Panaetius (changes requested):
Thanks, looks great! I've added a minor improvement suggestion

Panaetius (approved):

PR #2850, opened 2022-04-20
docs: add more reference docs

### LINKED ISSUE #2758: Expose more code documentation in read the docs reference documentation.
We currently have quite some documentation in https://renku-python.readthedocs.io/en/latest/reference/index.html but we should expose more of our docstrings there.

Namely:
- renku.core.commands are likely to be used by people, but aren't exposed in the documentation at all
- We have a section on "Gateways", but it's empty, we should expose the interface and implementation documentation.
- We've been moving things to `renku.core.management.X` (with X =[dataset, workflow, template, etc.]), but still include it in a section for the LocalClient documentation. We should have a general section for business logic with subsections for each of those in the docs, unrelated to LocalClient.

**The point of this issue is not to write new documentation, but to move existing documentation into a more appropriate structure.**

Panaetius:
closes #2758


mohammad-sdsc (approved):
LGTM, thank you!

PR #2852, opened 2022-04-21
docs: add plugin list

### LINKED ISSUE #2446: docs: list plugins
We need a page listing plugins to the renku CLI. So far there are two:

* [renku-mls](https://github.com/ratschlab/renku-mls): machine learning model metadata extraction
* [renku-aqs](https://github.com/oda-hub/renku-aqs): extraction of metadata from AstroQuery actions

Ideally the page would have links to example projects and a blog post or three.

Panaetius:
closes #2446 

I didn't include `renku-aqs` as it's not currently on pypi and hasn't had an update on the master branch in a while. We should add it once there's a proper release.


mohammad-sdsc (approved):
Thank you!

PR #2854, opened 2022-04-22
feat(service): restore optimized migration check

### LINKED ISSUE #2546: Fix optimized migration check for users with an invalid auth
https://github.com/SwissDataScienceCenter/renku-python/issues/2432 implements an optimized migration check by downloading only necessary files. However, if a user's access token is invalid, this optimization raises an exception even for public projects.

lorenzo-cavazzi:
This PR restores the optimized migration check for the service.
The problem was mainly caused by invalid GitLab tokens that trigger errors when passed to the `gitlab` python library -- quite unexpected considering the different outcomes when invoking similar GitLab APIs with the same expired/invalid tokens.

In short, I added logic to:
- handle the `gitlab` library common errors.
- fall back to full repository checkout in case of unforeseen errors.

/deploy
fix #2546


renku/ui/service/controllers/cache_migrations_check.py
```diff
@@ -71,11 +71,15 @@ def renku_op(self):
...
-        return result_response(self.RESPONSE_SERIALIZER, self.execute_op())
+        if "project_id" in self.context:
+            result = self.execute_op()
+        else:
+            # NOTE: use quick flow but fallback to regular flow in case of unexpected exceptions
+            try:
+                result = self._fast_op_without_cache()
+            except (AuthenticationError, ProjectNotFound):
+                raise
+            except Exception:
```
Panaetius, line 82:
```suggestion
            except BaseException:
```

This should be even lower-level and really catch everything. 

Or just this, which amounts to the same thing:
```suggestion
            except:
```

Panaetius (changes requested):
Thank you, looks a lot better now :slightly_smiling_face: I just had one really nitpicky comment

renku/ui/service/controllers/cache_migrations_check.py
lorenzo-cavazzi: Ah, I didn't know about `BaseException` :bulb: 
P.S. I actually went with just `except:` initially, but the linter complained.

lorenzo-cavazzi:
I added a test with a public repo, to be sure we don't break that for the optimized migrations checks (although that `except BaseException` should assure that will never be the case again)

PR #2855, opened 2022-04-22
chore: update telepresence script to support v2

lorenzo-cavazzi:
Add support to telepresence v2 to the `start-telepresence.sh` script


Panaetius (approved):
:rocket:

PR #2866, opened 2022-04-28
feat(cli): gc command

### LINKED ISSUE #2544: Command to delete temporary directories and caches
Renku stores some temporary files/repositories in `.renku/cache` and `.renku/tmp` directories. These directories can get too large for example when adding files from a remote git repositories.
We need a `renku gc` or `renku clean` command to delete these temporary directories.

mohammad-sdsc:
# Description

WIP

Fixes #2544


Panaetius (approved):
thank you!

PR #2876, opened 2022-04-29
refactor(core): remove unused git isolation code

### LINKED ISSUE #2862: Remove `set_git_isolation` and `get_git_isolation`
`set_git_isolation` and `get_git_isolation` are still in our code base https://github.com/SwissDataScienceCenter/renku-python/blob/628a019492a465d42cf2ec559f8c95f66d8ee308/renku/command/git.py#L62 but the environment variable they set isn't actually used anymore. We should remove that part of the code, update `option_isolation` in https://github.com/SwissDataScienceCenter/renku-python/blob/628a019492a465d42cf2ec559f8c95f66d8ee308/renku/command/options.py#L24 and move `option_isonlation` to `run.py` directly, the only place where it is used at the moment.

mohammad-sdsc:
Fixes #2862

### DIFF
--- a/renku/command/git.py
+++ b/renku/command/git.py
@@ -24,7 +24,6 @@ import click
 from renku.core import errors
 
 GIT_KEY = "renku.git"
-GIT_ISOLATION = "renku.worktree"
 
 
 def set_git_home(value: Path):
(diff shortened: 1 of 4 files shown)


Panaetius (approved):
Thank you!

PR #2877, opened 2022-05-02
fix(cli): add git user configuration to local session

### LINKED ISSUE #2857: git not properly configured in container started with `renku session`
**Describe the bug**

```
$ renku session start 
$ renku session open <session-id>
```

Open a terminal, do 

```
$ renku run wc README.md > readme.wc
Error: The user name and email are not configured. Please use the "git config" command to configure them.

        git config --global --add user.name "John Doe"
        git config --global --add user.email "john.doe@example.com"
```

To solve this, we should either 

a) forward the local git configuration to the container via git environment variables
b) do some post-start configuration in the container by executing git commands

Panaetius:
closes #2857 

Copies over `~/.gitconfig` if it exists, otherwise sets environment variables.

### DIFF
--- a/renku/core/session/docker.py
+++ b/renku/core/session/docker.py
@@ -17,6 +17,7 @@
 # limitations under the License.
 """Docker based interactive session provider."""
 
+import os.path
 from pathlib import Path
 from typing import Any, Dict, Iterable, List, Optional, Tuple, cast
 from uuid import uuid4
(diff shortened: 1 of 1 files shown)


renku/core/session/docker.py
```diff
@@ -158,6 +159,19 @@ def session_start(
                         docker.types.DeviceRequest(count=[gpu_request], capabilities=[["compute", "utility"]])
                     ]
 
+            # NOTE: set git user
+            environment = {}
+            volumes = [f"{str(client.path.resolve())}:/home/jovyan/work"]
```
rokroskar, line 164: careful here, the user isn't always `jovyan` (e.g. in RStudio images). Is it possible to do this after the container is up and running? Or inspect the image metadata for `USER`? It might be worth checking how vscode does this configuration when connecting a local editor to a container; they do all of it after the container is up and running afaict.

renku/core/session/docker.py
```diff
@@ -158,6 +159,19 @@ def session_start(
...
+
+            global_git_config_path = os.path.normpath(os.path.expanduser("~/.gitconfig"))
+
+            if os.path.exists(global_git_config_path):
+                volumes.append(f"{global_git_config_path}:/home/jovyan/.gitconfig")
```
rokroskar, line 169: same as above... keep paths generic

rokroskar (changes requested):

renku/core/session/docker.py
```diff
@@ -160,13 +160,19 @@ def session_start(
...
             # NOTE: set git user
+            image_data = self.docker_client().api.inspect_image(image_name)
+            working_dir = image_data.get("Config", {}).get("WorkingDir", None)
+
+            if working_dir is None:
+                working_dir = "/home/jovyan"
+
             environment = {}
-            volumes = [f"{str(client.path.resolve())}:/home/jovyan/work"]
+            volumes = [f"{str(client.path.resolve())}:{working_dir}/work"]
```
rokroskar, line 170: better to use pathlib or similar for path concatenation? To avoid e.g. double forward slashes.

rokroskar (commented):
I'd prefer to use something more intelligent for path concatenation than an f-string... unless there's good reason not to do it.

rokroskar (approved):
This works now as expected - thanks @Panaetius!!

mohammad-sdsc (approved):
🚀

PR #2880, opened 2022-05-03
feat(core): notebook_session provider

### LINKED ISSUE #2635: Create a renkulab interactive session provider
Using the `ISessionProvider` interface implement a `RenkulabSessionProvider` that uses the notebook API of renkulab

olevski:
# Description

Launching notebooks sessions with the cli.

Fixes #2635

## Type of change

Added a new session provider. It will work with both anonymous and registered users. As long as the renku deployment supports anonymous sessions.


tests/core/plugins/test_session.py
```diff
@@ -61,7 +62,13 @@ def fake_session_list(self, project_name, config):
     return ["0xdeadbeef"]
 
 
-@pytest.mark.parametrize("provider", ["docker"])
+@pytest.mark.parametrize(
```
olevski, line 65: Thes addition of this parametrization is the true change here. And it was added on all the tests. Github flags more stuff as big changes because the indentation changed on all tests because now the patching needs a parameter and we use `with patch(...):` rather than `@patch()` as it was before.

renku/core/session/notebook_service.py
```diff
@@ -0,0 +1,386 @@
...
+            sleep(5)
+        raise errors.NotebookServiceSessionError(
+            f"Waiting for the CI jobs for project {gitlab_project_name} to finish timed out."
+        )
+
+    def _commit_and_push_checks(self):
+        client = _get_client()
+        remote = get_remote(client.repository)
+        if self._is_user_registered():
+            if client.repository.is_dirty() or len(client.repository.untracked_files) > 0:
```
Panaetius, line 207:
```suggestion
            if client.repository.is_dirty(untracked_files=True):
```

renku/core/session/notebook_service.py
```diff
@@ -0,0 +1,386 @@
...
+                    "You have new uncommitted or untracked changes to your repository. "
+                    "Renku can automatically commit and push these changes so that it builds "
+                    "the correct image for your session. Do you wish to proceed?",
+                    abort=True,
+                )
+                client.repository.add(all=True)
+                client.repository.commit("Automated commit by Renku CLI.")
+                client.repository.push()
+            remote_head_hexsha = client.repository.run_git_command(
+                "rev-parse", f"{remote.name}/{client.repository.active_branch}"
```
Panaetius, line 218: We shouldn't call `run_git_command` directly, from outside repository.py, to keep separation of git from our codebase clean.

Maybe we could add a method to `Remote` and use `client.repository.remotes[...].reference` or something like that?

Do we need to fetch beforehand to make sure we have up-to-date information?

renku/core/session/notebook_service.py
```diff
@@ -0,0 +1,386 @@
...
+                communication.confirm(
+                    "You have new changes that are present only in your local repository. "
+                    "Renku can automatically push these changes so that it builds "
+                    "the correct image for your session. Do you wish to proceed?",
+                    abort=True,
+                )
+                client.repository.push()
+
+    @staticmethod
+    def _send_renku_request(req_type: str, *args, **kwargs):
```
Panaetius, line 230: We should probably use these methods instead: https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/core/util/requests.py#L63 they also have retry logic with exponential backoff

Panaetius (changes requested):

renku/core/session/notebook_service.py
olevski: @Panaetius _send_renku_request is using `renku.core.session.utils.requests`. The requests in the method is coming from `renku.core.session.utils`

renku/core/session/notebook_service.py
```diff
@@ -0,0 +1,386 @@
...
+            )
+        client.set_value(section="anonymous_tokens", key=renku_host, value=anon_token, global_only=True)
+    return anon_token
+
+
+@inject.autoparams()
+def _get_client(client_dispatcher: IClientDispatcher) -> LocalClient:
+    """Get the local client."""
+    client = client_dispatcher.current_client
+    return client
```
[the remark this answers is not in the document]
olevski: I had some issues with this but I managed to get by it. Hopefully the implementation is good enough. With or without the Proxy trying to inject in a class variable did not work. This is because the NotebookServiceSessionProvider is initialized when registered as a plugin and this causes the injection to occur when it is too early and fail.

Using a class method bypasses this.

renku/core/session/notebook_service.py
```diff
@@ -0,0 +1,386 @@
...
+
+
+def _split_project_name(project_name: str) -> Dict[str, str]:
+    """Split a full project name into the namespace and project components required by the notebooks API."""
+    parts = project_name.split("/")
+    return {
+        "namespace": "/".join(parts[:-1]),
+        "project": parts[-1],
+    }
```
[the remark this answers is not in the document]
olevski: The main reason I have to split the name is that sometimes the project name is taken from `client.path.name` when `client.remote["name"]` is not defined. When `client.path.name` is used then the owner/name is not split up.

But I did make this case more obvious in the code now. And I dod not split things if `client.remote["name"]` is defined.

renku/core/session/notebook_service.py
olevski: cool did not know about this

renku/core/session/notebook_service.py
olevski: Ok cool. I will add such a method.

> Do we need to fetch beforehand to make sure we have up-to-date information?

I am not sure what you mean by this. I think it is fine if the method fetches the remote ref when it is called. We do not have to do this in the background or periodically and save the ref somewhere... If that is what you meant. So just `client.repository.remotes[...].reference()` or `client.repository.remotes[...].head()` should just return the remote head commit when called.

Panaetius: >  think it is fine if the method fetches the remote ref when it is called
Yes this is what I meant. I think we should do that

olevski: Ok yeah this is not a `head` property on the `Remote` class as you suggested. It fetches also before running `rev-parse`.

renku/core/session/notebook_service.py
```diff
@@ -0,0 +1,381 @@
...
+
+
+class NotebookServiceSessionProvider(ISessionProvider):
+    """A session provider that uses the notebook service API to launch sessions."""
+
+    DEFAULT_TIMEOUT_SECONDS = 300
+    _client_dispatcher = None
+
+    @classmethod
+    def client_dispatcher(cls):
```
Panaetius, line 89:
```suggestion
    @property
    def client_dispatcher(self):
```

It being a property should work as well and make code accessing it look a bit cleaner

renku/core/session/notebook_service.py
```diff
@@ -0,0 +1,386 @@
...
+from renku.core.management.client import LocalClient
+from renku.core.plugin import hookimpl
+from renku.core.session.utils import get_renku_project_name
+from renku.core.util import communication, requests
+from renku.core.util.git import get_remote
+from renku.core.util.urls import parse_authentication_endpoint
+from renku.domain_model.session import ISessionProvider, Session
+
+
+@inject.autoparams("client_dispatcher")
```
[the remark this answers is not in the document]
olevski: Now there is no client injection in any of the methods in the notebook service. Thanks

renku/core/session/notebook_service.py
```diff
@@ -0,0 +1,386 @@
...
+                        "X-Forwarded-Uri": "/api/user",
+                    },
+                )
+            except (requests.exceptions.RequestException, requests.exceptions.ConnectionError):
+                pass
+            return session.cookies.get("anon-id")
+
+    client = client_dispatcher.current_client
+    renku_host = urllib.parse.urlparse(renku_url).netloc
+    anon_token = client.get_value(section="anonymous_tokens", key=renku_host)
```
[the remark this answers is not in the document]
olevski: Yes, done. I did the change my self in a batch with a bunch of others.

renku/core/session/notebook_service.py
```diff
@@ -0,0 +1,386 @@
...
+    if not anon_token:
+        anon_token = _get_anonymous_token()
+        if not anon_token:
+            raise errors.AuthenticationError(
+                "Could not get anonymous user token from Renku. "
+                f"Ensure the Renku deployment at {renku_url} supports anonymous sessions."
+            )
+        client.set_value(section="anonymous_tokens", key=renku_host, value=anon_token, global_only=True)
```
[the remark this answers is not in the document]
olevski: Yes, done. I did the change my self in a batch with a bunch of others.

renku/core/session/notebook_service.py
```diff
@@ -0,0 +1,386 @@
...
+                return
+            sleep(5)
+        raise errors.NotebookServiceSessionError(
+            f"Waiting for the CI jobs for project {gitlab_project_name} to finish timed out."
+        )
+
+    def _commit_and_push_checks(self):
+        client = _get_client()
+        remote = get_remote(client.repository)
+        if self._is_user_registered():
```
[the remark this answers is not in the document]
olevski: Yes, done. I did the change my self in a batch with a bunch of others.

renku/core/session/notebook_service.py
olevski: I did the change my self in a batch with a bunch of others.

renku/core/session/notebook_service.py
olevski: Ah cool, I wasn't sure if there were performance or other concerns in making this a property. I will make it a property then.

olevski: Ok so `@property` does not work with the injection or some other thing. But I could just make it a method that returns the client. And that works just fine. And I do not mind putting an extra set of brackets when calling the dispatcher.

olevski:
p.s. I think that the `documentation-pending` label can be removed since the use of the `renku session` cli command has been documented in this PR. But then I am not sure what is the exact meaning of the label and maybe it should stay on this PR. Let me know if there is something I can do here for the docs or anything else.

Panaetius:
> p.s. I think that the documentation-pending label can be removed since the use of the renku session cli command has been documented in this PR

That's the exact purpose of that label. Remind people to check if they added documentation. And once they did, they can remove the label.

Panaetius (approved):

Panaetius (changes requested):
We also need to link the session command here: https://github.com/SwissDataScienceCenter/renku-python/blob/develop/docs/reference/commands.rst

otherwise it won't end up in the docs

rokroskar:
There seems to be an issue with creating the image name: 

```
renku session start -p notebook_service                                                                                     
The container image 'registry.renkulab.io/k.roskar/jupytext:a83f00b' does not exists. Would you like to build it? 
```

It should be looking for `registry.renkulab.io/rok.roskar/jupytext:a83f00b`

rokroskar:
one more comment - instead of having `renku session open` go to `<renku-url>/sessions/<session>` could we instead go to `<renku-url>/projects/<namespace>/<project>/sessions/show/<session>`? The latter takes you to the framed notebook with a bit more of the renkulab context around it. The user can click on "open" to get the session in the full tab afterwards, but there is no way to go in the other direction as of now.

olevski:
> It seems to be a problem with the fact that I did `renku login --git` - when I skip `--git` it works.

What is the purpose of the `--git` flag? From what I could tell it just uses a slightly different remote name that gets routed some special way through the gateway or something? The explanation from the `--help` says `Log in to gitlab too.`. But a bit more context in this case would be definitely useful.

renku/core/session/utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+@inject.autoparams()
+def get_renku_project_name(client_dispatcher: IClientDispatcher) -> str:
+    """Get the full name of a renku project."""
+    client = client_dispatcher.current_client
+    project_name = (
+        f"{client.remote['owner']}/{client.remote['name']}" if client.remote["name"] else f"{client.path.name}"
+    )
+    if get_remote(client.repository, name="renku-backup-origin") and project_name.startswith("repos/"):
+        project_name = project_name.lstrip("repos/")
```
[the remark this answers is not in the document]
olevski: A yeah you are right Rok. `lstrip` replaces any of the characters in the string - not only the whole occurence of the string. That is why it stripped `r` and `o` from your name at the beginning. This should fix the problem you reported earlier.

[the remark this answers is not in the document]
rokroskar: yeah though we should probably be careful... if `repos/` is some other (legitimate) part of the project it will also get nuked. Not very likely, but still.

rokroskar:
> What is the purpose of the --git flag?

it replaces the remote with one that goes through the proxy so a JWT can be used

olevski:
@Panaetius I added a section in the file you mentioned so that things appear in the documentation.

@rokroskar I addressed your comments:
- now the provider (both internally and externally) is called `renkulab`
- the issues with wrong/incomplete project names have been resolved (it was all due to `lstrip`)
- the session is opened in the iframe view when the user is registered, I could not use the iframe view for anonymous sessions because I need to pass the anonymous user ID in the URL but the UI does not seem to accept this in the iframe view (in the iframe view the UI just relies on the anonymous ID being present in a cookie I think).

Let me know if you need anything else.

rokroskar:
Thanks @olevski! It's almost there - I still get this (after logging in with `renku login --git`:

```
$ renku session start -p renkulab
Error: Cannot start session via the notebook service because {"messages":{"error":"Cannot start server rok-2erosk-jupytext-ac25ab2f, because project jupytext does not exist, branch master does not exist, commit a83f00bc4f3d151358322b2799460f830b63e083 does not exist, image registry.renkulab.io/rok.roskar/jupytext:a83f00b does not exist or cannot be accessed"}}
```

Works like a charm if using the normal remote.

olevski:
Does anyone know how to make this docs linting error go away? I do not even have bullet lists in line 38 at `renku.ui.cli.session`. 

```
Warning, treated as error:
/opt/hostedtoolcache/Python/3.7.13/x64/lib/python3.7/site-packages/renku/ui/cli/session.py:docstring of renku.ui.cli.session:38:Bullet list ends without a blank line; unexpected unindent.
Error: Process completed with exit code 2.
```

olevski:
In addition to testing with a fork of your `jupytext` repo I also tested with the `jupytext` repo directly @rokroskar. That also works.

olevski:
@rokroskar I fixed the error you mentioned. I used `lstrip` in two places and I thought I fixed both of them but I hadn't. Now it should all work.

olevski:
Ok all the comments and linting errors have been addressed.

rokroskar (approved):
Looks great, thanks @olevski!

Panaetius (approved):

PR #2882, opened 2022-05-03
docs: add core-service architecture diagram

Panaetius:
Adds a plantuml diagram of the core service architecture.


docs/reference/service.rst
```diff
@@ -21,6 +21,50 @@ Renku Core Service
 The Renku Core service exposes a functionality similar to the Renku CLI via a
 JSON-RPC API.
 
+Components:
+
+.. uml::
```
rokroskar, line 26: why uml and not mermaid?

Panaetius: It's even uglier in mermaid. mermaid just doesn't really support this kind of diagram well and doesn't offer any control over layout. I tried for 1-2 hours and couldn't get a good result with mermaid. 

Since we use plantuml in the main `renku` repo, I decided to check that out, and [this](https://plantuml.com/component-diagram) looked much closer to what I was going for. And I can specify the cardinal direction of arrows at least. It even follows what I specify roughly 70% of the time!

rokroskar: I was hoping to harmonize all the diagrams to mermaid so we only ever have to use one...

Panaetius: That's ok for diagram types mermaid supports, but they don't support that many. You can thumbs-up https://github.com/mermaid-js/mermaid/issues/1253 if you'd like. That's what I'd really want. Mermaid is fine when layout doesn't matter or if you're lucky and the layout is fine, but the lack of control over the layout is a big hindrance imo.

The alternative is I make it with draw.io exactly how I want and store it as svg.

rokroskar: ok I see what you mean - I like this one, we can beautify it later :)

docs/conf.py
```diff
@@ -42,12 +42,14 @@
...
 # ones.
 extensions = [
+    "plantweb.directive",
     "sphinx.ext.autodoc",
     "sphinx.ext.coverage",
     "sphinx.ext.doctest",
     "sphinx.ext.intersphinx",
     "sphinx.ext.napoleon",
     "sphinx.ext.viewcode",
+    "sphinxcontrib.mermaid",
```
rokroskar, line 52: aren't actually using mermaid anywhere, right?

pyproject.toml
```diff
@@ -129,8 +130,9 @@ responses = { version = "<0.19.0,>=0.7.0", optional = true }
 rich = "<11.3.0,>=9.3.0"
 rq = { version = "==1.10.1", optional = true }
 rq-scheduler = { version = "==0.11.0", optional = true }
-sentry-sdk = { version = ">=1.0.0,<1.5.6", extras = ["flask"], optional = true }
+sentry-sdk = { version = ">=1.0.0,<1.5.12,!=1.5.10", extras = ["flask"], optional = true }
 shellingham = "1.4.0"
+sphinxcontrib-mermaid = {version = ">=0.7.1,<0.8.0", optional = true }
```
rokroskar, line 135: remove

pyproject.toml
```diff
@@ -203,7 +205,7 @@ tests = [
     "types-tabulate"
 ]
 toil = ["toil"]
-docs = ["renku-sphinx-theme", "sphinx-rtd-theme", "sphinxcontrib-spelling", "sphinx-tabs"]
+docs = ["plantweb", "renku-sphinx-theme", "sphinx-rtd-theme", "sphinx-contrig-mermaid", "sphinxcontrib-spelling", "sphinx-tabs"]
```
rokroskar, line 208: remove mermaid

pyproject.toml
```diff
@@ -242,6 +245,7 @@ all = [
     "rq",
     "rq-scheduler",
     "sentry-sdk",
+    "sphinx-contrig-mermaid",
```
rokroskar, line 248: remove

rokroskar (changes requested):

docs/conf.py
Panaetius: not yet but since I added it anyways and we can do more mermaid stuff in the future, I saw no reason to remove this.

rokroskar: there's also no reason to keep it here for some hypothetical future :)

rokroskar (dismissed):

docs/conf.py
Panaetius: I could just open an issue "Add mermaid support for docs" and change this to "closes #thatissue" and then it's a feature! :stuck_out_tongue_closed_eyes:

rokroskar (dismissed):
mermaid, no mermaid... mermaid, no mermaid... mermaid, no mermaid... 🤷

olevski (approved):

PR #2890, opened 2022-05-06
fix: remove old scheduled jobs

rokroskar:
Removes old jobs from the queue before scheduling new ones to avoid duplication. 

/deploy #persist


rokroskar:
I verified that this works as expected by checking in redis that the old jobs are removed when the pod is restarted.

Panaetius (approved):
Thanks!

PR #2891, opened 2022-05-09
fix(service): raise exception correctly

lorenzo-cavazzi:
/deploy


renku/ui/service/controllers/api/mixins.py
```diff
@@ -277,7 +277,7 @@ def reset_local_repo(self, project):
                         except GitCommandError as e:
                             project.purge()
                             raise IntermittentCacheError(e)
-                        else:
+                        except BaseException:
                             raise
```
Panaetius, line 281: There lines should just be removed alltogether

Panaetius (approved):

PR #2893, opened 2022-05-09
feat(api): project status

### LINKED ISSUE #2858: Add renku.api Project.status functionality

mohammad-sdsc:
# Description

- Adds a `status()` method to `api.Project`
- Some refactorings/removals

Fixes #2858


renku/core/workflow/run.py
```diff
@@ -0,0 +1,101 @@
...
+@inject.autoparams("client_dispatcher")
+def get_status(
+    client_dispatcher: IClientDispatcher, paths: Optional[List[Union[Path, str]]] = None, ignore_deleted: bool = False
+) -> Tuple[Dict[str, Set[str]], Dict[str, Set[str]], Set[str], Set[str]]:
+    """Return status of a project.
+
+    Args:
+        client_dispatcher(IClientDispatcher): Injected client dispatcher.
+        paths(Optional[List[Union[Path, str]]]): Limit the status to this list of paths (Default value = None).
+        ignore_deleted(bool): Whether to ignore deleted generations.
```
Panaetius, line 43:
```suggestion
        ignore_deleted(bool): Whether to ignore deleted generations (Default value = False).
```

renku/core/workflow/run.py
```diff
@@ -0,0 +1,101 @@
...
+
+    Returns:
+        Tuple[Dict[str, Set[str]], Dict[str, Set[str]], Set[str], Set[str]]: A quadruple containing a mapping of stale
+            outputs to modified inputs, a mapping of stale activities that have no generation to modified inputs, a set
+            of modified inputs, and a set of deleted inputs.
```
Panaetius, line 48: I wonder if we should use a typing.NamedTuple instead?

renku/ui/api/models/project.py
```diff
@@ -69,6 +81,24 @@ def path(self):
...
+            paths(Optional[List[Union[Path, str]]]): Limit the status to this list of paths (Default value = None).
+            ignore_deleted(bool): Whether to ignore deleted generations.
+
+        Returns:
+            Tuple[Dict[str, Set[str]], Dict[str, Set[str]], Set[str], Set[str]]: A quadruple containing a mapping of
+                stale outputs to modified inputs, a mapping of stale activities that have no generation to modified
+                inputs, a set of modified inputs, and a set of deleted inputs.
+
+        """
+        result = get_status_command().build().execute(paths=paths, ignore_deleted=ignore_deleted)
```
Panaetius, line 99: This will create a new LocalClient based on CWD when we already have one in `self._client`. I don't think that can cause problems, but I wonder if we should add a way to set a client in the CommandBuilder directly? Or at least set the path?

That would also allow us to refactor https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/ui/service/controllers/api/mixins.py#L314-L315 to not use click context. Using click to inject a created client in the service is a bit ugly. But changing the service could be done in a follow-up issue.

Panaetius (dismissed):
Thank you!

PR #2901, opened 2022-05-13
fix(service): create dummy metadata.yml when creating a project

### LINKED ISSUE #2900: add backwards compatibility metadata.yml on service as well
https://github.com/SwissDataScienceCenter/renku-python/pull/2444 only actually works when creating a project through the CLI, NOT when creating it through the service. This can cause users to create a project through the UI with 1.x.x and then accidentally work on it with 0.16.x, leading to weird project metadata

Panaetius:
closes #2900

### DIFF
--- a/renku/command/init.py
+++ b/renku/command/init.py
@@ -215,14 +215,6 @@ def _init(
 
     branch_name = create_backup_branch(path=path)
 
-    # add metadata.yml for backwards compatibility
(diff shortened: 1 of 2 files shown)


mohammad-sdsc (approved):
Thank you!

mohammad-sdsc (approved):

PR #2909, opened 2022-05-16
feat(api): plan

### LINKED ISSUE #2859: Add renku.api Plan functionality
Add `Plan` and `CompositePlan` classes and methods as per https://github.com/SwissDataScienceCenter/renku-rfc/blob/main/design/006-renku-api-activities-plans/006-renku-api-activities-plans.md#getting-activities-and-plans along with `Input`, `Output`, `Parameter`, `Mapping` and `Link` classes.

mohammad-sdsc:
# Description

Wraps an execution `Plan` in an API model class.

Fixes #2859 

# TODO
  - [x] Rename `renku/ui/api/models/run.py` to `renku/ui/api/models/parameter.py` once the review is done.


renku/ui/api/models/plan.py
```diff
@@ -0,0 +1,213 @@
...
+
+        Returns:
+            Plan: An API Plan model.
+        """
+        return cls(
+            command=plan.command,
+            date_created=plan.date_created,
+            description=plan.description,
+            inputs=[Input.from_parameter(i) for i in plan.inputs],
+            invalidated_at=plan.invalidated_at,
```
Panaetius, line 86: Should we just keep a boolean field `deleted` in the api? Not sure `invalidated_at` is meaningful for a user. Or change to `date_deleted`.

renku/ui/api/models/plan.py
```diff
@@ -0,0 +1,213 @@
...
+        invalidated_at: Optional[datetime] = None,
+        keywords: Optional[List[str]] = None,
+        links: List[Link] = None,
+        mappings: List[Mapping] = None,
+        name: Optional[str] = None,
+        plans: List[Union["CompositePlan", "Plan"]] = None,
+    ):
+        self.date_created: Optional[datetime] = date_created
+        self.description: Optional[str] = description
+        self.invalidated_at: Optional[datetime] = invalidated_at
```
Panaetius, line 128: Same as above

renku/ui/api/models/plan.py
```diff
@@ -0,0 +1,213 @@
...
+    def list(include_deleted: bool = False) -> List[Union["Plan", "CompositePlan"]]:
+        """List all plans in a project.
+
+        Args:
+            include_deleted(bool): Whether to include deleted plans.
+
+        Returns:
+            A list of all plans in the supplied project.
+        """
+        return _list_plans(include_deleted=include_deleted)
```
Panaetius, line 104: Should we filter so `Plan.list` only returns `Plan` and `CompositePlan.list` only returns `CompositePlan` ?

renku/ui/api/models/run.py
```diff
@@ -95,66 +95,233 @@ def __fspath__(self):
...
+    @ensure_project_context
+    def __init__(
+        self,
+        name: str,
+        value: Union[Path, str, bool, int, float],
+        project,
+        default_value: Union[Path, str, bool, int, float] = None,
+        description: str = None,
+        position: int = None,
+        prefix: str = None,
```
Panaetius, line 115: I think we don't want to allow users setting this, and we should only set it in the class method below. Setting e.g. position in my user script wouldn't do anything, so it'd be a useless option for a user to set.

Description might be nice but we probably don't want to change the indirect parameter logic in this PR.

I'm a bit torn here, letting a user set something that does nothing isn't nice, but if we let users use the API to construct Plans in the future, it might be useful again. It'd be nice if Python allowed constructor overloading...

renku/ui/api/models/run.py
```diff
@@ -95,66 +95,233 @@ def __fspath__(self):
...
         return get_indirect_inputs_path(project_path)
 
+    def __init__(
+        self,
+        name: str,
+        path: Union[str, Path],
+        description: str = None,
+        mapped_stream: str = None,
+        position: int = None,
+        prefix: str = None,
```
Panaetius, line 191: Same

renku/ui/api/models/run.py
```diff
@@ -95,66 +95,233 @@ def __fspath__(self):
...
         return get_indirect_outputs_path(project_path)
 
+    def __init__(
+        self,
+        name: str,
+        path: Union[str, Path],
+        description: str = None,
+        mapped_stream: str = None,
+        position: int = None,
+        prefix: str = None,
```
Panaetius, line 239: Same

renku/ui/api/models/run.py
```diff
@@ -95,66 +95,233 @@ def __fspath__(self):
...
+        self.default_value: Union[Path, str, bool, int, float] = default_value if default_value is not None else value
+        self.description: Optional[str] = description
+        self.position: Optional[int] = position
+        self.prefix: Optional[str] = prefix
+
+        if not skip_addition:
+            add_indirect_parameter(project.path, name=name, value=value)
+
+    @classmethod
+    def from_parameter(cls, parameter: core_parameter.CommandParameterBase) -> "Parameter":
```
Panaetius, line 134:
```suggestion
    def from_parameter(cls, parameter: core_parameter.CommandParameter) -> "Parameter":
```
Though that probably makes mypy sad

renku/ui/api/models/run.py
```diff
@@ -95,66 +95,233 @@ def __fspath__(self):
...
-    if not name:
-        raise errors.ParameterError("'name' must be set.")
-
-    if not name_pattern.match(name):
-        raise errors.ParameterError(
-            f"Name {name} contains illegal characters. Only characters, numbers, _ and - are allowed."
+        self.mapped_stream: Optional[str] = mapped_stream
+
+    @classmethod
+    def from_parameter(cls, output: core_parameter.CommandParameterBase) -> "Output":
```
Panaetius, line 258:
```suggestion
    def from_parameter(cls, output: core_parameter.CommandOutput) -> "Output":
```

Panaetius (changes requested):
Thank you! I left some comments, but some of them I'm unsure about

renku/ui/api/models/plan.py
mohammad-sdsc: I've removed it and added a `deleted` field instead. We should have named it `date_deleted` in the core `Plan` class as well.

renku/ui/api/models/plan.py
mohammad-sdsc: This is probably cleaner since CompositePlan doesn't inherit from Plan.

Panaetius (dismissed):

Panaetius (approved):

PR #2911, opened 2022-05-17
feat(api): activity

### LINKED ISSUE #2860: Add renku.api Activity functionality
Add `Activity` class and methods as per https://github.com/SwissDataScienceCenter/renku-rfc/blob/main/design/006-renku-api-activities-plans/006-renku-api-activities-plans.md#getting-activities-and-plans along with `CreatedOutput`, `UsedInput` and `FieldValue`.

mohammad-sdsc:
# Description

* Adds an Activity class to Renku Python API that allows listing and filtering activities.

Fixes #2860


mohammad-sdsc:
Based on the [RFC](https://github.com/SwissDataScienceCenter/renku-rfc/blob/main/design/006-renku-api-activities-plans/006-renku-api-activities-plans.md#getting-activities-and-plans), we use `parameters` to refer to all field values that are used in the `Activity` class. This is a bit misleading since `parameters` doesn't only include plan parameters, but it also includes its inputs and outputs. So, we need to come up with a better name for this field like `executed_values`, ...

I also suggest that we rename `created_outputs` to `generated_outputs` to be consistent in using generation and usage everywhere.

renku/ui/api/models/activity.py
```diff
@@ -0,0 +1,427 @@
...
+    @classmethod
+    def from_parameter_value(cls, parameter: core_parameter.ParameterValue, plan: core_plan.Plan):
+        """Create an instance from core ParameterValue."""
+        return cls(field=convert_parameter(plan.get_field_by_id(parameter.parameter_id)), value=parameter.value)
+
+    def __repr__(self):
+        if isinstance(self.field, Input):
+            return f"<Input '{self.value}'>"
+        elif isinstance(self.field, Output):
+            return f"<Output '{self.value}'>"
```
Panaetius, line 396: Should we include the name for inputs/outputs as well?

renku/ui/api/models/activity.py
```diff
@@ -0,0 +1,427 @@
...
+def get_activities(plan_id: str = None) -> List[Activity]:
+    """Return list of activities from a given ``plan_id``."""
+    activity_gateway = get_activity_gateway()
+    if not activity_gateway:
+        return []
+
+    activities = activity_gateway.get_all_activities()
+
+    if plan_id:
+        activities = [a for a in activities if a.association.plan.id == plan_id]
```
Panaetius, line 410: Plans can be derived from, in which case their id changes. But we only allow (basically) non-functional changes. so I wonder if we should check for all instances of a plan here.

tests/core/utils/test_os.py
```diff
@@ -0,0 +1,49 @@
...
+        ["path/in/sub-dir", "sub-dir", False],
+        ["path/in/sub-dir", "**/in", True],
+        ["path/in/sub-dir", "path/some-other-dir", False],
+        ["path/in/sub-dir", "path/**/in", False],
+        ["path/in/sub-dir", "**/sub-dir/**", False],
+        ["path/in/sub-dir", "in", False],
+    ],
+)
+def test_path_match(path, pattern, should_match):
+    """Test host is set correctly in a different Renku domain."""
```
Panaetius, line 48: Copy paste of docstring :slightly_smiling_face:

Panaetius (changes requested):

Panaetius:
Thank you, looks really nice. Did you get feedback from ordes on the two changes you proposed?

Panaetius:
Essentially what it's doing is overriding the default values of the Plan. So maybe we should use the word `override` in there somewhere? like `value_overrides` or `overridden_arguments`?

or simply just `values`, since `value` vs `default value` seems common in IT.

lokijuhy:
Yes, I suppose simpler is better. I wonder if sometimes the user will provide arguments not necessarily to override default values, but just because that's their habit, and therefore the `overridden` part may be non-intuitive. Perhaps just `values` is best, as you suggested.

renku/ui/api/models/activity.py
mohammad-sdsc: I've change this to return all activities that use a plan or any of its upstream plans. To have less overhead, it won't include downstream plans. Users can still get the latest version of a plan and then get its activities to have all version. I can change it to get the latest version automatically if you think that's a better approach.

Panaetius (approved):

PR #2913, opened 2022-05-18
chore: auto-add issues/prs to project

Panaetius:
Does what it says


.github/workflows/github-project.yml
```diff
@@ -0,0 +1,16 @@
...
+    types:
+      - opened
+jobs:
+  add_to_project:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@main
+        with:
+          project-url: https://github.com/orgs/SwissDataScienceCenter/projects/45
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
aledegano, line 16: Personally I find very annoying the idea of putting some real person's PAT in the secrets.
Perhaps this: https://github.com/navikt/github-app-token-generator might help in converting an APP authentication into a PAT.

Panaetius: `GITHUB_TOKEN` is automatically created by github, not someones personal token.

But anyways, I should use `RENKUBOT_GITHUB_TOKEN` since projects are at the organisation level. Changed it to that. And that one isn't personal but from the RenkuBot user as far as I understand.

And I wouldn't want to figure out setting an app token for this PR which is a temporary solution until github properly implements this in the UI. But maybe we should have an issue to try and use the token generator in all our repos? I don't think this PR is the right place for that discussion though.

Panaetius: Well it still needs to be merged and tested to see if it actually works 🙈

mohammad-sdsc (approved):
LGTM, thank you!

PR #2914, opened 2022-05-19
feat(cli): show spinner when cloning repo for dataset import/update

### LINKED ISSUE #2547: Enhance dataset update feedback
Currently, `dataset update` shows a progressbar to users. The progressbar is not updated uniformly when renku needs to clone a git repository (time required for cloning is not reflected in the progress). We should provide a better feedback to users when updating projects with lots of datasets or with large datasets.

Panaetius:
closes #2547


mohammad-sdsc (approved):
LGTM, thanks!

PR #2915, opened 2022-05-19
feat(cli): improve renku run error messages

### LINKED ISSUE #2836: Improve `renku run` error messages
https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/core/workflow/plan_factory.py#L329
--> Not that helpful. It should explain why this is an issue (Renku treats the whole directory as an output/as generated by this run, so it can't already contain files), and steps the user can take (delete the files, use `--param` to not treat it as a folder but a string instead or use a sub-subfolder).

https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/command/run.py#L181
--> Mention why the command cannot be executed, it wasn't found, so $PATH might be wrong or it was misspelled.

Panaetius:
closes #2836


renku/core/errors.py
```diff
@@ -212,28 +211,16 @@ class InvalidOutputPath(RenkuException):
...
-                "Once you have removed files that should be used as outputs,\n"
-                "you can safely rerun the previous command."
-                "\nYou can use --output flag to specify outputs explicitly."
-            )
-        else:
-            msg += "\n\nIf you want to track the command anyway use " "--no-output option."
+
+        msg = (
+            "There are not any detected outputs in the repository. This can be due to your command not creating "
+            "any new files or due to files that get created already existing before the command was run. In the "
```
mohammad-sdsc, line 219: I'm not sure about this: `already existing` -> `already exist`

renku/core/workflow/plan_factory.py
```diff
@@ -79,19 +79,21 @@ def __init__(
...
-            raise errors.UsageError("Command line can not be empty.")
+            raise errors.UsageError("Command line can not be empty. Please specify a command to execute.")
 
         if not directory:
             directory = os.getcwd()
         self.directory = Path(directory)
         if not self.directory.exists():
-            raise errors.UsageError("Directory must exist.")
+            raise errors.UsageError(
+                "Current directory doesn't exist. Please make sure you are inside an existing directory."
```
mohammad-sdsc, line 89: This message looks a bit weird to me. I'd change to `f"Directory doesn't exist: {self.directory}"

renku/core/workflow/plan_factory.py
```diff
@@ -79,19 +79,21 @@ def __init__(
...
+            )
 
         if not working_dir:
             working_dir = os.getcwd()
         self.working_dir = Path(working_dir)
         if not self.working_dir.exists():
-            raise errors.UsageError("Working Directory must exist.")
+            raise errors.UsageError("Repository path doesn't exist. Make sure you are inside a Renku repository.")
```
mohammad-sdsc, line 96: It's good to print the directory in the message: `f"Repository path doesn't exist: {self.working_dir} ..."

mohammad-sdsc (commented):

renku/core/errors.py
Panaetius: I think plural is correct, `due to files [...] already existing`

renku/core/workflow/plan_factory.py
Panaetius: This part of the code is really misleading. `directory` gets set to `os.getcwd()` and `working_dir` gets set to `client.repository.path`. And it's not something a user can set.

The error above only happens if someone does `cd somedir && rm -r ../somedir && renku run ...`, I think.

PR #2919, opened 2022-05-20
feat(cli): show dataset metadata for tag

### LINKED ISSUE #2902: Add --tag option to renku dataset show
The tags mechanism provide a user friendly way to version datasets for improved reproducibility.
Currently, the `renku dataset show` command only allows to inspect the latest version of a dataset.

To improve the UX around tags, user should be able to inspect dataset attributes at previous tags. This could be done through the addition of a `--tag` option to `renku dataset show`.

Panaetius:
closes #2902

### DIFF
--- a/renku/core/dataset/dataset.py
+++ b/renku/core/dataset/dataset.py
@@ -733,17 +733,30 @@ def update_datasets(
     return imported_datasets_view_models, dataset_files_view_models
 
 
-def show_dataset(name):
+def show_dataset(name: str, tag: Optional[str] = None):
     """Show detailed dataset information.
 
     Args:
-        name: Name of dataset to show details for.
+        name(str): Name of dataset to show details for.
+        tag(str, optional): Tags for which to get the metadata (Default value = None).
 
     Returns:
         dict: JSON dictionary of dataset details.
     """
-    dataset = DatasetsProvenance().get_by_name(name, strict=True)
+    datasets_provenance = DatasetsProvenance()
+    dataset = datasets_provenance.get_by_name(name, strict=True)
 
+    if tag is None:
+        return DatasetDetailsJson().dump(dataset)
+
+    tags = datasets_provenance.get_all_tags(dataset=cast(Dataset, dataset))
+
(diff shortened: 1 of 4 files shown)


renku/core/dataset/dataset.py
```diff
@@ -733,17 +733,30 @@ def update_datasets(
...
-        name: Name of dataset to show details for.
+        name(str): Name of dataset to show details for.
+        tag(str, optional): Tags for which to get the metadata (Default value = None).
 
     Returns:
         dict: JSON dictionary of dataset details.
     """
-    dataset = DatasetsProvenance().get_by_name(name, strict=True)
+    datasets_provenance = DatasetsProvenance()
+    dataset = datasets_provenance.get_by_name(name, strict=True)
```
mohammad-sdsc, line 747: This is not used if tag is not None. We should put it inside the if below.

tests/cli/test_datasets.py
```diff
@@ -128,6 +128,69 @@ def test_dataset_show(runner, client, subdirectory):
...
+    result = runner.invoke(cli, ["dataset", "tag", "my-dataset", "tag3"])
+    assert 0 == result.exit_code, format_result_exception(result)
+
+    result = runner.invoke(cli, ["dataset", "show", "my-dataset"])
+    assert 0 == result.exit_code, format_result_exception(result)
+    assert "description3" in result.output
+
+    result = runner.invoke(cli, ["dataset", "show", "--tag", "tag3", "my-dataset"])
+    assert 0 == result.exit_code, format_result_exception(result)
+    assert "description3" in result.output
```
mohammad-sdsc, line 183: We should also add `assert "description2" not in result.output`. Also we should do it for the other checks below.

mohammad-sdsc (commented):
Thanks Ralf, looks good! Just some minor comments.

renku/core/dataset/dataset.py
Panaetius: it is used both on line 750 in the if and 752 in the get_all_tags method

mohammad-sdsc (approved):
🚀

PR #2920, opened 2022-05-23
fix(ci): pin poetry-dynamic-versioning to 0.17.1

olevski:
Fix ci issues with poetry-dynamic-versioning


mohammad-sdsc (approved):
Thank you!

PR #2928, opened 2022-05-25
fix(core): automatically cleanup dangling git processes

### LINKED ISSUE #2889: Clean up dangling GitPython processes
Dangling GitPython subprocesses clogs up triples generation on vomcases during mgiration.

According to https://github.com/gitpython-developers/GitPython/issues/1333 deleting the `Repository` object should clean up dangling processes, as it triggers garbage collection.

We should check if there's a way we can reduce the dangling `git cat-file` processes, either by deleting the `Repository` object at regular intervals and recreating it, or maybe killing child `git cat-file` processes in between processing commits (or every n commits).

Timeboxed to two days.

Inform KG team if this can be fixed.

Panaetius:
closes #2889 

Fixes a bug where SubmoduleManager would create a new Repository on each access, which in turn would have a dangling git subprocess.

Force cleans/deletes the repository in many cases where it makes sense, reducing the risk of dangling processes


renku/infrastructure/repository.py
```diff
@@ -675,6 +684,22 @@ def get_user(self) -> "Actor":
...
+        if self._repository:
+            self._repository.close()
+            del self._repository
+            self._repository = None
+
+    def refresh_repository(self):
+        """Refreshes the underlying repository."""
+        self.close()
+
+        self._repository = git.Repo(self.path, search_parent_directories=False)
```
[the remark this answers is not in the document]
mohammad-sdsc: Also, this method is not used anywhere, so, should we remove it?

renku/infrastructure/repository.py
```diff
@@ -881,6 +907,9 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return f"<Submodule {self.relative_path}>"
 
+    def __del__(self) -> None:
+        self._repository.close()
```
mohammad-sdsc, line 911: Should we also do:
```
del self._repository
self._repository = None
```
like the `close` method above?

renku/infrastructure/repository.py
```diff
@@ -902,22 +931,32 @@ class SubmoduleManager:
 
     def __init__(self, repository: git.Repo):
         self._repository = repository
+        self._submodule_cache: Dict[git.Submodule, Submodule] = {}  # type: ignore
```
mohammad-sdsc, line 934: Is `Submodule` object reused by GitPython? Maybe its better to use submodule's name as the key to be on the safe side.

renku/version.py
```diff
@@ -24,7 +24,7 @@
 except ImportError:
     from importlib_metadata import distribution
 
-__version__ = "0.0.0"
+__version__ = "1.2.4.dev17+geb9386e8"
```
mohammad-sdsc, line 27: 1.2.4.dev17+geb9386e8 -> 0.0.0.

mohammad-sdsc (commented):

mohammad-sdsc (approved):
Thank you!

PR #2929, opened 2022-05-31
feat(cli): add custom metadata to Plan

### LINKED ISSUE #2888: Add custom metadata to a plan
Similar to how we can add custom metadata to an Activity with a plugin, we should support adding custom metadata to a Plan in `renku run` though a plugin hook

Panaetius:
closes #2888 

Also removes some old plugin hooks, which aren't used by smartsky or mls anyways.

### DIFF
--- a/renku/domain_model/workflow/plan.py
+++ b/renku/domain_model/workflow/plan.py
@@ -30,6 +30,7 @@ from werkzeug.utils import secure_filename
 
 from renku.core import errors
 from renku.core.util.datetime8601 import local_now
+from renku.domain_model.provenance.annotation import Annotation
 from renku.domain_model.workflow.parameter import CommandInput, CommandOutput, CommandParameter, CommandParameterBase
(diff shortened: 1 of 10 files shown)


renku/domain_model/workflow/plan.py
```diff
@@ -135,6 +136,8 @@ def is_derivation(self) -> bool:
 class Plan(AbstractPlan):
     """Represent a `renku run` execution template."""
 
+    annotations: List[Annotation] = list()
```
mohammad-sdsc, line 139: Is this needed?

mohammad-sdsc (approved):
Thanks Ralf! LGTM. Just made a single comment; feel free to ignore it.

renku/domain_model/workflow/plan.py
Panaetius: yes, this is for backwards compatibility. Plans created before this change will automatically have `annotations` set to an empty list when loaded from the database

PR #2932, opened 2022-06-02
feat(cli): add pager support to renku log

### LINKED ISSUE #2908: Add pager to `renku log`
Currently, `renku log` just outputs everything directly. We should check if there is a pager (e.g. `less`) available and use that to only show the first few entries, with the user being able to use the arrow keys/page down/enter to read further down.

Panaetius:
closes #2908 

uses pager unless text fits on a single page or `--no-pager` is passed

### DIFF
--- a/renku/ui/cli/log.py
+++ b/renku/ui/cli/log.py
@@ -55,6 +55,9 @@ You can select a format using the ``--format <format>`` argument.
    :extended:
 """
 
+import pydoc
+import shutil
+
 import click
 
 import renku.ui.cli.utils.color as color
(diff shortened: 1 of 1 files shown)


renku/ui/cli/log.py
```diff
@@ -208,14 +211,24 @@ def _print_dataset_log(log_entry: DatasetLogViewModel) -> str:
...
+def log(columns, format, workflows, datasets, no_pager):
     """Show a history of renku workflow and dataset commands."""
     from renku.command.log import log_command
 
+    def _show_text(text: str, pager: bool):
+
+        tty_size = shutil.get_terminal_size(fallback=(120, 120))
+
+        if pager and len(text.splitlines()) >= tty_size.lines:
+            pydoc.tempfilepager(text, "less --chop-long-lines -R --tilde")
```
mohammad-sdsc, line 224: It seems that `pipepager` is doing the same thing without creating a file. Is there a reason not to use it?

renku/ui/cli/log.py
```diff
@@ -208,14 +211,24 @@ def _print_dataset_log(log_entry: DatasetLogViewModel) -> str:
 )
 @click.option("-w", "--workflows", is_flag=True, default=False, help="Show only workflow executions.")
 @click.option("-d", "--datasets", is_flag=True, default=False, help="Show only dataset modifications.")
-def log(columns, format, workflows, datasets):
+@click.option("--no-pager", is_flag=True, help="Don't use pager (less) for output.")
+def log(columns, format, workflows, datasets, no_pager):
     """Show a history of renku workflow and dataset commands."""
     from renku.command.log import log_command
 
+    def _show_text(text: str, pager: bool):
```
mohammad-sdsc, line 219: I believe it's OK to have public names for inner functions since they are visible from the outside anyways.

mohammad-sdsc (commented):
Thanks Rlaf! Please see my comments. If we redirect the output of the `renku log` to a file we will get the special color characters. It would be nice if we could avoid this without using `--no-pager`. I did a quick search but couldn't find anything.

renku/ui/cli/log.py
```diff
@@ -208,14 +211,24 @@ def _print_dataset_log(log_entry: DatasetLogViewModel) -> str:
...
+
+        tty_size = shutil.get_terminal_size(fallback=(120, 120))
```
mohammad-sdsc, line 221: 120 is too much for number of lines. Maybe (120, 40)?

renku/ui/cli/log.py
```diff
@@ -208,14 +209,23 @@ def _print_dataset_log(log_entry: DatasetLogViewModel) -> str:
...
 
     result = log_command().with_database().build().execute(workflows_only=workflows, datasets_only=datasets).output
     if format == "detailed":
         entries = sorted(result, key=lambda e: e.date, reverse=True)
-        texts = [_print_log(e) for e in entries]
-        click.echo("\n\n".join(texts))
+        text = "\n\n".join([_print_log(e) for e in entries])
+
+        if no_color:
+            text = text = strip_ansi_codes(text)
```
mohammad-sdsc, line 224: `text = text =`

mohammad-sdsc (approved):
🚀

PR #2936, opened 2022-06-07
feat(cli): add renku template validate command

### LINKED ISSUE #2896: Addition of `renku template validate` command
As a Renku template developer, I want to be able to validate a template repository to make sure that they are usable in Renku.

### Acceptance Criteria
- The command should validate that:
  - All defined templates in the manifest file exist and are valid
  - No renku metadata file other than `renku.ini` is included in the `.renku` folder
  - Required files are included (currently, `Dockerfile` and `.renku/renku.ini`)
  - Template parameters are valid
  - No symlink resolved to outside the template directory
  - More???
- Print warning messages for deprecated/renamed features:
  - Use `id` instead of `folder` in the manifest
  - Old template parameter format
- Use the current head to validate but should accept a git reference as well (no need to accept a template id since it validates all templates)
- Use the same validation rules when creating a project from a template or when setting a template  

**Additional context**
Look at some validations that are already implemented: https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/domain_model/template.py.

Panaetius:
closes #2896


renku/domain_model/template.py
```diff
@@ -148,40 +149,64 @@ def get_raw_content(self) -> List[Dict]:
...
+            if not isinstance(template_entry, dict):
+                raise errors.InvalidTemplateError(f"Invalid template type: '{type(template_entry).__name__}'")
 
-            id = template.get("id") or template.get("folder")
+            id = template_entry.get("id") or template_entry.get("folder")
             if not id:
-                raise errors.InvalidTemplateError(f"Template doesn't have an id: '{template}'")
+                raise errors.InvalidTemplateError(f"Template doesn't have an id: '{template_entry}'")
+            if not template_entry.get("folder"):
+                warnings.append(f"Template '{id}' should use 'folder' attribute instead of 'id'.")
```
mohammad-sdsc, line 170: It's the other way around: `folder` should be replaced with `id` which is more meaningful imo.

renku/domain_model/template.py
```diff
@@ -148,40 +149,64 @@ def get_raw_content(self) -> List[Dict]:
...
 
-            parameters = template.get("parameters") or template.get("variables")
+            parameters = template_entry.get("parameters") or template_entry.get("variables")
             if parameters:
+                if template_entry.get("parameters"):
+                    warnings.append(f"Template '{id}' should use 'variables' instead of 'parameters' in manifest.")
```
mohammad-sdsc, line 175: It's actually the case in all templates. IMO `parameters` is a better name than `variables` but it's ok if we want to keep `variables`. In that case we don't need this check and we should remove `template_entry.get("parameters")`. What do you think?

renku/domain_model/template.py
```diff
@@ -148,40 +149,64 @@ def get_raw_content(self) -> List[Dict]:
...
+
+                if not template_entry.get("variables"):
+                    template_entry["variables"] = {}
+
                 if not isinstance(parameters, dict):
-                    raise errors.InvalidTemplateError(f"Invalid template variable type: '{type(parameters).__name__}'")
+                    raise errors.InvalidTemplateError(
+                        f"Invalid template variable type on template '{id}': '{type(parameters).__name__}'"
```
mohammad-sdsc, line 182: It's useful to say that the type should be a dict.

renku/domain_model/template.py
```diff
@@ -148,40 +149,64 @@ def get_raw_content(self) -> List[Dict]:
...
 class Template:
     """Template files and metadata from a template source."""
 
     REQUIRED_ATTRIBUTES = ("name",)
     REQUIRED_FILES = (os.path.join(RENKU_HOME, "renku.ini"), "Dockerfile")
+    PROHIBITED_PATHS = (
+        os.path.join(RENKU_HOME, "metadata"),
+        os.path.join(RENKU_HOME, "metadata.yml"),
+        os.path.join(RENKU_HOME, "template_checksums.json"),
+        os.path.join(RENKU_HOME, "cache"),
```
mohammad-sdsc, line 208: We should add `tmp`, `pointers`, `datasets`, ... here as well. Perhaps it's easier/safer to prohibit all paths in `RENKU_HOME` except `renku.ini`.

renku/ui/cli/template.py
```diff
@@ -238,6 +255,41 @@ def update_template(force, interactive, dry_run):
...
+
+    if json_format:
+        click.echo(json.dumps(result))
+    else:
+        if result["warnings"]:
+            click.echo("Manifest Errors:\n")
+            for warning in result["warnings"]:
+                click.echo(f"\t{warning}\n")
+        if result["manifest"]:
+            click.echo(f"Manifest Errors:\n\t{result['manifest']}\n")
```
mohammad-sdsc, line 280: These also include warnings; for example, https://github.com/SwissDataScienceCenter/contributed-project-templates.git is a valid template but has several warnings that are printed as errors.

renku/ui/cli/template.py
```diff
@@ -238,6 +255,41 @@ def update_template(force, interactive, dry_run):
...
+    """Validate a template repository and check for common issues."""
+    from renku.command.template import validate_templates_command
+
+    result = validate_templates_command().build().execute(reference=revision).output
+
+    if json_format:
+        click.echo(json.dumps(result))
+    else:
+        if result["warnings"]:
+            click.echo("Manifest Errors:\n")
```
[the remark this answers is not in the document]
mohammad-sdsc: It good to use color-coding here: Orange/yellow for warning and red for error. I mean for the title only and not the messages.

mohammad-sdsc (changes requested):
This looks great! Thanks Ralf! Please see my comments.

renku/ui/cli/template.py
```diff
@@ -263,29 +264,36 @@ def update_template(force, interactive, dry_run):
     type=click.STRING,
     help="Git revision/branch/tag to validate the template at.",
 )
-def validate_template(ctx, json_format, revision):
+@click.option(
+    "--source",
+    type=click.STRING,
+    help="Remote template repository to clone and check.",
+)
+def validate_template(ctx, json_format, revision, source):
```
mohammad-sdsc, line 272: To be in par with other `template` command (what Lorenzo said), we should rename `revision` to `reference` and also add `-r` and `-s`.

PR #2944, opened 2022-06-12
feat(dataset): export to a local directory

### LINKED ISSUE #2905: Add "temp folder" option to renku dataset export
Users can create tags to version their Renku datasets, but accessing the contents of previous tags requires exporting it to a new {renku,zenodo,dataverse} dataset and importing it.

To improve the UX around tags, users should be able to "checkout" previous dataset tags without publishing them online.
One way to achieve this could be to add a "temporary folder" option to `renku dataset export` to extract the dataset content at a specific tags into a directory.

mohammad-sdsc:
# Description

Adds a `local` provider that can be used to export a dataset to a local directory. If `-p`/`--path` is passed then dataset is copied to that path; otherwise, a directory is created.

Fixes #2905


renku/core/dataset/providers/dataverse.py
```diff
@@ -165,12 +165,13 @@ def supports_import():
...
+    def get_export_parameters() -> List[ProviderParameter]:
         """Returns parameters that can be set for export."""
-        return {
-            "dataverse-server": ("Dataverse server URL.", str),
-            "dataverse-name": ("Dataverse name to export to.", str),
-        }
+        return [
+            ProviderParameter("dataverse-server", description="Dataverse server URL.", type=str),
+            ProviderParameter("dataverse-name", description="Dataverse name to export to.", type=str),
+            ProviderParameter("publish", description="Publish the exported dataset.", is_flag=True),
```
Panaetius, line 173: Should we add some aliases here?

renku/core/dataset/providers/olos.py
```diff
@@ -48,19 +49,17 @@ def supports_export():
         return True
 
     @staticmethod
-    def export_parameters():
+    def get_export_parameters() -> List[ProviderParameter]:
         """Returns parameters that can be set for export."""
-        return {
-            "dlcm-server": ("DLCM server base url.", str),
-        }
+        return [ProviderParameter("dlcm-server", description="DLCM server base url.", type=str)]
```
Panaetius, line 54: Should we add some aliases here?

Panaetius (approved):
I added some comments, feel free to disregard.

Love the changes you did to the export provider API, the lock file and moving namespaces to more sane directories!

PR #2949, opened 2022-06-15
feat(core): allow partial updates on dataset and project edit

### LINKED ISSUE #2938: Change dataset and project edit to allow not changing values and properly unset values
There are two problems in the `dataset.edit` and `project.edit` endpoint in the service:

1. When not wanting to update a value, the original value has to be passed instead of just not passing the value
2. When wanting to unset a value, passing `null` is ignored instead

Case 1 happens for dataset `images` whereas case 2 happens for `keywords`, `creators` and `description` (both on projects and datasets).

Case 2 happens because we just ignore empty values passed for those fields, and we should not ignore those, instead unsetting the value if we get a null value. Fixing this will cause these fields to also have the problem in Case 1.

For Case 1, we should change the schema so that sending a parameter with null means unsetting, but not sending the parameter should leave the parameter untouched. I.e., changing `self.ctx.get('images')` to `if "images" in self.ctx: [...]` and having a fixed value (probably `UNMODIFIED = object()`) that we pass if a parameter is not to be changed.

On the service, this should be straight forward. For the cli, we might need additional flags to support this behavior.

An example, assume current keywords on a dataset are `['a', 'b']`, then 
- Not sending `keywords` in the request leaves them unchanged
- Sending `keywords=['c']` removes `a` and `b` and sets `c`
- Sending `keywords=['a','b','c']` appends `c` to the keywords
- Sending `keywords=None` deletes all keywords

All of the fields mentioned should behave this way.

Panaetius:
closes #2938

### DIFF
--- a/tests/service/views/test_dataset_views.py
+++ b/tests/service/views/test_dataset_views.py
@@ -791,7 +791,7 @@ def test_create_and_list_datasets_view(svc_client_with_repo):
 @pytest.mark.integration
 @retry_failed
 def test_list_dataset_files(svc_client_with_repo):
-    """Check listing of dataset files"""
+    """Check listing of dataset files."""
     svc_client, headers, project_id, _ = svc_client_with_repo
 
     file_name = uuid.uuid4().hex
@@ -1272,7 +1272,6 @@ def test_edit_datasets_view_unset_values(svc_client_with_repo):
     }
 
     response = svc_client.post("/datasets.create", data=json.dumps(payload), headers=headers)
-
     assert_rpc_response(response)
     assert {"name", "remote_branch"} == set(response.json["result"].keys())
     assert payload["name"] == response.json["result"]["name"]
@@ -1287,8 +1286,6 @@ def test_edit_datasets_view_unset_values(svc_client_with_repo):
     edit_payload = {
         "project_id": project_id,
         "name": name,
-        "creators": None,
-        "description": None,
         "keywords": None,
         "images": None,
         "custom_metadata": None,
@@ -1297,13 +1294,9 @@ def test_edit_datasets_view_unset_values(svc_client_with_repo):
 
     assert_rpc_response(response)
     assert {"warnings", "edited", "remote_branch"} == set(response.json["result"])
-    assert {
-        "keywords": [],
-        "creators": None,
-        "description": None,
-        "custom_metadata": None,
-        "images": [],
-    } == response.json["result"]["edited"]
+    assert {"keywords": [], "custom_metadata": None, "images": [],} == response.json[
+        "result"
+    ]["edited"]
 
     params_list = {
         "project_id": project_id,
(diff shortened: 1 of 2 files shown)


renku/ui/cli/dataset.py
```diff
@@ -627,40 +630,84 @@ def create(name, title, description, creators, metadata, keyword):
...
+    images = NO_VALUE
 
-    custom_metadata = None
+    if list(creators) == [NO_VALUE]:
+        creators = NO_VALUE
+
+    if list(keywords) == [NO_VALUE]:
+        keywords = NO_VALUE
+
+    if "c" in unset or "creators" in unset:
```
mohammad-sdsc, line 682: `unset` doesn't accept creators.

renku/ui/cli/project.py
```diff
@@ -55,6 +55,9 @@
 | -m, --metadata    | Path to json file containing custom metadata to be   |
 |                   | added to the project knowledge graph.                |
 +-------------------+------------------------------------------------------+
+| -u, --unset <type>| Remove a value from a project. One of ``creators``,  |
```
mohammad-sdsc, line 58: `unset` shouldn't accept creators (it doesn't in the code above). Since we need a creator for a dataset and we shouldn't allow unsetting it.

renku/ui/service/serializers/project.py
```diff
@@ -62,10 +62,10 @@ class ProjectShowResponseRPC(RenkuSyncSchema):
 class ProjectEditRequest(AsyncSchema, LocalRepositorySchema, RemoteRepositorySchema, MigrateSchema):
     """Project edit metadata request."""
 
-    description = fields.String(default=None, description="New description for the project")
+    description = fields.String(description="New description for the project")
     creator = fields.Nested(DatasetCreators, description="New creator for the project")
-    custom_metadata = fields.Dict(default=None, description="Custom JSON-LD metadata")
-    keywords = fields.List(fields.String(), default=None, Missing=None, description="Keyword(s) for the project")
+    custom_metadata = fields.Dict(allow_none=True, description="New custom JSON-LD metadata")
+    keywords = fields.List(fields.String(), allow_none=True, description="New keyword(s) for the project")
```
mohammad-sdsc, line 68: Why removing `missing=None`? Isn't it a bit convenient not to pass it instead of setting it to `None`?

mohammad-sdsc (commented):

renku/ui/cli/dataset.py
Panaetius: Right, I had that initially but then decided against allowing no creators

renku/ui/service/serializers/project.py
Panaetius: Not passing = `NO_VALUE` = Leave it unchanged
Passing None = Unset the value.

With `missing` it would always set it to None and we couldn't differentiate those two cases.

PR #2950, opened 2022-06-15
feat(dataset): filter ls-files by tag

### LINKED ISSUE #2918: Add --tag option to renku dataset ls-files
Currently, `renku dataset ls-files` only lists files for the current version of a dataset.

It should be extended to show files for a specific tag with a `--tag` option.

mohammad-sdsc:
# Description

Allows specifying a tag when running `renku dataset ls-files` that limits the results to those files that are in the given tag.

Fixes #2918

### DIFF
--- a/renku/core/dataset/dataset.py
+++ b/renku/core/dataset/dataset.py
@@ -243,10 +243,11 @@ def edit_dataset(
     return updated
 
 
-@inject.autoparams()
+@inject.autoparams("client_dispatcher")
 def list_dataset_files(
     client_dispatcher: IClientDispatcher,
-    datasets=None,
+    datasets: List[str] = None,
+    tag: Optional[str] = None,
     creators=None,
     include=None,
     exclude=None,
@@ -255,7 +256,8 @@ def list_dataset_files(
 
     Args:
         client_dispatcher(IClientDispatcher): Injected client dispatcher.
-        datasets: Datasets to list files for (Default value = None).
+        datasets(List[str]): Datasets to list files for (Default value = None).
+        tag(str): Tag to filter by (Default value = None).
         creators: Creators to filter by (Default value = None).
         include: Include filters for file paths (Default value = None).
         exclude: Exclude filters for file paths (Default value = None).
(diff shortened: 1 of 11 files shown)


renku/core/dataset/dataset.py
```diff
@@ -1154,16 +1158,33 @@ def filter_dataset_files(
...
-        exclude: Exclude files matching file pattern. (Default value = None).
-        ignore: Ignored datasets. (Default value = None).
-        immutable: Return immutable copies of dataset objects. (Default value = False).
+        names: Filter by specified dataset names (Default value = None).
+        tag(Optional[str]): Filter by specified tag (Default value = None).
+        creators: Filter by creators (Default value = None).
+        include: Tuple containing patterns to which include from result (Default value = None).
+        exclude: Tuple containing patterns to which exclude from result (Default value = None).
+        ignore: Ignored datasets (Default value = None).
+        immutable: Return immutable copies of dataset objects (Default value = False).
```
Panaetius, line 1167: Should we add types to other parameters as well?

renku/core/dataset/dataset.py
```diff
@@ -1174,54 +1195,40 @@ def filter_dataset_files(
 
     records = []
     unused_names = set(names)
+
     for dataset in dataset_gateway.get_all_active_datasets():
+        if (names and dataset.name not in names) or (ignore and dataset.name in ignore):
+            continue
+
+        if tag:
+            tags = dataset_gateway.get_all_tags(dataset)
```
Panaetius, line 1204: might be a bit late to realize this, but maybe we should have a `get_dataset_by_tag` method on the gateway? I think this pattern here sprung up in a couple of places

renku/ui/cli/dataset.py
```diff
@@ -498,6 +498,11 @@
     my-dataset          2020-02-28 16:49:02  data/my-dataset/weather/file1  *
     my-dataset          2020-02-28 16:49:02  data/my-dataset/weather/file2  *
 
+Dataset files can be listed for a specific version (tag) of a dataset using the
+``--tag`` option. In this case, file from datasets that have that specific tag are
```
Panaetius, line 502:
```suggestion
``--tag`` option. In this case, files from datasets that have that specific tag are
```

Panaetius (changes requested):
Looks good. Some minor comments/improvement suggestions.

renku/core/dataset/dataset.py
mohammad-sdsc: I've added it to `core/dataset/tag.py` since we use gateway to interface with the database.

Panaetius (approved):
Thank you!

PR #2952, opened 2022-06-16
chore(chart): run nginx as non-root

olevski:
This is messing with the BIT deployment. I don't see why we had to give special capabilities to nginx but @Panaetius  maybe there was a specific reason for this?

I deployed this and everything seems to be fine. The nginx server starts normally.

### DIFF
--- a/helm-chart/renku-core/templates/deployment-nginx.yaml
+++ b/helm-chart/renku-core/templates/deployment-nginx.yaml
@@ -20,13 +20,6 @@ spec:
           imagePullPolicy: IfNotPresent
           securityContext:
             allowPrivilegeEscalation: false
-            capabilities:
-              drop:
-                - ALL
-              add:
-                - chown
-                - setgid
-                - setuid
             runAsNonRoot: true
           ports:
           - containerPort: 8080


Panaetius:
@mohammad-sdsc you added this, do you remember?

mohammad-sdsc:
iirc the service starts but it won't work and that's why I had to give these special permissions. I believe I've got these permissions from errors messages given by nginx. Maybe try to open a project in the deployment and see if it works.

olevski:
>  Maybe try to open a project in the deployment and see if it works.

@mohammad-sdsc I did not see any errors in the nginx logs. Also I am not sure what you mean by this? Is creating a new project and starting a session from it sufficient to test this out? Or should I do more to confirm everything works correctly?

mohammad-sdsc:
> I did not see any errors in the nginx logs. Also I am not sure what you mean by this? Is creating a new project and starting a session from it sufficient to test this out? Or should I do more to confirm everything works correctly?

I'm not sure how I got to the errors but I didn't do much. So, just open an existing project and add a dataset and also create a new one and open a session. This should be enough to get the errors.

olevski:
I can confirm that running this on personal deployment causes no issues. I can successfully create projects, datasets and launch sessions. I also checked the logs of the nginx server and there are no errors there.

mohammad-sdsc (approved):
Thank you!

PR #2958, opened 2022-06-20
fix(service): correctly use passed namespace in project id generation

### LINKED ISSUE #2937: fix project id for projects in subgroups
For projects in subgroups in gitlab e.g. `group/subgroup/myproject`, the id created in our metadata is `host/projects/owner/myproject`, which is wrong.

To fix this, https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/core/management/repository.py#L261 should allow passing a namespace, which should override what we get on https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/domain_model/project.py#L70

The reason this fails is that, at the time when the core-service creates a project, there isn't a git remote yet, so the namespace can't be guessed from the remote, falling back to the user name. We do get the `namespace` field in the service request though, so we can set it directly.

We also need a `renku doctor` command to fix this on existing projects.

We need to coordinate with KG on how to deal with this, as projects suddenly changing id's might be an issue for them.

Panaetius:
closes #2937 

Fixes id generation for newly created projects.

For existing projects, `renku doctor --fix` will fix the id. Checked with KG and it shouldn't affect them, as they build the ID themselves based on gitlab.

### DIFF
--- a/renku/domain_model/project.py
+++ b/renku/domain_model/project.py
@@ -126,13 +126,13 @@ class Project(persistent.Persistent):
             annotations = [Annotation(id=Annotation.generate_id(), body=custom_metadata, source="renku")]
 
         if creator is None:
-            raise ValueError("Project Creator not set")
+            raise errors.ParameterError("Project Creator not set", "creator")
 
         if name is None:
-            raise ValueError("Project 'name' not set and could not be generated")
+            raise errors.ParameterError("Project 'name' not set and could not be generated", "name")
 
         if namespace is None:
-            raise ValueError("Project 'namespace' not set and could not be generated")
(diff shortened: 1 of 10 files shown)


renku/domain_model/project.py
```diff
@@ -115,20 +131,24 @@ def from_client(
         if name is None:
             raise ValueError("Project 'name' not set and could not be generated")
 
+        if namespace is None:
+            raise ValueError("Project 'namespace' not set and could not be generated")
```
mohammad-sdsc, line 135: I know that this was in the code before but I don't see anywhere if we catch `ValueError`s. Should we use a Renku exception instead? It also applies to other places in this method.

mohammad-sdsc (approved):
Thank you!

Issue #2945, opened 2022-06-13
bump git lfs to version 3.x in Dockerfile and Github Actions


Panaetius:
We should theoretically work with git lfs 3.x again, so we should bump the version in our Docker file and the Github Actions (for Macos) and then seriously test that everything still works (including a test deployment with the service image)

PR #2985, opened 2022-07-01
ignore: triggering tests

olevski:
I am not really sure that is is enough to make sure the tests run with git LFS version 3.2.0 (i..e the latest)

@Panaetius do you know how I can check or make sure that this is the case?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM python:3.9-slim as base
 
 # hadolint ignore=DL3008,DL3009,DL3013
 RUN apt-get update && \
-    apt-get install --no-install-recommends -y git git-lfs=2.* python3-dev && \
+    apt-get install --no-install-recommends -y git git-lfs=2.3.0 python3-dev && \
     pip install --no-cache-dir --upgrade pip poetry
 
 FROM base as builder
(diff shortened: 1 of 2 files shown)


Panaetius:
Just append `&& git lfs --version` and check what it prints?

Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM python:3.9-slim as base
 
 # hadolint ignore=DL3008,DL3009,DL3013
 RUN apt-get update && \
-    apt-get install --no-install-recommends -y git git-lfs=2.* python3-dev && \
+    apt-get install --no-install-recommends -y git git-lfs=2.3.0 python3-dev && \
```
Panaetius, line 5:
```suggestion
    apt-get install --no-install-recommends -y git git-lfs=3.2.0 python3-dev && \
```
I think?

Panaetius:
We don't use the Dockerfile in our tests, so you need to change all https://github.com/SwissDataScienceCenter/renku-python/blob/develop/.github/workflows/test_deploy.yml#L532 lines as well. I think the linux tests default to whatever comes from github, which should be a new version

Panaetius:
By the way, closes https://github.com/SwissDataScienceCenter/renku-python/issues/2945 ?

olevski:
This is not needed anymore. It was only used to trigger the tests.

Issue #2939, opened 2022-06-08
cli: renku session start


lokijuhy:
## Describe the bug
I tried to use `renku session start` on my [GeeMap evaluation](https://renkulab.io/projects/laura.kinkead1/geemap-evaluation) project, and it errored with "Ahhhh you have found a bug". 

```
❯ git clone git@renkulab.io:laura.kinkead1/geemap-evaluation.git
Cloning into 'geemap-evaluation'...
remote: Enumerating objects: 91, done.
remote: Counting objects: 100% (31/31), done.
remote: Compressing objects: 100% (23/23), done.
remote: Total 91 (delta 13), reused 20 (delta 8), pack-reused 60
Receiving objects: 100% (91/91), 24.39 KiB | 2.71 MiB/s, done.
Resolving deltas: 100% (24/24), done.
❯ cd geemap-evaluation
❯ renku session start
Ahhhhhhhh! You have found a bug. 🐞

1. Open an issue by typing "open";
2. Print human-readable information by typing "print";
3. See the full traceback without submitting details (default: "ignore").

Please select an action by typing its name (open, print, ignore) [ignore]: open
```

## Details
I can successfully run `renku session start` on the [Renku-stories/geemap](https://renkulab.io/projects/renku-stories/geemap) project I have cloned in the same directory.

Ralf suspects this is caused by the `.` in the project's namespace.

**Renku version:** 1.3.1
**OS:** Darwin (Darwin Kernel Version 21.4.0: Fri Mar 18 00:46:32 PDT 2022; root:xnu-8020.101.4~15/RELEASE_ARM64_T6000)
**Python:** 3.9.12
**Docker Desktop**: 4.8.2

### Traceback

```
Traceback (most recent call last):
  File "[...]/site-packages/docker/api/client.py", line 268, in _raise_for_status
    response.raise_for_status()
  File "[...]/site-packages/requests/models.py", line 960, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http+docker://localhost/v1.41/images/laura.kinkead1/geemap-evaluation:56c4aa5/json

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "[...]/site-packages/docker/utils/decorators.py", line 19, in wrapped
    return f(self, resource_id, *args, **kwargs)
  File "[...]/site-packages/docker/api/image.py", line 251, in inspect_image
[...]

PR #2991, opened 2022-07-05
feat(core): use existing remote image when starting sessions

### LINKED ISSUE #2934: feat: use existing remote image on local session launch
If an image has been built on the remote renkulab's CI, we should not need to build it again locally. Instead, we should just launch with the remote image. #2880 is already doing some image checks on the remote registry so we should be able to reuse some of the same logic. This would save a lot of time for users wanting to replicate cloud sessions on local machines. Caveat: not sure what to do about authentication with the docker registry for private projects.

mohammad-sdsc:
# Description

https://github.com/SwissDataScienceCenter/renku-python/pull/2880 already pulls a remote image if it exists when a local image is not found. However, it wasn't clear that it's doing this. This PR adds proper progress messages. It also solves some issues with stoping sessions and for cases where a session is starting/stoping/... for projects with no remote.

It also renames `session list` to `session ls` to be consistent with other commands.

Fixes #2934


Panaetius (approved):
Well, I went through this two times and still can't find anything to nitpick... Well done!

Let me know if there's some further changes needed to address #2939 , toherwise ready to merge 🚀

mohammad-sdsc:
> Let me know if there's some further changes needed to address #2939 , otherwise ready to merge 🚀

I've checked that issue and it seems that it's fixed by this PR. I wrote a comment there but left it open so that Laura can check it.

PR #2993, opened 2022-07-05
fix: prevent deletion of plans that are still used in composite plans

### LINKED ISSUE #2968: Removing workflows that are used in composite plans
When removing a workflow using `renku workflow remove`, there is no check if this workflow was used in composite plans. So, it's possible to have composite plans that are using deleted plans.

**Acceptance Criteria**
- `workflow remove` should fail if the plan is used in composite plans and should list such composite plans
- Add a `--force` flag to allow deleting such plans
- To be discussed: Add `--recursive` flag to delete all composite plans that use this plan

Panaetius:
closes #2968 

I didn't implement the `--recursive`, for that we'd have to recursively get containing plans (as a composite can contain a composite), warn the user about all of them and then remove them, which I think has a high chance of doing more than a user wants and makes the command more complicated, plus probably very rarely needed. We can add it if there is actual user need for it in the future


renku/core/workflow/plan.py
```diff
@@ -132,3 +145,24 @@ def is_plan_removed(plan: AbstractPlan) -> bool:
...
+
+    derivatives = {p.id for p in get_derivative_chain(plan=plan)}
+
+    all_plans = plan_gateway.get_all_plans()
+
+    child_plan: Optional[AbstractPlan] = plan
+    while child_plan is not None:
+        derivatives.add(child_plan.id)
+        plan = child_plan
+        child_plan = next((p for p in all_plans if p.derived_from == plan.id), None)
```
mohammad-sdsc, line 162: This `while` block is not needed since `get_derivative_chain` already returned all parents/children of the plan.

renku/core/workflow/plan.py
```diff
@@ -83,6 +84,18 @@ def remove_plan(name_or_id: str, force: bool, plan_gateway: IPlanGateway, when:
     if latest_version.deleted:
         raise errors.ParameterError(f"The specified workflow '{name_or_id}' is already deleted.")
 
+    composites_containing_child = get_composite_plans_by_child(plan)
+
+    composite_names = "\n\t".join([c.name for c in composites_containing_child])
+
+    if composites_containing_child and not force:
```
mohammad-sdsc, line 91: This can be simplified by using nested ifs:
```
if composites_containing_child:
    if force:
        communication.warn(f"Removing '{name_or_id}', which is still used in these workflows:\n\t{composite_names}")
    else:
    ....
```

mohammad-sdsc (commented):

mohammad-sdsc (approved):
Thank you!

PR #2999, opened 2022-07-06
docs: update commands

lorenzo-cavazzi:
"Renku command line" reference page:
* order commands alphabetically
* add missing commands 

Output of `renku --help`
* add missing commands


Panaetius (approved):
🔡

PR #3000, opened 2022-07-06
feat(cli): inform users about 'renku login' in related errors

### LINKED ISSUE #2964: Suggest login when renku dataset not found
Users may need to import their own private datasets in other projects.
This is only possible in the CLI if they ran `renku login` before.

**Context**

Currently, the CLI does not guide the user towards this solution.
When attempting to import a private dataset, it raises `Error: Resource not found in knowledge graph: <dataset-url>`.

**Objective**

Provide a hint to the user that they need to `renku login` before importing private datasets.

Panaetius:
closes #2964


mohammad-sdsc (approved):
Thank you!

PR #3001, opened 2022-07-06
refactor(dataset): dataset add/import refactoring

### LINKED ISSUE #2943: refactor dataset import/add code
The code for importing datasets or adding files to datasets has grown organically for a while and is becoming a bit unwieldy.

We should refactor it with dedicated classes for different strategies (Import from Git, Import from Provider, etc.) to make it easier to understand and clean it up

mohammad-sdsc:
Refactors dataset add for various cases (git, local, ...) to their own provider classes. Does the same thing for import providers by creating common methods for them.

Fixes #2943 

## TODO
Structural changes:
- [x] Move `ProviderFactory` to its own file
- [x] Restructure all providers' files to have similar order of classes
~- [ ] Rename `local.py` to `filesystem.py`~


renku/core/dataset/providers/web.py
```diff
@@ -0,0 +1,196 @@
...
+    uri = _provider_check(uri)
+
+    try:
+        start = time.time() * 1e3
+
+        tmp_root, paths = requests.download_file(
+            base_directory=client.renku_path / CACHE, url=uri, filename=filename, extract=extract
+        )
+
+        exec_time = (time.time() * 1e3 - start) // 1e3
```
Panaetius, line 123: I feel just using `round()` or `floor()` instead of the `1e3` code is cleaner

tests/cli/test_datasets.py
```diff
@@ -638,6 +638,7 @@ def test_usage_error_in_add_from_url(runner, client, params, message):
     assert message in result.output
 
 
+@pytest.mark.skip(reason="This warning is not shown anymore")
```
Panaetius, line 641: Should we just remove the test?

tests/cli/test_integration_datasets.py
```diff
@@ -189,14 +189,15 @@ def test_dataset_import_real_doi_warnings(runner, project, sleep_after):
 
     result = runner.invoke(cli, ["dataset", "ls"])
     assert 0 == result.exit_code, format_result_exception(result) + str(result.stderr_bytes)
-    assert "pyndl_naive_discr_v0.8.1" in result.output
+    assert "pyndl_naive_discr_v0.6.0" in result.output
```
Panaetius, line 192: I fixed this in one of my PRs as well, but I used 0.8.2 as that's the newest one

tests/core/commands/test_cli.py
```diff
@@ -57,10 +57,8 @@ def test_version_duration(run_shell):
         assert output[1] is None
         total_duration += duration
     total_duration /= 5
-    if sys.platform == "darwin":
-        assert total_duration < 2.0
-    else:
-        assert total_duration < 1.5
+
+    assert total_duration < 2.0
```
Panaetius, line 61: I removed this in my other PR

Panaetius (changes requested):

renku/core/dataset/dataset.py
```diff
@@ -441,154 +426,89 @@ def import_dataset(
...
+        is_latest_version = importer.is_latest_version()
     except KeyError as e:
         raise errors.ParameterError(f"Could not process '{uri}'.\nUnable to fetch metadata: {e}")
     except LookupError as e:
         raise errors.ParameterError(f"Could not process '{uri}'.\nReason: {e}")
 
-    if not files:
+    if not importer.provider_dataset_files:
         raise errors.ParameterError(f"Dataset '{uri}' has no files.")
+    if not is_latest_version:
```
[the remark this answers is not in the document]
mohammad-sdsc: The reason was that this check can raise exceptions in some providers. I moved the checks to the `try` block.

renku/core/dataset/providers/web.py
mohammad-sdsc: I've changed this rate-limit logic a bit.

tests/cli/test_integration_datasets.py
mohammad-sdsc: I'll resolve this once merging develop

renku/core/dataset/dataset.py
```diff
@@ -472,17 +471,16 @@ def remove_files(dataset):
     try:
         importer = provider.get_importer(uri, gitlab_token=gitlab_token, **kwargs)
         provider_dataset: ProviderDataset = importer.fetch_provider_dataset()
-        is_latest_version = importer.is_latest_version()
+
+        if not importer.provider_dataset_files:
+            raise errors.ParameterError(f"Dataset '{uri}' has no files.")
+        if not importer.is_latest_version():
+            communication.warn(f"Newer version found at {importer.latest_uri}")
```
Panaetius, line 478: I just remembered I always found this message a bit odd. It just informs the user that there is a new version that could be downloaded instead of the one they're currently downloading. But the way it's worded, it always left me confused, especially because the confirm_download method will print `Do you wish to download this version?`

So
```
Newer version found at .....
Do you wish to download this version?
```
implies "Do you want to download the new version instead of the one you're trying to download?" even though that's not what it does, it's purely a hint that there is a newer version. We should change that

Panaetius (approved):

renku/core/dataset/dataset.py
mohammad-sdsc: I moved the new version warning after the download confirmation. This makes it clear that which version will be downloaded.

Panaetius (approved):

PR #3004, opened 2022-07-11
chore: fix coverage paths

Panaetius:
Fixes coverage paths so files in ./renku/ and [...]/site-packages/renku/ are treated as the same.

Also changes the action to always run.

https://coveralls.io/builds/50753472


mohammad-sdsc (approved):
Thank you!

PR #3017, opened 2022-07-13
test: add tests for session, service and graph export

### LINKED ISSUE #2064: Fix renku service up tests
All the tests in `tests/cli/test_service.py` are currently broken and don't actually test what they seem to test.

The `renku service up` command depends on redis running somewhere on the machine. Without redis, the daemon will error out after ~1-2 seconds. The tests just randomly pass because they check if the daemon processes are running withing this 1-2s window when they are running.

We should write proper tests for this:
- Find a way to not have them error out without redis --> not sure how we could mock redis for a Popen subprocess
- Set `--runtime-dir` to a random directory on each execution of a test, check the renku.err file in that directory to make sure there's no error

### LINKED ISSUE #2174: Add more tests for `renku service` commands
Currently coverage of https://coveralls.io/builds/40593871/source?filename=renku%2Fcli%2Fservice.py is abysmal. We should add more tests to test the individual commands, especially since some of them are used to run the core service in production, making them critical commands.

### LINKED ISSUE #3005: Add tests for `renku session`, `renku service` and `renku graph export`
As seen in https://coveralls.io/builds/50753472 at the bottom, `renku session`, `renku service` and `renku graph export --format [dot|dot-landscape|dot-debug]` are the biggest offenders in not being covered by tests and we should add tests that cover the relevant pieces of code.

Also remove `test_version_duration` while we're at it, as it is to volatile on github actions and not really useful.

Panaetius:
closes #3005 
closes #2174 
closes #2064

### DIFF
--- a/tests/core/commands/test_graph.py
+++ b/tests/core/commands/test_graph.py
@@ -1,0 +1,534 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2017-2022 - Swiss Data Science Center (SDSC)
+# A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
+# Eidgenössische Technische Hochschule Zürich (ETHZ).
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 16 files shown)


tests/core/commands/test_graph.py
```diff
@@ -0,0 +1,534 @@
...
+from renku.domain_model.provenance.activity import Activity, Association
+from renku.domain_model.provenance.agent import Person
+from renku.domain_model.workflow.composite_plan import CompositePlan
+from renku.domain_model.workflow.plan import Plan
+from renku.infrastructure.database import Index
+
+
+@pytest.fixture()
+def input_objects(request, dataset_model, activity_model, plan_model):
+    """Use fixtures as input for paramtrize.
```
mohammad-sdsc, line 48: paramtrize -> parametrize

mohammad-sdsc (commented):
Thank you! Literally a single typo only.

mohammad-sdsc (approved):

PR #3025, opened 2022-07-15
feat(cli): option to skip metadata update

### LINKED ISSUE #3020: Skip metadata update and commit in `renku run` and `workflow execute`
In some cases (especially when using the argo provider) it is beneficial to not update the metadata in the local project when doing `renku workflow execute`.

In the case of argo this is useful because the metadata it fully changed and updated (and pushed to the repo) in the argo container and not in the local clone of the repo.

olevski:
This is a feature that can be useful in cases where the metadata update can happen elsewhere (i.e. argo workflow) or simply should not be committed.

For example running tests or troubleshooting things should not be tracked.

I was not sure where to exactly put the logic. I can move it up or down the chain of depedencies. I.e. we can have a whole new function called `run_without_metadata_update` instead of passing the skip metadata flag in the current function.

closes #3020


renku/command/run.py
```diff
@@ -37,9 +37,14 @@
 from renku.domain_model.provenance.activity import Activity
 
 
-def run_command():
+def run_command(skip_metadata_update: bool):
     """Tracking work on a specific problem."""
-    return Command().command(_run_command).require_migration().require_clean().with_database(write=True).with_commit()
+    command = Command().command(_run_command).require_migration().require_clean()
```
Panaetius, line 42: I'm a bit unsure about adding it for `renku run`. Since this does not allow using a backend, not updating metadata and not creating a commit is functionally the same as not running with `renku run`. Whereas for the other that execute an already recorded command, there's no alternative way to do that without committing .

Panaetius (changes requested):
This is nice, thank you!

We should probably add it for other execution commands as well (renku update/renku rerun/renku workflow iterate)

renku/command/run.py
olevski: Ah yeah that is a good point. I will remove it for `renku run` and add it for the other commands you mention.

Panaetius:
oh also, documentation of the parameter in the docstring of the files at the top of the renku.ui.cli.* files and we maybe should add a shorthand flag for the `--skip-metadata-update` flag (`-s`?) as it's a bit of a mouthful to write

Panaetius (approved):
Thank you!

olevski:
@Panaetius I am sorry for making you re-approve this. I had some issues with the style checks. My pre-commit hooks are broken it seems. Or that specific style check is not enforced by the pre-commit tests.

tests/cli/test_workflow.py
```diff
@@ -1253,7 +1252,7 @@ def test_workflow_cycle_detection(run_shell, project, capsys, client):
     assert b"Cycles detected in execution graph" in result[0]
 
 
-@pytest.mark.skipif(sys.platform == "darwin", reason="GitHub macOS image doesn't include Docker")
+# @pytest.mark.skipif(sys.platform == "darwin", reason="GitHub macOS image doesn't include Docker")
```
Panaetius, line 1255: I'm guessing this didn't work, looking at the failed test?

olevski: Oh my bad I did that so that it runs on my computer. I meant to put it back the way it was.

Panaetius (approved):

Panaetius (approved):

PR #3027, opened 2022-07-18
feat(core): allow per dataset data directory

### LINKED ISSUE #2977: Change `datadir` to be per dataset instead of per project
Currently, we allow setting `--data-dir` on a per-project basis. We should keep this to allow setting a default datadir, but then also add a `--data-dir` flag to `renku dataset create` to override this for a dataset. 

The default should still be `<project_data_dir>/<dataset_name>`.

We do not need to expose this property in JSON-LD, tracking it in internal metadata should be enough.

Make sure that dataset import still works with this.

Panaetius:
closes #2977 

Adds a `--datadir` option to `renku dataset create`, `renku dataset add --create` and `renku dataset import`)


tests/cli/test_integration_datasets.py
```diff
@@ -396,15 +396,21 @@ def test_dataset_import_renkulab_dataset_with_image(runner, project, client, cli
 @pytest.mark.integration
 @retry_failed
 @pytest.mark.vcr
-def test_import_renku_dataset_preserves_directory_hierarchy(runner, project, client, load_dataset_with_injection):
+@pytest.mark.parametrize(
+    "datadir_option,expected_datadir", [([], Path(DATA_DIR) / "remote"), (["--datadir", "mydir"], Path("mydir"))]
```
olevski, line 400: How does `DATA_DIR` get its value here? Is it a fixture?

renku/core/dataset/providers/local.py
```diff
@@ -84,21 +84,20 @@ def add(
...
+        dataset: Optional["Dataset"] = None,
         external: bool = False,
         **kwargs,
     ) -> List["DatasetAddMetadata"]:
         """Add files from a URI to a dataset."""
         from renku.core.dataset.providers.models import DatasetAddAction, DatasetAddMetadata
-        from renku.domain_model.dataset import get_dataset_data_dir
 
-        assert dataset_name is not None, "Dataset name is not passed"
+        assert dataset is not None, "Dataset name is not passed"
```
olevski, line 94: Should this raise an error instead?

olevski (dismissed):
I see that that `assert` statement was there already. So there is perhaps a good reason to leave it in. I just wanted to point it out. Everything else looks good.

renku/core/dataset/providers/renku.py
```diff
@@ -514,7 +514,7 @@ def _fetch_dataset(self, client_dispatcher: IClientDispatcher, database_dispatch
             database_dispatcher.pop_database()
             client_dispatcher.pop_client()
 
-        provider_dataset.data_dir = get_dataset_data_dir(self._remote_client, provider_dataset.name)
+        provider_dataset.data_dir = provider_dataset.get_datadir(self._remote_client)
```
mohammad-sdsc, line 517: We don't need this anymore; it was set so that we have the dataset's datadir in the model. Now that we have it we can directly call `dataset.get_datadir()` instead of this.

renku/domain_model/dataset.py
```diff
@@ -380,6 +381,7 @@ def __init__(
         self.title: Optional[str] = title
         self.version: Optional[str] = version
         self.annotations: List["Annotation"] = annotations or []
+        self._datadir: Optional[str] = str(datadir) if datadir else None
```
mohammad-sdsc, line 384: I suggest that we use the same pattern for new attributes that we add to a Persistent model. Before, we've added the attribute to the class and then set it for individual instances if it had a value. We should do the same here. Also, we can call it `datadir` instead of `_datadir`.

renku/domain_model/dataset.py
```diff
@@ -420,6 +422,13 @@ def keywords_csv(self):
         """Comma-separated list of keywords associated with dataset."""
         return ", ".join(self.keywords)
 
+    def get_datadir(self, client) -> Path:
```
mohammad-sdsc, line 425: Should we use injection here? Or maybe use `get_client()`? It's not very clear to me where we should use it and I believe we don't do it consistently.

renku/domain_model/dataset.py
```diff
@@ -664,14 +673,9 @@ class ImageObjectRequestJson(marshmallow.Schema):
...
-def get_dataset_data_dir(client, dataset_name: str) -> str:
-    """Return default data directory for a dataset."""
-    return os.path.join(client.data_dir, dataset_name)
-
-
 def get_file_path_in_dataset(client, dataset: Dataset, dataset_file: DatasetFile) -> Path:
     """Return path of a file relative to dataset's data dir."""
     try:
-        return (client.path / dataset_file.entity.path).relative_to(get_dataset_data_dir(client, dataset.name))
+        return (client.path / dataset_file.entity.path).relative_to(dataset.get_datadir(client))
```
mohammad-sdsc, line 679: I've just realized this function won't work correctly (no unit tests 😭) since `get_dataset_data_dir` doesn't return an absolute path. It should be like:
```
return (client.path / dataset_file.entity.path).relative_to(client.path / dataset.get_datadir(client))
```

mohammad-sdsc (changes requested):
Thanks Ralf! This looks great! Please see my comments.

renku/domain_model/dataset.py
Panaetius: yes I had it as a property first, then this made sense. I'll change it. :+1:

tests/cli/test_integration_datasets.py
Panaetius: It's imported at the top `from renku.core.management.repository import DEFAULT_DATA_DIR as DATA_DIR`

mohammad-sdsc:
We should also display dataset's data dir in `renku dataset ls` and `renku dataset show`.

tests/cli/test_integration_datasets.py
olevski: 🤦 my bad I completely missed that. I thought I fully expanded the diffs but I guess I did not.

Panaetius:
Don't merge until a new release is made

mohammad-sdsc (approved):
Tests are failing due to minimum version check. OK otherwise.

mohammad-sdsc (approved):
🚀

PR #3029, opened 2022-07-18
feat: add dot output for worflow visualize command

### LINKED ISSUE #2376: add option to produce dot output in `renku workflow visualize`
https://github.com/SwissDataScienceCenter/renku-python/pull/2372 introduced the `renku workflow visualize` command that produces Sugiyama formatted activity graphs. We should extend the command to allow outputting graphs in the dot/Graphviz format instead of ascii rendering to the console. We'd still want the Sugiyama layout information about node positions, if possible, so the graph looks the same, just in dot format.

lorenzo-cavazzi:
_Still a draft_

Add the option to output the graph from `worflow visualize` in `dot` format 

fix #2376

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -159,11 +159,11 @@ wcwidth = ">=0.1.4"
 
 [[package]]
 name = "blinker"
-version = "1.4"
+version = "1.5"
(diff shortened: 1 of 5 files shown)


lorenzo-cavazzi:
superseded by #3032

PR #3031, opened 2022-07-19
feat: add RDFGraph to API

rokroskar:
Adds an `RDFGraph` object to the API

Usage:

```
from renku.ui.api import RDFGraph

g = RDFGraph()
# get a list of contributors to the project
list(g.subjects(object=URIRef("http://schema.org/Person")))
```


Panaetius (changes requested):
Should probably add it to https://github.com/SwissDataScienceCenter/renku-python/blob/develop/docs/reference/api.rst as well.

Otherwise looks good, thanks!

rokroskar:
Thanks @Panaetius - done

PR #3032, opened 2022-07-19
feat(workflow): add dot output on workflow visualize

### LINKED ISSUE #2376: add option to produce dot output in `renku workflow visualize`

lorenzo-cavazzi:
# Description

Add a `--format` option on the `renku workflow visualize` command. The output can now be either `console` or `dot`.
Maybe instead of `-i` would it make sense to add an extra format? Something like `console`, `interactive`, `dot`?

-----

This also fixes an error when providing non-existing column names

E.G. `renku workflow visualize <path> --columns aaa ` was triggering
```
renku workflow visualize results.txt --columns aaa 
Ahhhhhhhh! You have found a bug. 🐞
[...]
File "/[...]/renku/command/view_model/activity_graph.py", line 113, in <listcomp>
    columns = [ACTIVITY_GRAPH_COLUMNS[c] for c in columns.split(",")]
KeyError: 'aaa'
```

Now it prints 

```
Error: Invalid parameter value for columns 'aaa': you can use any of 'command,id,date,plan'
```

fix #2376

### DIFF
--- a/renku/ui/cli/workflow.py
+++ b/renku/ui/cli/workflow.py
@@ -214,7 +214,7 @@ variable is going to be substituted with the iteration index (0, 1, 2, ...).
             --map output=output_{iter_index}.txt my-run
 
 This would execute ``my-run`` three times, where ``parameter-1`` values would be
-``10``, `20`` and ``30`` and the producing output files ``output_0.txt``,
+``10``, ``20`` and ``30`` and the producing output files ``output_0.txt``,
 ``output_1.txt`` and ``output_2.txt`` files in this order.
 
 In some cases it may be desirable to avoid updating the renku metadata
@@ -300,7 +300,7 @@ The basic usage is:
 
 .. code-block:: console
 
-   $ renku run --name step1-- cp input intermediate
+   $ renku run --name step1 -- cp input intermediate
    $ renku run --name step2 -- cp intermediate output
    $ renku workflow compose my-composed-workflow step1 step2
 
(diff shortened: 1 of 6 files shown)


renku/ui/cli/workflow.py
```diff
@@ -622,6 +622,14 @@
...
 by pressing the <Enter> key.
 
+If you prefer to elaborate the output graph further, or if you wish to export
+it for any reason, you can use the ``--format`` option to specify an output
+format.
+
+.. code-block:: console
+
+   $ renku workflow visualize --format dot <path>
+
```
Panaetius, line 632: Maybe write a sentence that this would print dot format, which could be accepted by the `dot` command

tests/cli/test_workflow.py
```diff
@@ -917,6 +917,24 @@ def test_workflow_visualize_non_interactive(runner, project, client, workflow_gr
...
+    '"Y" -> "bash -c \\"cat X Y | tee R S\\"";' in result.output
+    '"X" -> "bash -c \\"cat X Y | tee R S\\"";' in result.output
+    '"bash -c \\"cat X Y | tee R S\\"" -> "R";' in result.output
+    '"bash -c \\"cat X Y | tee R S\\"" -> "S";' in result.output
+    4 == result.output.count('"bash -c \\"cat X Y | tee R S\\"')
+
+    1 == result.output.count('"echo other > H" -> "H"')
+    1 == result.output.count('-> "H"')
+    0 == result.output.count('"H" -->')
+    1 == result.output.count('"H"')
```
Panaetius, line 935: All of these statements need an `assert` in front of them.

Panaetius (changes requested):
Thank you, this looks a lot simpler than the pygraphviz approach!

Panaetius (approved):
Thank you!

PR #3047, opened 2022-07-24
feat(dataset): dataset creation with s3 storage backend

### LINKED ISSUE #2970: Create dataset with a S3 storage backend
As a Renku user I want to be able to create dataset with a S3 storage backend so that I can load and store data on a S3
bucket.
```
renku dataset create my-dataset --storage s3://server/path/to/bucket
```

## Acceptance Criteria
- Add a `--storage` flag to `dataset create` which accepts a URI
- S3 URI is stored in dataset's metadata
- Dataset's data directory is added to .gitignore
- Renku asks for S3 credentials (if not available) and asks the user whether to store them or not. It should store them in `~/.renku/renku.ini`
- Renku must check if S3 bucket exists and is accessible. The operation must fail otherwise.

## Notes
- When implementing this consider the use-case of accessing files from an external storage but outside of a Renku dataset.

### LINKED ISSUE #2972: Show information about external storage
As a Renku user I want to be able to use `dataset show/ls-files` commands to know external storage sources for datasets and files.

## Acceptance Criteria
- Add a `Data Sources` section to `dataset show` that lists all used external sources in a dataset including its type, URI and location on the filesystem
- Add a `data source` columns to `dataset ls-files` to show the external source for each file

mohammad-sdsc:
# Description

Allows creating datasets with S3 storage backends. Currently, only supports `s3://<bucket-name>/<path>` URI format. Uses `RClone` to interact with S3.

### NOTE:
`RClone` cannot check if a directory exists for storage backends that don't support empty directories like S3. So, it's not possible to fail if the directory doesn't exists as was required in the story.

Fixes #2970 
Fixes #2972 

### TODO
Test with private S3 buckets and valid S3 credentials.

### DIFF
--- a/renku/core/dataset/providers/api.py
+++ b/renku/core/dataset/providers/api.py
@@ -261,6 +261,10 @@ class ProviderCredentials(abc.ABC, UserDict):
         """Return the associated provider instance."""
         return self._provider
 
+    def get_credentials_names_with_no_value(self) -> Tuple[str, ...]:
+        """Return a tuple of credential keys that don't have a valid value."""
+        return tuple(key for key, value in self.items() if value is NO_VALUE)
+
     def get_canonical_credentials_names(self) -> Tuple[str, ...]:
         """Return canonical credentials names that can be used as config keys."""
         return tuple(get_canonical_key(key) for key in self.get_credentials_names())
--- a/renku/core/dataset/providers/api.py
+++ b/renku/core/dataset/providers/api.py
@@ -1,4 +1,4 @@
-# Copyright 2020 - Swiss Data Science Center (SDSC)
+# Copyright 2017-2022 - Swiss Data Science Center (SDSC)
 # A partnership between École Polytechnique Fédérale de Lausanne (EPFL) and
 # Eidgenössische Technische Hochschule Zürich (ETHZ).
 #
@@ -19,10 +19,11 @@ import abc
 from collections import UserDict
 from enum import IntEnum
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 
 from renku.core import errors
 from renku.core.util.metadata import get_canonical_key, read_credentials, store_credentials
+from renku.core.util.util import NO_VALUE, NoValueType
 
 if TYPE_CHECKING:
     from renku.core.dataset.providers.models import (
@@ -53,8 +54,8 @@ class ProviderApi(abc.ABC):
     priority: Optional[ProviderPriority] = None
     name: Optional[str] = None
 
-    def __init__(self, uri: str):
-        self._uri: str = uri
+    def __init__(self, uri: Optional[str]):
+        self._uri: str = uri or ""
 
     def __init_subclass__(cls, **kwargs):
         for required_property in ("priority", "name"):
@@ -239,18 +240,20 @@ class ExporterApi(abc.ABC):
 class ProviderCredentials(abc.ABC, UserDict):
     """Credentials of a provider.
 
-    NOTE: A ``None`` value for a key means that it is not set; however, an empty value, "", means that the key was set.
+    NOTE: An empty string, "", is a valid value. ``NO_VALUE`` means that the value for a key is not set.
     """
 
     def __init__(self, provider: ProviderApi):
         super().__init__()
         self._provider: ProviderApi = provider
-        self.data: Dict[str, Optional[str]] = {key: None for key in self.get_canonical_credentials_names()}
+        self.data: Dict[str, Union[str, NoValueType]] = {
+            key: NO_VALUE for key in self.get_canonical_credentials_names()
+        }
 
     @staticmethod
     @abc.abstractmethod
     def get_credentials_names() -> Tuple[str, ...]:
-        """Return list of the required credentials for a provider."""
+        """Return a tuple of the required credentials for a provider."""
         raise NotImplementedError
 
     @property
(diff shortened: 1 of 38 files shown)


renku/infrastructure/storage/base.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def transform_kwarg(key: str, value: Any) -> List[str]:
+        if value in (False, None):
+            return []
+        else:
+            name = f"-{key}" if len(key) == 1 else f"--{key.replace('_', '-')}"
+            return [name] if value is True else [name, f"{value}"]
+
+    args = [transform_kwarg(k, v) for k, v in kwargs.items()]
+    return [a for arg in args for a in arg]
```
Panaetius, line 83:
```suggestion
    args =[]
    for k, v in kwargs.items():
        args.extend(transform_kwarg(k, v))
    return args
```

not that it matters at all, feel free to ignore, i just find it slightly more readable than the list comprehension in this case.

renku/core/dataset/providers/api.py
```diff
@@ -215,3 +234,54 @@ def get_access_token_url(self) -> str:
     def export(self, **kwargs) -> str:
         """Execute export process."""
         raise NotImplementedError
+
+
+class ProviderCredentials(abc.ABC, UserDict):
+    """Credentials of a provider.
+
+    NOTE: A ``None`` value for a key means that it is not set; however, an empty value, "", means that the key was set.
```
Panaetius, line 242: Together with the other comment on `None` instead of `""`. We do have `renku.core.util.util.NO_VALUE` (And `NoValueType`) for this purpose, which allows showing intent more clearly and also reflects this in the type signature.

renku/core/dataset/providers/api.py
```diff
@@ -215,3 +234,54 @@ def get_access_token_url(self) -> str:
...
+    """
+
+    def __init__(self, provider: ProviderApi):
+        super().__init__()
+        self._provider: ProviderApi = provider
+        self.data: Dict[str, Optional[str]] = {key: None for key in self.get_canonical_credentials_names()}
+
+    @staticmethod
+    @abc.abstractmethod
+    def get_credentials_names() -> Tuple[str, ...]:
```
Panaetius, line 252: Should this just return `List[str]` as per the docstring?

Panaetius (changes requested):
Looks really nice and should be easy to extend, thank you. Only some minor comments.

renku/infrastructure/storage/base.py
```diff
@@ -0,0 +1,96 @@
...
+import os
+import subprocess
+from typing import Any, List
+
+from renku.core import errors
+from renku.core.interface.storage import IStorage
+
+
+class BaseStorage(IStorage):
+    """Base external storage handler class."""
```
olevski, line 29: This currently has a lot of rclone-specific things. Should it then be called `RcloneBaseStorage` or something like that. Or if this is supposed to be more general then the rclone-specific stuff should then move someplace else?

renku/core/dataset/providers/s3.py
```diff
@@ -0,0 +1,96 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2020 - Swiss Data Science Center (SDSC)
```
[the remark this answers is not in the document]
mohammad-sdsc: I've changed all of them to `2017-2022`

renku/core/dataset/providers/api.py
mohammad-sdsc: `""` is a valid value. So, I've changed `None` to `NO_VALUE`.

renku/core/dataset/providers/api.py
mohammad-sdsc: This won't change and makes sense to be a tuple. I've updated the docstring.

renku/domain_model/dataset.py
```diff
@@ -318,6 +318,7 @@ class Dataset(Persistent):
     """Represent a dataset."""
 
     date_modified: Optional[datetime] = None  # type: ignore
+    storage: Optional[str] = None
```
olevski, line 321: Do we really need to have `storage` as both a class and instance variable? Isn't it simpler to just have the instance variable?

But maybe I am missing something. This also seems to happen with a few other class/instance variables on dataset.

olevski:
Also (if I am not mistaken) the credentials can only be added for a single s3 bucket. So I believe that the way this is setup is that I can only add buckets (with credentials) from a single region and provider.

If I am wrong about this lets touch base over slack @mohammad-sdsc because I probably misunderstood of how the credential provider is supposed to be used.

But my impression is that `s3://` as a prefix is only tied to one set of credentials. And we are allowing only `s3://` to be used as a prefix for s3 buckets.

renku/domain_model/dataset.py
mohammad-sdsc: We need the class variable for backward compatibility: When we load an object from an older project that doesn't have the `storage` property, the class variable makes sure that this property is set to `None`. The library that we use to store Renku metadata (i.e. ZODB/zope/persistent) doesn't call the `__init__` method when deserializing an object (it creates an empty object and calls `__setstate__` upon the first access to the object). So, we need the class variable for every new attribute that we add to our models.

mohammad-sdsc:
> Also (if I am not mistaken) the credentials can only be added for a single s3 bucket. So I believe that the way this is setup is that I can only add buckets (with credentials) from a single region and provider.
> 
> If I am wrong about this lets touch base over slack @mohammad-sdsc because I probably misunderstood of how the credential provider is supposed to be used.
> 
> But my impression is that `s3://` as a prefix is only tied to one set of credentials. And we are allowing only `s3://` to be used as a prefix for s3 buckets.

That's true, atm we only can have one set of credentials for S3. We can have multiple credentials by overriding `ProviderCredentials::get_credentials_section_name` for S3 (the comment there reads: `NOTE: This methods should be overridden by subclasses to allow multiple credentials per providers if needed.`). I added bucket name as the part of key before but then we needed to enter the credentials for each bucket so I removed it. I can easily put it back or implement any other mechanism.

olevski (approved):

olevski:
This LGTM but if you want to wait also for Ralf's approval that is def fine by me.

Panaetius (approved):

PR #3049, opened 2022-07-25
feat(core): copy files to datadir

### LINKED ISSUE #2980: Put files into a dataset's datadir when adding a local file
Change `renku dataset add` to put a file into a datasets datadir.

There are three ways of handling this:
- `--move`: Move the file into the datadir
- `--copy`: Copy the file into the datadir
- `--ln`: create a symlink into the datadir

By default, use the move strategy. Prompt the user to confirm, unless `--force` is passed.

For the ln/symlink case, ensure that the checksum used is that of the original file, not of the symlink.

Panaetius:
closes #2980

### DIFF
--- a/renku/core/dataset/providers/local.py
+++ b/renku/core/dataset/providers/local.py
@@ -27,7 +27,7 @@ from renku.core import errors
 from renku.core.dataset.providers.api import ExporterApi, ProviderApi, ProviderPriority
 from renku.core.util import communication
 from renku.core.util.dataset import check_url
-from renku.core.util.os import get_absolute_path, is_path_empty, is_subpath
+from renku.core.util.os import get_absolute_path, is_path_empty
 
 if TYPE_CHECKING:
     from renku.core.dataset.providers.models import DatasetAddMetadata, ProviderParameter
@@ -69,6 +69,27 @@ class FilesystemProvider(ProviderApi):
             ProviderParameter(
                 "external", flags=["e", "external"], help="Creates a link to external data.", is_flag=True
             ),
+            ProviderParameter(
+                "copy",
+                flags=["cp", "copy"],
+                help="Copy files to the dataset's data directory. Mutually exclusive with --move and --link.",
+                is_flag=True,
+                default=False,
+            ),
+            ProviderParameter(
+                "move",
+                flags=["mv", "move"],
+                help="Move files to the dataset's data directory. Mutually exclusive with --copy and --link.",
+                is_flag=True,
+                default=False,
+            ),
+            ProviderParameter(
+                "link",
+                flags=["ln", "link"],
+                help="Symlink files to the dataset's data directory. Mutually exclusive with --copy and --move.",
+                is_flag=True,
+                default=False,
+            ),
         ]
 
     @staticmethod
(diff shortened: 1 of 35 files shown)


renku/core/dataset/providers/local.py
```diff
@@ -86,22 +107,40 @@ def add(
...
+        force: bool = False,
         **kwargs,
     ) -> List["DatasetAddMetadata"]:
         """Add files from a URI to a dataset."""
         from renku.core.dataset.providers.models import DatasetAddAction, DatasetAddMetadata
 
+        if sum([move, copy, link]) > 1:
+            raise errors.ParameterError("--move, --copy and --link are mutually exclusive.")
+
+        default_action = DatasetAddAction.COPY
```
olevski, line 122: @Panaetius the issue for this (#2980) says "by default use the move strategy".

I guess things changed in the meantime and the default strategy is now copy.

olevski (approved):
Looks good to me. One question you do not have to address. I was just mostly curious.

renku/core/dataset/providers/local.py
Panaetius: When we discussed this, we said move should be the default, but it wasn't a very strong decision (I think mostly because I pushed for it).

It's essentially a trade-off, with move you don't have an issue with potentially duplicating a large file and running out of space in a session, so it's safer in regards to disk usage. But it does risk breaking users code (like if an input file or script is suddenly not around anymore, because it was moved, which you don't have with copy. Of course with copy a user might run into an issue that he updates the original file and is surprised that the file in the dataset doesn't update (since he'd have to edit the copy).
Symlinks would have neither of those problems, but then they wouldn't really put a file into the data dir and have other errors cases (like user moves the original file and inadvertently breaks the link, but we don't see that on the dataset), so I personally prefer to have physical files in the data dir, it's cleaner imo.

 But so I don't think there is a right answer, I opted for copy as it doesn't break existing code. Maybe @cmdoret has an opinion/input?

cmdoret: A scenario I have seen multiple times is:
1. Creates a workflow
2. Add workflow output to a dataset
3. Periodically rerun the workflow + update dataset

I guess in this scenario symlinks would work, but all three solutions remove the lineage from dataset files (I think) ?
I agree with @Panaetius that copy as a default action makes sense, but I also think the prompt should encourage users to put the files in the dataset folder from the start to retain lineage, if that makes sense.

Panaetius: Right now with the default action it shows a message like `The following files will be copied to my-dataset/ (use '--move' or '--link' to move or symlink them instead, '--copy' to not show this warning): [...files...] Proceed?` but we could also add a warning "Consider putting files into a datasets data directory directly and running `renku dataset update` to add them instead" or something like that? (I'm still working on support for `renku dataset update` to pick up files added to the datadir atm).

olevski: Ok cool. My approval still stand then. Thanks for the info and background @Panaetius.

Panaetius: I'll add the "Consider putting files..." message in the `renku dataset update` change when I finish it and merge this, then. Thanks for the feedback!

PR #3052, opened 2022-07-26
chore: change CI test versions of python from 3.7,[3.8,3.9] to 3.9,[3.7,3.10]

Panaetius:
Since 3.9 is the most common version and we should test with 3.10 as well.


lorenzo-cavazzi (approved):
The tests failed in the predicted way so... lgtm! :smile: 

I guess we need to change the required tests for the PRs in the repo's settings to merge this since the test names changed

Issue #2971, opened 2022-06-28
Add data to datasets from S3


mohammad-sdsc:
As a Renku user I want to be able to add data from S3 to my Renku datasets so that I can track their provenance.
```
renku dataset add my-dataset s3://server/path/to/bucket
```
## Acceptance Criteria
- Renku update datasets metadata to include info about the added file and its origin (using `DatasetFile.based_on`)
- Use the checksum that S3 provides as the file checksum.
- Only files from the same S3 bucket that the dataset was created for, can be added to the dataset.

## Notes
- Local data cannot be added to a dataset with external storage backend
- All data are from the same external source of the dataset.
- Wildcards won't be supported.
- Refactor dataset code to have a plugin architecture to allow different providers managing different sources
- Since dataset's data directory is git-ignored, we must guard data that is not pushed to the remote yet against `git clean` or similar operations.
- Make sure to abstract away the interface to external storages in case we want to change the underlying library/tool.
- Checksum calculation can be done locally or by S3: https://aws.amazon.com/blogs/aws/new-additional-checksum-algorithms-for-amazon-s3/. If we want to use checksum calculated by S3, then we must put the correct checksum in the metadata once we push the object to S3.

PR #3055, opened 2022-07-26
feat(dataset): make dataset providers plugins

olevski:
As I am working through #2971 one of the subtasks is to make the dataset providers plugins.

I isolated the changes for this and thought it would be a good idea to submit this PR separately from the rest.

Also wanted to get a feeling of whether my way of going about this is acceptable. I decided to not try to separate `ProviderApi` and the similar importer and exporter abstract classes and try to move some portion of them in the domain_model. But if you think this is necessary then I am happy to do it. 

For now the plugins simply provide the defintion of a specific ProviderApi. And the provider factory is now using pluggy to list all the providers rather than importing what is only locally available.

### DIFF
--- a/renku/core/dataset/providers/dataverse.py
+++ b/renku/core/dataset/providers/dataverse.py
@@ -74,7 +74,7 @@ DATAVERSE_SUBJECTS = [
 ]
 
 
-class DataverseProvider(ProviderApi):
+class DataverseProvider(ProviderApi, IDatasetProviderPlugin):
     """Dataverse API provider."""
 
     priority = ProviderPriority.HIGH
@@ -184,6 +184,12 @@ class DataverseProvider(ProviderApi):
 
         set_export_parameters()
         return DataverseExporter(dataset=dataset, server_url=self._server_url, dataverse_name=self._dataverse_name)
+
+    @classmethod
+    @hookimpl
+    def dataset_provider(cls) -> "Type[DataverseProvider]":
+        """The defintion of the provider."""
+        return cls
 
 
 class DataverseImporter(RepositoryImporter):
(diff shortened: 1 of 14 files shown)


renku/core/dataset/providers/dataverse.py
```diff
@@ -580,3 +582,12 @@ def make_file_url(file_id, base_url):
     args_dict = {"persistentId": file_id}
     url_parts[4] = urllib.parse.urlencode(args_dict)
     return urllib.parse.urlunparse(url_parts)
+
+
+class DataverseProviderPlugin(IDatasetProviderPlugin):
```
Panaetius, line 587: Could we add this as a mixin on e.g. `DataverseProvider` directly?

Like 
```
class DataverseProvider(ProviderApi, IDatasetProviderPlugin):
[...]
    @hookimpl
    @classmethod  # not sure on the correct order of decorators
    def dataset_provider(cls) -> "Type[ProviderApi]":
        """The defintion of the provider."""
        return cls
[...]
```
Same for the other providers.

Panaetius (approved):
Looks all great. I left on comment on how we could simplify it by adding the method on the provider directly (then we don't need extra classes), but that doesn't change anything functionality-wise, so feel free to ignore.

renku/core/dataset/providers/dataverse.py
olevski: Oh cool. A class method will do exactly what I was looking for. Don't know why I did not think of this.

I will also fix the mypy and style problems. Just wanted to see if this approach was acceptable or not.

Panaetius (dismissed):
Thank you, glad it worked with the classmethod :slightly_smiling_face:

olevski:
@Panaetius can I get a re-approve here? There were merge conflicts I had to resolve and that cleared the approval.

Panaetius (approved):

PR #3056, opened 2022-07-28
chore: add vscode settings.json

olevski:
I am ok if people would like this to not be checked in. 

But if we do want it I replicated the stuff from pyproject.toml in the settings.json file.

The most important thing tbh is that vscode properly uses 120 as the line length after using this file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,8 @@ __pycache__/
 *.py[cod]
 
 # VSCode
-.vscode/
+.vscode/*
+!.vscode/settings.json
 
 # Idea software family
 .idea/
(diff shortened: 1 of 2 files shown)


Panaetius (approved):
:+1:

.gitignore
```diff
@@ -4,7 +4,8 @@ __pycache__/
 *.py[cod]
 
 # VSCode
-.vscode/
+.vscode/*
```
rokroskar, line 7: maybe don't need the `*`?

olevski: @rokroskar if you remove the `*` then simply the whole folder is gitignored and the line below `!.vscode/settings.json` does have any effects at all. So then the settings.json will not be checked in and it will be ignored.

rokroskar: then does it need `.vscode/**/*` or similar to include subdirectories?

Issue #3061, opened 2022-07-29
prevent `renku mv` from moving files outside of a datasets data directory


Panaetius:
`renku mv` will happily move a datasets file to a location outside of its datadir, which we do not allow.

We should either make it so it symlinks to the destination instead or raise an error and prevent the move.

Reenable `test_move_dataset_file` when this is done

PR #3062, opened 2022-07-29
feat(cli): enable renku dataset update to pick up files in datasets data directory

### LINKED ISSUE #2978: add check for new/modified/deleted dataset files in datasets' datadir to pre-commit hook
We should extend the pre commit hook in Renku to check if any files were added to/removed from a datasets datadir or if they were modified and fail the commit if it's out date.

There should be a config option to disable this check on a per project basis.

Add a command to check the status of a dataset.

Make sure that `renku move` cannot break this relation.

### LINKED ISSUE #2981: Add a doctor check that ensures all local dataset files are inside the dataset's datadir
As per the title, check that local files are inside a datasets datadir. With `--fix`, symlink files into the datasets data directory and update metadata accordingly.

Panaetius:
closes #2978 
closes #2981

Followup: https://github.com/SwissDataScienceCenter/renku-python/issues/3061

Dropping a file in a datasets data directory will pick it up with `renku dataset update`.  `renku doctor` will now check for files outside the data dir and symlink them in when used with `--fix`


renku/ui/cli/dataset.py
```diff
@@ -1135,23 +1167,39 @@ def get_dataset_files(records):
...
-            click.echo(f"The following files will be deleted{message}:\n\n{files}\n")
+            if plain:
+                files = [f"r {f.path} {f.dataset.name}" for f in deleted_files]
+                click.echo(os.linesep.join(files))
+            else:
+                files = get_dataset_files(deleted_files)
+                message = " (pass '--delete' to remove them from datasets' metadata)" if not delete else ""
+                click.echo(f"The following files will be deleted{message}:\n\n{files}\n")
+
+        ctx.exit(1)
```
olevski, line 1205: Just making sure this `ctx.exit(1)` is intentional and not leftover from troubleshooting/testing.

tests/cli/test_datasets.py
```diff
@@ -1694,6 +1694,50 @@ def test_lfs_hook_can_be_avoided(runner, project, subdirectory, large_file):
...
+    with pytest.raises(errors.GitCommandError) as e:
+        client.repository.commit("datadir files not in dataset")
+
+    assert "Files in datasets data directory that aren't up to date" in e.value.stderr
+
+    result = runner.invoke(cli, ["config", "set", "check_datadir_files", "false"])
+    assert 0 == result.exit_code, format_result_exception(result)
+
+    client.repository.add(all=True)
+    client.repository.commit("datadir files not in dataset")
```
olevski, line 1738: Is this line and the one immediately above really necessary if you are not doing any assertions afterward?

tests/cli/test_datasets.py
```diff
@@ -1694,6 +1694,50 @@ def test_lfs_hook_can_be_avoided(runner, project, subdirectory, large_file):
...
+    assert 0 == result.exit_code, format_result_exception(result)
+
+    file = datadir / "new_file"
+    file.write_text("some updates")
+    file2 = datadir / "another_file"
+    file2.write_text("some updates")
+
+    client.repository.add(all=True)
+
+    # Commit fails when file is not tracked in LFS
```
olevski, line 1712: If the file was committed but not tracked in LFS then things would have passed, right?

olevski (changes requested):
Just a few minor things.

renku/ui/cli/dataset.py
Panaetius: Yes, if datasets are not up to date, it'll raise the status. That's what enables using it in e.g. the pre-commit hook

tests/cli/test_datasets.py
Panaetius: Commit would raise an exception here if there was a file in the data dir that is not in the datasets metadata

tests/cli/test_datasets.py
Panaetius: ah that's a copy paste error, the comment was from a similar piece of code in another test :see_no_evil:

olevski (approved):

olevski (approved):

Issue #2973, opened 2022-06-28
Copy data from S3


mohammad-sdsc:
As a Renku user I want to copy data from S3 into my local machine so that I can use this data in Renku workflows.
```
renku dataset stage my-dataset
```
## Acceptance Criteria
- Data is copied to dataset's data directory.
- A `--location` flag allows copying data to a different directory. This directory must be outside the repository. A link from the dataset's data directory to this alternative directory must be made. The alternative directory might be on another filesystem, so, we should use symlinks. The location must be persisted in the project (in `.renku/renku.ini` for example).

## Notes
- This command should be hidden from users since it's an experimental feature.
- See https://github.com/SwissDataScienceCenter/renku-python/issues/2912 for some notes on implementation. TL;DR: [RClone](https://rclone.org/downloads/) seems to be a good solution. Also check [fsspec](https://filesystem-spec.readthedocs.io/en/latest/index.html), [pyfilesystem](https://github.com/PyFilesystem/pyfilesystem2), and [smart_open](https://github.com/RaRe-Technologies/smart_open).

PR #3063, opened 2022-08-01
feat(dataset): add data from s3

### LINKED ISSUE #2971: Add data to datasets from S3

olevski:
Fixes #2971 

Should be reviewed and merged only after #3047 is merged.

Not really sure how to tackle a few situations:

1. what happens if someone simply adds a file into the mounted s3 bucket outside of renku?
2. what happens if for some reason the s3 bucket is unmounted?

For the 2nd question I think that a possible solution is the following:
- when a `Dataset` instance is created check if the storage is s3
- if the storage is s3 check if the bucket is mounted
- if the bucket is mounted then (optionally and expensively) check if the hashes of files have changed -> if yes update hashes in metadata
- if the bucket is not mounted then mount it (also check for hash changes and update metadata)

Another issue is that s3 does not provide hashes for certain files. For example `.gz` or similar tars/archives do not have hashes. These can be very large and I have set things up so that I download these files and then do the hashes. So that I can fill in any missing hashes. Is it acceptable to simply have missing hashes in some cases then?


renku/core/dataset/dataset_add.py
```diff
@@ -92,7 +93,9 @@ def add_to_dataset(
                     "Ignored adding paths under a .git directory:\n\t" + "\n\t".join(str(p) for p in paths_to_avoid)
                 )
 
-            files_to_commit = {f.get_absolute_commit_path(client.path) for f in files}
+            files_to_commit = {
+                f.get_absolute_commit_path(client.path) for f in files if not str(f.url).startswith("s3://")
```
mohammad-sdsc, line 97: This is too specific to S3; we should make it a bit more general in case we add other external storages in future. Maybe by adding a hook function to `ProviderApi` to filter dataset files or some other approaches.

renku/core/dataset/dataset_add.py
```diff
@@ -155,6 +159,21 @@ def _download_files(
     **kwargs,
 ) -> List["DatasetAddMetadata"]:
     """Process file URLs for adding to a dataset."""
+    if dataset.storage and dataset.storage.lower().startswith("s3://"):
+        if len(urls) == 0:
+            raise errors.ParameterError("No URL is specified")
+        if len(urls) > 1:
+            raise errors.ParameterError("Can only add 1 uri with S3 datasets.")
```
mohammad-sdsc, line 166: Is there a problem to have more than one URL? We can check if they belong to the same storage as the dataset in the `for` loop below.

renku/core/dataset/dataset_add.py
```diff
@@ -155,6 +159,21 @@ def _download_files(
...
+        if not urls[0].lower().startswith("s3://"):
```
mohammad-sdsc, line 167: This check can be moved to the for loop if we allow more than one URL

mohammad-sdsc: We should generalize this check to allow other providers. Maybe by adding a function to the dataset class.

renku/core/dataset/dataset_add.py
```diff
@@ -155,6 +159,21 @@ def _download_files(
     **kwargs,
 ) -> List["DatasetAddMetadata"]:
     """Process file URLs for adding to a dataset."""
+    if dataset.storage and dataset.storage.lower().startswith("s3://"):
```
mohammad-sdsc, line 162: I think `and dataset.storage.lower().startswith("s3://")` is not needed.

renku/core/dataset/dataset_add.py
```diff
@@ -155,6 +159,21 @@ def _download_files(
...
+        if len(urls) == 0:
+            raise errors.ParameterError("No URL is specified")
+        if len(urls) > 1:
+            raise errors.ParameterError("Can only add 1 uri with S3 datasets.")
+        if not urls[0].lower().startswith("s3://"):
+            raise errors.ParameterError("Only files from S3 buckets can be added to S3 datasets.")
+        if sources:
+            raise errors.ParameterError("Can not use '--source' with S3 datasets.")
+        if len(dataset.files) > 0:
```
mohammad-sdsc, line 171: This is too limiting: It means that we can only add one file to a dataset. Is there a reason not to allow adding more files?

renku/core/dataset/providers/s3.py
```diff
@@ -59,6 +65,43 @@ def supports_create() -> bool:
...
+
+        storage = get_storage(provider=provider, credentials=credentials)
+        dataset = kwargs.get("dataset")
+        if dataset and dataset.storage and not is_uri_subfolder(dataset.storage, uri):
+            raise errors.ParameterError(
+                f"S3 uri {uri} should be located within or at the storage uri {dataset.storage}. "
+            )
+        if not storage.exists(uri):
+            raise errors.ParameterError(f"S3 bucket '{uri}' doesn't exists.")
+        storage.mount(uri, mount_location=destination)
```
mohammad-sdsc, line 90: We shouldn't mount or download files at this stage. This operation is only about adding files to the dataset and not copying them. I guess this is required for hash calculation if rclone cannot hash a file; in this case, it's better not to have a hash and then add it later when we pull/copy/stage the data.

renku/infrastructure/storage/base.py
```diff
@@ -47,12 +51,86 @@ def exists(self, uri: str) -> bool:
...
+            output.append(
+                FileHash(
+                    base_uri=uri,
+                    path=hash["Path"],
+                    hash=hash_content,
+                    hash_type=hash_type if hash_content else None,
+                    modified_datetime=hash.get("ModTime"),
+                )
+            )
+        output = self._get_missing_hashes(output, hash_type=hash_type)
```
mohammad-sdsc, line 98: Please see my comment above about not mounting/downloading files at this point.

mohammad-sdsc (changes requested):
Thanks Tasko! Please see my comments. In general I believe we don't need to mount or download data when adding data to the dataset. We just need to add the metadata and the actual data will be downloaded/pulled in the other story: https://github.com/SwissDataScienceCenter/renku-python/issues/2973. Also, you need to make some parts of the code more general since we will add more external storage types in future.

mohammad-sdsc:
> Should be reviewed and merged only after #3047 is merged.
> 1. what happens if someone simply adds a file into the mounted s3 bucket outside of renku?

I believe it's not related to this story but in that case we probably can ignore the file because we have no metadata for the file in renku.

> Another issue is that s3 does not provide hashes for certain files. For example `.gz` or similar tars/archives do not have hashes. These can be very large and I have set things up so that I download these files and then do the hashes. So that I can fill in any missing hashes. Is it acceptable to simply have missing hashes in some cases then?

I'd say it's ok to have an empty hash at this point. We can add the hash once we download the file.

renku/core/dataset/dataset_add.py
olevski: Yes if you mount s3 data then you cannot mount more than one uri to the same location.

If you copy stuff too it is a problem. What is there are conflicts/duplicates between the two uris? Which files will end up in the dataset? Which will be overwritten? But this is definitely more solvable if you are copying and not mounting.

Perhaps to keep things simple to start we enforce this? I am happy to remove it. You will have to deal with it in the PR on copying stuff.

renku/core/dataset/dataset_add.py
olevski: This is a limitation if you are mounting. If you are copying it becomes easier to handle. I will remove it and we can deal with it in the PR about copying. 

IMO copying kind of defeats the whole purpose of using S3 to some extent. Because the biggest motivation for using s3 is to not need the same storage in the session (or your local machine).

renku/core/dataset/providers/s3.py
olevski: I will remove this. I did not understand the issue.

renku/core/dataset/providers/s3.py
```diff
@@ -76,7 +119,10 @@ def on_create(self, dataset: "Dataset") -> None:
             raise errors.ParameterError(f"S3 bucket '{self.bucket}' doesn't exists.")
 
         repository = get_repository()
-        repository.add_ignored_pattern(pattern=str(dataset.get_datadir()))
+        repository.add_ignored_pattern(
+            pattern=str(dataset.get_datadir().absolute().relative_to(repository.path.absolute()))
```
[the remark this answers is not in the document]
olevski: I will double check. I think some of my tests were failing when I used only `str(dataset.get_datadir())`

renku/core/dataset/dataset_add.py
olevski: I ended up adding `gitignored` field on the `DatasetAddMetadata` class. Is that an acceptable way to do this? It is `False` by default. And will be set to `True` for S3 and any other provider we have in the future can decide to do the same. Then this function becomes pretty general and understandable.

renku/infrastructure/storage/base.py
olevski: This will also be removed.

renku/core/dataset/dataset_add.py
mohammad-sdsc: URIs comes from the user: For example, when a user runs `renku dataset add s3-data s3://bucket/path/to/file-1 s3://bucket/path/to/file-2`, we will receive `s3://bucket/path/to/file-1` and `s3://bucket/path/to/file-2` in this function. In this case, it's quite convenient to allow multiple URIs.
It's users' responsibility if they add files that take a long time to process; we can't/shouldn't prevent users from doing that. Also, AFAIK we never check for addition on duplicate data. It's possible to run `renku dataset add my-data /path/to/dir /path/to/dir/a/file/inside`. They won't conflict in such cases because they go in different sub-directories.

renku/core/dataset/dataset_add.py
mohammad-sdsc: Copying is for use-cases where a user want to have faster access to data; mounting is more space-efficient but we always need to stream data from the external storage (unless the underlying library implements a caching mechanism).

renku/core/dataset/dataset_add.py
mohammad-sdsc: Yes, that would work as well. You could also add an `IGNORE` action to `DatasetAddAction` which is included in `DatasetAddMetadata`.

olevski: I will stick to the `gitignored` stuff I added. I think it is not impossible to use one of the existing actions and to have the file be "gitignored".

renku/core/dataset/dataset_add.py
olevski: I removed/changed it so that it is more general.

renku/core/dataset/dataset_add.py
olevski: I kinda changed/removed this whole section. I did this by requiring that the scheme of `dataset.storage` matches the scheme of the uri being added.

renku/core/dataset/dataset_add.py
olevski: Ok yeah I removed this.

renku/core/dataset/dataset_add.py
olevski: Ok. I removed this limitation too.

renku/core/dataset/providers/s3.py
```diff
@@ -70,25 +69,55 @@ def supports_add() -> bool:
...
+
+        return [
+            ProviderParameter(
+                "sources",
+                flags=["s", "source"],
+                default=None,
+                help="Path(s) within the s3 bucket that should be added",
+                multiple=True,
+                type=str,
+            ),
```
mohammad-sdsc, line 85: I'm not sure whats the benefit of this option; we can simply pass the value that we pass here as a URI and things should work as normal. As I understood, this is supposed to act as a prefix to URI so that if we run `renku dataset add s3 -s s3://giab/ Aspera_download_from_ftp.README` we would get `s3://giab/Aspera_download_from_ftp.README`. However, this won't work as intended because in this case URI is `Aspera_download_from_ftp.README` which won't be processed by the S3 provider. Running the above command result in:
```
Error: Invalid parameter value - The scheme of some urls ('Aspera_download_from_ftp.README',) does not match the defined storage url s3://giab/.
```

mohammad-sdsc (changes requested):
Thanks Tasko for the changes. Please see my comment regarding the `-s` option. I believe we should remove it.

renku/core/dataset/providers/s3.py
olevski: @mohammad-sdsc I have removed the --source option for s3 datasets as you suggested. It is def. better to simply not allow this.

olevski:
@mohammad-sdsc thanks for removing the "source parameter" stuff I missed. Let me know if we need anything else.

mohammad-sdsc:
> @mohammad-sdsc thanks for removing the "source parameter" stuff I missed. Let me know if we need anything else.

Sorry to edit with your PR. I thought it's easier this way to get things going. Thanks! It works like a charm.

PR #3065, opened 2022-08-03
docs: add command line full reference using sphinx-click

### LINKED ISSUE #3033: Add CLI command documentation via sphinx-click
Our CLI documentation does not currently include a structured description of the command line arguments and parameters available. So, when a user is wondering what parameters are available for each command, they don't have an easy reference to use. This structured documentation can be created via the sphinx plugin [sphinx-click](https://sphinx-click.readthedocs.io/en/latest/usage/).

I recommend that the `sphinx-click` documentation be added _above_ the current long-form documentation, so that the structured information is easily findable. For example:

```
``renku session``
-----------------

.. click:: renku.ui.cli.session:session
   :prog:  renku session
   :nested: full

.. automodule:: renku.ui.cli.session
```

image (3)

**A note on dynamic argument options:** I tested it with `renku dataset`:

```
``renku dataset``
-----------------

.. click:: renku.ui.cli.dataset:dataset
   :prog:  renku dataset
   :nested: full


.. automodule:: renku.ui.cli.dataset
```

And the `sphinx-click` documentation showed `dataverse-name`, which comes from the provider dynamically, so that's great.

image (4)

**Acceptance Criteria**
- [ ] click entries added for each CLI command, above the `automodule call`. 
- [ ] long-form documentation updated (slimmed down) wherever it is redundant after the click documentation. For example, in some places we do already have tables with the options so we'd need to remove those.

lorenzo-cavazzi:
I adopted a solution similar to the one suggested in #3033 . The main difference is the click-generated docs being included in the ui.cli files to better position the content.

I (loosely) took inspiration from the [Git documentation](https://git-scm.com/docs/git-checkout) where every command has the following sections:

- description
- options
- command-specific section(s) [optional]
- examples [optional]

In our case, after the description, I added a "Commands and options" section, moving relevant explanations and examples at the end in a similar fashion.
If we like this approach, I would separate each command into a different page because the "Renku Command Line" page has grown way too big -- Git does the same.

P.S.
The repetition of the command as a title (just after "Commands and options") is quite annoying, but luckily it will soon be possible to remove it https://github.com/click-contrib/sphinx-click/pull/100 -- we could speed up the process by forking the repo and finalizing that PR, or just waiting (I actually reviewed it since there are a couple of minor inconsistencies).

Here is an example of what we could simply remove with that PR (left is the current output with the additional `renku project` title). We could also remove the `Project commands` description.

Screenshot_20220805_154547

P.P.S we discussed other approaches, but the content we wrote seems useful: it would be a pity to throw it away, and IMHO integrating sphinx-click led to a decent result.
Of course, I'm open to explore other solutions if you think the result is not as good as it should be.

fix #3033


renku/ui/cli/config.py
```diff
@@ -55,33 +69,6 @@
...
-configuration take precedence over global values. Use ``--local`` or
-``--global`` flag to read corresponding configuration only.
-You can provide a KEY to display only its value:
-
-.. code-block:: console
-
-    $ renku config show interactive.default_url
-    default_url = /lab
-
-.. cheatsheet::
```
Panaetius, line 79: Did you intend to remove the cheatsheet entry?

renku/ui/cli/graph.py
```diff
@@ -17,17 +17,30 @@
...
+~~~~~~~~~~~~~~~~~~~~
+
+.. click:: renku.ui.cli.graph:graph
+   :prog: renku graph
+   :nested: full
+
+Examples
+~~~~~~~~
+
+Here we create a new dataset and then ``renku graph export`` export the
```
Panaetius, line 41:
```suggestion
Here we create a new dataset and then use ``renku graph export`` to export the
```

renku/ui/cli/init.py
```diff
@@ -17,32 +17,18 @@
...
 r"""Create an empty Renku project or reinitialize an existing one.
 
-Start a Renku project
-~~~~~~~~~~~~~~~~~~~~~
+Description
+~~~~~~~~~~~
 
-If you have an existing directory which you want to turn into a Renku project,
-you can type:
+Use an existing renku templates, either official or from the community, to
```
Panaetius, line 23:
```suggestion
Use an existing renku template, either official or from the community, to
```

or
```suggestion
Use existing renku templates, either official or from the community, to
```

renku/ui/cli/mergetool.py
```diff
@@ -17,17 +17,24 @@
...
-To support users when doing this, Renku provides a custom merge tool that takes care of merging the metadata.
-
-The merge tool is set up automatically when creating a new project or when using ``renku clone`` to clone a Renku
-project.
+When working in multiple branches with Renku, this metadata needs to be merged
+when a git merge is made. To support users when doing this, Renku provides a
+custom merge tool that takes care of merging the metadata.
 
+The merge tool is set up automatically when creating a new project or when
+using ``renku clone`` to clone a Renku project.
```
Panaetius, line 29: Wasn't there before either, but maybe we should write something along the lines of "Merge tool configuration cannot be shared through remote repositories and has to be set up on each clone separately" ? `renku clone` covers it but something to let the user know that if they do a normal clone, this will be missing.

renku/ui/cli/project.py
```diff
@@ -17,47 +17,17 @@
...
-| -m, --metadata    | Path to json file containing custom metadata to be   |
-|                   | added to the project knowledge graph.                |
-+-------------------+------------------------------------------------------+
-| -u, --unset <type>| Remove a value from a project. One of ``creators``,  |
-|                   | ``m``, ``keywords``, ``k``, ``metadata``, ``m``      |
-+-------------------+------------------------------------------------------+
+Description
+~~~~~~~~~~~
+
+Visualize and edit project's metadata.
```
Panaetius, line 23: `Show and edit`? Visualize to me implies some graphic display.

renku/ui/cli/remove.py
```diff
@@ -17,16 +17,25 @@
 # limitations under the License.
 """Remove a file, a directory, or a symlink.
 
-Removing a file that belongs to a dataset will update its metadata. It also
-will attempt to update tracking information for files stored in an external
+Description
+~~~~~~~~~~~
+
+Remove a file that belongs to a dataset and update its metadata. It also
```
Panaetius, line 23: This command is for removing any file (not just ones belonging to a dataset). but if they are in a dataset, it takes care of not having inconsistent metadata.

Panaetius (changes requested):
I think separating into different pages makes a lot of sense. And this whole change makes the docs much more usable, thank you!

I wonder if we should make sub sub commands italic or differentiate them further? At least to me, `renku config` and `remove` look almost the same in regards to size:
image
Same with `Options` which is just 1.6px smaller than the name of the subcommand.
Looking at e.g. https://git-scm.com/docs/git-remote the sub-commands are in their own `Commands` section and italic. Not that we need to do it this way but it looks cleaner to me. And the different colors help readability a lot as well. Probably also goes well with splitting it to individual files. 

Not really related, we should create an issue to investigate https://github.com/click-contrib/click-man to make docs even better. No need to open a browser.

olevski:
@lorenzo-cavazzi @Panaetius I know this is more work but how about adding a specific section for each subcommand. Like this:
image

I did the following to achieve this:
```rst
Description
~~~~~~~~~~~

Create, edit and manage the datasets in your Renku project.

This is a core feature of Renku. You might want to go through the examples
listed below to get an idea of how you can create, import, and edit datasets.


``add``
~~~~~~~

.. click:: renku.ui.cli.dataset:add
   :prog: renku dataset add
   :nested: full

``create``
~~~~~~~~~~

.. click:: renku.ui.cli.dataset:create
   :prog: renku dataset create
   :nested: full


Examples
~~~~~~~~
```

This requires more work but it makes the docs even nicer and much easier to navigate IMO. If we have one section for all subcommands then it is very hard to see where one command ends and another begins. Also it is very hard to understand what subcommands are available.

I think this is def worth doing. Either here or in a followup PR. What do you think?

renku/ui/cli/config.py
lorenzo-cavazzi: Ouch! Nope, that was not intended. I'll restore it

renku/ui/cli/mergetool.py
lorenzo-cavazzi: Yes, that makes sense. Otherwise, users might believe that `renku clone` keeps everything alligned

Panaetius:
> @lorenzo-cavazzi @Panaetius I know this is more work but how about adding a specific section for each subcommand. Like this: image
> 
> I did the following to achieve this:
> 
> ```
> Description
> ~~~~~~~~~~~
> 
> Create, edit and manage the datasets in your Renku project.
> 
> This is a core feature of Renku. You might want to go through the examples
> listed below to get an idea of how you can create, import, and edit datasets.
> 
> 
> ``add``
> ~~~~~~~
> 
> .. click:: renku.ui.cli.dataset:add
>    :prog: renku dataset add
>    :nested: full
> 
> ``create``
> ~~~~~~~~~~
> 
> .. click:: renku.ui.cli.dataset:create
>    :prog: renku dataset create
>    :nested: full
> 
> 
> Examples
> ~~~~~~~~
> ```
> 
> This requires more work but it makes the docs even nicer and much easier to navigate IMO. If we have one section for all subcommands then it is very hard to see where one command ends and another begins. Also it is very hard to understand what subcommands are available.
> 
> I think this is def worth doing. Either here or in a followup PR. What do you think?

That does look nicer, yes. Though I still think using colors/style could improve it even further. Not sure how easy that'd be to do, might need updating in https://github.com/SwissDataScienceCenter/renku-sphinx-theme

renku/ui/cli/remove.py
```diff
@@ -20,9 +20,11 @@
 Description
 ~~~~~~~~~~~
 
-Remove a file that belongs to a dataset and update its metadata. It also
-attempts to update tracking information for files stored in an external
-storage (using Git LFS).
+Remove a file from then project, updating:
+
+* its metadata, if the file belong to a dataset
```
Panaetius, line 25: belongs

lorenzo-cavazzi:
P.S: sorry for force pushing. I messed up badly when merging `develop` back :facepalm:

lorenzo-cavazzi:
Each command now has its own separate page.
I tried a few combinations based on your suggestions.

In the end, I think using @olevski approach to separate the single commands might be too much when there are many sub-commands. You can check it on the `renku dataset` command

Screenshot_20220810_085514

Otoh the final output when using a different styling (closer to the Git documentation) as suggested by @Panaetius looks good to me. You can check it on the `renku template` command.
P.S. further personalizing some elements is really easy using the `override-them.css`.

Screenshot_20220810_085502

Panaetius:
I like the second version with italics :+1:

lorenzo-cavazzi:
Ok great, I changed all the commands to adopt the italics+tabbed style and dropped the links to the subcommands.

The following is an example of how the dataset page changed compared to the previous screenshot.

Screenshot_20220811_112312

lokijuhy:
This looks awesome! I agree having each command on its own page makes it easier to navigate. 

One potential suggestion: I could see it being useful to users to have a list of all the options for a command, without having to scroll the length of the page and find each heading to see what's available. I see that it's a bit much in the sidebar, but what about a table-of-contents-esque entry at the top of the "Commands and options" section, where the reader can see all commands available, and click to scroll down to the one they want? (add, create, edit, ...)

lorenzo-cavazzi:
Do you mean something like this? (ignore the style, I just made it on the fly, image it styled as a proper "table of contents")

Screenshot_20220815_110858

lokijuhy:
Yes, precisely!

lorenzo-cavazzi:
Ok great, I agree it would be useful.

The anchors get automatically generated by the library, which is a good start. Since the commands and options documentation is automatically created, I feel we should automatize the table-of-contents creation as well -- in a similar fashion to what we do with the cheat sheet. Doing it manually seems more error-prone.

That might require a bit of time though, and I have no capacity for this (and the following) sprint on renku-python. Would it make sense to create a follow-up issue so we don't keep this pending for too long?

lokijuhy:
Yes, that sounds like a good plan!

lokijuhy (approved):
📚 ✨

lorenzo-cavazzi:
Here is the follow-up issue: #3086

rokroskar:
This PR broke the renku docs build, btw. The only reason we didn't notice was because the current submodule still points to a release from June...

PR #3066, opened 2022-08-04
feat(dataset): copy/pull data from external storage

### LINKED ISSUE #2973: Copy data from S3

mohammad-sdsc:
# Description

Adds a `dataset pull` command to download/copy external data to the local project.

Fixes #2973 

### TODO:
- Add documentation once this feature is finalized
- Test with a private S3 bucket


renku/core/util/dispatcher.py
```diff
@@ -52,6 +53,16 @@ def get_database_helper(database_dispatcher: IDatabaseDispatcher):
     return get_database_helper()
 
 
+def get_dataset_gateway() -> "DatasetGateway":
```
Panaetius, line 56: I'm honestly not sure I'm a big fan of these types of methods getting added everywhere. It might make the method body more readable, but it does lose the big plus where in testing you can just pass in a mocked instance of an injected class directly to test the method in isolation. Whereas this makes it so we must always set up fake injection when testing single methods.

I quite like being able to do https://github.com/SwissDataScienceCenter/renku-python/blob/develop/tests/core/commands/test_graph.py#L318-L323

Especially if the mocked instances have to be changed between calls, updating injection/replacing existing entries is a nightmare compared to just passing another instance.

renku/core/util/os.py
```diff
@@ -165,32 +193,38 @@ def delete_file(filepath: Union[Path, str], ignore_errors: bool = True, follow_s
...
     with open(path, "rb") as f:
         for byte_block in iter(lambda: f.read(BLOCK_SIZE), b""):
-            sha256_hash.update(byte_block)
+            hash_value.update(byte_block)
+
+    return hash_value.hexdigest()
 
-    return sha256_hash.hexdigest()
 
+def hash_str(content: str, hash_type: str = "sha256"):
```
Panaetius, line 213: I wonder if we could unify these two functions using StringIO for strings and `open()` for files? As in, have `hash_str` call `hash_file` or both call the same helper method?

renku/core/dataset/dataset.py
```diff
@@ -1168,3 +1170,90 @@ def should_include(filepath: Path) -> bool:
...
+            if create_symlinks:
+                symlink_path = client.path / file.entity.path
+                symlink_path.parent.mkdir(parents=True, exist_ok=True)
+                create_symlink(path=path, symlink_path=symlink_path, overwrite=True)
+
+    # NOTE: Store location in metadata in case where we want to mount the external storage in the same location
+    store_dataset_data_location(dataset=dataset, location=location)
+
+    if updated_files:
+        _update_datasets_files_metadata(client, updated_files=updated_files, deleted_files=[], delete=False)
```
Panaetius, line 1241: Should we only update metadata if a file was actually changed during the pull?

Panaetius (changes requested):
Looks really nice, I have some suggestions/questions but the functionality is all good

renku/core/dataset/dataset.py
```diff
@@ -1168,3 +1170,90 @@ def should_include(filepath: Path) -> bool:
...
+
+    Args:
+        name(str): Name of the dataset
+        location(Optional[Path]): A directory to copy data to (Default value = None).
+    """
+    client = get_client()
+    datasets_provenance = DatasetsProvenance()
+
+    dataset = datasets_provenance.get_by_name(name=name, strict=True)
+    assert dataset  # NOTE: To make mypy happy
```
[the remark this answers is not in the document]
Panaetius: e.g. https://github.com/SwissDataScienceCenter/renku-python/pull/3071/files#diff-44504c88307d8f6a0c6e125d4d1de71e45b73c90fb6109dc95c27c62194cc9c9R60 works. It's annoying that we need to add `*,` so it's a keyword arg, because we can't define two default values in the overloads, otherwise mypy complains. But it works pretty nicely otherwise.

renku/core/util/dispatcher.py
mohammad-sdsc: TBH, this is the first test that I see which doesn't use injection. Testing like this is possible only if the method doesn't call other method that use injection. Otherwise, we still need to set up a fake injection.

I've changed its usage in my code since it doesn't call other injected method.

renku/core/dataset/dataset.py
mohammad-sdsc: When we add data from S3, it's not always possible to get the checksum from S3. So, we leave an empty checksum in the metadata and then we calculate the checksum here once we've downloaded the data. This is for such cases.

renku/core/dataset/dataset.py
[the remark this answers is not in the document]
mohammad-sdsc: This won't work in Python 3.7 because `Literal` was added in 3.8. I'm not sure why we changed the base Python for testing from 3.7 to 3.9.

PR #3071, opened 2022-08-10
feat(cli): change renku mv to respect datasets' datadir

### LINKED ISSUE #3061: prevent `renku mv` from moving files outside of a datasets data directory

Panaetius:
closes #3061 

Removes file from datasets metadata if moving out of datadir (and warns the user), correctly handles that moving to a dataset means the file needs to end up in the data dir.


mohammad-sdsc (commented):
Thanks Ralf! I just have two questions:
1. Why not treating external files (pointers) like other files?
2. Should we move a file to a dataset if the target is that dataset's data dir? We can do this in another story if needed.

Panaetius:
@mohammad-sdsc 
> Why not treating external files (pointers) like other files?

Well if I move a file, it gets removed from the dataset. Then if it is inside the datasets datadir still, it gets re-added with the new path. If it is not, it's not added. But for external files it doesn't matter, they can be anywhere, so we add them to the dataset anyways. Since we can't determine if a user wants them in the dataset still or not, we err on the side of including them. They can always manually remove them if they don't want them.

> Should we move a file to a dataset if the target is that dataset's data dir? We can do this in another story if needed.
:+1:

mohammad-sdsc (approved):

PR #3075, opened 2022-08-12
fix(cli): allow display of ssh password prompt in clone operations

### LINKED ISSUE #2987: dataset import hangs with passphrase-protected SSH key
If the git repository uses an SSH key with a passphrase, `renku dataset import` hangs on the loading animation until the passphrase is entered.

**To Reproduce**
1. Setup renku project using local ssh-key with passphrase for authentication
2. renku login
3. Attempt to import a renku dataset
4. Loading animation runs indefinitely (or until user enters the passphrase)

**Expected behavior**
Renku should explicitely prompt for the passphrase before starting the loading animation.

 - renku-python 1.4.0
 - Python 3.9.7
 - local terminal

Panaetius:
closes #2987 

We can't use `yaspin` / `communication.busy` with subprocess calls that need a prompt, as yaspin will swallow all shell output while it is running.

Also updates communication to allow printing while yaspin is up

### DIFF
--- a/renku/ui/cli/utils/callback.py
+++ b/renku/ui/cli/utils/callback.py
@@ -37,41 +37,36 @@ class StandardOutput(CommunicationCallback):
         self._progress_types = ["download"]
         self._spinner = None
 
+    @contextmanager
+    def pause_spinner(self):
+        """Pause Spinner so output gets actually written."""
+        if self._spinner:
+            self._spinner.hide()
+
+        yield
+
+        if self._spinner:
+            self._spinner.show()
+
     def echo(self, msg, end="\n"):
         """Write a message."""
-        with CommunicationCallback.lock:
-            if self._spinner:
-                self._spinner.hide()
+        with CommunicationCallback.lock, self.pause_spinner():
             print(msg, end=end)
-            if self._spinner:
-                self._spinner.show()
 
     def info(self, msg):
         """Write an info message."""
-        with CommunicationCallback.lock:
-            if self._spinner:
-                self._spinner.hide()
+        with CommunicationCallback.lock, self.pause_spinner():
             print(msg)
-            if self._spinner:
-                self._spinner.show()
 
     def warn(self, msg):
         """Write a warning message."""
-        with CommunicationCallback.lock:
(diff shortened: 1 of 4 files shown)


renku/ui/cli/utils/callback.py
```diff
@@ -35,26 +35,43 @@ def __init__(self):
         super().__init__()
         self._progress_bars = {}
         self._progress_types = ["download"]
+        self._spinner = None
 
     def echo(self, msg, end="\n"):
         """Write a message."""
         with CommunicationCallback.lock:
+            if self._spinner:
+                self._spinner.hide()
```
mohammad-sdsc, line 44: Why do we need to hide `_spinner` here? Also, it seems that we can use `hidden` context manager to temporarily hide the spinner if needed (just a matter of taste):
```
with self._spinner.hidden():
    print(...)
```

mohammad-sdsc: We should also refactor all these methods to a single one.

renku/ui/cli/utils/callback.py
```diff
@@ -115,19 +137,35 @@ def echo(self, msg, end="\n"):
         if end != "\n":
             msg = msg + end
             new_line = False
+        if self._spinner:
+            self._spinner.hide()
         click.echo(msg, nl=new_line)
+        if self._spinner:
+            self._spinner.show()
```
mohammad-sdsc, line 144: Same as above: We should refactor these to a common method.

mohammad-sdsc (commented):
Please see my comments.

renku/ui/cli/utils/callback.py
Panaetius: Spinner swallows all print output while active

Panaetius: also using the `hidden` contextmanager doesn't work because _spinner might be None. I'll make a custom context manager

mohammad-sdsc (dismissed):
Thank you!

mohammad-sdsc (approved):

PR #3077, opened 2022-08-12
feat: add meaningful data to auto branch names 

### LINKED ISSUE #2374: Protected branch handling -- name of branch
# Description

PR #1222 introduced support for projects in which the default branch does not allow pushes. To handle this, a new branch is created and the changes are pushed to the new branch. The name of this new branch is a UUID.

Thought the UUID is a robust solution, naming a branch this way makes it difficult to read. As an alternative, I suggest using a timestamp. For example, `YYYY-MM-DD_HH-MM_[3 digit random number]`. This introduces a potential race condition if multiple branches are created simultaneously, but this is 1. unlikely, and 2. should be solved by generating a new branch name if a push fails.

Having branches names this way makes the branch name easier to interpret.

lorenzo-cavazzi:
# Description

Change name of branches automatically generated when trying to push to a remote protected branch.
The final branch name will be `renku/autobranch/<username>/<protected_branch_name>/<commit_sha>`.
The username is the email address to keep it in line with what renku notebooks does with autosaves.

Screenshot_20220812_141331

P.S. I'm not 100% sure that using `get_git_user` is the best solution. I assume it's fine most of the time, but I can imagine it being potentially wrong when working on the cli. Should I instead pass the local user to `push_changes` as a parameter? I don't like that solution either and, to be honest, I feel adding that user here is somewhat unnecessary (it's fine for autosaves since they contain personal content, but in general having a username/email in a branch name looks very weird).

This also includes a minor fix for a service error.

Screenshot_20220812_104253

Fixes #2374


renku/core/util/git.py
```diff
@@ -545,8 +546,11 @@ def push_changes(repository: "Repository", remote: Optional[str] = None, reset:
 
         if merge_conflict or push_failed:
             # NOTE: Push to a new remote branch and reset the cache.
+            last_short_sha = repository.head.commit.hexsha[0:8]
             old_active_branch = repository.active_branch
-            pushed_branch = uuid4().hex
+            user = get_git_user(repository)
+            email = getattr(user, "email", "unknown")
+            pushed_branch = f"{PRETECTED_BRANCH_PREFIX}/{email}/{old_active_branch}/{last_short_sha}"
```
Panaetius, line 553: I think we should check that the branch name length does not get too long. According to https://stackoverflow.com/questions/60045157/what-is-the-maximum-length-of-a-github-branch-name it's anywhere between 60 and 250 characters. We probably want to shorten `old_active_branch` down to a minimum of 6 if it's too long or so and if it's still too long, chorten the email?

Panaetius (changes requested):

renku/core/util/git.py
lorenzo-cavazzi: This makes sense. A shortened email might not be very useful after all. Let me check with @ciyer if dropping it would work.
I'll still add the check for the `old_active_branch` just in case

ciyer: Then let us drop the email.

lorenzo-cavazzi: Ok, I dropped the email and added the logic to limit branch names to 250 chars

renku/core/util/git.py
```diff
@@ -37,7 +37,9 @@
     from renku.infrastructure.repository import Commit, Remote, Repository
 
 
+BRANCH_NAME_LIMIT = 250
 CLI_GITLAB_ENDPOINT = "repos"
+PRETECTED_BRANCH_PREFIX = "renku/autobranch"
```
Panaetius, line 42:
```suggestion
PROTECTED_BRANCH_PREFIX = "renku/autobranch"
```

Panaetius (changes requested):

Panaetius (approved):

Panaetius (approved):

PR #3087, opened 2022-08-16
chore: switch to poetry 1.3

### LINKED ISSUE #2553: Download templates in poetry build
In https://github.com/SwissDataScienceCenter/renku-python/pull/2552 downloading templates was moved to the Makefile, because poetry doesn't properly support build scripts/hooks/plugins.

Add the functionality back once poetry properly supports plugins (or custom build tags). Most likely in poetry 1.2 or 1.3

### LINKED ISSUE #2921: Unpin poetry-dynamic-versioning
This is a reminder that poetry-dynamic-versioning is pinned in the `test_deploy.yml` CI workflows and in the Dockerfile.

See #2920 

This should be updated or preferrably undone when the problems with the package (when it is not pinned) are resolved.

Panaetius:
Moves everything over to work with poetry 1.2.0b3. This is a beta version but we've waited long enough for a proper 1.2 release and the non-deterministic lock files are taking a lot of time and are a nuisance.

Switches extras for dependency groups where it makes sense (so now it's e.g. `poetry install --with all` instead of `pip install .[all]`). Only `toil` is still a proper extra.

~Note, poetry's plugin system is currently not really supported from pip, see https://github.com/python-poetry/poetry/issues/5701 I've added a plugin on our side that downloads templates on `poetry build` but this doesn't work when installing through `pip` because of this issue. Since we use `poetry build` for creating packages, this would only affect installing from a local repository, which I think we can live with for now.~ I added back the old makefile way of doing things, it just works better. Otherwise we'd hafr to make a separate package for the plugin that we install first.

closes #2921 
closes #2553


.github/workflows/test_deploy.yml
```diff
@@ -192,14 +188,12 @@ jobs:
         if: steps.dependency-cache.outputs.cache-hit != 'true' || 'refs/heads/master' == github.ref || 'refs/heads/develop' == github.ref || startsWith(github.ref, 'refs/tags/')
         run: |
           python -m pip install --upgrade pip
-          python -m pip install coveralls poetry==1.1.15 wheel twine poetry-dynamic-versioning==0.17.1
-          poetry-dynamic-versioning
+          python -m pip install coveralls poetry wheel twine poetry-dynamic-versioning-plugin
           make download-templates
           pip install .[all]
```
mohammad-alisafaee, line 193: Should we use poetry here as well?

.github/workflows/test_deploy.yml
```diff
@@ -629,15 +626,17 @@ jobs:
...
+        env:
+          POETRY_VIRTUALENVS_CREATE: false
         run: |
           brew update || true
           brew unlink git-lfs || true
           curl -L https://raw.githubusercontent.com/Homebrew/homebrew-core/43842898fd3ff43273466052722f5ba2789196cb/Formula/git-lfs.rb > git-lfs.rb && brew install git-lfs.rb && rm git-lfs.rb
           brew install shellcheck node || brew link --overwrite node
           python -m pip install --upgrade pip
-          python -m pip install wheel
+          python -m pip install wheel poetry
```
mohammad-alisafaee, line 637: Aren't `coveralls`, `twine` and `poetry-dynamic-versioning-plugin` needed here?

.github/workflows/test_deploy.yml
```diff
@@ -845,15 +844,17 @@ jobs:
...
```
mohammad-alisafaee, line 855: `coveralls`, `twine` and `poetry-dynamic-versioning-plugin` are missing.

Dockerfile
```diff
@@ -13,29 +13,28 @@ RUN apt-get install --no-install-recommends -y build-essential && \
...
 # renku-python is installed in the image. This is the default for chartpress builds.
 ARG CLEAN_INSTALL
 RUN if [ -n "${CLEAN_INSTALL}" ]; then git reset --hard ; fi && \
     make download-templates
 
-# set the BUILD_CORE_SERVICE to non null to install additional service dependencies
 ARG BUILD_CORE_SERVICE
-RUN if [ -n "${BUILD_CORE_SERVICE}" ]; then export EXT_BUILD="[service]" ; fi && \
-    pip wheel --wheel-dir /wheels ".${EXT_BUILD}" && \
+RUN pip wheel --wheel-dir /wheels ".[toil]" && \
```
mohammad-alisafaee, line 34: Just to be sure, we don't need `.[service]` here, right?

mohammad-alisafaee (commented):

.github/workflows/test_deploy.yml
Panaetius: We only run coveralls on linux (no point checking on mac as well). 
In the linux steps, we just include twine so the install is the same in all steps, because of the dependency cache (all installs need to be identical), but on the mac action we don't use a cache, so it's not needed

Dockerfile
Panaetius: I think this was a leftover from debugging things

Dockerfile
```diff
@@ -13,29 +13,29 @@ RUN apt-get install --no-install-recommends -y build-essential && \
...
-    pip install --no-cache-dir -r /tmp/requirements.txt && \
-    pip install --no-cache-dir poetry-dynamic-versioning==0.17.1
-
 COPY .git /code/renku/.git
 COPY renku /code/renku/renku
+WORKDIR /code/renku
+# hadolint ignore=SC2086
+RUN if [ -n "${BUILD_CORE_SERVICE}" ]; then export EXT_BUILD="-E service" ; fi && \
+    poetry export --without-hashes ${EXT_BUILD} -f requirements.txt --output /tmp/requirements.txt && \
+    pip install --no-cache-dir -r /tmp/requirements.txt
```
olevski, line 25: Why export the `requirements.txt` when you can make poetry install everything in a virtualenv in the builder docker image stage and then just copy that virtualenv over in the final stage. That is what I do in the notebook service. It saves space on the final image because you can tell poetry to not install a bunch of uneeded things like setuptools, poetry itself and others.

See here for example: https://github.com/SwissDataScienceCenter/renku-notebooks/blob/master/git_services/Dockerfile.sidecar

You do not need poetry, pip or setuptools to run the core service.

You do not have to change this now if you dont want to. Just open an issue about it.

olevski (changes requested):
One thing about including poetry in the core service image.

But if you want to address it in another PR that is ok. Just let me know.

mohammad-alisafaee (dismissed):

Dockerfile
Panaetius: I switched to using poetry, though I couldn't go with exactly the same approach as in the git_services sidecar.

olevski (approved):

PR #3090, opened 2022-08-24
chore: release v1.7.1

### LINKED ISSUE #3035: add files to dataset when creating a dataset with existing datadir
When doing `renku dataset create --datadir X [...]`, we should check if X already exists and contains files. 
If so, we should prompt the user if he wants to add the existing files to the dataset as well.

Panaetius:
closes #3035

automatically adds files in a datasets datadir on creation
allows adding files to datasets that are already in its datadir


renku/core/dataset/dataset.py
```diff
@@ -791,11 +793,38 @@ def show_dataset(name: str, tag: Optional[str] = None):
...
+        client(LocalClient): The ``LocalClient``.
+        dataset(Dataset): The dataset to add data dir files to.
+    """
+    datadir = get_safe_relative_path(dataset.get_datadir(), client.path)
+
+    if datadir.exists():
+        # NOTE: Add existing files to dataset
+        dataset_files: List[DatasetFile] = []
+        files: List[Path] = []
+        for file in datadir.rglob("*"):
```
mohammad-sdsc, line 809: This returns directories as well. You should use `renku.core.util.os.get_files`.

renku/core/dataset/dataset.py
```diff
@@ -174,6 +174,8 @@ def create_dataset(
     if images:
         set_dataset_images(client, dataset, images)
 
+    add_datadir_files_to_dataset(client, dataset)
```
mohammad-sdsc, line 177: We shouldn't add files if dataset has a remote storage like S3.

tests/cli/test_datasets.py
```diff
@@ -475,6 +496,10 @@ def test_add_and_create_dataset(
     assert 1 == result.exit_code
     assert 'Dataset "new-dataset" does not exist.' in result.output
 
+    existing_file = client.path / datadir / "myfile"
```
mohammad-sdsc, line 499: Maybe put `myfile` in a sub-dir to test that directories aren't added.

mohammad-sdsc (changes requested):

tests/cli/test_datasets.py
Panaetius: You mean empty directories? Or why not add subdirectories?

mohammad-sdsc: I mean `existing_file = client.path / datadir / sub-dir / "myfile"`. Then checking that `datadir / sub-dir` is not added to dataset's metadata. This test should fail now because of using `rglob`.

Panaetius: ok. That should be done (the test checks for files so if there was a folder as well, there'd be more than 4 paths returned

mohammad-sdsc (approved):
Thank you!

PR #3094, opened 2022-08-25
feat: scaling core service with sticky sessions

### LINKED ISSUE #3039: POC: sticky core service
Deploy two core services side-by-side with traefik load balancer in front with sticky sessions (cookie based) enabled and test it (use multiple browsers?).

Does it behave as expected? Are there any errors?
How is the latency?
What happens if one of the core-services goes down?

Timeboxed to two days to get some initial impression of sticky load-balancing in traefik.

olevski:
closes #3039 

Still WIP.

### DIFF
--- a/helm-chart/renku-core/templates/configmap.yaml
+++ b/helm-chart/renku-core/templates/configmap.yaml
@@ -4,14 +4,6 @@ metadata:
   name: {{ include "renku-core.fullname" . }}-nginx
 data:
   nginx-server-blocks.conf: |
-    {{- range $version := .Values.versions }}
-    upstream {{ $version.name }} {
-      server {{ include "renku-core.fullname" $ }}-{{ $version.name }};
-      keepalive 32;
-      keepalive_timeout 60s;
-    }
-    {{ end }}
-
     server {
       listen 8080;
 
@@ -21,31 +13,10 @@ data:
       tcp_nopush on;
       client_max_body_size 0; # Required for uploading large files
 
-      location /renku/version {
+      location /api/renku/version {
         root /;
         add_header Content-Type application/json;
         try_files /usr/share/nginx/html/version.json =404;
-      }
-
-      {{- range $version := .Values.versions }}
-      location /renku/{{ $version.prefix }} {
-        rewrite /renku/{{ $version.prefix }}/(.*) /renku/$1  break;
-        proxy_set_header Host $host;
-        proxy_pass http://{{ $version.name }};
-        proxy_send_timeout {{ $.Values.requestTimeout }}s;
-        proxy_read_timeout {{ $.Values.requestTimeout }}s;
-        proxy_http_version 1.1;
-        proxy_set_header "Connection" "";
-      }
-      {{- end }}
-
-      location /renku {
-        proxy_set_header Host $host;
-        proxy_pass http://{{ .Values.versions.latest.name }};
-        proxy_send_timeout {{ $.Values.requestTimeout }}s;
-        proxy_read_timeout {{ $.Values.requestTimeout }}s;
-        proxy_http_version 1.1;
-        proxy_set_header "Connection" "";
       }
     }
   version.json: |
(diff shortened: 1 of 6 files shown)


olevski:
So here is how I generated the following chart:
- ensured that the same session is used for one round of tests. One "round" are two results in the same row that belong either to the sticky session or "latest" scenario
- latest is what we have at dev.renku.ch (i.e. no sticky sessions)
- every time I ran a test I deleted the core service pods so that the cache is fully blank, I could not reliably purge the cache by deleting /svc/cache/*/*
- measured the response time for calling `/api/renku/datasets.list`

| Sticky sessions No cache | Sticky sessions With cache | Latest No cache | Latest With cache |
|--------------------------|----------------------------|-----------------|-------------------|
| 1.382287                 | 0.202083                   | 1.719212        | 0.365914          |
| 1.253002                 | 0.246281                   | 1.813227        | 0.455418          |
| 1.31778                  | 0.238344                   | 1.878961        | 0.393304          |

Conclusions:
- as expected caching saves about 1 to 1.5 seconds from the response time
- the sticky sessions do not introduce additional latency, they actually seem to have lower latency

olevski:
Ok the following changes were required to make the routing through the ui-server work (these are not part of this PR at all):
- the ui-server should not strip the cookies from incoming requests
- the gateway (i.e. traefik proxy) should not strip cookies from incoming requests for the core service
- the sticky session cookie path should be `/`, this is because the cookies should be used either for `/api/renku` or `/ui-server/api/renku` requests, the only common path for this is `/`
- the pathing when using the ui-server is different (see below)

The pathing for a request when it goes through the ui-server is as follows:
- renku k8s ingress
- ui-server
- gateway proxy (here the proxy redirects to outside of the cluster so that the sticky session cookie can be applied)
- renku-core k8s ingress
- core service

Here are 3 sets of response times when requests go through the ui-server through the route described above. These can be compared to the response times in the table in the previous comment.

| Sticky sessions No cache (w/ ui-server) | Sticky sessions With cache (w/ ui-server) | Latest No cache | Latest With cache |
|--------------------------|----------------------------|-----------------|-------------------|
| 1.488981                 | 0.254608                   | 1.719212        | 0.365914          |
| 1.513667                 | 0.323505                   | 1.813227        | 0.455418          |
| 1.532614                  | 0.2941                   | 1.878961        | 0.393304          |

olevski:
I also confirmed that the following scenario works out as expected:
- start core service with 1 replica
- send requests through browser to initiate sticky session cookie
- scale the core service out to 4 replicas
- sending new requests with the sticky session cookie results in all requests (that have the same cookie) going to the same replica and none of the new replicas get any requests

olevski:
@Panaetius should I clean up this PR and implement this? What do you think?

lorenzo-cavazzi:
> the ui-server should not strip the cookies from incoming requests

Just a tiny comment on this: there are some cookies that we want to strip out before forwarding the request from the UI server.
The best would be knowing what cookies need to be preserved for sticky sessions -- I assume it will be one (or more) cookie with a specific name.
It would be great to get that from the values files -- as a bonus feature we get almost for free, we might just add an array so that admins can extend the cookies that won't be stripped out.

Panaetius:
Could the decrease in latency be that with the PoC, requests don't go through the UI server?

Moving forward with this makes sense, though there are some things that we probably also want to address (maybe in follow-up issues):
- autoscaling (follow-up?)
- dealing with the RQ workers
  * Should we separate out the normal workers with their own ephemeral cache for background jobs and a common queue?
  * Some queues need to be local to each service (those clean up the local cache after ~30 minutes and also clean up the uploaded file cache)
  * if we decide to do the above in a follow-up, we need at least separate redis queues per core-service instance so they don't interfere with each other
- Dealing with error cases if e.g. a service goes down. For the project cache, this is easy, as a new service will just clone again on a new request. For uploaded files, it's more complicated. A user could upload, then the service goes down, then they click dataset create and the new service they hit will say "Sorry, the file you want to add doesn't exist". Might just need some adjustment of the wording.

olevski:
@Panaetius I will add issues for the next sprint(s) to address the points you brought up.

@lorenzo-cavazzi yeah I agree we do not have to simple avoid stripping all cookies. I just quickly typed up all the things I changed to make things work so we do not forget later. A list of cookie names that are left would be perfect. And you are absolutely correct `stickyCoreIngressCookie` is the name of the cookie that enables sticky sessions.

olevski:
@Panaetius I am mostly confused as to why there is a difference between the POC (labeled "sticky sessions" in tables above) and what is deployed at dev.renku.ch (these cases are called "latest" in the tables with response times). The difference is not that great but it is not 0. I am not sure if that is what you referred in your comment. But I expected that this POC would be worse or the same when compared to the latest stable renku release, not better.

If you look at the two cases of going through the ui-server and not going through the ui-server (solely for the POC) then it makes sense that skipping the ui-server is quicker.

Panaetius:
> I am not sure if that is what you referred in your comment.
Yes that's what I meant. I wondered if the ui-server or gateway (which isn't used in this POC) is responsible for the overhead.

olevski:
superseded by https://github.com/SwissDataScienceCenter/renku-python/pull/3178

PR #3105, opened 2022-08-31
chore: add project properties class and replace client.path

### LINKED ISSUE #3081: add a global config object and dispatcher
Add a global config object (Like `RenkuConfig`) that can hold configuration values a and a `(I)ConfigDispatcher` that can hold these config objects and return the current one from an internal stack. Set up injection for the dispatcher in the command builder.

The object does not need any properties for now, those will be added further down the line.

Panaetius:
closes #3081 

I didn't go with a Dispatcher/Interface/inject approach as I think this is easier to use, reduces clutter and and fits out needs. Plus we don't really need to mock this as it's pretty fixed and should be relatively simple.

I also fixed some issues that caused warnings in pytest. Some further warnings should be fixed by https://github.com/SwissDataScienceCenter/calamus/pull/79 .


mohammad-sdsc (commented):
Thanks Ralf! I believe `RenkuConfig` is not a fully descriptive name for this. It's also can get confused with a Renku project config (`renku.ini`). The path is not a configuration; we cannot change it as normally we do with configuration values. It's a core property and has a one-to-one correspondence with a project (when it changes we are in a new project). Maybe `ProjectProperties` or `ProjectInfo` is a more descriptive name. What do you think?

Also, it's good if we move it out of the `management` module since we are going to remove this namespace once we get rid of the `LocalClient`.

mohammad-sdsc (approved):
Thank you!

mohammad-sdsc (dismissed):

mohammad-sdsc (approved):

PR #3106, opened 2022-08-31
feat(dataset): allow mounting s3 data

### LINKED ISSUE #2974: Mount data from S3
As a Renku user I want to mount data from S3 so that I can use this data in Renku workflows without copying them to my machine.
```
renku dataset stage my-dataset --mount /mount/point
```
## Acceptance Criteria
- Add a command/subcommand to mount data for datasets that have a S3 storage backend.
- Include an `--existing` flag that allows using an existing mount-point

## Notes
- We don't support multiple storage sources.

mohammad-sdsc:
# Description

Adds a `mount` command to allow mounting data from S3 storages. 

Fixes #2974

### DIFF
--- a/renku/core/dataset/dataset.py
+++ b/renku/core/dataset/dataset.py
@@ -1285,6 +1285,9 @@ def pull_external_data(
         with communication.busy(f"Copying {file.entity.path} ..."):
             storage.copy(file.source, path)
 
+            # NOTE: Make files read-only since we don't support pushing data to the remote storage
+            os.chmod(path, 0o400)
+
             if file.based_on and not file.based_on.checksum:
                 md5_hash = hash_file(path, hash_type="md5") or ""
                 file.based_on = RemoteEntity(checksum=md5_hash, url=file.based_on.url, path=file.based_on.path)
@@ -1327,9 +1330,7 @@ def read_dataset_data_location(dataset: Dataset) -> Optional[str]:
 def mount_external_storage(
     name: str,
     existing: Optional[Path],
-    unmount: bool,
     yes: bool,
-    client_dispatcher: IClientDispatcher,
     storage_factory: IStorageFactory,
 ) -> None:
     """Mount an external storage to a dataset's data directory.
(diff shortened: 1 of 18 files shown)


renku/core/dataset/dataset.py
```diff
@@ -1315,3 +1321,60 @@ def store_dataset_data_location(dataset: Dataset, location: Optional[Path]) -> N
...
+    client_dispatcher: IClientDispatcher,
+    storage_factory: IStorageFactory,
+) -> None:
+    """Mount an external storage to a dataset's data directory.
+
+    Args:
+        name(str): Name of the dataset
+        existing(Optional[Path]): An existing mount point to use instead of actually mounting the external storage.
+        unmount(bool): Unmount the external storage.
+        yes(bool): Don't prompt when removing non-empty dataset's data directory.
```
Panaetius, line 1341: I know it's annoying but to satisfy the pre-commit check (I think `pydocstyle` one), we have to document every argument in the docstring, including the two injected ones. I guess you haven't updated/installed the `pre-commit` config?

I prefer adding everything but then being sure that we don't miss a relevant parameter instead of not checking but not having to document the injected parameters, as it's safer that way.

renku/core/dataset/dataset.py
```diff
@@ -1315,3 +1321,60 @@ def store_dataset_data_location(dataset: Dataset, location: Optional[Path]) -> N
...
+    if not dataset.storage and not unmount:
+        communication.warn(f"Cannot mount dataset '{name}': Dataset doesn't have a storage backend")
+        return
+
+    datadir = client.path / dataset.get_datadir()
+
+    # NOTE: Try to unmount the path in case it was mounted before
+    unmount_path(datadir)
+
+    if unmount:
```
Panaetius, line 1357: it's a bit weird that `mount_external_storage(...)` does the exact opposite of its name if `unmount` is set true. Would be nicer if it's two separate methods (and `mount_external_storage` can call `unmount_external_storage`)

renku/core/util/os.py
```diff
@@ -144,6 +147,28 @@ def delete_path(path: Union[Path, str]) -> None:
...
+
+    path = str(path)
+
+    # NOTE: A symlink means that the path is not mounted itself but it's a link to a mount-point; just delete the link.
+    if os.path.islink(path):
+        os.remove(path)
+        return
+
+    # NOTE: ``fusermount`` is available on linux and ``umount`` is for macOS
+    execute_command("fusermount", "-u", "-z", path) or execute_command("umount", path)
```
Panaetius, line 169: I'd use https://docs.python.org/3/library/shutil.html#shutil.which instead, seems cleaner.

Panaetius (changes requested):
Looks good overall. I was expecting a much larger PR :slightly_smiling_face:

renku/core/dataset/dataset.py
mohammad-sdsc: > I guess you haven't updated/installed the pre-commit config?

No,  I don't.

Panaetius (dismissed):

Panaetius (approved):

PR #3111, opened 2022-09-02
fix(cli): make file size display consistent

### LINKED ISSUE #3072: make file size display in the CLI consistent
`renku dataset import` shows a header like `Size (B)` and then sizes like `0.5`, whereas `renku dataset ls-files` has a header `Size` and then shows sizes like `0.5 mb` on each entry.

We should make sure all our file size display is consistent and make sure it is correct (e.g. for dataset import the size shown is not actually in Bytes). To this end we should remove the unit from the headers and show it on each entry, with the applicable highest unit that makes sense (MB if bigger than 1MB, kb if bigger than 1 KB etc.).

Panaetius:
closes #3072


mohammad-sdsc (approved):
LGTM, thank you!

PR #3125, opened 2022-09-06
test: fixes flaky test with timing issues

Panaetius:
if two reruns happened in the same second, renku couldn't create an order between them. this fixes that.


mohammad-sdsc (approved):
Thanks!

PR #3132, opened 2022-09-12
docs: support json builder for cheatsheet

### LINKED ISSUE #2960: add missing commands to autogenerated cheatsheet
We should add `renku session` commands to the cheatsheet (in their own section).

We should also add `renku template` commands to the cheatsheet (also in their own section)

`renku rollback` should be added to the cheatsheet as well.

Check if there are any other non-internal commands missing

### LINKED ISSUE #3098: add JSON output for cheatsheet to be used in UI
Initial support was added in a branch in the branch https://github.com/SwissDataScienceCenter/renku-python/tree/feature/cheatsheet-json-output (not merged to develop/master).

The structure of that commit should already match the structure in the UI issue https://github.com/SwissDataScienceCenter/renku-ui/issues/1920 

We should rename/-arrange our groups as outlined in that UI section, add the new commands mentioned and consolidate the cheatsheets, with our PDF showing all entries and UI being able to only show `extended=False` entries.

Probably also makes sense to address https://github.com/SwissDataScienceCenter/renku-python/issues/2960 as there is a big overlap.

This file should be automatically generated in our CI, same as the `cheatsheet.pdf` that we already generate.

Panaetius:
Fixes #3098 
Fixes #2960 

### TODO:
- Having a `Typical Workflow` title and a `Workflows` title is a bit confusing. Should we rename one of them?


Panaetius:
`Typical Workflow` is for UI and is just renku commands in general, typical steps a user might do in a session, `Workflows` is actual `renku workflow` type stuff.

Thank you for your changes, looks good to go :+1:

mohammad-sdsc:
> `Typical Workflow` is for UI and is just renku commands in general, typical steps a user might do in a session, `Workflows` is actual `renku workflow` type stuff.

I know the difference and that's my point here: We are using the same word in two different contexts.

I believe AFA this PR is concerned we can merge it and then discuss that issue with other teams.

mohammad-sdsc (approved):

PR #3134, opened 2022-09-14
feat(svc): add workflow_plans.list endpoint

Panaetius:
activity list endpoint

/deploy #persist

### DIFF
--- a/renku/core/workflow/plan.py
+++ b/renku/core/workflow/plan.py
@@ -19,8 +19,9 @@
 
 import itertools
 from datetime import datetime
+from functools import lru_cache
 from pathlib import Path
-from typing import Dict, Generator, List, Optional, Union, cast
+from typing import Dict, Generator, List, Optional, Set, Tuple, Union, cast, overload
 
 from renku.command.command_builder import inject
 from renku.command.format.workflow import WORKFLOW_FORMATS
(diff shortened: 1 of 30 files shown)


renku/core/workflow/plan.py
```diff
@@ -644,3 +684,90 @@ def get_composite_plans_by_child(plan: AbstractPlan, plan_gateway: IPlanGateway)
     composites_containing_child = [c for c in composites if {p.id for p in c.plans}.intersection(derivatives)]
 
     return composites_containing_child
+
+
+@inject.autoparams("activity_gateway", "plan_gateway")
+def get_active_plans(activity_gateway: IActivityGateway, plan_gateway: IPlanGateway) -> List[AbstractPlan]:
+    """Get all plans that are active in the project.
+
+    Active means not deleted and with at least one of the used/created files existing.
```
mohammad-sdsc, line 693: I don't see that we filter the results based on the `touches_existing_files`; we add all the plans regardless. If that's intended then this description should change and wise-versa.

renku/core/workflow/plan.py
```diff
@@ -644,3 +684,90 @@ def get_composite_plans_by_child(plan: AbstractPlan, plan_gateway: IPlanGateway)
...
+    """
+    from renku.core.workflow.activity import filter_overridden_activities
+
+    latest_plan_cache: Dict[AbstractPlan, AbstractPlan] = {}
+
+    def _cached_latest_plan(plan: AbstractPlan):
+        return latest_plan_cache.setdefault(plan, get_latest_plan(plan))
+
+    all_activities = activity_gateway.get_all_activities()
+    relevant_activities = filter_overridden_activities(all_activities)
```
mohammad-sdsc, line 703: I believe all plans which their activity is in this list are a superset of the results that we want (we are filtering the plans using `relevant_activities` in line 733). We just need to filter out those that don't have an existing output. So, we can remove the code that follows this line and replace it with this filtering. Does that sound right to you?

mohammad-sdsc (commented):

renku/core/workflow/plan.py
Panaetius: Right, we decided to return everything and have the UI filter based on `touches_existing_files`, I have to rename/reword that.

renku/core/workflow/plan.py
Panaetius: as mentioned above, we want all latest activities, not just "active" ones. 

And we want to also have composite plans, which never have an activity associated with them

mohammad-sdsc (approved):

PR #3135, opened 2022-09-16
feat(svc): workflow ui support

Panaetius:
feature branch for workflow ui feature


renku/command/view_model/composite_plan.py
```diff
@@ -43,12 +44,13 @@
...
 
 
-def _parameter_id_to_plan_id(parameter_id: str):
+def _parameter_id_to_plan_id(parameter: CommandParameterBase, parent_plan: AbstractPlan, latest: bool = False):
     """Extract plan id from a parameter id."""
-    parts = parameter_id.split("/")
-    parts = parts[:-2]
-    id = "/".join(parts)
-    return f"/{id}"
+    containing_plan = parent_plan.get_parameter_path(parameter)[-1]
```
mohammad-alisafaee, line 49: We are sure that this never returns `None`, right?

renku/command/view_model/plan.py
```diff
@@ -69,10 +71,11 @@ def from_input(cls, input: CommandInput, plan_id: Optional[str] = None):
             name=input.name,
             description=input.description,
             default_value=str(input.default_value),
-            position=str(input.position),
+            position=str(input.position) if input.position else None,
```
mohammad-alisafaee, line 74: `if input.position is not None`!? `position` can be 0 too, although it might not be assigned to parameters. I still think we should check against `None` since the value is an integer.

renku/command/view_model/plan.py
```diff
@@ -157,7 +163,7 @@ def from_parameter(cls, parameter: CommandParameter, plan_id: Optional[str] = No
             name=parameter.name,
             description=parameter.description,
             default_value=str(parameter.default_value),
-            position=str(parameter.position),
+            position=str(parameter.position) if parameter.position else None,
```
mohammad-alisafaee, line 166: Same as above.

renku/core/workflow/plan.py
```diff
@@ -141,10 +141,12 @@ def show_workflow(
     Returns:
         Details of the Plan.
     """
+    from renku.command.view_model.plan import plan_view
+
     workflow = cast(Union[Plan, CompositePlan], plan_gateway.get_by_name_or_id(name_or_id))
 
     if is_plan_removed(workflow):
-        raise errors.ParameterError(f"The specified workflow '{name_or_id}' cannot be found.")
+        raise errors.ParameterError(name_or_id)
```
mohammad-alisafaee, line 149: Should we raise `WorkflowNotFoundError` instead of `ParameterError`?

mohammad-alisafaee (commented):
Thank Ralf! Please see my comments.

mohammad-alisafaee (approved):

PR #3149, opened 2022-09-26
chore: add automated release workflow

Panaetius:
Add a manually triggered workflow that creates a release PR and draft release.

Example PR can be seen here: https://github.com/SwissDataScienceCenter/renku-python/pull/3148
Example release draft can be seen here: https://github.com/SwissDataScienceCenter/renku-python/releases/tag/untagged-bd9c50875f99cdb6d673 (please delete once this is merged)

yq does reformat our values.yaml when setting the version, but I think we can live with that.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,61 @@
+name: "Create Release PR"
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: New release version
+        type: string
+        required: true
+
+jobs:
+  create-release-pr:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


olevski:
@Panaetius it is not clear to me how this workflow triggers? On the branch name?

Panaetius:
@olevski it's `workflow_dispatch`, so you'll be able to go something similar to https://github.com/SwissDataScienceCenter/renku-python/actions/workflows/combine-dependabot-updates.yml and click the `run workflow` button, then it should prompt for the `version` parameter and then run.

olevski (approved):
Looks great! Thanks!

PR #3150, opened 2022-09-26
chore: fix cheatsheet action step ids

Panaetius:
the id's were copy pasted but you can't have the same step id twice.


mohammad-sdsc (approved):
Thank you!

PR #3151, opened 2022-09-28
chore: fix cleanup of service test branches

Panaetius:
the branch was deleted in the fixture but the deletion was not pushed to the remote


mohammad-sdsc (approved):
Thank you!

PR #3159, opened 2022-10-10
feat(dataset): support non-AWS S3 URI

### LINKED ISSUE #3133: Support for non-AWS S3 providers
As a Renku user, I want to be able to add data from all S3 providers, so that I can access my data in S3 storages other than AWS.

## Notes
- Currently, we only support AWS S3 storage with a URI format of `s3://<bucket>/path/to/data`
- We should adapt the URI to include an endpoint/host that specifies the provider: `s3://<host>/<bucket>/path/to/data`
- We need the `s3://` scheme since in future we want to support other storage backends in addition to S3
- If a host is not provided, then the default provider will be AWS. If this is not feasible to implement, it's ok to always require the host.
- See https://swiss-data-science.slack.com/archives/C0133UVATF1/p1662388842159379 for some background and a working non-AWS provider for testing

## Questions:
- Should we add a `renku storage create` command that allows users to define all info for a provider and then use this `storage` in the dataset commands?

mohammad-sdsc:
# Description

Accepts `s3://<endpoint>/<bucket-name>/<path>` URI format for s3 backends. Credentials is stored per `endpoint` instead of for all `s3` backends.

Fixes #3133


rokroskar:
When I do 

```
renku dataset add <dataset-name> s3://<endpoint>/<bucket>/<path>
```

I see a 

```
Warning: No new file was added to project
```

is that expected?

mohammad-sdsc:
> is that expected?

Yes. This is because we don't add those files to the git repo and actually we don't download them at this point. You need to run `renku dataset pull` to get the files.

renku/core/interface/storage.py
```diff
@@ -90,6 +95,5 @@ def mount(self, path: Union[Path, str]) -> None:
         raise NotImplementedError
 
     @abc.abstractmethod
-    def set_configurations(self) -> None:
-        """Set required configurations to access the storage."""
-        raise NotImplementedError
+    def run_rclone_command(self, command: str, uri: str, *args, **kwargs) -> Any:
+        """Run a RClone command by possibly add storage-specific flags."""
```
Panaetius, line 99: No `raise NotImplementedError` here?

Panaetius (changes requested):

renku/core/interface/storage.py
mohammad-sdsc: I've refactored this in https://github.com/SwissDataScienceCenter/renku-python/pull/3163 and it's removed from the interface. So, this will be addressed once that PR is merged if it's ok with you.

Panaetius (approved):

PR #3163, opened 2022-10-11
feat(dataset): copy data to s3

### LINKED ISSUE #3137: Copy data to S3
As a Renku user, I want to add local/modified data to a dataset with S3 backend and then copy the data to the S3 storage, so that I have full read/write S3 support in Renku.

## Acceptance Criteria
- Local data can be added to a dataset with S3 backend
- Data must be copied to the remote S3 storage when adding the data to the dataset. This should be a blocking operation.

## Notes
- Data that is not yet pushed to the remote storage, should be protected against deletion/modification. We can also assume that it's user's responsibility to protect the data.
- Checksum calculation should be consistent with the remote S3.
- ATM, we don't have a writable S3 storage for testing this feature.
- The data can be in a different location if dataset was pulled with the `--location` option. Make sure to support it as well.

mohammad-alisafaee:
# Description

Fixes #3137 

Some refactoring in the dataset code, mainly, splitting the `ProviderApi` interface into multiple interfaces. Also, moving the `S3Storage` code to `renku.core`.

# TODO:
- Rename `dataset_provider` to `data_provider`
- Move `providers` directory from `core.dataset` to `core`


renku/core/dataset/dataset_add.py
```diff
@@ -246,73 +204,198 @@ def _create_destination_directory(dataset: Dataset, destination: Optional[Union[
...
             shutil.move(file.source, file.destination, copy_function=shutil.copy)  # type: ignore
         elif file.action == DatasetAddAction.SYMLINK:
             create_external_file(target=file.source, path=file.destination)
+            # NOTE: Don't track symlinks to external files in LFS
+            track_in_lfs = False
+        elif file.action == DatasetAddAction.DOWNLOAD:
+            assert file.provider, f"Storage provider isn't set for {file} with DOWNLOAD action"
+            storage = file.provider.get_storage()
+            storage.download(file.url, file.destination)
+        elif file.action == DatasetAddAction.METADATA_ONLY:
```
Panaetius, line 337:
```suggestion
        elif file.metadata_only:
```

renku/core/dataset/dataset_add.py
```diff
@@ -246,73 +204,198 @@ def _create_destination_directory(dataset: Dataset, destination: Optional[Union[
...
+            DatasetAddAction.DOWNLOAD,
+        ):
+            # NOTE: Remove existing file if any; required as a safety-net to avoid corrupting external files
+            delete_dataset_file(file.destination, follow_symlinks=True)
+            file.destination.parent.mkdir(parents=True, exist_ok=True)
+
+        track_in_lfs = True
+
+        # NOTE: If file is in a sub-directory of a dataset's remote storage URI, only update the metadata
+        if file.action == DatasetAddAction.REMOTE_STORAGE:
```
Panaetius, line 319:
```suggestion
        if file.remote_storage:
```

tests/cli/test_datasets.py
```diff
@@ -2723,3 +2723,32 @@ def test_update_with_no_dataset(runner, project):
...
+    assert 3 == len(dataset.files)
+    assert {f"{uri}/directory_tree/file1", f"{uri}/directory_tree/dir1/file2", f"{uri}/directory_tree/dir1/file3"} == {
+        f.based_on.url for f in dataset.files
+    }
+
+    assert {
+        "9d98eede4ccb193e379d6dbd7cc1eb86",
+        "7bec9352114f8139c2640b2554563508",
+        "dacb3fd4bbd9ab5a17e2f7686b90c1d2",
+    } == {f.based_on.checksum for f in dataset.files}
```
Panaetius, line 2754: Should we test that `upload` was called on the mocked `storage_factory` ?

Panaetius (changes requested):

renku/core/dataset/dataset_add.py
```diff
@@ -246,73 +204,198 @@ def _create_destination_directory(dataset: Dataset, destination: Optional[Union[
...
+    files = list(itertools.chain(local_files, remote_files))
+
+    return files
 
-def move_files_to_dataset(files: List["DatasetAddMetadata"]):
+
+def get_dataset_file_path_within_dataset(dataset: Dataset, entity_path: Union[Path, str]) -> Path:
+    """Return a dataset file's path relative to the dataset's datadir.
+
+    NOTE: Dataset files can reside outside the dataset's datadir.
```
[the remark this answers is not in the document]
mohammad-alisafaee: Yes. AFAIK, we don't force moving files outside the datadir to the datadir (via a migration). It's a `doctor` fix which might not be called by users. So, projects can still have files that are outside their dataset's data directory.
Although we call this function only on the new datasets with storage, we might use it at some point for the old datasets and I wanted it to be functionally correct in those cases.

Panaetius (approved):

PR #3165, opened 2022-10-13
feat(cli): set list of custom metadata for project and dataset

### LINKED ISSUE #3120: allow a list of annotations(custom_metadata) to be passed to `project_edit` and `dataset_edit`
We currently only accept dict's for the `custom_metadata` parameter in `renku project edit` and `renku dataset edit`, which replace existing annotations if set. This does not allow adding more than one annotation to one of those entities, which is necessary for e.g. the renku-bio plugin. Additionally, we set the `source` hardcoded to `renku` but it would be nice to allow setting it to something else, like `renku-bio`.

We should add a new `--metadata-source` parameter that defaults to `renku` and should expand the edit/update logic as follows:
- If custom_metadata is a dictionary -> We create one annotation object for it, replacing any existing annotation with the same `source`
- If custom_metadata is a list of dicts -> we create one annotation per dict, replacing all existing annotations with the same `source`

and we should verify that it is one of the two, not something else.

Additionally, it'd be nice if we could verify the dict(s) to check that they contain valid JSON-LD (maybe with pyld?), if possible.

olevski:
closes #3120 

I was not able to figure out how to validate jsonld.

### DIFF
--- a/renku/core/project.py
+++ b/renku/core/project.py
@@ -34,7 +34,8 @@ def edit_project(
     description: Optional[Union[str, NoValueType]],
     creator: Union[Dict, str, NoValueType],
     keywords: Optional[Union[List[str], NoValueType]],
-    custom_metadata: Optional[Union[Dict, NoValueType]],
+    custom_metadata: Optional[Union[Dict, List[Dict], NoValueType]],
+    custom_metadata_source: str,
     project_gateway: IProjectGateway,
 ):
     """Edit dataset metadata.
@@ -43,7 +44,8 @@ def edit_project(
         description(Union[Optional[str], NoValueType]): New description.
         creator(Union[Dict, str, NoValueType]): New creators.
         keywords(Union[Optional[List[str]]): New keywords.
-        custom_metadata(Union[Optional[Dict]): Custom JSON-LD metadata.
+        custom_metadata(Union[Optional[Dict, List[Dict]]): Custom JSON-LD metadata.
+        custom_meatadata_source(Optional[str]): Custom metadata source.
         project_gateway(IProjectGateway): Injected project gateway.
 
     Returns:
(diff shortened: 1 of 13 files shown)


renku/core/project.py
```diff
@@ -43,7 +44,8 @@ def edit_project(
         description(Union[Optional[str], NoValueType]): New description.
         creator(Union[Dict, str, NoValueType]): New creators.
         keywords(Union[Optional[List[str]]): New keywords.
-        custom_metadata(Union[Optional[Dict]): Custom JSON-LD metadata.
+        custom_metadata(Union[Optional[Dict, List[Dict]]): Custom JSON-LD metadata.
+        custom_meatadata_source(Optional[str]): Custom metadata source.
```
Panaetius, line 48:
```suggestion
        custom_metadata_source(Optional[str]): Custom metadata source.
```
:meat_on_bone:

renku/domain_model/project.py
```diff
@@ -187,15 +187,16 @@ def update_metadata(self, custom_metadata=None, custom_metadata_source=None, **k
...
-                        id=Annotation.generate_id(),
-                        body=icustom_metadata,
-                        source=custom_metadata_source,
+                if icustom_metadata is not None:
+                    existing_metadata.append(
+                        Annotation(
+                            id=Annotation.generate_id(),
+                            body=icustom_metadata,
+                            source=custom_metadata_source,
+                        )
```
Panaetius, line 199: I think you just want

```
        if custom_metadata is not None:
            if not isinstance(custom_metadata, list):
                custom_metadata = [custom_metadata]
            for icustom_metadata in custom_metadata:
                existing_metadata.append(
                    Annotation(
                        id=Annotation.generate_id(),
                        body=icustom_metadata,
                        source=custom_metadata_source,
                    )
```

olevski:
@Panaetius this is ready now

Panaetius (approved):
Thank you so much! looks great :pray:

PR #3171, opened 2022-10-17
chore: pin poetry in github actions

Panaetius:
the action to publish the `renku_lock` package didn't work properly with poetry 1.2.x. Other actions seem to work, but just in case, it's probably safer to pin it everywhere until we switch.


mohammad-sdsc (approved):
Thank you!

PR #3179, opened 2022-10-21
fix(service): file upload memory leak

olevski:
Trying to fix memory usage when uploading files.


Panaetius:
if you do `pre-commit install` it'll format code for you before committing, then you don't end up with the style-check errors you had here

PR #3187, opened 2022-10-27
chore: update precommit

Panaetius:
Adds some more pre-commit checks, updates revisions of checks, properly formats file

New checks are:
- check filename case for python files (`MyPythonFile.py` is invalid)
- ignore helmcharts in check-yaml (it's not proper yaml usually because of templating)
- check no big files are committed to git
- check TOML file formatting
- check not `>>>> HEAD` style git merge markers are committed
- check not `breakpoint()` style debug statements are committed
- add some more options for mypy checks
- use `yesqa` to warn about unneeded `# noqa` comments
- use `bandit` to run security checks
- use `poetry check` to check pyproject.toml


renku/core/util/metadata.py
```diff
@@ -42,7 +42,7 @@ def construct_creators(creators: List[Union[dict, str]], ignore_email=False) ->
     """Parse input and return a list of Person."""
     creators = creators or []
 
-    if not isinstance(creators, Iterable) or isinstance(creators, str):
+    if not isinstance(creators, Iterable):
```
mohammad-alisafaee, line 45: Why is this removed? Here we except a list of creators and a string would mean a single creator.

mohammad-alisafaee (commented):
LGTM, thanks! Please see my comment.

renku/core/util/metadata.py
Panaetius: That part is a bit annoying, mypy will complain about dead code, since as per the type signature, this code is nonsensical. Actually the whole `if` would be, I'm not sure why it doesn't complain about the first part.

But of course in python you could pass whatever since the types aren't enforced, and mypy will pretend they're enforced (since they are for our code, by mypy) and complain

This discrepancy is what led me to open/write https://github.com/SwissDataScienceCenter/renku-python/issues/3188 , which got completed first. We don't want that everywhere for performance reasons, but having it around our core layer to protect from the outside world makes sense to me. So I think the whole `if` above could be removed now.

mohammad-alisafaee (approved):

mohammad-alisafaee (approved):

PR #3189, opened 2022-11-01
tests: fix test repo urls to match new gitlab cloudnative form

Panaetius:
With the cloud-native chart, urls are now `gitlab.dev.renku.ch` instead of `dev.renku.ch/gitlab`

### DIFF
--- a/tests/cli/test_init.py
+++ b/tests/cli/test_init.py
@@ -148,12 +148,12 @@ def test_init_initial_branch(isolated_runner, project_init):
     "remote",
     [
         (
-            "https://user:password@dev.renku.ch/gitlab/group/subgroup/project.git",
+            "https://user:password@gitlab.dev.renku.ch/group/subgroup/project.git",
             "https://dev.renku.ch/projects/group/subgroup/project",
         ),
         ("ssh://@dev.renku.ch:group/subgroup/project.git", "https://dev.renku.ch/projects/group/subgroup/project"),
         (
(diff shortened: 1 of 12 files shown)


mohammad-alisafaee (commented):
Thanks Ralf for the changes! I was wondering if there are places in code that we rely on the previous format to parse Renku gitlab URL. A search for "gitlab" (with whole-word filter) shows some places in the code that use this:
https://github.com/SwissDataScienceCenter/renku-python/blob/42aee90b893fd24b0b7cf968f7958ce46905e88f/renku/domain_model/git.py#L49
https://github.com/SwissDataScienceCenter/renku-python/blob/42aee90b893fd24b0b7cf968f7958ce46905e88f/renku/domain_model/project_context.py#L353
There are few more places. Also in some places we use the namespace (e.g. `"gitlab/renku-qa"`)

Panaetius:
> `_RE_PATHNAME_WITH_GITLAB`
The gitlab is optional here (`(gitlab/){0,1}`) so it's not affected I think. And there might be an external gitlab deployment someone uses that still has /gitlab.

> if len(hostname_parts) > 2 and hostname_parts[0] == "gitlab":
that code just removes `gitlab` from the path if it's the first part, so also doesn't break things

> There are few more places. Also in some places we use the namespace (e.g. "gitlab/renku-qa")
I did run those test and they passed, so weren't affected. But I've removed the gitlab part now and they still work. I think it's just dummy metadata anyways that's not actually used by the test

There's some other tests failing that I'll look into

mohammad-alisafaee (approved):
LGTM, thanks!

PR #3193, opened 2022-11-04
fix(service): fix import of private datasets with gitlab cloud native

Panaetius:
Fixes https://github.com/SwissDataScienceCenter/renku-python/actions/runs/3387590517/jobs/5628534121#step:12:1140


mohammad-alisafaee (approved):
Thank you!

PR #3195, opened 2022-11-04
fix(service): remove project cache directory before clone

Panaetius:
If two simultaneous requests happen and a project is not in cache, but try to clone, with one waiting for the lock of the other to be released. If the first one fails but leaves data in the cache directory, the second wouldn't be able to clone because the target directory isn't empty.


olevski:
@Panaetius I can confirm that this fixes the problem we saw in https://github.com/SwissDataScienceCenter/renku/pull/2771. After pointing the CI deployment to this fix the issue we saw before is gone.

PR #3203, opened 2022-11-14
feat(load-test): Add load test for classroom scenario

li-il-li:
This PR adds a load test that simulates a starting class:

1. Login
2. Fork an example project
3. Start session
4. Wait for session to start
5. Let session run for n seconds
6.  Shutdown session
7. Cleanup/delete forked project

Additionally a custom metric + trend is introduced which measures the servers
startup time.

I was running into some problems, where I was getting 401
errors forking the project, even after successfull login.
This has not been resolved yet.


olevski:
@li-il-li can you move this over to the [notebooks repo](https://github.com/SwissDataScienceCenter/renku-notebooks)? I think it targets pretty much exclusively renku-notebooks components.

load-tests/class-scenario.js
```diff
@@ -0,0 +1,244 @@
...
+      vus: 10, // tested up to 30
+      iterations: 1,
+    },
+  },
+};
+
+// k6 custom metrics
+const sessionStartupTrend = new Trend("session_startup");
+
+function httpRetry(httpRequest, n, logMessage) {
```
olevski, line 28: @li-il-li sorry for the long delay in reviewing this. But I think that this function hides exactly the problems that we want the load tests to uncover.

Can you please remove it and I will approve.

If requests need to be retried in the load tests then please submit issues for it in the corresponding components.

olevski (changes requested):
One thing about the retries. I think they are valuable insights (i.e. bugs) from the load testing. We should fail the load tests if API requests need to be retried.

olevski:
Ok lets just get this in because the added tests are definitely useful.

I opened an issue to look a bit more into why some requests need retries: #3240

PR #3210, opened 2022-11-14
fix(workflow): set plan creation date to activity start date

### LINKED ISSUE #3208: Plan's creation date is after Activity's start date
Creation date of Plans that are created by `renku run` is set after the start date of the created activity. This should be fixed so that Plan's creation date is equal or less than Activity's start date.

mohammad-alisafaee:
# Description

Set Plan's creation date to be equal to Activity's start date.

Fixes #3208


Panaetius (approved):
Thank you

PR #3212, opened 2022-11-16
feat(service): add workflow export endpoint

### LINKED ISSUE #3190: add a workflow export endpoint on the core service
We should add a `workflow_plans.export` endpoint to the core service that exposes `renku workflow export` functionality. It should take a git url, branch, workflow name or id and target format (limited to CWL at the moment) and return valid CWl code in its payload.

This feature is useful for smartsky, as it makes it easy to use in other service for executing workflows.

olevski:
closes #3190 

Adds a workflow export endpoint.

I had to also re-format a big part of the cwl export methods to:
- not use absolute paths
- not use multiple workflow files (but just one with embedded workflows)

/deploy #persist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -123,7 +123,7 @@ rdflib = "<7.0,>=6.0.0"
 redis = { version = ">=3.5.3,<4.2.0", optional = true }
 renku-sphinx-theme = { version = ">=0.2.0", optional = true }
 requests = ">=2.23.0,<2.28.2"
-responses = "^0.22.0"
+responses = { version = ">=0.7.0,<=0.22.0", optional = true }
 rich = ">=9.3.0,<12.6.0"
 rq = { version = "==1.11.0", optional = true }
 rq-scheduler = { version = "==0.11.0", optional = true }
@@ -343,6 +343,7 @@ files = [
     "renku/**/*.py",
     "tests/**/*.py"
 ]
+implicit_optional = true
 
 [[tool.mypy.overrides]]
 module = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -123,7 +123,7 @@ rdflib = "<7.0,>=6.0.0"
 redis = { version = ">=3.5.3,<4.2.0", optional = true }
 renku-sphinx-theme = { version = ">=0.2.0", optional = true }
 requests = ">=2.23.0,<2.28.2"
-responses = { version = ">=0.7.0,<0.22.0", optional = true }
+responses = "^0.22.0"
 rich = ">=9.3.0,<12.6.0"
 rq = { version = "==1.11.0", optional = true }
 rq-scheduler = { version = "==0.11.0", optional = true }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,7 +83,7 @@ importlib-resources = { version = ">=5.4.0,<5.10.0", python = "<3.9.0" }
 inject = "<4.4.0,>=4.3.0"
 isort = { version = "<5.10.2,>=5.3.2", optional = true }
 jinja2 = { version = ">=2.11.3,<3.1.3" }
-marshmallow = { version = ">=3.14.0,<3.18.0", optional = true }
+marshmallow = { version = ">=3.18.0", optional = true }
 marshmallow-oneofschema = { version=">=3.0.1,<4.0.0", optional = true }
 mypy = {version = ">=0.942,<1.0", optional = true}
 networkx = "<2.7,>=2.6.0"
--- a/renku/core/workflow/converters/cwl.py
+++ b/renku/core/workflow/converters/cwl.py
@@ -21,7 +21,7 @@ import os
 import re
 import tempfile
 from pathlib import Path
-from typing import Any, Dict, Optional, Tuple, Union
+from typing import Any, cast, Dict, List, Optional, Tuple, Union
 from uuid import uuid4
 
 import cwl_utils.parser.cwl_v1_2 as cwl
@@ -109,6 +109,7 @@ class CWLExporter(IWorkflowConverter):
         if output:
             if output.is_dir():
                 tmpdir = output
+                filename = None
             else:
                 tmpdir = output.parent
                 filename = output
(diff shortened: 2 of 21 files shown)


renku/ui/service/controllers/workflow_plans_export.py
```diff
@@ -0,0 +1,63 @@
...
+    def renku_op(self):
+        """Renku operation for the controller."""
+        result = (
+            export_workflow_command()
+            .build()
+            .execute(
+                name_or_id=self.ctx["plan_id"],
+                format=self.ctx["format"].value,
+                output=None,
+                values=None,
```
Panaetius, line 52: Should we support passing a values dict? Might be especially useful in a service endpoint that's meant for re-use of workflows.

tests/utils.py
```diff
@@ -465,3 +466,10 @@ def assert_rpc_response(response, with_key="result"):
     assert response and 200 == response.status_code
 
     assert with_key in response.json.keys(), str(response.json)
+
+
+def assert_valid_cwl(cwl: Path) -> str:
+    """Run cwltool --validate --strict on a cwl file and checks if the file is valid."""
+    output = str(check_output(["cwltool", "--validate", "--strict", str(cwl.resolve())]))
```
Panaetius, line 473: cwltool is written in Python, I think you can just use 
```
from cwltool.context import LoadingContext
from cwltool.workflow import default_make_tool
from cwltool.resolver import tool_resolver
from cwltool.load_tool import load_tool

context = LoadingContext({"construct_tool_object": default_make_tool,
                               "resolver": tool_resolver})
parsed = load_tool(path, context)
```

and that will raise an exception

renku/core/workflow/converters/cwl.py
```diff
@@ -94,38 +94,64 @@ def workflow_format(self):
...
         else:
             tmpdir = Path(tempfile.mkdtemp())
 
         if isinstance(workflow, CompositePlan):
-            path = CWLExporter._convert_composite(
-                workflow, tmpdir, basedir, filename=filename, output_format=output_format
-            )
+            cwl_workflow = CWLExporter._convert_composite(
+                workflow, basedir, resolve_paths=resolve_paths
+            )  # type: ignore
```
Panaetius, line 122: don't use `# type: ignore` unless absolutely necessary.

Doing `cwl_workflow: Union[cwl.Workflow, CommandLineTool]` above before the if should resolve the type error.

pyproject.toml
```diff
@@ -123,7 +123,7 @@ rdflib = "<7.0,>=6.0.0"
 redis = { version = ">=3.5.3,<4.2.0", optional = true }
 renku-sphinx-theme = { version = ">=0.2.0", optional = true }
 requests = ">=2.23.0,<2.28.2"
-responses = { version = ">=0.7.0,<0.22.0", optional = true }
+responses = "^0.22.0"
```
Panaetius, line 126: You should always pin dependencies with some maximum version, so upstream updating doesn't suddenly break things on our end

Panaetius (changes requested):

tests/utils.py
olevski: Ok great. I tried to check how I can do this but gave up too easily. Thanks for figuring it out.

renku/core/workflow/converters/cwl.py
```diff
@@ -94,10 +94,18 @@ def workflow_format(self):
...
     def workflow_convert(
-        self, workflow: Union[CompositePlan, Plan], basedir: Path, output: Optional[Path], output_format: Optional[str]
-    ):
+        self,
+        workflow: Union[CompositePlan, Plan],
+        basedir: Path,
+        output: Optional[Path],
+        output_format: Optional[str],  # only YAML is generated
+        resolve_paths: bool,  # if True will make paths absolute and resolve symlinks
+        nest_workflows: Optional[bool],  # if True, generate a single workflow spec and file
```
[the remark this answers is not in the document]
olevski: I think with `nest_workflows: bool = False` even when I supplied values they were ignored and the default was always used. So I made sure the types all match but they all use `Optional` now. I think the hook/pluggy stuff does this weird defaults thing.

renku/ui/service/controllers/workflow_plans_export.py
olevski: Done. We accept a values file in the endpoint now.

pyproject.toml
olevski: The reason for that is because mypy was broken in the our CI. And updating it and the responses library fixed it. But now we get more errors from mypy. The funny thing is that I do not even see those errors when I run it locally. And mypy always worked for me locally.

Not sure if you have seen this before... I will try to fix it tomorrow.

olevski: So in mypy 0.991 or 0.990 they added a breaking change for implicit optional types. I.e. before if you just assign None as a default for an argument mypy considered that to be `Optional[x]` but not anymore. So I added the flag to get us the old behaviour. Then we can fix this up in a future PR. See the link below on the optional implicit stuff. 

https://mypy-lang.blogspot.com/2022/11/mypy-0990-released.html

Panaetius: Can you create a follow-up issue to address this?

olevski: Yes: https://github.com/SwissDataScienceCenter/renku-python/issues/3219

olevski:
This is ready now @Panaetius I am done with all the fixes/edits.

renku/ui/service/views/workflow_plans.py
```diff
@@ -30,7 +31,7 @@
 
 
 @workflow_plans_blueprint.route(
-    "/workflow_plans.list", methods=["GET"], provide_automatic_options=False, versions=[V1_4]
+    "/workflow_plans.list", methods=["GET"], provide_automatic_options=False, versions=[VERSIONS_FROM_V1_4]
```
Panaetius, line 34:
```suggestion
    "/workflow_plans.list", methods=["GET"], provide_automatic_options=False, versions=VERSIONS_FROM_V1_4
```
:wink:

Panaetius (approved):

PR #3213, opened 2022-11-16
fix(core): fix git url regex matching taking too long

Panaetius:
The `gitlab_re` that we build has an issue with [catastropic backtracking](https://www.regular-expressions.info/catastrophic.html).

essentially, we matched a path like `/~something/owner/repo` and the part that matches `~something`, the part that matches `owner` and the part that matches `repo` were greedy matches for a variable number of path fragments separated by `/` so there's a ton of valid match combinations which regex searches exhaustively.

with this change, it only matches a single part with `~` in front, `owner` matches a variable path like `a/b/c/d` and repo only matches the last part. so the only variable-length thing is `owner` (in the case of subgroups in gitlab). This makes it run fast again.

I also added a test case that, with the old regex, takes 14 seconds and the duration check fails, with the new regex takes 2.3 seconds (for 14 tests) and passes.

### DIFF
--- a/renku/domain_model/git.py
+++ b/renku/domain_model/git.py
@@ -44,7 +44,7 @@ _RE_HOSTNAME = (
 
 _RE_PORT = r":(?P<port>\d+)"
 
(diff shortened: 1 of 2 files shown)


olevski (approved):

olevski:
Awesome that we have a test for this now!

PR #3221, opened 2022-11-18
chore: add type validation to core commands

### LINKED ISSUE #3188: add runtime type-checking
we have added types in most of our code and check it with `mypy`, but these checks are not enforced at runtime. But since they are there, we don't usually code defensively to make sure that what we get actually matches the expected type.

We should add pydantic and check types at runtime as well, to make the code more robust. We should also check if this has a performance impact and, if so, if we can live with it.

Panaetius:
adds pydantic type checks to all `renku.core.command` commands. This way we don't need to check if something decorated as one type is actually another type.

It already helped to find some bugs that I fixed here as well.

closes #3188


olevski:
I had to resolve conflicts but the conflicts were only in the poetry.lock file which were fixed by simply re-locking the file. So my approving this PR is still OK IMO.

olevski:
@Panaetius now mypy fails in a bunch of places. Not sure how it did not fail before.

Panaetius:
There's only 3 errors, the rest `note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]` is just a warning

```
renku/command/login.py:138: error: Argument "repository" to "_set_git_credential_helper" has incompatible type "Optional[Repository]"; expected "Repository"  [arg-type]
renku/command/login.py:148: error: Argument "repository" to "_set_renku_url_for_remote" has incompatible type "Optional[Repository]"; expected "Repository"  [arg-type]
renku/command/init.py:184: error: Argument "input_parameters" to "set_template_parameters" has incompatible type "Optional[Dict[str, str]]"; expected "Dict[str, str]"  [arg-type]
```

mohammad-alisafaee (changes requested):
Thanks Ralf! It should have been a huge headache to deal with all `mypy` error messages! I've just found a tiny typo which existed before.

mohammad-alisafaee (approved):

PR #3232, opened 2022-12-05
fix(cli): fix setting cpu request in renku.ini for renku session start

### LINKED ISSUE #3191: cli: renku session start error: nano_cpus > 4300?
## Describe the bug
Local renku docker session fails to start - apparently an issue with requesting too many 'nano_cpus', not sure what this is.

**Renku version:** 1.8.1
**OS:** Linux (#1 SMP PREEMPT_DYNAMIC Wed Oct 26 15:55:21 UTC 2022)
**Python:** 3.10.7

### Traceback

```
  File "[...]/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "[...]/renku/ui/cli/session.py", line 222, in start
    session_start_command()
  File "[...]/renku/command/command_builder/command.py", line 252, in execute
    output = self._operation(*args, **kwargs)  # type: ignore
  File "[...]/renku/core/session/session.py", line 107, in session_start
    session_name = provider_api.session_start(
  File "[...]/renku/core/session/docker.py", line 150, in session_start
    resource_requests["nano_cpus"] = int(cpu_request * 10**9)
ValueError: Exceeds the limit (4300) for integer string conversion: value has 1000000000 digits
```

## Additional context

If you go to line 150 in docker.py and change:
`resource_requests["nano_cpus"] = int(cpu_request * 10**9)`
for 
`resource_requests["nano_cpus"] = int(4300)`
The container will start up fine and seems to work no issue - obviously this hack is not a long-term solution though.

Panaetius:
closes #3191


renku/core/session/session.py
```diff
@@ -103,6 +103,10 @@ def session_start(
 
     # set resource settings
     cpu_limit = cpu_request or get_value("interactive", "cpu_request")
+
+    if cpu_limit is not None:
+        cpu_limit = float(cpu_limit)
```
mohammad-alisafaee, line 108: We should catch a `ValueError` to avoid the `you've found a bug` message if `cpu_limit` is not a valid float str. I believe the passed-in `cpu_request` is checked by `validate_arguments` but the one that we read from the config is not checked.

mohammad-alisafaee (commented):
Thanks Ralf! Please read my comment to see if it makes sense to have a conversion check.

mohammad-alisafaee (approved):

PR #3236, opened 2022-12-09
feat(core): v10 migrations

### LINKED ISSUE #2807: metadata version 10 changes
Once we have to add a migration to metadata 10, there are some things that need to be cleaned up. This issue is meant to keep track of them.

- https://github.com/SwissDataScienceCenter/renku-python/pull/2805 introduced patching of python namespaces so old namespaces get automatically redirected to new namespaces. We should migrate all database objects that still point to an old namespace to the corresponding new one, so this isn't necessary anymore.
- https://github.com/SwissDataScienceCenter/renku-python/pull/2748/files#diff-1f777de0e82e75241448bca5bba8ab018557867063bd98d4e3a9fee45f90b8d5R50 we had broken activity ids and now fix them on the fly. we should fix them in the migration instead and remove this code.

### LINKED ISSUE #2813: Refactor template metadata in project
Template metadata is stored in the `Project` class in various fields. We should refactor it to have all template metadata in a single field. This requires a metadata migration.

### LINKED ISSUE #2957: `date_modified` on datasets is set incorrectly
**Describe the bug**
`date_modified` property on the datasets is always set to current date/time which is not correct: https://github.com/SwissDataScienceCenter/renku-python/blob/add433e3b52854b5546bf9c57a09e893c3ff0d7b/renku/domain_model/dataset.py#L367
A manifestation of this bug is visible in `renku log`.

**To Reproduce**
1. `renku clone git@renkulab.io:rok.roskar/atmospheric-vertical-profiles.git`
2. `renku log`
3. See `Date` on each entry

**Expected behavior**
- `date_modified` should get its value from the persisted value
- migration must set proper `date_modified` on the datasets; we can use `date_created` or commit date/time
- current values for this field should be fixed either by a `renku doctor` fix or by using `date_created`

### LINKED ISSUE #3209: Fix Plan creation date for old projects
As said in https://github.com/SwissDataScienceCenter/renku-python/issues/3208, Plan's creation date should be fixed for older projects to be equal to Activity's start date. This would require a migration on the metadata.

Panaetius:
closes #2807 
closes #2813
closes #2957 
closes #3209


renku/core/migration/m_0010__metadata_fixes.py
```diff
@@ -0,0 +1,310 @@
...
+        modification_date = dataset.date_created or dataset.date_published
+        if modification_date is not None:
+            dataset.unfreeze()
+            dataset.date_modified = modification_date
+            dataset.freeze()
+
+        for child in found_datasets:
+            child.unfreeze()
+            child.date_created = dataset.date_created
+            child.date_published = dataset.date_published
```
mohammad-alisafaee, line 307: Derivative datasets shouldn't have a `date_published` so we should remove this and modify the upper statement to `child.date_created = dataset.date_created or dataset.date_published` or just `child.date_created = modification_date`

renku/core/migration/m_0010__metadata_fixes.py
```diff
@@ -0,0 +1,310 @@
...
+                dataset.unfreeze()
+                dataset.date_modified = modification_date
+                dataset.freeze()
+                previous_modification_date = modification_date
+            found_datasets.append(dataset)
+            dataset = dataset_gateway.get_by_id(dataset.derived_from.value)
+
+        # NOTE: first dataset in chain
+        modification_date = dataset.date_created or dataset.date_published
+        if modification_date is not None:
```
mohammad-alisafaee, line 299: Shouldn't this be an error? Same goes for the one in the while loop.

renku/domain_model/dataset.py
```diff
@@ -491,9 +492,9 @@ def derive_from(
         self.initial_identifier = dataset.initial_identifier
         self.derived_from = Url(url_id=dataset.id)
         self.same_as = None
-        self.date_created = date_created or local_now()
+        self.date_created = date_created or dataset.date_created
         self.date_modified = local_now()
-        self.date_published = None
+        self.date_published = dataset.date_published
```
mohammad-alisafaee, line 497: Did we decide to keep `date_published` on the derivatives of an imported dataset? I don't recall this.

renku/domain_model/workflow/plan.py
```diff
@@ -66,6 +67,7 @@ def __init__(
         self.description: Optional[str] = description
         self.id: str = id
         self.date_created: datetime = date_created or local_now()
+        self.date_modified: datetime = date_modified or local_now()
         self.invalidated_at: Optional[datetime] = invalidated_at
```
mohammad-alisafaee, line 71: This is really minor thing, so, don't bother with it if it's a headache: Should we rename `invalidate_at` to `date_removed` (like datasets)?

mohammad-alisafaee: Same for composite plans.

mohammad-alisafaee (changes requested):

renku/core/migration/m_0010__metadata_fixes.py
Panaetius: There's some weird broken projects out there where I don't know how we could fix the dates, so I gated it here. They don't really seem to break anything but have a really strange provenance chain

renku/domain_model/dataset.py
Panaetius: I had a discussion with KG on this and we thought it would be consistent. date_created or date_published are kept on each dataset on the chain, so one does not have to go up the chain to know these things, which can improve some queries. The other option would be to remove both properties on all derivations/set to `None`.

It didn't seem like it'd cause an issue so it seemed sensible to me.

renku/command/schema/project.py
```diff
@@ -37,17 +37,26 @@ class Meta:
 
     agent_version = StringList(schema.agent, load_default="pre-0.11.0")
     annotations = Nested(oa.hasTarget, AnnotationSchema, reverse=True, many=True)
-    automated_update = fields.Boolean(renku.automatedTemplateUpdate, load_default=True)
+    automated_update = fields.Boolean(renku.automatedTemplateUpdate, load_default=True, dump_only=True)
```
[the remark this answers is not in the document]
Panaetius: I didn't want to change the KG metadata. But I think KG just ignores these properties anyways.

mohammad-alisafaee (dismissed):

mohammad-alisafaee (approved):

PR #3243, opened 2022-12-19
fix(service): use temporary directory to clone project templates on project creation

### LINKED ISSUE #3237: do not use project cache for getting template manifests in core-svc
currently we use `user_project_clone` for cloning templates in the core service. This leads to problems that the templates are in a users cache and might not be up-to-date, among other things. 

We really should just clone the templates directly to a temp directory that we clean up after the request, to ensure they're always up to date and so we don't have to worry about their lifecycle (like keeping the cache in sync, for potentially different branches etc.)

we can still use the `clone_from` machinery, but without all the `user_project_clone` things like storing it in redis etc.

Panaetius:
closes #3237

### DIFF
--- a/renku/core/template/template.py
+++ b/renku/core/template/template.py
@@ -403,7 +403,7 @@ class EmbeddedTemplates(TemplatesSource):
         else:
             return (self.reference, self.version) if current_version < Version(self.version) else (reference, version)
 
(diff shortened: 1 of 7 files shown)


mohammad-alisafaee (approved):
LGTM, Thanks!

PR #3245, opened 2022-12-19
chore: fix tests

mohammad-alisafaee:
Fix tests on `develop`. Set python version for tests to `3.7`.


.github/workflows/test_deploy.yml
```diff
@@ -49,10 +49,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3.1.0
       - run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
-      - name: Set up Python 3.8
+      - name: Set up Python 3.7
         uses: actions/setup-python@v4
         with:
-          python-version: "3.8"
+          python-version: "3.7"
```
Panaetius, line 55: was leaving this at 3.7 an oversight? Same below in the mypy action

mohammad-alisafaee: Not really. We use 3.7 in more places in this file and I wanted them to be consistent. I've changed all to 3.9 now. Let's see if the tests pass.

Panaetius: funnily enough, I did the same over in the poetry 1.3 PR

Panaetius (approved):

PR #3254, opened 2023-01-10
docs: fix readthedocs build not installing docs group

Panaetius:
since dev and docs are now groups, not extras, `pip install .[docs]` doesn't work anymore. This fixes that


mohammad-alisafaee (approved):
Thank you!

PR #3268, opened 2023-01-17
chore: update release action to support arbitrary refs

Panaetius:
in case we want to make hotfix releases or similar.


rokroskar:
@Panaetius I think the step `Merge master` needs to be removed, otherwise there is no point in releasing from an arbitrary ref (that may not include master)

mohammad-alisafaee (approved):
LGTM, thanks@

PR #3287, opened 2023-01-20
feat(cli): login to remote gitlab by default

### LINKED ISSUE #3249: Make renku login enable GitLab auth access by default
### Problem
Currently, when a user does `renku login`, and then clones and tries to push to a repo, they get an auth error, which is very confusing, since this is the point of using `renku login` .

Example: When you do
```
renku login dev.renku.ch
renku clone <>
<modify files>
git push
```
you get an auth error:
```
remote: HTTP Basic: Access denied
fatal: Authentication failed for 'https://dev.renku.ch/repos/laura.kinkead/workflow-file-demo.git/'
```
This can be fixed by using `renku login --git`, however this is not obvious to the user as the solution.

### Solution
We should make the `--git` flag True by default, so that this flow works seamlessly out of the box. 

### Notes
There is a security concern that applying the `--git` flag changes the repo's remote URL, however we can sufficiently manage this risk by displaying a warning to the user that the URL has been updated the first time the user runs the `renku login` command.

mohammad-alisafaee:
# Description

Makes `--git` the default in `renku login`. Adds a `--no-git` option to avoid git login. Moves business logic from `renku/command/login.py` to `renku/core/login.py`

Fixes #3249


renku/core/login.py
```diff
@@ -0,0 +1,283 @@
...
+        repository = project_context.repository
+    except ValueError:
+        repository = None
+
+    parsed_endpoint = _parse_endpoint(endpoint)
+
+    remote_name, remote_url = None, None
+    if git_login:
+        if repository is None:
+            communication.warn("Cannot log in to git outside a project.")
```
Panaetius, line 60: Should this mention `--no-git` ?

Panaetius (approved):

renku/core/login.py
mohammad-alisafaee: I don't think so. This is just a warning and won't hurt if they don't pass it. `--no-git` is mentioned in the docs.

PR #3294, opened 2023-01-24
feat(core): add ssh_supported flag to template and migrations_check

### LINKED ISSUE #3293: return ssh status in migrations_check
Add support in `/cache.migrations_check` endpoint on whether a project supports SSH or not.

The basic version of this is just seeing if the template version used is higher than x.

A nice to have would be checking if port 22 is open/exposed in a session, maybe using https://github.com/NVIDIA/container-canary . Or checking that an image is based on an appropriate base image.

Panaetius:
closes #3293


renku/ui/cli/config.py
```diff
@@ -97,13 +97,19 @@
 | ``lfs_threshold``              | Threshold file size below which     | ``100kb`` |
 |                                | files are not added to git LFS      |           |
 +--------------------------------+-------------------------------------+-----------+
-| ``show_lfs_message``           | Whether to show messages about      | ``True``  |
+| ``show_lfs_message``           | Whether to show messages about      | ``true``  |
```
mohammad-alisafaee, line 100: We either need to change `True` to `true` whenever we use this config (with a `.lower()` to be compatible with older projects) or just undo this change here.

renku/ui/cli/config.py
```diff
@@ -97,13 +97,19 @@
...
 |                                | files being added to git LFS or not |           |
 +--------------------------------+-------------------------------------+-----------+
-| ``show_login_warning``         | Whether to warn when logging in to  | ``True``  |
+| ``show_login_warning``         | Whether to warn when logging in to  | ``true``  |
```
mohammad-alisafaee, line 103: Same as above; I've just realized that I've made an error when reading this config in https://github.com/SwissDataScienceCenter/renku-python/blob/fff20ba40bed3a3dcc5a0c1620f0ee9cc6570161/renku/core/login.py#L67. It return a str not a bool.

mohammad-alisafaee (commented):

mohammad-alisafaee (approved):

PR #3318, opened 2023-02-14
feat(core): add support for ssh connections into sessions

### LINKED ISSUE #3289: add `renku session ssh-setup` command
Add a new subcommand `renku session ssh-setup` that
- checks that a user is logged in
- checks that a compatible ssh client is installed in a project
- (optional) checks if ssh-agent is available
- checks if a users keys are already installed for a project
  * In this case, command exits and nothing is changed, unless called with the `--yes` flag
- generates private and public keys for the user (or use existing one if called with `--existing-key=....`
- sets up the users ssh config with the jump host for that renku deployment (1)
- Adds the users public key to the projects authorized_keys file
- checks if VSCode is installed and sets it up to use that ssh connection
- informs the user that they are ready to start ssh enabled sessions and how to do so

(1): Example config for this looks like
```
Host renkulab-jumphost
  HostName renkulab.io
  Port 2022
 
Host user-project-*
  ProxyJump  renkulab-jumphost
  IdentityFile ~/.ssh/renku-user-project.key
  User jovyan
```

where `project-user-session` is the session name and `renkulab.io` is the URL of the jump host for the logged in renku instance. 
`user-project-*` has to be the proper beginning of the session name, e.g. `john-2edoe-test-2dprotected-2dbranc-*`, but without commit hash.

the config file is stored in `~/.ssh/renku/jumphost` and `~/.ssh/renku/user-project-main` with an `Include ~/.ssh/renku/` directive added to the main `~/.ssh/config` file.

When a session gets started later, it only needs an entry for the `HostName` and all settings such as key are persisted in the `user-project-*` entry.

### LINKED ISSUE #3290: add `--ssh` option to `renku session start`
blocked by https://github.com/SwissDataScienceCenter/renku-python/issues/3289

- checks that a users ssh config is correct for a project (otherwise performs the steps outline in https://github.com/SwissDataScienceCenter/renku-python/issues/3289)
- Raises an error if run with local (docker) provider.
- Starts the session with SSH turned on
- Create a new ssh config entry on the users system

SSH config entry looks like
```
Host renku-session-john-2edoe-test-2dprotected-2dbranc-4e9e126a-0
    HostName john-2edoe-test-2dprotected-2dbranc-4e9e126a-0
```
stored in `~/.ssh/renku/renku-session-john-2edoe-test-2dprotected-2dbranc-4e9e126a-0`
prints out the command to connect via SSH, e.g. `ssh renku-session-john-2edoe-test-2dprotected-2dbranc-4e9e126a-0`

### LINKED ISSUE #3293: return ssh status in migrations_check

Panaetius:
closes #3293
closes #3289
closes #3290

/deploy #persist renku-notebooks=master extra-values=notebooks.ssh.enabled=true,notebooks.ssh.service.type=NodePort,notebooks.ssh.service.port=32023


rokroskar:
I've been playing around with this a little bit - it's looking really great already! Some comments in no particular order:

- I found this a little confusing:

```
Info: Added public key. Changes need to be pushed and remote image built for changes to take effect.
The container image 'registry.dev.renku.ch/rokroskar/test-session-ssh:9845c1b' does not exist. Would you like to build it using renkulab? [y/N]: y
Image registry.dev.renku.ch/rokroskar/test-session-ssh:9845c1b built successfully.
```

Basically we created the keypair and pushed - so now, by definition, the image is building - but we tell the user "image is not there, would you like to build it with Renkulab?" - What other option is there? We should just say "image building" and start the session when it's ready. 

- Why can't I connect with ssh to a session that I started in the UI?
- this sort of hostname is... big! `renku-ci-rp-3318.dev.renku.ch-test-session-ssh-rokroskar--test-2dsession-2dssh-5c3af687` Why can't we make it shorter? A config like this works fine:

```
Host jumphost-renku-ci-rp-3318.dev.renku.ch
    HostName renku-ci-rp-3318.dev.renku.ch
    Port 2022
    User jovyan
    
Host test-session-ssh-5c3af687
    HostName rokroskar--test-2dsession-2dssh-5c3af687
    RemoteCommand cd work/test-session-ssh/ || true && exec $SHELL --login
    RequestTTY yes
    ServerAliveInterval 15
    ServerAliveCountMax 3
    ProxyJump  jumphost-renku-ci-rp-3318.dev.renku.ch
    IdentityFile /Users/rok/.ssh/renku/renku-ci-rp-3318.dev.renku.ch-key
    User jovyan
```
This would have the added benefit of not showing the jumphost configuration inside the VSCode remote explorer.

Panaetius:
> Basically we created the keypair and pushed - so now, by definition, the image is building - but we tell the user "image is not there, would you like to build it with Renkulab?"

No, when it asks you this, the keypair is added, but not pushed. So, by definition, it isn't building yet

This code is provider independent, so this code is executed with docker and renkulab sessions

> Why can't I connect with ssh to a session that I started in the UI?

That's added now

> this sort of hostname is... big! renku-ci-rp-3318.dev.renku.ch-test-session-ssh-rokroskar--test-2dsession-2dssh-5c3af687 Why can't we make it shorter? A config like this works fine:

The original thinking was that this way you can distinguish what deployment a session is on based on the name, but since most users won't be on more than 1 deployment anyways, that's now a hard requirement.
One technical reason is that we have automated garbage collection based on the connection name (removing old connection configs if the corresponding session isn't running anymore) and that uses the connection name to identify the connections and I didn't want to delete SSH connections from other deployments just because we can't find them on the current deployment. We could probably also do that through the filename instead of through the connection name.
It's a bit of a leftover from before I realized that `Include ...` works well with many files, when I didn't want to litter a users SSH config with too much stuff, so the Wildcard match entry deduplicated a lot of the clutter.
If we want to change this then I would just use the session ID as connection name directly instead of having two different identifiers.

renku/core/session/renkulab.py
```diff
@@ -148,8 +157,46 @@ def pre_start_checks(self):
...
             repository.commit("Automated commit by Renku CLI.")
 
-    @staticmethod
-    def _remote_head_hexsha():
+        if ssh:
+            system_config = SystemSSHConfig()
+
+            if not system_config.is_configured:
+                if communication.confirm(
+                    "Your system is not set up for SSH connections to Renku. Would you like to set it up?"
```
lokijuhy, line 165:
```suggestion
                    "Your system is not set up for SSH connections to RenkuLab. Would you like to set it up?"
```

olevski (dismissed):

rokroskar (commented):
I found two questionable doc strings... 😆

PR #3319, opened 2023-02-14
fix(svc): properly use Redis sentinel

### LINKED ISSUE #3204: properly use Redis sentinel in core service
We currently only ask Sentinel once for a master address [here](https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/ui/service/cache/config.py#L30-L37) and then use that to create a normal Redis connection [here](https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/ui/service/cache/base.py#L41-L42). This connection is not Sentinel-aware and if Redis master changes, core-service doesn't work anymore until it is restarted.

We should change the code so if `REDIS_IS_SENTINEL==true`, we do something like

```
sentinel = redis.SentinelManager((REDIS_HOST, REDIS_PORT), **config)
cache = sentinel.master_for("renku") # change to appropriate app name
model_db = Database(connection_pool=cache.connection_pool) # maybe also pass **config?
```

in `BaseCache`

Also check about changing the chart so we can pass more than one sentinel address for the `REDIS_HOST`/`REDIS_PORT` above.

Panaetius:
closes #3204 

/deploy #persist

### DIFF
--- a/renku/ui/cli/service.py
+++ b/renku/ui/cli/service.py
@@ -176,7 +176,7 @@ def read_logs(log_file, follow=True, output_all=False):
 @click.pass_context
 def service(ctx, env):
     """Manage service components."""
-    import redis  # noqa: F401
+    import redis
     import rq  # noqa: F401
     from dotenv import load_dotenv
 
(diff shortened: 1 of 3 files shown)


Panaetius:
I randomly killed redis nodes and scaled them up/down and it kept working.

olevski (approved):

PR #3320, opened 2023-02-15
chore: fix coveralls comments

Panaetius:
Coveralls should be posting code coverage on the repo, but doesn't I think it's because the actions only run on push, not on PR. So from the POV of coveralls, it's never run on a PR and doesn't comment. This should fix that.

### DIFF
--- a/.github/workflows/test_deploy.yml
+++ b/.github/workflows/test_deploy.yml
@@ -3,9 +3,13 @@ name: Test, Integration Tests and Deploy
 on:
   push:
     branches:
-      - "**"
+      - master
+      - develop
     tags:
       - "v*.*.*"
+  pull_request:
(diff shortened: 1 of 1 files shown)


olevski:
@Panaetius so we should see a comment on this PR that shows the test coverage? If the fix works.

Panaetius:
> @Panaetius so we should see a comment on this PR that shows the test coverage? If the fix works.

Well, since it doesn't change code, I'm not sure. But this is the last thing I could think of why it's not writing those comments.

Panaetius:
Oh there it is

olevski (dismissed):
This is great! I thought you were supposed to dig through the actions logs to see the coverage.

olevski (approved):

PR #3338, opened 2023-02-28
fix: parallelize dataset add for performance

### LINKED ISSUE #3332: Enhance `dataset add`
`renku dataset add` takes a long time when adding a large number of files (thousands). We should do all git operations in batches to make this operation faster (e.g. https://github.com/SwissDataScienceCenter/renku-python/blob/4a5206971fb8212a399d06132fe811d948edb96f/renku/core/dataset/dataset_add.py#L405).

See https://renku.discourse.group/t/renku-dataset-import-is-extremely-slow/759 for an example.

Panaetius:
closes #3332 

I tested it with the mscoco dataset the user mentioned and it took 55 seconds to add.


mohammad-alisafaee (approved):
LGTM, Thank you!

PR #3339, opened 2023-03-02
feat(dataset): store s3 credentials per bucket

### LINKED ISSUE #3280: Allow storing S3 credentials per bucket
At the moment, we store S3 credentials per S3 provider endpoint. We should make it finer grain by allowing credentials per bucket so that multiple accounts with the same provider can be used concurrently.

mohammad-alisafaee:
Fixes #3280

### DIFF
--- a/renku/core/dataset/providers/s3.py
+++ b/renku/core/dataset/providers/s3.py
@@ -145,22 +145,29 @@ class S3Credentials(ProviderCredentials):
 
         NOTE: This methods should be overridden by subclasses to allow multiple credentials per providers if needed.
         """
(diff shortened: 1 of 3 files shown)


Panaetius:
This would fail for anyone already having stored credentials, correct?

mohammad-alisafaee:
> This would fail for anyone already having stored credentials, correct?

Yes. The section name in `renku.ini` is changed for S3 credentials and we don't look in the old place.

Panaetius (approved):

Panaetius (approved):

PR #3341, opened 2023-03-02
fix(cli): fix starting local session as user without uid 1000

### LINKED ISSUE #3321: No write permissions when running renku on an own machine
**Describe the bug**
When a renku session on an own machine the ids of the files in the container remain the same as on the host machine, which makes them impossible to modify and to create new files.

**Expected behavior**
When I start a renku session on my own machine I can modify and create files inside that session.

renku 2.2.0

Panaetius:
closes #3321


mohammad-alisafaee (approved):
LGTM, thank you!

PR #3342, opened 2023-03-02
feat(core): default sensible flags for update and export

### LINKED ISSUE #3276: feat: `renku update` should default to `--all` with a warning/prompt
Having to always remember to add `--all` when running `renku update` is quite annoying. Instead, we should do something like 

```
renku update
Warning: This could trigger expensive computations - are you sure? (Y/n)
```

which can be overridden with `--all`, i.e. if the flag is supplied the warning is suppressed.

mohammad-alisafaee:
Fixes #3276 

### TODO
* [x] Once review is done, move `_update` from `renku.command.update` to `renku.core.workflow.update`.


renku/ui/cli/update.py
```diff
@@ -190,6 +191,11 @@ def update(update_all, dry_run, paths, provider, config, ignore_deleted, skip_me
 
     communicator = ClickCallback()
 
+    if not paths and not update_all and not dry_run:
+        message = f"{WARNING}Updating all outputs could trigger expensive computations. Are you sure?"
+        if not communicator.confirm(message, default=True):
```
Panaetius, line 196: why not use `abort=True` ?

Panaetius (changes requested):

renku/ui/cli/update.py
Panaetius: @mohammad-alisafaee You marked the above as resolved but I don't see a corresponding change

mohammad-alisafaee: Sorry, forgot to push the latest version 🤦

PR #3343, opened 2023-03-03
feat(cli): remove CLI version check

### LINKED ISSUE #3311: Don't suggest updating renku version past releases deployed on Renkulab
When we change the renku metadata version (e.g. v10)/bump the major version, users get immediately prompted to update the CLI even though our deployment might not be compatible with this yet.

To prevent users from shooting themselves in the foot, the CLI should not prompt the user to update the CLI if the new version is ~a major version ahead of the currently installed one~ not deployed on Renkulab yet. We should check the Version endpoint for the core-service for this (set a short timeout so it doesn't hang the command).

Panaetius:
closes #3311


renku/command/version.py
```diff
@@ -17,13 +17,7 @@
 # limitations under the License.
 """Versioning utilities."""
 
-import datetime
-import json
-import sys
-
-import attr
 import click
-import lockfile
```
mohammad-alisafaee, line 26: `lockfile` isn't used anywhere else and can be remove from dependencies.

mohammad-alisafaee (changes requested):

renku/command/version.py
mohammad-alisafaee: We also need to do a `poetry lock`.

mohammad-alisafaee (approved):

PR #3347, opened 2023-03-04
feat(template): allow renaming templates by using aliases

### LINKED ISSUE #3305: Allow updating renamed Renku templates
Renku project template IDs (directories) have been renamed in version `0.4.1` (`minimal` postfix was removed). This causes the template update logic to ignore newer template versions since it cannot find previous IDs. This needs to be fixed by searching for the new IDs when updating.

**Additional context**
UI uses Renku template repository's URL as template source (instead of `"renku"`) for Renku templates.

**Workaround**
We can use `renku template set -f <new-id>` to manually enable updates.

mohammad-alisafaee:
# Description
Adds an `aliases` attribute to templates that allows having additional IDs for templates. This can be used to rename templates.

## TODO:
* Rename `renku/core/template/template.py` to `renku/core/template/model.py`
* Rename `renku/core/template/usecase.py` to `renku/core/template/template.py`
* Move required functions between these two files

Fixes #3305


Panaetius:
We should add docs for this here: https://renku.readthedocs.io/en/stable/reference/templates.html#create-a-template-repository

mohammad-alisafaee:
> We should add docs for this here: https://renku.readthedocs.io/en/stable/reference/templates.html#create-a-template-repository

Docs were added in https://github.com/SwissDataScienceCenter/renku/pull/3027

Panaetius (approved):

PR #3351, opened 2023-03-07
fix(core): update template checksum when updating Dockerfile

### LINKED ISSUE #3346: Make sure that template migration of dockerfile works correctly
There are issues with old projects and the new SSH feature, because the Dockerfile doesn't get update from the template.

What I think happens is
- user migrates the project
- Dockerfile is updated from template
- Renku version is set in template --> checksum changes
- some time later, user migrates a gain
- Template update sees that Dockerfile was changed since the last update from template, so since there's "local" modifications, Dockerfile isn't updated.

We should fix this to handle the checksums for the Dockerfile differently

mohammad-alisafaee:
# Description

We check Dockerfile changes in each commit and if they are only Renku version updates, we mark the Dockerfile as not being modified by users. Before doing this, we first check its checksum and checksum at time that the template was set to see if they match with the stored checksum; if so, we skip the git history traversal.

This PR also refactors `templates_source` fixture to use actual business models instead of a dummy template source.

Fixes #3346

### DIFF
--- a/renku/core/template/usecase.py
+++ b/renku/core/template/usecase.py
@@ -48,7 +48,7 @@ from renku.core.util.tabulate import tabulate
 from renku.domain_model.project import Project
 from renku.domain_model.project_context import project_context
 from renku.domain_model.template import RenderedTemplate, Template, TemplateMetadata, TemplatesSource
-from renku.infrastructure.repository import Repository
+from renku.infrastructure.repository import DiffLineChangeType, Repository
 
 
 def update_dockerfile_checksum(new_checksum: str):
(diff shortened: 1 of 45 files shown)


renku/core/template/usecase.py
```diff
@@ -38,16 +37,78 @@
...
+        raise errors.DockerfileUpdateError("Project doesn't have a Dockerfile")
+    if is_dockerfile_updated_by_user():
+        raise errors.DockerfileUpdateError("Cannot update Dockerfile checksum because it was update by the user")
+
+    checksums = read_template_checksum()
+    checksums["Dockerfile"] = new_checksum
+    write_template_checksum(checksums)
+
+
+def does_dockerfile_contains_only_version_change() -> bool:
```
Panaetius, line 66:
```suggestion
def does_dockerfile_contain_only_version_change() -> bool:
```

renku/core/template/usecase.py
```diff
@@ -38,16 +37,78 @@
...
 from renku.domain_model.template import RenderedTemplate, Template, TemplateMetadata, TemplatesSource
 from renku.infrastructure.repository import Repository
 
 
+def update_dockerfile_checksum(new_checksum: str):
+    """Update ``Dockerfile`` template checksum if possible."""
+    if not project_context.dockerfile_path.exists():
+        raise errors.DockerfileUpdateError("Project doesn't have a Dockerfile")
+    if is_dockerfile_updated_by_user():
+        raise errors.DockerfileUpdateError("Cannot update Dockerfile checksum because it was update by the user")
```
Panaetius, line 59:
```suggestion
        raise errors.DockerfileUpdateError("Cannot update Dockerfile checksum because it was updated by the user")
```

renku/infrastructure/repository.py
```diff
@@ -1259,6 +1259,28 @@ def __hash__(self) -> int:
...
 
 
+class DiffLine(NamedTuple):
+    """A single line in a patch."""
+
+    text: str
+    """
+    Possible values:
+        A = Added
+        D = Deleted
```
Panaetius, line 1269: Could use an enum

Panaetius (commented):
Minor comments, but looks good to go

Panaetius (approved):

Panaetius (approved):

Panaetius (approved):

PR #3361, opened 2023-03-13
chore: change CI to create pre-releases

### LINKED ISSUE #3329: Create renku pre-releases
We need a better way of releasing renku python to Pypi without it being installable on users machines immediately even though their deployment might not support it yet.

To this end, our CI could be changed to push pre-releases to Pypi instead on release, and only make a normal release using a manually triggered action (Either run when it's deployed or when KG has adapted and we have a combined KG+RP release PR)

Panaetius:
closes #3329


mohammad-alisafaee (approved):
LGTM, thank you!

PR #3362, opened 2023-03-14
feat(cli): support rsync-style syntax using / in dataset add

### LINKED ISSUE #3283: Use `rsync` syntax when adding data to a dataset
We should use `rsync` syntax when adding data from a local file system to a dataset: No trailing slash means add the directory (only the last level) and trailing slash means add files inside the directory.

**Additional context**
`renku dataset add` has a `--destination` flag to specify a target directory inside the dataset's data directory to copy files to.

Panaetius:
closes #3283


mohammad-alisafaee (approved):
LGTM, thanks!

PR #3380, opened 2023-03-29
chore: fix install on python 3.11 due to version regex

Panaetius:
Python 3.11 requires global flags (like `(?x)`) to be at the beginning of a regex, having this after some white-space caused installation to fail on Python 3.11.1+

This fixes all tests on `develop`/`master` not running


mohammad-alisafaee (approved):
Thank you!

PR #3381, opened 2023-03-30
chore: separate shaky tests

### LINKED ISSUE #3364: Separate category for failing integration tests
Many times, our integration tests because of an issue on dataset provider (either `dev.renku.ch` or external provides like Zenodo). These frequent failures might hide other tests that failed because of an actual bug and we would miss them (happened two times). We should move those frequent failing tests to a separate category so that failures of other tests become more visible.

mohammad-alisafaee:
# Description

Separates tests that fail frequently due to external providers.
Refactors GH actions.
Fixes some tests.

Fixes #3364


renku/domain_model/template.py
```diff
@@ -530,7 +531,8 @@ def from_project(cls, project: Optional["Project"]) -> "TemplateMetadata":
         # NOTE: Always set __renku_version__ to the value read from the Dockerfile (if available) since setting/updating
         # the template doesn't change project's metadata version and shouldn't update the Renku version either
         renku_version = metadata.get("__renku_version__")
-        metadata["__renku_version__"] = str(read_renku_version_from_dockerfile() or renku_version or __version__)
```
mohammad-alisafaee, line 533: All of these values can be `None` and if we use `str` on the result we'll get `"None"`.

Panaetius: before it was `str(read_renku_version_from_dockerfile()) or renku_version or __version__` and that led exactly to that error

.github/actions/install-linux/action.yml
```diff
@@ -0,0 +1,78 @@
+name: Install dependencies on Linux
```
Panaetius, line 1: Really cool that you can compose actions like this, I didn't know that :+1:

Panaetius (approved):
Nice work cleaning up the tests!

PR #3386, opened 2023-04-05
test: fix integration tests fii

Panaetius:
Fixes integration tests with migrate core-integration-test repo


tests/service/jobs/test_config.py
```diff
@@ -67,5 +66,6 @@ def test_delay_config_set_failure(svc_client_cache, it_remote_repo_url, view_use
 
     from renku.ui.service.jobs.delayed_ctrl import delayed_ctrl_job
 
-    with pytest.raises(MigrationRequired):
-        delayed_ctrl_job(context, view_user_data, job.job_id, renku_module, renku_ctrl)
+    updated_job = delayed_ctrl_job(context, view_user_data, job.job_id, renku_module, renku_ctrl)
```
mohammad-alisafaee, line 69: I believe this test will fail once we increase metadata version, right? This isn't very nice but we can fix it then.

tests/service/views/test_exceptions.py
```diff
@@ -99,6 +99,7 @@ def test_content_type_headers_exc(svc_client_with_repo):
 
 @pytest.mark.service
 @pytest.mark.integration
+@pytest.mark.remote_repo("old")
```
mohammad-alisafaee, line 102: We should add `remote_repo` to the list of valid markers in https://github.com/SwissDataScienceCenter/renku-python/blob/35f9d22f571b53df0ac5f52ded5ebabd828b68d5/pyproject.toml#L249

mohammad-alisafaee (changes requested):

tests/service/jobs/test_config.py
Panaetius: yes, now there's two repos, one with old metadata and one with new one, and we need to update the new one when we update the metadata. But that's much better than what it was, where a lot of these tests just checked that "MigrationRequired" is raised and not that setting the config or creating the dataset etc. works

Panaetius: I think it makes sense to have integration tests that work on an actual repo and having to keep that up to date isn't great but also not a big deal

mohammad-alisafaee (approved):

PR #3393, opened 2023-04-12
feat(core): raise proper errors when Renku metadata is corrupt

### LINKED ISSUE #3374: Better error message on corrupt metadata files
It sometimes happens that users don't resolve merge conflicts and don't have the renku mergetool installed and inadvertently check in metadata files containing `>>>>>>>` git merge markers.

When the database then tries to load these files, it just fails with some JSON error.

We should catch that error and instead create an error like "Couldn't load file '{path'}: metadata is corrupt" and maybe even check if there's `>>>>>>>` markers in there and also say that this was due to an unresolved merge conflict.

Panaetius:
closes #3374


renku/ui/service/errors.py
```diff
@@ -301,6 +301,17 @@ def __init__(self, exception=None):
         super().__init__(exception=exception)
 
 
+class UserProjectMetadataCorruptError(ServiceError):
+    """The target repository is valid but it is not a Renku project."""
+
+    code = SVC_ERROR_USER + 120
+    userMessage = "The target project is not a Renku project."
+    devMessage = "Cannot work on a non Renku project."
```
mohammad-alisafaee, line 309: I believe these error message and functions doc-str need to be reworded; they don't mention metadata corruption.

mohammad-alisafaee (changes requested):
Looks good, thank! Please see my comment.

mohammad-alisafaee (approved):

PR #3396, opened 2023-04-13
feat(core): do not update template if dockerfile is modified

### LINKED ISSUE #3354: Template update with user-modified Dockerfile
When updating a project's template, if user has modified the Dockerfile we don't overwrite it by default. This results in a template with the latest version but with an incomplete Dockerfile. If latest template version has a feature that requires Dockerfile to be updated then project will be in a (somehow) broken state.

The update operation should fail and users should resolve Dockerfile conflicts manually.

Add a Renku-specific section to the Dockerfile and use it for Renku-related features similar to what we do for git hooks: https://github.com/SwissDataScienceCenter/renku-python/blob/4220f4eb807b8ce51cb374dcbc435012fcdcda75/renku/data/pre-commit.sh#L21

mohammad-alisafaee:
# Description
A Renku-specific section can be added to a Dockerfile.  Any lines of the Dockerfile that are between begin and end end markers must not be modified by users. This section is used to calculate Dockerfile checksum (empty lines and whitespaces at the end of lines are ignored). If it's missing then we use the whole Dockerfile.

When updating a template, Renku only update this section if it exists in both project and template Dockerfile; otherwise, the whole Dockerfile is overridden.

```
########################################################
#        Renku-specific section - DO NOT MODIFY        #
########################################################


########################################################
#              End Renku-specific section              #
########################################################
```

Fixes #3354

### DIFF
--- a/renku/command/migrate.py
+++ b/renku/command/migrate.py
@@ -131,9 +131,9 @@ def _dockerfile_migration_check():
         Dictionary of Dockerfile migration status.
     """
     from renku import __version__
-    from renku.core.migration.migrate import is_docker_update_possible
+    from renku.core.migration.migrate import update_dockerfile
 
-    automated_dockerfile_update, newer_renku_available, dockerfile_renku_version = is_docker_update_possible()
+    automated_dockerfile_update, newer_renku_available, dockerfile_renku_version = update_dockerfile()
 
     return {
         "automated_dockerfile_update": automated_dockerfile_update,
@@ -194,14 +194,13 @@ def _check_project():
             # NOTE: v10 migration not done
             return MIGRATION_REQUIRED
 
-    # NOTE: ``project.automated_update`` is deprecated and we always allow template update for a project
+    # NOTE: ``project.automated_update`` is deprecated. We always allow template update for a project
     status = AUTOMATED_TEMPLATE_UPDATE_SUPPORTED
 
     if check_for_template_update(project_context.project)[0]:
         status |= TEMPLATE_UPDATE_POSSIBLE
-    if is_docker_update_possible()[0]:
+    if is_docker_update_possible():
         status |= DOCKERFILE_UPDATE_POSSIBLE
-
     if is_migration_required():
         return status | MIGRATION_REQUIRED
 
--- a/renku/command/migrate.py
+++ b/renku/command/migrate.py
@@ -133,7 +133,7 @@ def _dockerfile_migration_check():
     from renku import __version__
     from renku.core.migration.migrate import update_dockerfile
 
-    automated_dockerfile_update, newer_renku_available, dockerfile_renku_version = update_dockerfile()
+    automated_dockerfile_update, newer_renku_available, dockerfile_renku_version = update_dockerfile(check_only=True)
 
     return {
         "automated_dockerfile_update": automated_dockerfile_update,
--- a/renku/domain_model/template.py
+++ b/renku/domain_model/template.py
@@ -603,7 +603,7 @@ def find_renku_section(lines: List[str]) -> Tuple[int, int]:
     return start, end
 
 
-def get_renku_section_from_dockerfile(content: str) -> str:
+def get_renku_section_from_dockerfile(content: str) -> Optional[str]:
     """Return the Renku-specific section of the Dockerfile or the whole Dockerfile if it doesn't exist."""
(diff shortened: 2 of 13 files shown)


renku/command/migrate.py
```diff
@@ -131,9 +131,9 @@ def _dockerfile_migration_check():
         Dictionary of Dockerfile migration status.
     """
     from renku import __version__
-    from renku.core.migration.migrate import is_docker_update_possible
+    from renku.core.migration.migrate import update_dockerfile
 
-    automated_dockerfile_update, newer_renku_available, dockerfile_renku_version = is_docker_update_possible()
+    automated_dockerfile_update, newer_renku_available, dockerfile_renku_version = update_dockerfile()
```
olevski, line 136: Just an optional naming suggestion. `update_dockerfile` has a confusing name when you consider its only parameter actually skips the update and only checks. Seeing the function name at face value implies that the dockerfile will be updated and it does not indicate that it is possible to avoid overwriting the dockerfile. 

I propose 2 alternatvies to address this:
- Rename the function to clarify the intent - especially that the default is only a check OR
- Make the `check_only` function parameter be `False` by default force it to be used only with a keyword, i.e. `def update_dockerfile(*, check_only=False): ...`

I prefer the 2nd one. With the 2nd one you would have to say `update_checks = update_dockerfile(check_only=True)`. So the intent and consequences of the function are clearer.

EDIT: I realize after going through more code that the `update_dockerfile` function used to have the `check_only` parameter be `False` by default. I think this is another reason to not change it because someone may not realize we have changed it and introduce bugs.

renku/domain_model/template.py
```diff
@@ -586,3 +588,73 @@ def update(self, template: Template):
         self.metadata["__template_id__"] = template.id
         self.metadata["__automated_update__"] = template.allow_update
         self.immutable_files = template.immutable_files
+
+
+def find_renku_section(lines: List[str]) -> Tuple[int, int]:
+    """Return start and end line numbers of the Renku-specific section."""
+    start = end = -1
+    for index, line in enumerate(lines):
+        if line.startswith("#        Renku-specific section - DO NOT MODIFY        #"):
```
olevski, line 597: Because we often go back and forth on naming, configuration and similar things for templates can we make this a bit more general by looking in the template metadata for what is the start and end of this section and if there isn't anything in the template metadata then we default to what we have here.

This way if we change things in the future we can do so. Is that an easy thing to do? I think it would be nice if the template itself tells you what the "protected" start and end is.

renku/domain_model/template.py
```diff
@@ -586,3 +588,73 @@ def update(self, template: Template):
...
+            start = index
+        elif line.endswith("#              End Renku-specific section              #"):
+            end = index
+            break
+
+    return start, end
+
+
+def get_renku_section_from_dockerfile(content: str) -> str:
+    """Return the Renku-specific section of the Dockerfile or the whole Dockerfile if it doesn't exist."""
```
olevski, line 607: I think it is better to return None, empty string or raise an error if the renku section does not exist. As a naive user of this function that is the behaviour that I would expect.

renku/domain_model/template.py
```diff
@@ -586,3 +588,73 @@ def update(self, template: Template):
...
+
+def update_dockerfile_content(source: Path, destination: Path) -> None:
+    """Update the Renku-specific section of the destination Dockerfile with the one from the source Dockerfile."""
+    source_lines = [line.rstrip() for line in source.read_text().splitlines()]
+    source_start, source_end = find_renku_section(source_lines)
+
+    destination_lines = [line.rstrip() for line in destination.read_text().splitlines()]
+    destination_start, destination_end = find_renku_section(destination_lines)
+
+    # NOTE: If source or destination Dockerfiles doesn't have Renku-specific section, we overwrite the whole file
```
olevski, line 645: Is this potentially problematic from a user perspective? I.e. if I have an old project where I have done custom changes but I do not have the renku section then renku will tell me that I can upgrade automatically but then also wipe out my custom changes. But maybe I misunderstand.

Should we avoid updating altogether in this case?

olevski (changes requested):
A few suggestions and questions.

Do we need more extensive documentation changes to explain the formating and content of the Dockerfile sections and logic?

olevski (approved):
Thanks Mohammad.

PR #3398, opened 2023-04-17
feat(dataset): default dataset add action in configuration

### LINKED ISSUE #3333: Remove `--move`, `--copy`, and `--link` from `dataset add` options
I found `--move`, `--copy`, and `--link` flags complicated when adding files to a dataset. They also make the implementation much more complex since we need to address all these cases. 

In most cases, users want to copy files to a dataset's data directory when adding files to it. We should always copy files and remove all these options. The only drawback is that files that are parts of more than one dataset are duplicated. This only results in consuming more local disk space; if those files are large then they reside in LFS but then there is only one copy of these files in LFS because it uses file hash to store them. Moving files to another dataset is also possible using `renku mv` command.

We also need to migrate old projects to have the same structure everywhere.

mohammad-alisafaee:
# Description

Fails when adding files from outside the project with `--link`.
Allow configuring `copy`, `link`, or `move` in Renku's local/global configuration. A default value of `link` will be ignored for files outside the project and we'll prompt users.

Fixes #3333

### DIFF
--- a/renku/core/dataset/providers/local.py
+++ b/renku/core/dataset/providers/local.py
@@ -132,10 +132,12 @@ class LocalProvider(ProviderApi, AddProviderInterface, ExportProviderInterface):
                     default_action = DatasetAddAction.COPY
                 elif action.lower() == "move":
                     default_action = DatasetAddAction.MOVE
+                elif action.lower() == "link":
+                    default_action = DatasetAddAction.SYMLINK
                 else:
                     raise errors.ParameterError(
                         f"Invalid default action for adding to datasets in Renku config: '{action}'. "
-                        "Valid values are 'copy' and 'move'."
+                        "Valid values are 'copy', 'link', and 'move'."
                     )
             else:
                 default_action = DatasetAddAction.COPY
@@ -178,9 +180,6 @@ class LocalProvider(ProviderApi, AddProviderInterface, ExportProviderInterface):
         def get_file_metadata(src: Path) -> DatasetAddMetadata:
             in_datadir = is_subpath(src, destination)
 
-            if not is_subpath(src, project_context.path) and link:
-                raise errors.ParameterError(f"Cannot use '--link' for files outside of project: '{uri}'")
-
             relative_path = src.relative_to(source_root)
             dst = destination_root / relative_path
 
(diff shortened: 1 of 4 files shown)


renku/core/dataset/providers/local.py
```diff
@@ -114,14 +115,30 @@ def get_metadata(
...
+        else:
+            prompt_action = True
+            action = get_value("renku", "default_dataset_add_action")
+            if action:
+                prompt_action = False
+                if action.lower() == "copy":
+                    default_action = DatasetAddAction.COPY
+                elif action.lower() == "move":
+                    default_action = DatasetAddAction.MOVE
+                else:
```
Panaetius, line 135: should we allow setting `link` as a default as well, but prompt with `--copy` for external files? I think that's nicer from a UX perspective, if users can set all three actions as a default.

But U wonder if would could split it, so LocalProvider only handles paths within a repo and ExternalProvider is used for files outside of a project? That might be cleaner.

renku/ui/cli/config.py
```diff
@@ -93,6 +93,10 @@
 | ``dataverse.server_url``       | URL for the Dataverse API server    | ``None``  |
 |                                | to use                              |           |
 +--------------------------------+-------------------------------------+-----------+
+| ``default_dataset_add_action`` | Default action when adding files to | ``None``  |
```
Panaetius, line 96: Shouldn't the default be `copy` ?

mohammad-alisafaee: We'll prompt the user if nothing is configured here; to me, that's a default value of `None`. If the default value is `copy` then we shouldn't prompt when nothing is configured.

renku/core/dataset/providers/local.py
mohammad-alisafaee: A default value of `link` for external files will prompt users what to do.

Panaetius: > We don't allow linking for external files anymore. We have a whole dataset as external and we either mount it or symlink its data directory (not individual files).

Yes but I wonder if we should always use the External provider even for adding individual files (right now ExternalProvider is not an AddProvider). That would maybe help solve some of the confusion that could happen here

PR #3399, opened 2023-04-20
fix(core): fix error in metadata loading and v10 migration

Panaetius:
fixes the issue reported in https://renku.discourse.group/t/pipeline-takes-ages-and-or-fails/785/8

The database returned uninitialized objects when loading objects from the deserialization cache and the by usage/by generation indices weren't updated in the migration and sometimes pointed at old (deleted) activities.


mohammad-alisafaee (approved):
Thank you!

PR #3450, opened 2023-05-09
feat(core): shell completion for sessions

### LINKED ISSUE #3385: `renku sessions` UX improvement
Renku CLI should support tab completion for sessions names in `renku session` command.

mohammad-alisafaee:
# Description

Adds shell completion for session names and session providers.
If there is only one session, we open/stop that one session. NOTE that we don't check if there is only one session in total; we only check if there is one session in a provider. We search local providers before remote ones to make this behavior deterministic and a bit faster.

Fixes #3385


renku/core/session/renkulab.py
```diff
@@ -402,14 +405,23 @@ def session_stop(self, project_name: str, session_name: Optional[str], stop_all:
...
             for session in sessions:
                 responses.append(
                     self._send_renku_request(
                         "delete", f"{self._notebooks_url()}/servers/{session.id}", headers=self._auth_header()
                     )
                 )
                 self._wait_for_session_status(session.id, "stopping")
+        elif not session_name:
+            sessions = self.session_list(project_name=project_name)
+            if len(sessions) == 1:
```
Panaetius, line 418: We should print some warning if there's more than 1 session and not name/stop_all is specified. like `More than one session found, please specify a name of the session to stop or use --stop-all`. Or maybe in renku.ui.cli.session if the result is false?

renku/core/session/renkulab.py
```diff
@@ -422,21 +434,27 @@ def session_stop(self, project_name: str, session_name: Optional[str], stop_all:
...
+        sessions = self.session_list(project_name="")
         system_config = SystemSSHConfig()
         name = self._project_name_from_full_project_name(project_name)
         ssh_prefix = f"{system_config.renku_host}-{name}-"
 
+        if not session_name:
+            if len(sessions) == 1:
+                session_name = sessions[0].id
+            else:
+                return False
```
Panaetius, line 454: Same here with a warning

Panaetius (changes requested):

renku/core/session/docker.py
```diff
@@ -304,22 +304,24 @@ def session_stop(self, project_name: str, session_name: Optional[str], stop_all:
                 self._get_docker_containers(project_name)
                 if stop_all
                 else self.docker_client().containers.list(filters={"id": session_name})
+                if session_name
+                else self.docker_client().containers.list()
             )
 
-            if len(docker_containers) == 0:
+            if len(docker_containers) == 0 or (not session_name and len(docker_containers) > 1):
```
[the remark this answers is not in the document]
mohammad-alisafaee: Right. We can also simplify it, since we should have one container if `stop_all` is False.

renku/core/session/renkulab.py
mohammad-alisafaee: We raise an error in `core/session/session.py` in this case: `raise errors.ParameterError("Session name is missing")`

renku/core/session/renkulab.py
mohammad-alisafaee: I've changed the `session_stop` method to return a status of what has been stopped. Based on that, I show the error in `core/session/session.py`.

renku/domain_model/session.py
```diff
@@ -162,7 +172,7 @@ def session_stop(self, project_name: str, session_name: Optional[str], stop_all:
 
 
         Returns:
-            bool: True in case session(s) has been successfully stopped
+            bool: The status of running and stopped sessions
```
Panaetius, line 175:
```suggestion
            SessionStopStatus: The status of running and stopped sessions
```

Panaetius (changes requested):

Panaetius (approved):

PR #3472, opened 2023-05-22
fix(service): fix working on branches in the core-svc

### LINKED ISSUE #3458: Config API does not support the branch parameter
**Describe the bug**

Viewing a project's config with the `config.show` API does not respect the `branch` parameter, e.g.
```
https://dev.renku.ch/ui-server/api/renku/10/config.show?git_url=https://gitlab.dev.renku.ch/johann.thiebaut1/another-playground-project&branch=1-try-out-config-api
```
Shows the config from the `master` branch, not from the `1-try-out-config-api` branch.

**Expected behavior**

The `config.show` and `config.set` API endpoints should expose a parameter to view/update the config for a specific branch.

- renku-core:v2.3.2

**Additional context**

### Notes from @Panaetius 
- https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/ui/service/serializers/common.py#L55 this says "branch"
- https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/ui/service/controllers/api/mixins.py#L172 wants "ref"
- https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/ui/service/controllers/utils/remote_project.py#L48 also uses "ref" as "branch"

Panaetius:
closes #3458


renku/ui/service/serializers/common.py
```diff
@@ -52,7 +52,18 @@ def validate_git_url(self, value):
...
 
-    branch = fields.String(metadata={"description": "Remote git branch."})
+    branch = fields.String(load_default=None, metadata={"description": "Remote git branch (or tag or commit SHA)."})
+
+    @pre_load
+    def set_branch_from_ref(self, data, **kwargs):
+        """Set `branch` field from `ref` if present."""
+        if "ref" in data and not data.get("branch"):
+            # Backward compatibility: branch and ref were both used. Let's keep branch as the exposed field
+            # even if interally it gets converted to "ref" later.
```
mohammad-alisafaee, line 62: `internally`

renku/ui/service/controllers/api/mixins.py
```diff
@@ -169,8 +169,8 @@ def execute_op(self):
                     "git_url": self.request_data["git_url"],
                 }
 
-                if "ref" in self.request_data:
-                    clone_context["ref"] = self.request_data["ref"]
+                if "branch" in self.request_data:
+                    clone_context["ref"] = self.request_data["branch"]
```
mohammad-alisafaee, line 173: Shouldn't this be `clone_context["branch"]` (or include it too)? We don't use `"ref"` after this but we use `"branch"`.

mohammad-alisafaee (changes requested):
Thanks Ralf! Please see my comment.

mohammad-alisafaee (approved):

PR #3476, opened 2023-05-23
chore: remove unused code in renku lfs migrate

### LINKED ISSUE #3440: Make sure `renku lfs migrate` works as intended
Currently, while we do collect the hexsha's before and after the `git lfs migrate` command [here](https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/core/storage.py#L580), the [methods for updating checksums](https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/core/storage.py#L585-L624) are never actually called.

We should check that `renku lfs migrate` actually migrates both files and our metadata correctly and change it to work properly.

We should also check if git-lfs ever fixed the change where we can't set a filesize limit for `git lfs migrate`

Panaetius:
closes #3440

### DIFF
--- a/renku/core/storage.py
+++ b/renku/core/storage.py
@@ -15,7 +15,6 @@
 # limitations under the License.
 """Logic for handling a data storage."""
 
-import csv
 import functools
 import itertools
 import os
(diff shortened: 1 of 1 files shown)


mohammad-alisafaee (approved):
LGTM, thanks!

PR #3483, opened 2023-05-25
chore: remove numpy dependency

### LINKED ISSUE #3452: Remove numpy as a dependency of the cli
We use numpy in `renku workflow visualize` but this adds a lot of dependncies and it makes renku images larger.

Re-implement all the functionality we use with numpy to do this but with plain python.

Panaetius:
closes #3452


olevski (approved):
❤️

olevski (approved):

olevski (approved):

PR #3506, opened 2023-06-05
test: add cleanup for autobranches created in tests

Panaetius:
branches are piling up in renku-python-integration-tests/core-it-protected I'm cleaning them up now but this should make the tests clean up themselves in the future.


tests/core/fixtures/core_models.py
```diff
@@ -100,7 +100,15 @@ def protected_git_repository(tmp_path):
...
     with repository.get_configuration(writable=True) as config:
         config.set_value("pull", "rebase", "false")
 
     yield repository
+
+    branches_after = set(b.name for b in repository.branches)
+
+    for branch in branches_after - branches_before:
+        # delete created branches
+        repository.push("origin", branch, delete=True)
```
mohammad-alisafaee, line 114: It's good to catch errors so that a single failure doesn't prevent deletion of other branches.

mohammad-alisafaee: Should we have a `remove_branch` on the `Remote` class instead? It will be a bit inconsistent with the branch deletion in `Repository` (there we do `repo.branches.remove(branch_name)`) but makes it easier to search for all remote branch deletion.
```
def remote_branch(name):
     _run_git_command(self._repository, "push", self.name, name, delete=True)
```

tests/core/fixtures/core_models.py
```diff
@@ -100,7 +100,15 @@ def protected_git_repository(tmp_path):
 
     repository = Repository.clone_from(url=parsed_url, path=tmp_path)
 
+    branches_before = set(b.name for b in repository.branches)
```
mohammad-alisafaee, line 103: `{b.name for b in repository.branches}` is nicer.

mohammad-alisafaee (changes requested):
Thank you! Please see my comments.

olevski (approved):

mohammad-alisafaee (approved):

PR #3508, opened 2023-06-05
fix(cli): don't swallow command output in renku run

### LINKED ISSUE #3507: renku run doesn't show command output of wrapped command
We should show the output of the wrapped command if the user didn't redirect stdout/stderr.

Panaetius:
closes #3507


mohammad-alisafaee (approved):
Thank you!

Issue #3510, opened 2023-06-06
add renku doctor fix for wrong workflow file plan ids


Panaetius:
(Composite)Plans from Workflow files have a wrong id when the files are rerun, like `/plans//plans/<uuid>`. This was fixed in a hotfix, but we should add a doctor check/fix to fix it in repositories.

PR #3511, opened 2023-06-06
fix(cli): fix wff plan ids

Panaetius:
doctor fix/proper fix can be added with https://github.com/SwissDataScienceCenter/renku-python/issues/3510 later


Panaetius:
closed in favor of https://github.com/SwissDataScienceCenter/renku-python/pull/3512 so actions run

PR #3512, opened 2023-06-06
fix(cli): fix wff plan ids

Panaetius:
doctor fix/proper fix can be added with https://github.com/SwissDataScienceCenter/renku-python/issues/3510 later

### DIFF
--- a/renku/command/schema/workflow_file.py
+++ b/renku/command/schema/workflow_file.py
@@ -33,6 +33,23 @@ class WorkflowFilePlanSchema(PlanSchema):
         model = WorkflowFilePlan
         unknown = marshmallow.EXCLUDE
 
+    @marshmallow.pre_dump(pass_many=True)
+    def fix_ids(self, objs, many, **kwargs):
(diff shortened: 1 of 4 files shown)


olevski (dismissed):

olevski (approved):

renku/command/schema/workflow_file.py
```diff
@@ -33,6 +33,23 @@ class Meta:
         model = WorkflowFilePlan
         unknown = marshmallow.EXCLUDE
 
+    @marshmallow.pre_dump(pass_many=True)
+    def fix_ids(self, objs, many, **kwargs):
```
mohammad-alisafaee, line 37: Should we do this for Plan's inputs, outputs, and parameters as well?

Panaetius: I didn't see any input/output/param ids that had this issue, but I can add it just to be safe

mohammad-alisafaee (approved):
Thank you!

PR #3523, opened 2023-06-13
fix(service): fixes importing private datasets in deployments with external gitlab

### LINKED ISSUE #3500: Dataset import in core-svc should check against actual host URL to determine token safety
In https://github.com/SwissDataScienceCenter/renku-python/blob/2e3a0977dfeb65748ea549247ed092b78ec50432/renku/ui/service/jobs/datasets.py#L88 we currently check if a datasets host matches the projects gitlab URL host to see if it's safe to forward the gitlab token. We should instead check the datasets host vs the configured RENKU_DOMAIN to determine if it's safe.
otherwise, dataset import does not work on deployments where gitlab is hosted under a different URL (e.g. CI deployments)

Panaetius:
closes #3500

### DIFF
--- a/renku/ui/service/jobs/datasets.py
+++ b/renku/ui/service/jobs/datasets.py
@@ -15,6 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Dataset jobs."""
+import os
 import urllib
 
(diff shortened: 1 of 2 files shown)


mohammad-alisafaee (approved):
LGTM, thank you!

PR #3534, opened 2023-06-20
feat(service): replace/refactor internal repository cache

### LINKED ISSUE #2242: Refactor/untangle RenkuOperationMixin
Currently, `renku.service.controllers.api.mixins.RenkuOperationMixin` has a bit too many hats. It handles:

- delayed/asynchronous or synchronous execution of a request
- adding/getting projects to/from the cache
- "remote" execution (cloning to a temp directory, bypassing the cache), used for anonymous sessions
- on-the-fly migration of a project
- Switching between branches in cached projects
- resetting/fetching remote changes in cached projects so they are in sync with gitlab

This makes this component very fragile, as there are many reasons why it might change and many changes that can break other things. It also makes it almost impossible to test all the different combinations of situations.

This functionality should be split up into several classes, with the service controls specifying which ones are relevant to them.

Namely:
- async execution: most endpoints don't make sense to be called in an async way (e.g. dataset.list), whereas others probably always should be async (`cache.migrate`, `dataset.import`). This should be handled by a dedicated control that only handles async execution, and we should determine which endpoints should be async. It is unclear if there are any good candidates for endpoints that should support both?
- A separate class for "local" and for "remote" execution so the two aren't mixed
- The cache itself should offer cache management functionality or there should be a `CacheManager` class that deals with fetching/resetting cached projects. 
- cached projects should be per branch/ref, not per project, so there's no conflict between them and no need to switch branches
- Some endpoints can work on refs (branch/tag/commit) while other (mainly write requests) can only really work on a branch. Service controls need to be able to specify which of the two they support, and the API should have a consistent naming for it (`ref` vs. `branch` ?)

It is unclear to me if we want on-the-fly migrations. We don't want them on write requests, sow e don't inadvertently migrate a repo, and they probably only make sense in async requests, as they can take quite long. And I'm not aware of any long running read requests that we have at the moment. Additionally, there is the issue that these implicit migrations poison the cache, as the cached project is migrated but not in sync with the remote anymore, where a read request shouldn't change the state of the repo.

Architecture wise, what we have right now with `renku_op` that is wrapped by different classes isn't actually bad, if it weren't just a single class doing the heavy lifting. Maybe using the Decorator pattern would be a good fit for this (maybe with a Factory class to return a decorator for Local or Remote execution?).

### LINKED ISSUE #2797: Referencing a non existing branch in the service APIs doesn't fail
**Describe the bug**
If I try to invoke some of the APIs exposed by the service that allows specifying a branch (E.G. `GET /config.show`), and I put a non-existing branch name, I get a successful response with the content of a different branch.

**To Reproduce**
I think the lowest level API to reproduce it is `POST /cache.project_clone`. But anything containing `branch` in the parameters leads to the same result.
To reproduce the problem, invoke the API using a valid renku project as `git_url` and a non-existing branch as `branch`.

**Expected behavior**
I would expect the API to fail with a `UserRepoReferenceInvalidError` error.

**Screenshots and/or execution output**
Here is an example:
image

image

 - renku-python version: >= 1.1.4 are affected, as well as `develop`
 [removed the others, not relevant for the renku-core service]

### LINKED ISSUE #3482: renku-core does not handle expired git credentials in its cached repositories
It seems that the `git` cache of `renku-core` is not invalidated when the credentials are not valid anymore.
* Manually deleting the pod (to force the cache to go away) removes the issue for a while
* The `config.show` endpoint seems to store the wrong value (or maybe serves it from its cache but it's not synced to GitLab)

Sentry Issue: [RENKU-PYTHON-T6](https://sentry.dev.renku.ch/organizations/sentry/issues/2075/?referrer=github_integration)

```
GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git push --set-upstream origin renku/autobranch/master/74c39556
  stderr: 'remote: HTTP Basic: Access denied
fatal: Authentication failed for 'https://gitlab.dev.renku.ch/johann.thiebaut1/another-playground-project.git/''
  File "renku/infrastructure/repository.py", line 1891, in _run_git_command
    return getattr(repository.git, command)(*args, **kwargs)
  File "git/cmd.py", line 639, in <lambda>
    return lambda *args, **kwargs: self._call_process(name, *args, **kwargs)
  File "git/cmd.py", line 1184, in _call_process
    return self.execute(call, **exec_kwargs)
  File "git/cmd.py", line 984, in execute
    raise GitCommandError(redacted_command, status, stderr_value, stdout_value)

GitCommandError: Git command failed: Cmd('git') failed due to: exit code(128)
  cmdline: git push --set-upstream origin renku/autobranch/master/74c39556
  stderr: 'remote: HTTP Basic: Access denied
fatal: Authentication failed for 'https://gitlab.dev.renku.ch/johann.thiebaut1/another-playground-project.git/''
  File "renku/core/util/git.py", line 602, in push_changes
    repository.push(pushed_remote, pushed_branch, set_upstream=True)
  File "renku/infrastructure/repository.py", line 333, in push
    self.run_git_command(
  File "renku/infrastructure/repository.py", line 414, in run_git_command
    return _run_git_command(self._repository, command, *args, **kwargs)
  File "renku/infrastructure/repository.py", line 1893, in _run_git_command
    raise errors.GitCommandError(

GitError: Cannot push changes
(8 additional frame(s) were not displayed)
...
  File "renku/ui/service/views/config.py", line 89, in set_config
    return SetConfigCtrl(cache, user_data, dict(request.json)).to_response()  # type: ignore
  File "renku/ui/service/controllers/config_set.py", line 56, in to_response
    op_result, remote_branch = self.execute_and_sync()
  File "renku/ui/service/controllers/api/mixins.py", line 365, in execute_and_sync
    return result, self.sync(remote=remote)
  File "renku/ui/service/controllers/api/mixins.py", line 350, in sync
    return push_changes(repository, remote=remote)
  File "renku/core/util/git.py", line 611, in push_changes
    raise errors.GitError("Cannot push changes") from e

ProgramRenkuError:
```

### LINKED ISSUE #3493: make sure directory exists in service project lock
Sentry Issue: [RENKU-PYTHON-WM](https://sentry.dev.renku.ch/organizations/sentry/issues/2357/?referrer=github_integration)

```
FileNotFoundError: [Errno 2] No such file or directory: '/svc/cache/projects/Y3JhbWFrcmlldGh6LmNo/cramakri/a-2549-test-1.lock'
(11 additional frame(s) were not displayed)
...
  File "renku/ui/service/controllers/api/mixins.py", line 225, in execute_op
    self.reset_local_repo(project)
  File "renku/ui/service/controllers/api/mixins.py", line 248, in reset_local_repo
    with lock:
  File "portalocker/utils.py", line 157, in __enter__
    return self.acquire()
  File "portalocker/utils.py", line 239, in acquire
    fh = self._get_fh()
  File "portalocker/utils.py", line 289, in _get_fh
    return open(self.filename, self.mode, **self.file_open_kwargs)

ProgramInternalError:
```

we should make the core service more resilient to the project directory being missing when a request comes in. Like checking it exists and otherwise recloning.

Panaetius:
Replaces the repository cache function contained in RenkuOperationMixin with a dedicated repository cache class.

closes #2242 
closes #3493 
closes #3482 
closes #2797

/deploy #persist renku-gateway=master renku=core-svc-horizontal-scaling extra-values=core.sentry.dsn=https://0cc02f3a02d3474d9f7e90915873ab4d@sentry.dev.renku.ch/2,core.sentry.environment=renku-dev,core.sentry.enabled=true


renku/ui/service/gateways/repository_cache.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def get(
+        self, cache: ServiceCache, git_url: str, branch: Optional[str], user: User, shallow: bool = True
+    ) -> Project:
+        """Get a project from cache (clone if necessary)."""
+        if git_url is None:
+            raise ValidationError("Invalid `git_url`, URL is empty", "git_url")
+
+        try:
+            project = Project.get(
```
olevski, line 54: I am not really familiar with Walrus. But the author of the package says this in the docs about using `Models`:

> Walrus models should not be considered production-grade code and I strongly advise against anyone actually using it for anything other than experimenting or for inspiration/learning.

> My advice: just use hashes for your structured data. If you need ad-hoc queries, then use a relational database.

Should we worry about this?

See here: https://walrus.readthedocs.io/en/latest/models.html#

renku/ui/service/gateways/repository_cache.py
```diff
@@ -0,0 +1,233 @@
...
+        except ValueError:
+            # project not found in DB
+            return self._clone_project(cache, git_url, branch, user, shallow)
+
+        if not project.abs_path.exists():
+            # cache folder doesn't exist anymore
+            project.delete()
+            return self._clone_project(cache, git_url, branch, user, shallow)
+
+        if not shallow and project.clone_depth is not None and project.clone_depth <= 1:
```
olevski, line 66: I am assuming that clone_depth set to `0` means to actually not have a shallow clone? Maybe we should clarify that in the docs here or for the `Project`.

renku/ui/service/gateways/repository_cache.py
```diff
@@ -0,0 +1,233 @@
...
+        except Exception as e:
+            service_log.error(f"Couldn't purge project {project.project_id}:{project.name} from cache", exc_info=e)
+
+    def evict_expired(self):
+        """Evict expired projects from cache."""
+        for project in Project.all():
+            if project.ttl_expired():
+                self.evict(project)
+
+    def _update_project_access_date(self, project: Project):
```
olevski, line 95: Just an optional suggestion... Is this something that should be a decorator used on other methods here so that it updates the accessed date when specific methods are called?

renku/ui/service/gateways/repository_cache.py
```diff
@@ -0,0 +1,233 @@
...
+                        service_log.debug(f"project already cloned, skipping clone: {git_url}")
+                        self._update_project_access_date(found_project)
+                        return found_project
+
+                # clean directory in case of previous failed state
+                for root, dirs, files in os.walk(project.abs_path):
+                    for f in files:
+                        os.unlink(os.path.join(root, f))
+                    for d in dirs:
+                        shutil.rmtree(os.path.join(root, d))
```
olevski, line 153: Why not `shutil.rmtree` on the root folder?

olevski (changes requested):
A few questions or minor (in most cases optional) changes.

renku/ui/service/gateways/repository_cache.py
Panaetius: We don't want to delete the root folder, only its contents. And we can't recreate the root folder afterwards, otherwise file handles pointing to the previous one  (e.g. the lock) will fail.

renku/ui/service/gateways/repository_cache.py
olevski: Ok 2 years without issues justifies leaving this alone then.

renku/ui/service/gateways/repository_cache.py
olevski: Ok sounds good. I thought it may not work. Just thought I'd bring it up.

olevski (approved):

PR #3548, opened 2023-07-05
feat(cli): allow disabling automated parameter detection in renku run

### LINKED ISSUE #3504: Unclear messaging for error in using workflows
Tracking inputs, outputs and parameters, a user had this error many times, causing the workflow to not be tracked 

```
finished.
Error: Invalid parameter value - Duplicate input, output or parameter names found: use_embeddings
```

This is the definition of that parameter:

```
_ = Parameter(name="use_embeddings", value=args.use_embeddings)
```

where `args.use_embeddings` is a `str`. The user has tried a few things: cleaning with `renku gc`, commenting this (and then it gives the same error with another parameter), but could not understand what is going on. In the beginning, as value I was providing `True`, and I thought that might be the reason. But now, I just provide a `str`. 

Finally, the user could make it work with the command:
`renku run python src/bert_topic_extraction.py --nr_topics 300 --use_embeddings False --remove_entities False`

Hence, these 3 parameters may be tracked twice: internally, and from the command use to run. Hence, the error is understood.  However, it is still en extremely annoying this. We propose: (i) could this be checked before actually running /executing the files? And (ii) Renku prompting: `Duplicated inputs, please do something.`

Panaetius:
closes #3504

### DIFF
--- a/renku/core/workflow/plan_factory.py
+++ b/renku/core/workflow/plan_factory.py
@@ -412,21 +412,7 @@ class PlanFactory:
         existing_parameter = next((p for p in self.inputs if p.name == inp_param.name), None)
 
         if existing_parameter is not None and existing_parameter.default_value == inp_param.default_value:
-            # merge with new parameter
-            if existing_parameter.prefix is None:
-                existing_parameter.prefix = inp_param.prefix
-
-            if existing_parameter.position is None:
-                existing_parameter.position = inp_param.position
-
-            if existing_parameter.postfix is None:
-                existing_parameter.postfix = inp_param.postfix
-
-            if existing_parameter.encoding_format is None:
-                existing_parameter.encoding_format = inp_param.encoding_format
-
-            if existing_parameter.mapped_to is None:
-                existing_parameter.mapped_to = inp_param.mapped_to
+            existing_parameter.update_from(inp_param)
         elif existing_parameter is not None:
             # duplicate with different values!
             raise errors.ParameterError(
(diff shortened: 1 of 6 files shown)


renku/core/workflow/plan_factory.py
```diff
@@ -392,23 +394,48 @@ def add_command_input(
...
+            if existing_parameter.mapped_to is None:
+                existing_parameter.mapped_to = inp_param.mapped_to
+        elif existing_parameter is not None:
+            # duplicate with different values!
+            raise errors.ParameterError(
+                f"Duplicate input '{inp_param.name}' found with differing values ('{inp_param.default_value}'"
+                f" vs. '{existing_parameter.default_value}')"
+            )
+        else:
+            self.inputs.append(inp_param)
```
olevski, line 437: This is repeated futher but for an output parameter. I guess if we repeat it one or two more times we should factor it out... Just an observation I dont think we need to do it now.

renku/ui/cli/run.py
```diff
@@ -132,6 +132,14 @@
     This only affects files and directories; command options and flags are
     still treated as inputs.
 
+.. topic:: Disabling parameter detection (``--no-parameter-detection``)
+
+    Inputs that aren't files or directories are automatically detected as
+    parameters in ``renku run``. You can disable this feature by passing the
+    ``--no-parameter-detection`` flag. You can still manually specify
+    parameters using ``--param`` arguments mentioned above or using the
+    ``renku.api.Parameter`` class in Python code.
```
olevski, line 141: Maybe we should very briefly mention what happens to the part of a command passed to renku run that is not a file or directory and the `--no-parameter-detection` flag is used? I am assuming then this part of the command is fully ignored? Or more like it is recorded in the metadata but then it cannot be dynamically modified with rerun/run?

olevski (approved):
Just a few optional suggestions. If you commit changes ping me and I will re-approve.

renku/core/workflow/plan_factory.py
Panaetius: they're slightly different, one also has a `create_folder` property (and CommandParameter doesn't have encoding_format or mapped_to).

But it's a good point anyways, it shouldn't be in this code. I refactored it into methods on the type itself.

olevski (approved):
🚢

PR #3556, opened 2023-07-11
fix(service): isolate core-service into one redis-namespace each

Panaetius:
makes it so each core-service instance has its own redis namespace, to prevent caches from overlapping


olevski (dismissed):

olevski (approved):

Panaetius:
Actually this is not needed anymore since I had to merge back master anyways, though it was helpful for resolving the merge conflicts

PR #3591, opened 2023-08-15
fix(service): return project version as int

### LINKED ISSUE #3569: fix /apiversion endpoint format and redirect in gateway
currently the /apiversion endpoint returns `supported_metadata_version: 10.0` but this should be an int or string `"10"`, not a float.

Additionally, when doing e.g. `/api/renku/6/apiversion` (with no v6 renku-core pod deployed), the gateway forwards the call to the newest core-service instance (e.g. v10) instead of giving a 404

olevski:
/deploy #persist

closes #3569

### DIFF
--- a/renku/ui/service/serializers/version.py
+++ b/renku/ui/service/serializers/version.py
@@ -21,7 +21,7 @@ class VersionResponse(Schema):
     """Version response schema."""
 
     latest_version = fields.String()
-    supported_project_version = fields.Number()
+    supported_project_version = fields.Integer()
     minimum_api_version = fields.String()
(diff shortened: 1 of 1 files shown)


olevski:
If you visit: https://renku-ci-rp-3591.dev.renku.ch/api/renku/apiversion

The result has an integer instead of a float:

```json
{"result":{"latest_version":"2.7.0.dev3+g12f73dd5","maximum_api_version":"2.1","minimum_api_version":"1.0","supported_project_version":10}}
```

Panaetius (approved):

PR #3595, opened 2023-08-18
fix(core): setting non-existing config property to null

### LINKED ISSUE #3544: Setting a project setting to `null` more than once results in an error from `renku-core`
Setting a project setting to `null` more than once results in an error from `renku-core`:

Endpoint:

```
POST https://renku-ci-ui-2620.dev.renku.ch/ui-server/api/renku/10/config.set
```

Payload:

```json
{
  "git_url":"https://gitlab.dev.renku.ch/johann.thiebaut1/another-playground-project",
  "config": {
    "interactive.lfs_auto_fetch": null
  }
}
```

Response:

```json
{
  "error": {
    "code": 3112,
    "devMessage": "Unexpected error on setting 'interactive.lfs_auto_fetch', possibly caused by concurrent actions.",
    "devReference": "https://renku-python.readthedocs.io/en/latest/service_errors.html#renku.ui.service.errors.IntermittentSettingExistsError",
    "userMessage": "There was an error with the setting 'interactive.lfs_auto_fetch'. Please refresh the page and try again."
  }
}
```

**Expected outcome**

Setting a config value more than once should be idempotent. This is already the case for non-null values. Resetting a config value more than once should behave the same and not result in an error from `renku-core`.

olevski:
fixes #3544 

/deploy #persist

### DIFF
--- a/renku/ui/service/views/apispec.py
+++ b/renku/ui/service/views/apispec.py
@@ -78,7 +78,7 @@ spec = APISpec(
     openapi_version=OPENAPI_VERSION,
     version=API_VERSION,
     plugins=[MultiURLFlaskPlugin(), MarshmallowPlugin()],
-    servers=[{"url": SERVICE_API_BASE_PATH}],
+    servers=[{"url": SERVICE_API_BASE_PATH}, {"url": "/ui-server"+SERVICE_API_BASE_PATH}],
     security=[{"oidc": []}, {"JWT": [], "gitlab-token": []}],
     info={"description": TOP_LEVEL_DESCRIPTION},
 )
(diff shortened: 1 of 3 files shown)


Panaetius (approved):

renku/ui/service/views/apispec.py
```diff
@@ -78,7 +78,7 @@ def path_helper(self, path, operations, *, view, app=None, **kwargs):
     openapi_version=OPENAPI_VERSION,
     version=API_VERSION,
     plugins=[MultiURLFlaskPlugin(), MarshmallowPlugin()],
-    servers=[{"url": SERVICE_API_BASE_PATH}],
+    servers=[{"url": SERVICE_API_BASE_PATH}, {"url": "/ui-server"+SERVICE_API_BASE_PATH}],
```
mohammad-alisafaee, line 81: We can write this a f-string and avoid using string addition: `f"/ui-server{SERVICE_API_BASE_PATH}"`

olevski: Sure I will make the change. Just FYI this is still WIP and a draft.

Panaetius (approved):

PR #3608, opened 2023-08-25
feat(service): accept commit SHA in read endpoints

### LINKED ISSUE #3579: allow passing non-branch ref to read-only core-svc endpoints
For readonly endpoints like config.show, passing a commit sha or tag should be fine alongside a branch. This should not work for write endpoints, as writing on top of a tag or detached commit doesn't make sense.

The Ui needs this to get config values when launching sessions from a commit and some other cases.

olevski:
closes #3579 

/deploy #persist

This is ready for review.


renku/core/util/git.py
```diff
@@ -823,17 +823,27 @@ def clone(branch, depth):
...
+                raise
+            raise error_from_progress(progress, url)
 
         # NOTE: clone without branch set, in case checkout_revision was not a branch or a tag but a commit
         try:
             repository = clone(branch=None, depth=None)
         except errors.GitCommandError:
-            handle_git_exception()
-            raise
+            if re.match("^.* destination path .* already exists and is not an empty directory", err.stderr):
```
Panaetius, line 836: Should we just unconditionally clean the directory before each clone attempt instead of the complicated error handling mechanism here?

renku/ui/service/gateways/repository_cache.py
```diff
@@ -120,6 +132,8 @@ def _clone_project(
             "branch": branch,
             "git_url": git_url,
             "user_id": user.user_id,
+            "branch": branch,
```
Panaetius, line 135: branch is already there 4 lines above

renku/ui/service/gateways/repository_cache.py
```diff
@@ -182,6 +197,9 @@ def _clone_project(
 
     def _unshallow_project(self, project: Project, user: User):
         """Turn a shallow clone into a full clone."""
+        if project.commit_sha is None:
```
Panaetius, line 200: Shouldn't this be `is not None` ?

renku/ui/service/gateways/repository_cache.py
```diff
@@ -204,6 +222,10 @@ def _maybe_update_cache(self, project: Project, user: User):
         if project.fetch_age < PROJECT_FETCH_TIME:
             return
 
+        if project.commit_sha is None:
```
Panaetius, line 225: Shouldn't this be `is not None` ?

Panaetius (changes requested):

Panaetius (approved):

PR #3615, opened 2023-09-12
fix(cli): ssh setup and key usage

### LINKED ISSUE #3584: session ssh-setup command does not add public ssh keys to a project
To replicate:
1. Create a new project 
2. Use renku clone ... to clone it locally
3. Navigate to the root of the project and run `renku session ssh-setup`
4. `ls -la` shows that no `.ssh` folder was created (but one should have been created)

Tested on renkulab.io with renku cli version 2.6.1 and with renku version 0.33.0.

### LINKED ISSUE #3585: Streamline ssh access to sessions via the CLI
See https://renku.discourse.group/t/error-cannot-start-session-via-the-notebook-service/870/10

Two main issues:
- When the CLI is used to access the session via ssh then only a single specific (to that renku deployment) key pair from `~/.ssh/renku` is being used to log in, but we should simply try all key pairs in `~/.ssh` and in `~/.ssh/renku`.
- Revise/change the logic for how the CLI reports that a session supports SSH. This flag should be true/yes if the ssh daemon is running in the session. Not based on whether whoever is listing the session has access to it via ssh. From testing this feature out it seems that the SSH supported field starts out being set to false/no and when you successfully log in once with ssh then it shows yes.

olevski:
closes #3585 
closes #3584 

I am not sure that getting into people's ssh configurations to such extent (as we currently do) is maintainable and feasible without a lot of edge cases or bugs. So I decided to simply give a warning when the `renku session ssh-setup` command is used. When I saw this command (including other users from discord) I expected that it would inject ssh keys into the current project. But it does not do this. It only generates/updates stuff in `~/.ssh/renku`. 

I think a more sustainable solution is to step back from doing all this ssh configuration manipulation for users. Because the biggest problem is that it is hard to make things work both with and without using the renku CLI. But we can discuss this further in a separate issue.

### DIFF
--- a/renku/core/session/session.py
+++ b/renku/core/session/session.py
@@ -353,3 +353,8 @@ def ssh_setup(existing_key: Optional[Path] = None, force: bool = False):
             """
         )
         f.write(content)
+
+    communication.warn(
+        "This command does not add any public SSH keys into your project. "
+        "Keys have to be added manually or by using the renku session start command with the --ssh flag."
+    )
--- a/renku/core/session/session.py
+++ b/renku/core/session/session.py
@@ -355,6 +355,6 @@ def ssh_setup(existing_key: Optional[Path] = None, force: bool = False):
         f.write(content)
 
     communication.warn(
-        "This command does not add any public SSH keys into your project. "
-        "Keys have to be added manually or by using the renku session start command with the --ssh flag."
+        "This command does not add any public SSH keys to your project. "
+        "Keys have to be added manually or by using the 'renku session start' command with the '--ssh' flag."
     )
--- a/renku/command/session.py
+++ b/renku/command/session.py
@@ -39,7 +39,7 @@ def search_session_providers_command():
 
 def session_list_command():
     """List all the running interactive sessions."""
-    return Command().command(session_list)
+    return Command().command(session_list).with_database(write=False)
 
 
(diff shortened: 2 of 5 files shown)


renku/core/session/session.py
```diff
@@ -353,3 +353,8 @@ def ssh_setup(existing_key: Optional[Path] = None, force: bool = False):
             """
         )
         f.write(content)
+
+    communication.warn(
+        "This command does not add any public SSH keys into your project. "
```
mohammad-alisafaee, line 358: I'm not sure if we should use `into` here or just `to`.

renku/core/session/session.py
```diff
@@ -353,3 +353,8 @@ def ssh_setup(existing_key: Optional[Path] = None, force: bool = False):
...
+        "Keys have to be added manually or by using the renku session start command with the --ssh flag."
```
mohammad-alisafaee, line 359: The convention for such messages is to put commands/options/flags inside single quotes to differentiate them with the rest of the message: `... by using the 'renku session start' command with the '--ssh' flag.`

mohammad-alisafaee (commented):
Thanks Tasko! I've just added 2 comments regarding a message formatting/wording.

mohammad-alisafaee (approved):

Issue #3563, opened 2023-07-13
make name and slug consistent for datasets and projects


Panaetius:
There's some inconsistency with "name", "title", "slug" on the platform, with different services/the cli using different name for the same things.

To make this consistent across the whole platform, we should change the service and the CLI to have

name: The human readable (e.g. with spaces) name
slug: the machine readable (e.g. spaces replaced with `-`) name

The cli should then use `slug` as parameter in datasets and projects.

Also inform the omnibenchmark people once this is done.

PR #3620, opened 2023-09-18
fix: make usage of name and slug consistent

mohammad-alisafaee:
# Description
Replaces `name`/`title` with `slug`/`name` in datasets. Related command flags are hidden and deprecation warning is printed upon their usage (e.g. `--name`/`--short-name` when importing a dataset).

Fixes: #3563


Panaetius (changes requested):
Thank you! 

One question, do we need to bump the minimum renku version? I assume old CLi won't be able to work on a project written to by this version?

Panaetius (approved):

PR #3623, opened 2023-09-27
feat(cli, service): support for project image

### LINKED ISSUE #3576: add support for project images in the cli and core-svc
Users should be able to set an image for a project in the core-svc and the cli, analogous to how they can set one for datasets (but only a single image).

mohammad-alisafaee:
# Description

Adds `-i` and `--image` flags to `init` and `project edit` commands that accepts a path/URL to an image. The image is copied to `.renku/images/project/0.png` and added to the project.

## NOTE
* The image is always saved with a PNG extension; this shouldn't matter since UI loads the image correctly regardless (tested with dataset images).
* We don't allow adding non-image files from the local filesystem

Fixes #3576


renku/core/project.py
```diff
@@ -56,6 +64,11 @@ def edit_project(
         "description": description,
         "keywords": keywords,
         "custom_metadata": custom_metadata,
+        "image": NO_VALUE
+        if image_request is NO_VALUE
+        else None
+        if image_request is None
+        else ImageObjectRequestJson().dump(image_request),
```
Panaetius, line 71:
```suggestion
        "image": image_request
        if image_request is NO_VALUE or image_request is None
        else ImageObjectRequestJson().dump(image_request),
```
may be a bit easier to follow

Panaetius (approved):
looks great, tiny suggestion, feel free to ignore

Panaetius (approved):

PR #3626, opened 2023-09-28
chore(tests): fix template name check

olevski:
The test checks for the template name. But we changed the template name so the test fails even though everything works.


olevski:
The cache date is in the future because I accidentally committed todays date as the cache key in another branch so the things that exactly I did not want to cache got cached under todays date.

Panaetius (approved):

PR #3632, opened 2023-10-04
fix: don't fail on exotic git diff types

### LINKED ISSUE #3616: missing "U" ChangeType for git diffs
"U" is a valid git diff change type, but it's missing in the enumhttps://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/infrastructure/repository.py#L1282 This can cause issues in the check if the repository is dirty.

We should add it. "C" and "X" are also missing (see https://git-scm.com/docs/git-diff#_raw_output_format for all the codes)

Panaetius:
fixes #3616


mohammad-alisafaee (approved):
LGTM, thank you!

PR #3633, opened 2023-10-09
feat(cli): session pause and resume

### LINKED ISSUE #3621: Pause and resume sessions from CLI
We implemented session hibernation in https://github.com/SwissDataScienceCenter/renku-notebooks/pull/1518 that allows users to pause and resume sessions from the UI. We should implement proper CLI commands to have the similar functionality for sessions that are managed by CLI.
```
renku session pause <session-name>

renku session resume/start/restart <session-name>
```

As nice-to-haves: 
* When a user runs `renku session stop` on a session that has unsaved work, the cli warns them and prompts to ask if they really want to delete the session and lose unsaved changes, or if they want to pause it.
* If possible, there is no difference to the user between running `renku session start` and `renku session resume` on a paused session (so the user doesn't have to remember whether they have a paused session or are creating a new one).

mohammad-alisafaee:
# Description

Adds `session pause` and `session resume` commands.

Fixes #3621


Panaetius (approved):
Thank you, looks all good!

PR #3640, opened 2023-10-13
feat: add prometheus metrics

### LINKED ISSUE #3434: Add prometheus metrics to core service and rq
Check what metrics we already have and add more metrics to allow seeing what the state/load of the core svc and redis-queue is.

Check with YAT on which metrics make sense for them and what they'd like to see.

Panaetius:
Fixes #3434 

/deploy extra-values=core.metrics.enabled=true


mohammad-alisafaee (commented):
Thank you! Does this collect a set of default metrics which are good enough for us?

Panaetius:
> Thank you! Does this collect a set of default metrics which are good enough for us?

Yes, I think it's pretty comprehensive, for each endpoint. You can see the default metrics here: https://github.com/rycus86/prometheus_flask_exporter#default-metrics

mohammad-alisafaee (dismissed):

mohammad-alisafaee (approved):

PR #3646, opened 2023-10-20
feat(svc): rename git_url to template_git_url and return git_url on all endpoints

### LINKED ISSUE #3564: Return git_url in core-service responses
Currently we return `project_id` in core-svc responses like `project.show`, but that id is not globally unique, for instance it remains unchanged if a project is forked.

We should return the git_url in responses instead, as this would help the ui to differentiate responses better.

Panaetius:
closes #3564


Panaetius:
This should be ready. Feel free to merge. Need to coordinate with UI when it is merged

renku/ui/service/serializers/datasets.py
```diff
@@ -131,7 +132,7 @@ class DatasetDetailsResponse(DatasetDetails):
     images = fields.List(fields.Nested(ImageObject))
 
 
-class DatasetListResponse(Schema):
+class DatasetListResponse(GitUrlResponseMixin):
```
olevski, line 135: Adding mixins and using inheritance in all the schemas like this makes the code really hard to read and follow imo. To see what fields the `DatasetListResponse` contains you have to click through several levels of inheritance until you find a specific field. And there is no single place where you can look at to see what the full payload is supposed to be.

I know this is not something new but I think we should try to use it a lot more sparingly.

olevski: For example for `ManifestTemplatesRequest_v2_2` in `renku/ui/service/serializers/v1/templates.py` I was wondering why we have these lines in the `pre_load` section of the schema (because they do not apper in the code in `ManifestTemplatesRequest_v2_2`):

```
data["git_url"] = data.get("url")
data["template_git_url"] = data["git_url"]
```

So to find them I had to step through:
- `ManifestTemplatesRequest_v2_2`
- `RepositoryCloneRequest`
- `RemoteRepositorySchema`
- `RemoteRepositoryBaseSchema`

I think we should stick to just 1 level of inheritance or 2 levels in extreme situations. And if we need to reuse things we should reuse/compose field definitions that are common across different schemas rather than have this many levels of inheritance.

And even after all this I cannot see where we need `template_git_url`

olevski: We should not try to fix this now. But perhaps it is something we can consider simplifying in the cooldown period.

olevski: Ok I found `template_git_url` it seems I was not at the end of the inheritance chain at `ManifestTemplatesRequest_v2_2` - as I assumed.

The whole chain is:
- `ProjectTemplateRequest_v2_2`
- `ManifestTemplatesRequest_v2_2`
- `RepositoryCloneRequest`
- `RemoteRepositorySchema`
- `RemoteRepositoryBaseSchema`

I am sorry for the long comments - I just wanted to illustrate my point.

renku/ui/service/serializers/templates.py
```diff
@@ -17,30 +17,51 @@
...
 
 
-class ManifestTemplatesRequest(RepositoryCloneRequest):
+class ManifestTemplatesRequest(Schema):
     """Request schema for listing manifest templates."""
 
-    url = fields.String(required=True)
+    template_git_url = fields.String(required=True, metadata={"description": "Template git repository url."})
     depth = fields.Integer(load_default=TEMPLATE_CLONE_DEPTH_DEFAULT)
+    branch = fields.String(load_default=None, metadata={"description": "Remote git branch (or tag or commit SHA)."})
```
olevski, line 38: If I look at the swagger page on dev.renku.ch currently we have `url`, `ref` and `depth` as the parameters for `templates.read_manifest`. Why do we need to switch to branch. I think `ref` is a much more accurate and clearer name for this. Calling it `branch` and using it just like `ref` is really confusing.

renku/ui/service/serializers/templates.py
```diff
@@ -17,30 +17,51 @@
...
+
+    @pre_load
+    def set_fields(self, data, **_):
+        """Set `branch` field from `ref` if present and set template url."""
+        if "ref" in data and not data.get("branch"):
+            # Backward compatibility: branch and ref were both used. Let's keep branch as the exposed field
+            # even if internally it gets converted to "ref" later.
```
olevski, line 45: I loooked through all versins of `templates.read_manifest` on swagger and not a single one has `branch`. Where did we use `branch` before?

olevski (changes requested):

renku/ui/service/serializers/datasets.py
Panaetius: I completely agree, as it currently is, it's terrible. There's also some endpoints that take parameters that they don't use, just because a schema was inherited that doesn't fully match. I removed some of that here, where the template endpoints took project clone values that they just ignored for no good reason.
I think cooldown is a good fit for addressing this

renku/ui/service/serializers/templates.py
Panaetius: the swagger gets it from here: https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/ui/service/views/templates.py#L56
but the code has always used branch: https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/ui/service/controllers/templates_read_manifest.py#L52
which is ultimately defined here: https://github.com/SwissDataScienceCenter/renku-python/blob/develop/renku/ui/service/serializers/common.py#L56
it never used/worked with ref.
so I merely cleaned it up without a breaking change, swagger was just wrong before

olevski (approved):

PR #3648, opened 2023-11-01
feat(service): date_published in datasets.list response

### LINKED ISSUE #3647: Core service `/datasets.edit` endpoint should return `date_published`
Imported datasets have a `date_published` property that is missing from the response when listing datasets in the core service. This field should be added to the `/datasets.list` response.

mana-alisafaee:
# Description

Fixes #3647


Panaetius:
what's the state of this pr?

Panaetius (approved):

PR #3651, opened 2023-11-08
fix(service): set job timeouts correctly

Panaetius:
adds timeouts to rq jobs to cleanup metrics dashboard.
ttl=how long a job stays enqueued before failing
failure_ttl=how long a failed job is kept around


olevski (approved):
Sorry for the delay Ralf.

PR #3660, opened 2023-11-29
fix(cli): output proper session link and only check registry if logged in

### LINKED ISSUE #3658: Open the right link for Jupyterlab when running in docker
See https://renku.discourse.group/t/what-is-the-correct-renkulab-container-registry-address/958

### LINKED ISSUE #3659: CLI uses the wrong gitlab registry
See the beginning of https://renku.discourse.group/t/what-is-the-correct-renkulab-container-registry-address/958

The CLI assumes the registry is at `registry.gitlab.renkulab.io` but it is actually at `registry.renkulab.io`.

Panaetius:
closes #3659 
closes #3658


mohammad-alisafaee (approved):
LGTM, thank you!

PR #3670, opened 2023-12-13
fix(service): dataset import error

mohammad-alisafaee:
# Description

Renaming `name` to `slug` was not done for a service function which caused an exception when importing datasets from UI.


Panaetius (approved):
thank you for fixing this

PR #3678, opened 2024-01-15
fix(service): fix clone depth not being respected

Panaetius:
we always did a full clone instead of shallow when a user is logged in.

/deploy extra-values=global.renku.cli_version=2.8.0


olevski:
Wow so this is because the parameters to the git python function we call changed? Like from `depth` to `clone_depth`?

Panaetius:
> Wow so this is because the parameters to the git python function we call changed? Like from `depth` to `clone_depth`?

yes, and because we had the unknown=ignore part, we just never noticed...

PR #3690, opened 2024-01-22
fix(core): migrate Dockerfile after metadata

Panaetius:
properly fixes migration issues.
Docker/Template migration is now done after metadata migration and double checks if the Dockerfile can be migrated.


mana-alisafaee (approved):
LGTM. Thank you!

